PR #94, opened 2020-06-01
docs(deploy): deploy docs to gh-pages on master branches

dcwither:
### Summary:

Enable automatic deploys to docs site at https://chanzuckerberg.github.io/lp-design-system/?path=/story/tokens--page. This makes access by non engineers to design system docs easier as we're building.

### Test Plan:

Example passing workflow build at https://github.com/chanzuckerberg/lp-design-system/runs/728640162?check_suite_focus=true


.github/workflows/docs.yml
```diff
@@ -0,0 +1,36 @@
...
+          npm ci
+          npx lerna bootstrap
+
+      - name: Build Dist 🔧
+        run: npx lerna run build --stream
+
+      - name: Deploy Storybook Docs 🚀
+        run: npx lerna run deploy:docs
+        env:
+          GH_TOKEN: "${{ github.actor }}:${{ secrets.GITHUB_TOKEN }}"
```
dcwither, line 35: This was necessary to get the push access for the storybook-deploy script. It's a bit of a hack.

shitchcock:
Seems fine, hacking the GH token can be resolved later if we think of a better solution. No need to let it be blocking now.

PR #122, opened 2020-07-14
Add .nvmrc file and documentation

ahuth:
### Summary:

This PR adds an .nvmrc file to

1. Document which version of node is required
2. Make it easy to use [nvm](https://github.com/nvm-sh/nvm) to manage the node version (can install with `nvm install`, and switch to the correct version via `nvm use`)

If we decide to convert over to nodeenv, we can delete the .nvmrc file and update the readme.

### Test Plan:

n/a

@dcwither , what do you think about this?


dcwither:
@ahuth I didn't notice this before, but the build failed because the commit convention failed. I've accidentally done this a few times with the npm publishing. Something to keep an eye out for.

dcwither:
> **EDIT:** It seems it passed here, but not when it rebuilt in master. I'll do some digging. 

Ah, it seems you probably did a squash and merge which added a header above it. Maybe I should disable squash and merge?

ahuth:
Yeah, I was going to ask about this. I picked squash and merge since it was the default option. Do we prefer the rebase option?

dcwither:
I disabled squash and merge, we have a commitlint, so commits are already checked for their adherence to those requirements.

PR #123, opened 2020-07-16
improvement(color): move white to grey-50 and change white to #FFFFFF

dcwither:
### Summary:

Color migration still needs a true white (#FFFFFF) available, so we're setting that to white, and moving the current white to grey-50.

### Test Plan:
`npm start` and check http://localhost:6006/?path=/story/tokens--page to ensure the token is present
`lerna run build`

### DIFF
--- a/packages/tokens/properties/color/base.json
+++ b/packages/tokens/properties/color/base.json
@@ -2,6 +2,7 @@
   "legacy": {
     "color": {
       "gray": {
+        "50": { "value": "#F8F9FC" },
         "100": { "value": "#EBEBEC" },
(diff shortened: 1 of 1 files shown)


shitchcock:
Does this mean the whitest of whites on the site is currently?
image
and now the whitest of whites will be:
image

dcwither:
> Does this mean the whitest of whites on the site is currently?

There's already #FFFFFF in the site, but it wasn't accounted for here. I think the desire was to have a softer default background color (which was assumed to be white), but several of our backgrounds are white, and in the future we'll need white for inputs etc. This actually better reflects reality.

PR #167, opened 2020-09-12
Add axe-storybook

ahuth:
### Summary:

axe-storybook in edu-design-system 2

The current version is pretty slow (~500ms  per story). Not a big deal for this project currently.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [8.x, 10.x, 12.x]
+        node-version: [10.x, 12.x]
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,11 +52,17 @@ jobs:
 
       - name: Build Storybook 🔧
         run: npx lerna run build:storybook
-        
+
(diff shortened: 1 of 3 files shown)


packages/components/package.json
```diff
@@ -46,6 +48,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.6.4",
+    "@chanzuckerberg/axe-storybook": "^0.2.0",
```
ahuth, line 51: This uses playwright, instead of puppeteer, so we can run the tests in different browsers. Just noticed, though, that it doesn't install successfully on Node 8.

Can we drop Node v8 support from this project?

.github/workflows/test.yml
```diff
@@ -28,6 +28,9 @@ jobs:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
           CI: true
 
+      - name: A11y Testing ❤️
+        run: npx lerna run axe --stream
```
ahuth, line 32: Doesn't seem to finish the storybook build, which is weird: https://github.com/chanzuckerberg/lp-design-system/pull/167/checks?check_run_id=1120208796

ahuth:
Closing for now. Storybook will need to be upgraded to v6, and I had some issues doing that here.

PR #195, opened 2020-10-19
feat(tokens): replace eds token numbers

dcwither:
### Summary:
Rename tokens to better match the existing tokens. I think this will make a cleaner mapping.

### Test Plan:
`npm start`, check the storybook colors docs
image


ahuth (approved):
Makes sense, and LGTM 🤘

PR #203, opened 2020-11-04
Proposal: Use Babel to process TypeScript (and other updates)

ahuth:
### Summary:

Couple of updates, but the biggest is a proposal (discussed with Devin previously) to use Babel to process TypeScript, instead of tsc.

**Pros**: We have a much wider array of transformations that can be done. Also, the build step will be faster (offset by the added declaration file generation step, though).

**Cons**: Our build will be a bit more complicated. Babel can be slightly confusing to configure, and adds some dependencies. Also there's now an added step to generate declaration files.

There's nothing wrong with using tsc to compile this code, though, if we want. So a valid option is to not use Babel for this.

Additionally, this also
- Upgrades TypeScript from v3 to v4
- Adds a .node-version file, so that people like me who use nodenv can have a smoother experience
- Removes unnecessary namespaced React imports
- Makes defining default props more consistent

### Test Plan:

- CI
- Local testing


.github/workflows/test.yml
```diff
@@ -84,4 +84,26 @@ jobs:
...
+          fetch-depth: 0
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Install Dependencies ⬆️
+        run: |
+          npm ci
+          npx lerna bootstrap
+      - name: Run TypeScript type checking ✅
```
ahuth, line 108: Babel just removes the TypeScript types, and doesn't do any actual type checking. Because of that, we now need an explicit type-checking step.

packages/components/.babelrc.js
```diff
@@ -0,0 +1,16 @@
...
+    "@babel/preset-typescript",
+    "@babel/preset-react",
+    [
+      "@babel/preset-env",
+      {
+        targets: [
+          "last 2 chrome versions",
+          "last 2 firefox versions",
+          "last 2 safari versions",
+        ],
```
ahuth, line 12: tsc wasn't previously doing any crazy transformations, or polyfilling anything. So targeting recent versions here feels pretty safe.

We can always tweak this as necessary.

packages/components/package.json
```diff
@@ -20,9 +20,10 @@
     "url": "git+https://github.com/chanzuckerberg/lp-design-system.git"
   },
   "scripts": {
-    "build": "npm run build:clean && npm run build:ts -- --sourceMap false",
+    "build": "npm run build:clean && npm run build:ts && npm run build:types",
     "build:clean": "rm -rf dist/",
-    "build:ts": "tsc -project . --rootDir src/ ",
+    "build:ts": "babel --extensions '.js,.ts,.tsx' --ignore '**/*.stories.tsx','**/*.spec.tsx' --out-dir dist/ src/",
```
ahuth, line 25: This is ignoring '**/*.stories.tsx' instead of '**/*.stories.*' because of https://github.com/babel/babel/issues/12008, which is unfortunate.

packages/components/src/button/button.stories.tsx
```diff
@@ -1,6 +1,5 @@
-import * as React from "react";
-
 import Button from "./button";
+import React from "react";
```
ahuth, line 2: No need for the * import. We used to need that in Flow, but not anymore (and certainly not in TypeScript).

Not sure why eslint made me re-order this import, but not the one in the component file, though 🤔

packages/components/src/button/button.tsx
```diff
@@ -7,11 +7,11 @@ type ButtonProps = React.DetailedHTMLProps<
   React.ButtonHTMLAttributes<HTMLButtonElement>,
   HTMLButtonElement
 > & {
-  variant: "primary" | "secondary";
+  variant?: "primary" | "secondary";
 };
 
 function Button(props: ButtonProps): JSX.Element {
-  const { className = "", variant, ...rest } = props;
+  const { className = "", variant = "primary", ...rest } = props;
```
ahuth, line 14: I know this component is really just here for testing purposes, but figured we may as well be consistent in defining default props.

ahuth:
I think I messed up something in the package lock. Closing for now. We can always revisit later if we want.

PR #206, opened 2020-11-13
feat(components): adds Typography base component

anniehu4:
### Summary:
V1 Typography component! Uses styled-components and twin.macro.

[Figma type ramp](https://www.figma.com/file/sKEhNrLEdexPeVVn58dkWx/SDL-(Summit-Design-Language)-Explorations?node-id=1518%3A2925)

Type ramp in Storybook:
Screen Shot 2020-12-15 at 1 29 56 PM

### Test Plan:
- Jest tests pass
- `npm start` to run Storybook
- Confirm visual changes in Percy match Figma


packages/components/src/typography/typography.tsx
```diff
@@ -0,0 +1,102 @@
+import React, { ReactNode } from "react";
+
+import cx from "classnames";
+
+type TypographyElement = "h1" | "h2" | "h3" | "h4" | "h5" | "h6" | "span" | "p";
+
+type TypographySize =
```
anniehu4, line 7: would it be better to use semantic names?? this ties closely to the token names

packages/styles/src/typography/typography.scss
```diff
@@ -0,0 +1,18 @@
...
+body {
+  color: theme('colors.black');
+  font-family: 'Poppins', sans-serif;
+  font-size: 16px;
+}
+
+.text-xl {
+  font-size: 2rem;
+  line-height: 2.5rem;
+}
```
anniehu4, line 14: I would add a block for each text size

packages/components/src/typography/typography.tsx
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * Components that wrap typography sometimes requires props such as event handlers
+     * to be passed down into the element. One example is the tooltip component.  It
+     * attaches a onHover and onFocus event to the element to determine when to
+     * trigger the overlay.
+     */
+    ...rest
+  } = props;
+
+  const className = cx(`czedi-text-${size}`, color);
```
anniehu4, line 91: I'm not sure the best way to handle color ... I didn't want to enumerate every option here as a potential class. But this current approach doesn't feel great

dcwither: this is where we need to [start applying BEM](http://getbem.com/naming/) `.czedi-text--${color}`. Enumerating them may be cumbersome, but it'll allow us to be explicit - and we could iterate over them with the right `scss` syntax.

anniehu4: ahh ok will try

anniehu4: updated! that definitely makes more sense

packages/styles/src/typography/typography.scss
```diff
@@ -0,0 +1,33 @@
+@import url("https://fonts.googleapis.com/css2?family=Arimo:wght@400;600&display=swap");
+
+// Set font family and base size at a root level
+body {
+  color: theme("font.color.default");
+  font-family: theme("font.family"), sans-serif;
+  font-size: theme("font.size.default");
+}
+
+$sizes: ("h1", "h2", "h3", "h4", "h5", "default", "sm", "xs");
```
anniehu4, line 10: Is there a way for this and `$colors` to read from the JSON config instead? I've only tried using the `theme()` function and that doesn't work

packages/components/package.json
```diff
@@ -22,7 +22,7 @@
   "scripts": {
     "build": "npm run build:clean && npm run build:types && npm run build:js",
     "build:clean": "rm -rf dist/",
-    "build:js": "babel src -d dist --extensions '.js,.jsx,.ts,.tsx' --ignore src/**/*.spec.* --ignore src/**/*.stories.*",
```
anniehu4, line 25: this didn't work after all 😞 it would ignore the playground/ folder but not any other folders. looks like https://github.com/babel/babel/issues/12008 is still open too

packages/components/tailwind.config.js
```diff
@@ -3,7 +3,11 @@ const tokens = require("@chanzuckerberg/czedi-kit-tokens/json/variables-nested.j
 module.exports = {
   theme: {
     colors: tokens.eds.color,
-    extend: {},
+    fontFamily: {
+      sans: "Arimo, sans-serif",
```
anniehu4, line 7: note that this isn't actually imported to storybook yet, I'm thinking I should create a global style file and include it there -- that can also be the file that's exported to Traject?

packages/components/src/typography/typography.tsx
```diff
@@ -0,0 +1,169 @@
...
+   * actually be rendered, in the case that you want to render an element
+   * as one heading but style it as if it were another. If both an `as` prop
+   * and a `size` prop are passed, the `as` will be used to determine the element
+   * and the `size` will be used to determine the styling.
+   */
+  as?: TypographyElement;
+  /**
+   * Specifies font weight as either bold or normal (defaults to normal).
+   */
+  bold?: boolean;
```
anniehu4, line 122: Per Sean, originally all headers were only going to be bold. But designers said they needed non-bold options, so now most sizes have a bold and non-bold design in Figma. We don't have strict usages tied to each one yet, so I think we should leave this as a prop for now

packages/components/src/playground/button.tsx
```diff
@@ -1,29 +1,34 @@
...
+  const ButtonComponent =
+    variant === "primary" ? PrimaryButton : SecondaryButton;
 
-  return variant === "primary" ? (
-    <PrimaryButton {...rest} />
-  ) : (
-    <SecondaryButton {...rest} />
+  return (
+    <ButtonComponent {...rest}>
+      <Typography bold color="white" size="button">
```
anniehu4, line 28: wasn't 100% sure why there was a `Button` size in Figma, but I'm guessing this is how we'd use it? also fine with reverting this if that's not standard

packages/components/twin.d.ts
```diff
@@ -0,0 +1,16 @@
+import "twin.macro";
```
anniehu4, line 1: @ahuth fyi moved it here

packages/components/tailwind.config.js
anniehu4: alright ... here's my attempt at that: https://github.com/chanzuckerberg/lp-design-system/pull/231

packages/components/src/typography/typography.tsx
```diff
@@ -0,0 +1,161 @@
...
+      return tw`text-white`;
+    default:
+      return tw`text-font-base`;
+  }
+}
+
+function styleFromSize(size: TypographySize): TwStyle {
+  switch (size) {
+    case "h1":
+      return tw`text-h1 leading-h1 capitalize`;
```
dcwither, line 63: disable capitalize because articles

packages/components/src/typography/typography.tsx
```diff
@@ -0,0 +1,161 @@
...
+      return tw`text-sm leading-sm`;
+    case "xs":
+      return tw`text-xs leading-xs`;
+    case "button":
+      return tw`text-sm leading-body`;
+    case "caption":
+      return tw`text-xs leading-sm`;
+    case "overline":
+      return tw`text-xs leading-sm uppercase`;
+    default:
```
dcwither, line 84: probably don't need default because typescript, feel free to remove this rule from eslint

packages/components/src/typography/typography.tsx
```diff
@@ -0,0 +1,161 @@
...
+     * Components that wrap typography sometimes requires props such as event handlers
+     * to be passed down into the element. One example is the tooltip component.  It
+     * attaches a onHover and onFocus event to the element to determine when to
+     * trigger the overlay.
+     */
+    ...rest
+  } = props;
+
+  return (
+    <TypographyComponent
```
dcwither, line 149: TODO: mutually exclusive props e.g. headings + bold

packages/components/tailwind.config.js
dcwither: I think I'd be more comfortable calling this `arimo`

dcwither (approved):

packages/components/src/typography/typography.tsx
anniehu4: added to the `bold` prop comment

PR #207, opened 2020-11-13
feat(tokens): add initial font size/color tokens

anniehu4:
### Summary:
Adds font tokens for color, size, and line-height according to [current Figma type ramp](https://www.figma.com/file/sKEhNrLEdexPeVVn58dkWx/SDL-(Summit-Design-Language)-Explorations?node-id=1518%3A2925).

~I changed the existing file structure a bit to move things under a `font/` folder because the token naming made more sense to me that way~ Followed existing file structure because it's consistent with style dictionary docs.

### Test Plan:
- `npx lerna run build` compiles
- Variables in JSON and CSS files look correct

Also, I'll test these tokens in https://github.com/chanzuckerberg/lp-design-system/pull/206


packages/tokens/properties/font/base.json
```diff
@@ -0,0 +1,17 @@
+{
+    "eds": {
+      "font": {
+        "family": { "value": "Poppins" },
```
anniehu4, line 4: Based on DS meeting today, maybe this should just be `Arimo`?

packages/tokens/properties/font/size.json
```diff
@@ -0,0 +1,66 @@
+{
+  "eds": {
+    "font": {
+      "size": {
+        "xl": {
+          "value": "2rem",
+          "comment": "Font size for Heading 3 sized text"
```
anniehu4, line 7: Two notes:
- This is skipping Headings 1&2 ([slack thread w/ Sean](https://chanzuckerbergteam.slack.com/archives/CPJN8SDFE/p1605212242002000?thread_ts=1605212052.001800&cid=CPJN8SDFE)) for now
- The namings don't totally match Figma ... the problem is `H4/H5` are marked as `S/XS`, but we also have body text that is Small/Xsmall 🤔 

I think when Sean finalizes guidance in the next week it will clarify both

anniehu4:
eek `postcss src/*.scss src/*/[a-z\-]*.scss --no-map --ext=css --dir=dist/ --base=src/` is failing here but passes locally 😢

packages/tokens/properties/font/size.json
dcwither: I'm also inclined to name the sizes `h1, h2, ...` so that it's communicated when no element is explicitly provided.

anniehu4:
syncing with Sean tomorrow on the tokens

packages/tokens/config.js
```diff
@@ -0,0 +1,66 @@
+// eslint-disable-next-line no-undef
+const StyleDictionary = require("style-dictionary");
```
anniehu4, line 2: error was "`require` is not defined" which I think is expected here / fine to ignore?

packages/tokens/config.js
```diff
@@ -0,0 +1,66 @@
...
+});
+
+EDSStyleDictionary.registerTransform({
+  name: "name/kebabCase",
+  type: "name",
+  transformer: function ({ path }) {
+    return path.join("-");
+  },
+});
+
```
anniehu4, line 65: This makes sure the heading tokens are output as `size-font-h1` instead of split into `size-font-h-1`. This is an open issue for style-dictionary: https://github.com/amzn/style-dictionary/issues/300, so I used the solution in thread. Looks like they might address it at some point, but haven't yet

packages/tokens/properties/size/font.json
```diff
@@ -1,44 +1,68 @@
...
-          "comment": "Line height for medium text"
+        "h3": {
+          "value": "2.5rem",
+          "comment": "Line height for Heading 3 sized text"
+        },
+        "h4": {
+          "value": "2rem",
+          "comment": "Line height for Heading 4 sized text"
+        },
+        "base": {
```
anniehu4, line 55: Note that the `h5` line height is the same as `base`, so I opted to not duplicate that token

packages/tokens/config.js
ahuth: Hmmm, I think this isn't expected. Looks like eslint doesn't know that this file is in a node environment.

Our eslintrc is trying to set that up, though, which we can see in https://github.com/chanzuckerberg/lp-design-system/blob/925402f0b91581f3fdaf1078b823b9b6b43eaa10/.eslintrc#L45-L51.

Not sure why it doesn't seem to be working. But we'll want to dig into why that isn't working as expected.

dcwither: I think instead of that you should make its env `node` this is probably defined in the top level eslint, we can make an override for some files.

packages/tokens/config.js
dcwither: should we maybe to `path.join('-').toLowerCase()`?

dcwither (commented):
overall lgtm, just a few nits - also tempted to include built tokens into the repo since it's hard to tell what the impact will be here.

packages/tokens/config.js
dcwither: @ahuth beat me to it

anniehu4: good call, thanks :)

anniehu4:
> also tempted to include built tokens into the repo since it's hard to tell what the impact will be here

added -- agree it'll make reviews easier

packages/tokens/properties/size/font.json
```diff
@@ -1,44 +1,68 @@
...
-          "comment": "Font size for small text"
+        "h4": {
+          "value": "1.5rem",
+          "comment": "Font size for Heading 4 sized text"
+        },
+        "h5": {
+          "value": "1.125rem",
+          "comment": "Font size for Heading 5 sized text"
         },
         "base": {
```
dcwither, line 25: we can leave base since we'll use it in the root/body to set 1rem = 16px, but we should probably also have a body value.

.eslintrc
```diff
@@ -43,7 +43,7 @@
       }
     },
     {
-      "files": ["*.config.js"],
+      "files": ["config.js", "*.config.js"],
```
dcwither, line 46: 😭 so close

dcwither (approved):
changes are good, just a few nits

PR #213, opened 2020-12-03
Dcwither/update node to lts

dcwither:
### Summary:
This should fix issues with building on node version 14.x.  I've also updated the workflows to run on 10,12,14 instead of 8,10,12.

### Test Plan:
Local:

- install node 14.5.0
- reinstall dependencies
- `npm start`

CI: run it


lesliecodes:
Hm.. I just tried this out but it fails 😢

dcwither:
@lesliecodes can you share the failure? Was it the arg types one, or a node-sass one? For node sass you need to uninstall all the dependencies again.

.github/workflows/size-limit.yml
```diff
@@ -6,6 +6,11 @@ on:
 jobs:
   size:
     runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [12.x]
```
anniehu4, line 12: why is this one running on 12 instead of 14?

anniehu4 (commented):
is the bundle size failure expected still?

lesliecodes:
It works! I did have to remove all the old /node_modules

.github/workflows/size-limit.yml
dcwither: Didn't want to mess with that one yet, was hoping it would prevent it from failing. I don't think it has.

dcwither: Once this is merged, if the next one fails, I'll look into it more.

anniehu4 (approved):

PR #218, opened 2020-12-08
WIP linaria typography prototype

anniehu4:
### Summary:

haven't set up Babel or anything so `npm start` doesn't work yet :')


packages/components/tsconfig.json
```diff
@@ -1,25 +1,26 @@
...
+    "lib": ["es5", "es6", "dom"],
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "noImplicitAny": false,
     "noImplicitReturns": true,
     "noImplicitThis": true,
-    "noImplicitAny": false,
+    "noUnusedLocals": true,
+    "outDir": "dist/",
+    "resolveJsonModule": true,
```
anniehu4, line 19: added this line

packages/components/src/typography/typography.tsx
```diff
@@ -0,0 +1,124 @@
...
+     * attaches a onHover and onFocus event to the element to determine when to
+     * trigger the overlay.
+     */
+    ...rest
+  } = props;
+
+  const TypographyComponent = styledComponentFromSize(size);
+  return (
+    <TypographyComponent
+      as={as || elementFromSize(size)}
```
anniehu4, line 115: pulled this from the Along component, but I can't tell how it works 🤔

ahuth: Are you referring to the `as` prop, specifically?

If so, `as` is a prop from styled-components, so I wouldn't expect it to work here.

The intent is that someone can specify that they want something to look like an h2 (or h1, h3, whatever), but actually be something else (to fit into the document structure).

anniehu4: ohhh ok that makes much more sense. I was going crazy trying to figure out how `as` was actually being used

packages/components/src/typography/typography.tsx
```diff
@@ -0,0 +1,124 @@
...
+  color?: TypographyColor;
+  /**
+   * The size of the html element. If no `as` prop is provided, then
+   * the component uses this value to determine which html tag to render
+   * (e.g. 'h1', 'h2', etc.)
+   */
+  size: TypographySize;
+};
+
+const typography = styled.span`
```
ahuth, line 53: Linaria does have this `styled` api. Something to think about: when does it make sense to use it over its `css` api?

I like to use `css` when possible, to encourage folks to write non-dynamic CSS. But `styled` is a good option when we definitely need to change styles based on props.

Up to you, though, for what makes sense for the EDS components here 😄

packages/tokens/properties/font/base.json
```diff
@@ -0,0 +1,17 @@
+{
```
anniehu4, line 1: these are from #207 -- still planning to land separately, but I had to update the base branch since I split the linaria/twin.macro imports to another PR

anniehu4:
revisit when we switch to linaria + twin.macro

PR #220, opened 2020-12-08
build: use babel for compilation

anniehu4:
### Summary:
Following https://github.com/chanzuckerberg/lp-design-system/issues/212, switches to using babel. This is a prereq for using Linaria

### Test Plan:
```
npm run build
```
should successfully compile 1 file (`dist/button/button.js` ), ignoring `spec` and `stories` files, and 1 declarations file (`dist/button.d.ts`)

### DIFF
--- a/packages/components/.babelrc
+++ b/packages/components/.babelrc
@@ -1,5 +1,19 @@
 {
-  "presets": ["@babel/preset-typescript"],
+  "presets": [
+    "@babel/preset-typescript",
+    "@babel/preset-react",
+    [
+      "@babel/preset-env",
+      {
+        "targets": [
+          "last 2 chrome versions",
+          "last 2 firefox versions",
+          "last 2 safari versions",
+          "last 2 edge versions"
+        ]
+      }
+    ]
+  ],
   "plugins": [
     "@babel/plugin-transform-modules-commonjs",
     [
--- a/packages/components/.babelrc
+++ b/packages/components/.babelrc
@@ -1,0 +1,13 @@
+{
+  "presets": ["@babel/preset-typescript"],
+  "plugins": [
+    "@babel/plugin-transform-modules-commonjs",
+    [
+      "module-resolver",
+      {
+        "root": ["./src"],
+        "extensions": [".js", ".jsx", ".ts", ".tsx"]
+      }
+    ]
+  ]
+}
--- a/packages/components/.storybook/preview.tsx
+++ b/packages/components/.storybook/preview.tsx
@@ -10,7 +10,7 @@ import { addParameters } from "@storybook/react";
 import { withA11y } from "@storybook/addon-a11y";
 
 addDecorator(withA11y);
-addDecorator(storyFn => <div dir="ltr">{storyFn()}</div>);
+addDecorator((storyFn) => <div dir="ltr">{storyFn()}</div>);
 addParameters({
   docs: {
     container: DocsContainer,
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -67,6 +67,7 @@
     "@testing-library/react": "^9.3.0",
     "@types/jest": "^24.0.19",
     "@types/jest-axe": "^3.2.0",
+    "babel-jest": "^26.6.3",
     "babel-loader": "^8.0.6",
     "babel-plugin-module-resolver": "^4.0.0",
     "jest": "^24.9.0",
@@ -76,7 +77,6 @@
     "react-docgen-typescript-loader": "^3.3.0",
     "react-dom": "^16.10.2",
     "sass-loader": "^8.0.2",
-    "ts-jest": "^24.1.0",
     "ts-loader": "^6.2.0",
     "typescript": "^4.1.2"
   },
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -49,6 +49,8 @@
     "@babel/cli": "^7.12.8",
     "@babel/core": "^7.12.9",
     "@babel/plugin-transform-modules-commonjs": "^7.12.1",
+    "@babel/preset-env": "^7.12.7",
+    "@babel/preset-react": "^7.12.7",
     "@babel/preset-typescript": "^7.12.7",
     "@commitlint/cli": "^8.3.5",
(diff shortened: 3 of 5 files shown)


packages/components/.storybook/preview.tsx
```diff
@@ -10,7 +7,7 @@ import { addParameters } from "@storybook/react";
 import { withA11y } from "@storybook/addon-a11y";
 
 addDecorator(withA11y);
-addDecorator(storyFn => <div dir="ltr">{storyFn()}</div>);
+addDecorator((storyFn) => <div dir="ltr">{storyFn()}</div>);
```
dcwither, line 10: does prettier not care one way or the other?

packages/components/package.json
```diff
@@ -40,12 +41,14 @@
     "react": ">= 16.8.0"
   },
   "dependencies": {
-    "@chanzuckerberg/czedi-kit-styles": "^0.0.1-alpha.5",
```
anniehu4, line 43: I kept hitting a build error where it couldn't find this package 😭  that's why I removed lines from `.storybook/preview.tsx`, and why there are Percy changes

I don't know if it's an easy fix, but soon we won't need this dependency anyways (and the button is more of a playground component) so I felt ok about it

packages/components/package.json
```diff
@@ -20,10 +20,11 @@
     "url": "git+https://github.com/chanzuckerberg/lp-design-system.git"
   },
   "scripts": {
-    "build": "npm run build:clean && npm run build:ts -- --sourceMap false",
+    "build": "npm run build:clean && npm run build:js",
     "build:clean": "rm -rf dist/",
-    "build:ts": "tsc -project . --rootDir src/ ",
-    "build:watch": "npm run build:ts -- --watch",
+    "build:js": "babel src --out-dir dist --extensions .js,.jsx,.ts,.tsx --ignore src/**/*.spec.tsx,src/**/*.stories.tsx",
```
dcwither, line 25: can you try `spec.*` and `stories.*` since we want to account for `spec.ts` as well. I'm pretty sure I've tried full regex or blob though and it didn't work, so it won't take `src/**/*.(spec|stories).(tsx?|jsx?)` or `src/**/*.{spec,stories}.{js,jsx,ts,tsx}` IIRC.

packages/components/package.json
```diff
@@ -20,10 +20,11 @@
     "url": "git+https://github.com/chanzuckerberg/lp-design-system.git"
   },
   "scripts": {
-    "build": "npm run build:clean && npm run build:ts -- --sourceMap false",
+    "build": "npm run build:clean && npm run build:js",
```
anniehu4, line 23: should this step run `build:types` also?

packages/components/package.json
dcwither: locally, or in CI? This should be fixed by a `npx lerna bootstrap`

packages/components/package.json
anniehu4: hmm I'm trying
```
--ignore src/**/*.spec.*,src/**/*.stories.*
```
but that's not ignoring either type of file :/ 

just including one path `--ignore src/**/*.spec.*` works though

packages/components/.storybook/preview.tsx
anniehu4: my prettier did this on save, but if we hate it can add a rule to `.prettierrc`? it looks like the default changed from avoid -> always on the last update

packages/components/package.json
anniehu4: hm I saw it on CI but can try doing those and pushing up again

anniehu4: ah wild, running that locally fixed it on CI also 👍

packages/components/package.json
anniehu4: I can get around it by including `--ignore` twice 😬

anniehu4:
@ahuth just noticed you had opened https://github.com/chanzuckerberg/lp-design-system/pull/203 earlier -- lmk if there's anything you had that you want to see added here!

packages/components/.storybook/preview.tsx
ahuth: I like it 😄 . But it is a change, and either way is fine.

packages/components/package.json
```diff
@@ -71,7 +76,7 @@
     "sass-loader": "^8.0.2",
     "ts-jest": "^24.1.0",
```
ahuth, line 77: We may want to use `babel-jest` for tests, so that the same compilation happens in code as happens in builds.

packages/components/.babelrc
```diff
@@ -0,0 +1,13 @@
+{
+  "presets": ["@babel/preset-typescript"],
+  "plugins": [
+    "@babel/plugin-transform-modules-commonjs",
```
ahuth, line 4: Is JSX still being transformed? I would expect to need to use either an explicit transform for that, or babel/preset-env. But let me know if is still somehow being transformed 🤔 

We may want to use babel/preset-env anyway, to ensure that we're being explicit about what environments we support (and transforming all syntaxes that we need to for those environments).

ahuth (commented):
Nice, looks pretty good

packages/components/.storybook/preview.tsx
dcwither: It's not a problem, I'm just surprised it wasn't failing prettier before. Probably because it's in .storybook instead of `src/` I'll look into that.

packages/components/package.json
dcwither: I'll play around with it. Thanks testing it out 👍

packages/components/.babelrc
dcwither: Yeah, I was pairing with Annie, but was unable to see the git ignored files so I couldn't confirm. Since I'm taking over while Annie is out sick, I'll look into it.

packages/components/.storybook/preview.tsx
ahuth: Ah, that would make sense

ahuth (approved):
LGTM

PR #221, opened 2020-12-09
feat(components): add twin.macro for styling

anniehu4:
### Summary:
Builds on #224 

Adds twin.macro and updates the playground example. The tw styles are copied from `button.scss` so they _should_ look the same

### Test Plan:
**NOTE**: both Jest tests & storybook are currently broken 😬 

Jest error:
```
FAIL src/playground/button.spec.tsx
● Test suite failed to run
     TypeError: /Users/ahu/Documents/lp-design-system/packages/components/src/playground/button.tsx: Cannot destructure property 'end' of 'ex.node.loc' as it is undefined.
       at end (node_modules/@linaria/babel-preset/src/evaluators/templateProcessor.ts:123:17)
           at Array.forEach (<anonymous>)
       at forEach (node_modules/@linaria/babel-preset/src/evaluators/templateProcessor.ts:92:18)
       at process (node_modules/@linaria/babel-preset/src/extract.ts:201:41)
           at Array.forEach (<anonymous>)
       at PluginPass.forEach (node_modules/@linaria/babel-preset/src/extract.ts:201:23)
       at newFn (node_modules/@babel/core/node_modules/@babel/traverse/lib/visitors.js:175:21)
       at NodePath._call (node_modules/@babel/core/node_modules/@babel/traverse/lib/path/context.js:55:20)
       at NodePath.call (node_modules/@babel/core/node_modules/@babel/traverse/lib/path/context.js:42:17)
       at NodePath.visit (node_modules/@babel/core/node_modules/@babel/traverse/lib/path/context.js:92:31)
```

Storybook: same errors as #224

Once that's fixed, will confirm that:
- Percy doesn't show visual changes on the buttons
- `npm start` works
- Jest tests pass

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -74,6 +74,7 @@
     "@types/jest-axe": "^3.2.0",
     "babel-jest": "^26.6.3",
     "babel-loader": "^8.0.6",
+    "babel-plugin-macros": "^3.0.0",
     "babel-plugin-module-resolver": "^4.0.0",
     "jest": "^24.9.0",
     "jest-axe": "^3.2.0",
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -72,6 +72,7 @@
     "@testing-library/react": "^9.3.0",
     "@types/jest": "^24.0.19",
     "@types/jest-axe": "^3.2.0",
+    "autoprefixer": "^10.1.0",
     "babel-jest": "^26.6.3",
     "babel-loader": "^8.0.6",
     "babel-plugin-macros": "^3.0.0",
@@ -79,11 +80,13 @@
     "jest": "^24.9.0",
     "jest-axe": "^3.2.0",
     "node-sass": "^4.14.1",
+    "postcss": "^8.2.0",
     "react": "^16.10.2",
     "react-docgen-typescript-loader": "^3.3.0",
     "react-dom": "^16.10.2",
     "sass-loader": "^8.0.2",
     "ts-loader": "^6.2.0",
+    "twin.macro": "^2.0.5",
     "typescript": "^4.1.2"
   },
   "gitHead": "8680dac1fa3c685e2a35f80d9ceb157f5f5343bd"
--- a/packages/components/src/playground/button.tsx
+++ b/packages/components/src/playground/button.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import { css } from "@linaria/core";
 import { styled } from "@linaria/react";
+import tw from "twin.macro";
 
 // Temporary props for button to test out css options for variants
 type ButtonProps = React.DetailedHTMLProps<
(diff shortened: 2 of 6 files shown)


packages/components/src/playground/button.spec.tsx
```diff
@@ -3,7 +3,7 @@ import { render } from "@testing-library/react";
 import { variants } from "./button.stories";
 
 describe("<Button />", () => {
-  it("renders the component", () => {
+  it.skip("renders the component", () => {
```
anniehu4, line 6: the stories don't compile right now :/ still looking into it

packages/components/package.json
```diff
@@ -65,18 +69,23 @@
...
     "react": "^16.10.2",
     "react-docgen-typescript-loader": "^3.3.0",
     "react-dom": "^16.10.2",
     "sass-loader": "^8.0.2",
     "ts-jest": "^24.1.0",
     "ts-loader": "^6.2.0",
-    "typescript": "^4.1.2"
+    "twin.macro": "^2.0.5",
+    "typescript": "^4.1.2",
+    "vue": "^2.6.12"
```
anniehu4, line 88: I saw warnings that `vue`, `autoprefixer`, and `postcss` were peer dependencies

anniehu4:
These are the Storybook errors I'm seeing:
```
@chanzuckerberg/czedi-kit-components: ERROR in ./node_modules/resolve/lib/async.js
@chanzuckerberg/czedi-kit-components: Module not found: Error: Can't resolve 'fs' in '/Users/ahu/Documents/lp-design-system/packages/components/node_modules/resolve/lib'
```

This thread shows the same errors and warnings: https://github.com/ben-rogerson/twin.macro/issues/125, and one of the answers says
```
Those errors pop up when twin is run through the typescript compiler rather than the babel compiler.
You'll need to replace the typescript loader with babel-loader.
```

It's fixed in storybook@6.0.17, but I'm not sure what the version we're on (5.3.18) does

packages/components/src/playground/button.tsx
```diff
@@ -0,0 +1,35 @@
...
+  `;
+  const primary = css`
+    ${tw`bg-blue-300 hover:bg-blue-400`}
+  `;
+  const secondary = css`
+    ${tw`bg-gray-600 hover:bg-blue-800`}
+  `;
+
+  return (
+    <Button className={variant === "primary" ? primary : secondary} {...rest} />
```
anniehu4, line 27: there are a lot of ways to use Linaria's API -- this is based on their [Theming example](https://github.com/callstack/linaria/blob/master/docs/THEMING.md#css-custom-properties) but isn't exactly the same. I didn't have a strong feel for what's best

packages/components/package.json
dcwither: we definitely shouldn't need vue - I'm guessing this is a weird peer dependency that is meant to define what versions it's compatible with. We should try to exclude it even if there are warnings.

PR #224, opened 2020-12-10
feat(components): add Linaria for styling

anniehu4:
### Summary:
Split off from #221

Adds Linaria and changes the existing `button` component to a playground example. 

### Test Plan:
- Jest tests pass
- confirm that `npm start` loads Storybook


packages/components/.storybook/main.js
```diff
@@ -3,11 +3,27 @@ const path = require("path");
...
-  webpackFinal: config => {
+  webpackFinal: (config) => {
     config.module.rules.push({
       test: /\.scss$/,
-      use: ["style-loader", "css-loader", "sass-loader"],
       include: path.resolve(__dirname, "../"),
+      use: ["style-loader", "css-loader", "sass-loader"],
+    });
+
+    config.module.rules.push({
```
anniehu4, line 13: Follows the [Linaria guide for webpack setup](https://github.com/callstack/linaria/blob/master/docs/BUNDLERS_INTEGRATION.md#webpack)

Something isn't compiling correctly though :( I get a bunch of TS errors from `button.tsx` that don't make any sense

packages/components/src/playground/button.tsx
```diff
@@ -0,0 +1,35 @@
...
+function Button(props: ButtonProps): JSX.Element {
+  const { variant = "primary", ...rest } = props;
+  const primary = css`
+    --color-background: blue;
+  `;
+  const secondary = css`
+    --color-background: gray;
+  `;
+
+  const ButtonComponent = styled.button`
```
anniehu4, line 22: this follows the Linaria theming example closely, but i'm not sure it's appropriate for button variants. fine for testing at least

dcwither:
@anniehu4 ok to close for now?

Issue #212, opened 2020-12-03
Switch to CSS-in-JS for Styling Purposes


dcwither:
# ~Linaria~ Styled Components

~Linaria allows us to use CSS-in-JS while maintaining zero runtime by adding a compilation step to the application to pull out all the styles into CSS files for runtime usage.~

Styled components provides CSS-in-JS. While it has a runtime, it will allow us to use twin.macro in the design system, and in apps making our CSS more consistent.

### Automatic Bundling

CSS-in-JS makes bundling the design system according the an applications existing bundles easier. It has no need for importing split out CSS in the exact right place since it's already colocated with the JS.

### Customization/Theming

Working with ~Linaria~ Styled Components would allow us to integrate theming at the compilation phase using an import of styles, as well as at runtime using CSS Variables https://levelup.gitconnected.com/tailwindcss-with-css-variables-513abe2e9a5.

We could accomplish theming with tailwind using CSS variables as well, but Linaria gives us a framework to provide a single point for theming as well as some alternatives if we need to expand beyond CSS variables for theming.

Other options for theming with Linaria are outlined in the [Linaria Docs](https://github.com/callstack/linaria/blob/master/docs/THEMING.md#theming).

#### Along - Styled Components

Linaria and styled components mostly have similar APIs, so migrating styled-components to Linaria wouldn't be difficult. We would have to make some transitions from themes to CSS variables for greater consistency.

#### Traject - Global SCSS

We've planned to move away from Traject's global scss for some time (that's part of what the design system is for), so adopting Linaria + Tailwind for future styling would actually be a good option - we may even be able to export the configuration in place of the current `styles` project.

### Grepability of Classes

Using Linaria will remove the grepability of css classes since it uses a hashed classname and even minifies them. This is mitigated by using _react dev tools_ to inspect the components to find the origin of the styles as easily as we did before, and even ensure that we can easily track the full styles applied to these components. We can improve this further by adding explicit component names so that it survives compilation.

### Requires additional build step in dependent projects

Because we want the css+js bundling that comes with css-in-js, we would need to provide guidance eon how to build compile out the linaria setup in the app instead of on package publishing.

### No Support for CSS Only

Linaria exports minified and hashed classes. While this makes the CSS more performant, it removes the ability to style components using the CSS only options. Currently we are prioritizing the ability to theme, and the bundling benefits over this concern.

## Browser Support

[...]

PR #226, opened 2020-12-11
proposal: add styled-components and button example

anniehu4:
### Summary:
Overlaps a lot with #224, but uses styled-components instead (see #212 for motivation). Setting this up was wayyyy easier 😭 

#227 integrates twin.macro ~, haven't added reviewers yet b/c there are some CI errors~

### Test Plan:
- tests pass
- `npm start` loads Storybook
- Percy shows changes because of filename changes, ~but the buttons should match the old Tailwind styles~ and button colors now use EDS instead of legacy colors

**Note:** Bundle size script failure is expected when running on `master` because Typescript is not properly ignoring `node_modules` there


.stylelintrc
```diff
@@ -1,5 +1,8 @@
 {
-  "extends": "stylelint-config-standard",
+  "extends": [
+    "stylelint-config-standard",
+    "stylelint-config-styled-components"
```
anniehu4, line 4: needed to ignore rules like `no-missing-end-of-source-newline` in JS

packages/components/.storybook/preview.tsx
```diff
@@ -10,7 +10,7 @@ import { addParameters } from "@storybook/react";
 import { withA11y } from "@storybook/addon-a11y";
 
 addDecorator(withA11y);
-addDecorator((storyFn) => <div dir="ltr">{storyFn()}</div>);
+addDecorator((storyFn) => <div>{storyFn()}</div>);
```
anniehu4, line 13: removing this fixes the style overriding on buttons, idk if we need it for something else though

packages/components/src/playground/button.tsx
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import styled from "styled-components";
+
+// Temporary props for button to test out css options for variants
+type ButtonProps = React.DetailedHTMLProps<
+  React.ButtonHTMLAttributes<HTMLButtonElement>,
+  HTMLButtonElement
+> & {
+  ref?: React.Ref<HTMLButtonElement>;
```
anniehu4, line 9: Adding this to resolve a TS error where there were two conflicting declarations of `ref`?

packages/components/tsconfig.json
```diff
@@ -1,25 +1,26 @@
...
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "noImplicitAny": false,
     "noImplicitReturns": true,
     "noImplicitThis": true,
-    "noImplicitAny": false,
+    "noUnusedLocals": true,
+    "outDir": "dist/",
+    "sourceMap": true,
+    "skipLibCheck": true,
```
anniehu4, line 20: alphabetized + added this line to actually skip `node_modules` (https://github.com/Microsoft/TypeScript/issues/14293)

packages/components/.storybook/preview.tsx
ahuth: Removing it seems perfectly reasonable. My guess is storybook will apply this to a wrapping div on its own based on UI options.

packages/components/tsconfig.json
```diff
@@ -1,25 +1,26 @@
...
-    "lib": ["es5", "es6", "dom"],
-    "sourceMap": true,
     "allowJs": false,
-    "jsx": "react",
-    "moduleResolution": "node",
+    "allowSyntheticDefaultImports": true,
     "baseUrl": "src",
+    "declaration": true,
+    "emitDecoratorMetadata": true,
+    "experimentalDecorators": true,
```
ahuth, line 8: I'd remove this unless we're using decorators.

As far as I can tell, decorators is in [stage 2 of TC39](https://tc39.es/proposal-decorators/), which means it's not yet accepted as part of standard JS.

Having said that, if we think we're going to use it, then it's fine.

packages/components/tsconfig.json
```diff
@@ -1,25 +1,26 @@
...
     "forceConsistentCasingInFileNames": true,
+    "jsx": "react",
+    "lib": ["es5", "es6", "dom"],
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "noImplicitAny": false,
```
ahuth, line 14: I know you just alphabetized, but I'd recommend

- Turn on `noImplicitAny`
- Remove all these `noImplicitFoo` config options, and set `strict: true` (which turns them all on)

packages/components/src/playground/button.tsx
ahuth: This is pretty interesting. I think you're right: the `ref` type coming from `React.DetailedHTMLProps` seems to be conflicting with what we're getting from `styled.button`

Specifically, it seems like:
- `React.DetailedHTMLProps` is expecting "legacy" refs (object, callback, and the deprecated string ref)
- `styled.button` is expecting non-legacy refs (object and callback refs only)

(Verified by clicking into each definition. There are many layers to go through, but they're there)

This raises an interesting issue, though: we can't use `ref` as a prop like this 😬 .

A `ref` is handled separately from normal props (kind of hard to see, but [note here that "reserved" props [which ref is, along with key] are not copied over to the props object](https://github.com/facebook/react/blob/daf38ecdfcda32a1050231fcc6dea364f8b102e3/packages/react/src/ReactElement.js#L375-L380)).

That's why normally we see something like:
- Use a name other than `ref`, such as `buttonRef`
- Use `React.forwardRef`, which adds `ref` to custom components

This case doesn't seem like we quite need `ref` exposed, yet, and I'd recommend dropping the `DetailedHTMLProps`:

```ts
type Props = React.ButtonHTMLAttributes<HTMLButtonElement> & {
  variant?: "primary" | "secondary";
}
```

Let me know if that makes any sense at all. Also happy to talk this through in person, if that's more helpful.

---

Unrelated, but this also highlighted that I think we should have an explicit dependency on `@types/react` somewhere. This works because other dependencies are installing it, but (if we haven't yet), we'll want our own dependency on it.

ahuth: Actually, the example I linked to in the React source may not be a good example of why we can't use `ref` in this way.

A better reference is probably https://reactjs.org/docs/forwarding-refs.html, which says

> Regular function or class components don’t receive the ref argument, and ref is not available in props either.

anniehu4: 🤯 wow, thanks for digging into that! I agree we don't need `ref` here, since it's just a test component. When we build the actual Button, might need to revisit. And I'll add `@types/react` 👍

packages/components/src/playground/button.tsx
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+import styled from "styled-components";
+
+// Temporary props for button to test out css options for variants
+type ButtonProps = React.ButtonHTMLAttributes<HTMLButtonElement> & {
+  variant?: "primary" | "secondary";
+};
+
+const ButtonComponent = styled.button<ButtonProps>`
+  border: none;
```
anniehu4, line 10: hard-coding these styles to match `button.scss`

ahuth (approved):

packages/components/.storybook/preview.tsx
dcwither: I think to fix the overrides we can remove the all.css imports, which are no longer necessary

dcwither: ~We will want to import `<GlobalStyles />` from twin.macro though~, oh, this is pre-twin.macro right?

In that case, you can remove the rtl step at https://github.com/chanzuckerberg/lp-design-system/blob/master/packages/styles/postcss.config.js#L16

packages/components/src/playground/button.spec.tsx
```diff
@@ -8,12 +8,12 @@ describe("<Button />", () => {
     expect(container.firstChild).toMatchInlineSnapshot(`
       <div>
         <button
-          class="czedi-btn czedi-btn-primary"
+          class="sc-bdfBwQ sc-gsTCUz fvVoTP bMZeDH"
```
dcwither, line 11: to get unchanging output here, we can use https://github.com/chanzuckerberg/along/blob/cc2a3feac8ab7464014126bf3b1301a88b327d47/client/src/test/removeStylesJestSerializer.ts, though we can definitely wait until another PR for that.

packages/components/tsconfig.json
dcwither: Nice!

packages/components/.storybook/preview.tsx
anniehu4: oh nice, thanks!

packages/components/src/playground/button.spec.tsx
anniehu4: ooh cool, yeah I want to show the className change once we add twin.macro but this is good to consider after that

PR #227, opened 2020-12-11
feat(components): add twin.macro for styling

anniehu4:
### Summary:
Builds on #226 

Adds `twin.macro` so we can style components with Tailwind! Also removes the `styles/` package completely -- adds `<GlobalStyles />` instead to use Tailwind base styles.

### Test Plan:
- tests pass
- `npm start` loads Storybook
- Percy shows no visual changes


packages/components/package.json
```diff
@@ -41,8 +41,9 @@
     "react": ">= 16.8.0"
   },
   "dependencies": {
-    "@chanzuckerberg/czedi-kit-styles": "^0.0.1-alpha.5",
     "@chanzuckerberg/czedi-kit-tokens": "^0.0.1-alpha.5",
+    "autoprefixer": "^10.1.0",
+    "postcss": "^8.2.1",
```
anniehu4, line 46: peer dependencies

packages/components/.storybook/preview.tsx
```diff
@@ -1,16 +1,19 @@
...
 
 import { addDecorator } from "@storybook/react";
 import { addParameters } from "@storybook/react";
 import { withA11y } from "@storybook/addon-a11y";
 
 addDecorator(withA11y);
-addDecorator((storyFn) => <div dir="ltr">{storyFn()}</div>);
+addDecorator((storyFn) => (
+  <div dir="ltr">
+    <GlobalStyles />
```
anniehu4, line 13: hm just thinking ahead -- will we need to export components with this attached? or import this in traject?

packages/components/src/twin.d.ts
```diff
@@ -0,0 +1,28 @@
+import "twin.macro";
```
anniehu4, line 1: followed https://github.com/ben-rogerson/twin.examples/tree/master/react-styled-components#complete-the-typescript-setup

packages/components/.storybook/preview.tsx
ahuth: We'll want to re-export, so that traject (and along) can do something like:

```js
// In Traject...

import { GlobalStyleShenanigansThing } from '@chanzuckerberg/eds';
```

packages/components/package.json
```diff
@@ -22,7 +22,7 @@
   "scripts": {
     "build": "npm run build:clean && npm run build:types && npm run build:js",
     "build:clean": "rm -rf dist/",
-    "build:js": "babel src -d dist --extensions '.js,.jsx,.ts,.tsx' --ignore src/**/*.spec.* --ignore src/**/*.stories.*",
+    "build:js": "babel src -d dist --extensions '.js,.jsx,.ts,.tsx' --ignore src/**/*.spec.* --ignore src/**/*.stories.* --ignore src/**/*.d.ts",
```
ahuth, line 25: Were we running into issues with babel processing .d.ts files?

I ask because it seems (without understanding deeply the build system here :smile:) slightly odd, since I would expect Babel to produce these files in the first place. Is it an issue because they're left around from subsequent builds?

Or am I misunderstanding everything?

packages/components/src/twin.d.ts
ahuth: 🤔 Interesting. Would be nice if tw shipped with type definitions for its different integrations.

packages/components/src/twin.d.ts
```diff
@@ -0,0 +1,28 @@
...
+
+declare module "react" {
+  // The css prop
+  interface HTMLAttributes<T> extends DOMAttributes<T> {
+    css?: CSSProp;
+  }
+  // The inline svg css prop
+  interface SVGProps<T> extends SVGProps<SVGSVGElement> {
+    css?: CSSProp;
+  }
```
ahuth, line 18: Dang, adding a `css` attribute to all JSX elements is a bummer.

Any chance we can plan on not using those, and leave these types out? Or does that restrict the API too much?

anniehu4: yeah after filling out the Type component, I feel pretty sure that we don't need it. will leave `styled` and `as` though

packages/components/package.json
anniehu4: it's only because I added the `twin.d.ts` file 😞 I might remove from this one and add it to the Typography PR where I'm actually using `styled`

anniehu4: oh lol I put it in the wrong spot 🤦‍♀️ either way, moving it out of this one

.github/dependabot.yml
```diff
@@ -1,34 +1,26 @@
...
-    include: scope
-- package-ecosystem: npm
-  directory: "/packages/components"
-  schedule:
-    interval: daily
-  open-pull-requests-limit: 10
-  commit-message:
-    prefix: chore
-    include: scope
+  - package-ecosystem: npm
```
dcwither, line 3: I'm trying to figure out what changed here - tabs to spaces? At the very least, it's a whitespace only change

packages/components/.storybook/main.js
```diff
@@ -15,11 +15,6 @@ module.exports = {
       use: ["style-loader", "css-loader", "sass-loader"],
     });
 
-    config.watchOptions = {
```
dcwither, line 18: 👍

packages/components/package.json
dcwither: Is this because postcss is part of twin-macro, or styled-components? I'm mainly concerned about this becoming part of the included dependencies, makes more sense as a dev dependency if anything

packages/components/src/playground/button.tsx
```diff
@@ -1,33 +1,20 @@
...
+  font-bold py-2 px-4 rounded text-white border-none
 `;
 
-const PrimaryButton = styled(ButtonComponent)`
-  background-color: #574eff;
-
-  &:hover {
-    background-color: #433bde;
-  }
+const PrimaryButton = tw(ButtonComponent)`
```
dcwither, line 13: Not necessary for this PR, but I actually liked the way you did variants for linaria with the one `tw.button` and the others were variant classes that were appended

packages/components/src/playground/button.tsx
```diff
@@ -1,33 +1,20 @@
...
 };
 
-const ButtonComponent = styled.button<ButtonProps>`
-  border: none;
-  border-radius: 0.25rem;
-  color: white;
-  font-weight: bold;
-  padding: 0.5rem 1rem;
+const ButtonComponent = tw.button`
+  font-bold py-2 px-4 rounded text-white border-none
```
dcwither, line 10: Is stylelint already testing these? The way I checked was to run `npm run lint:styles -- -f verbose` https://stylelint.io/user-guide/usage/cli#example-i---print-on-success

packages/components/src/playground/button.spec.tsx
```diff
@@ -8,12 +8,12 @@ describe("<Button />", () => {
     expect(container.firstChild).toMatchInlineSnapshot(`
       <div>
         <button
-          class="sc-bdfBwQ sc-gsTCUz fvVoTP enkIjw"
+          class="button__ButtonComponent-ak17f2-0 button__PrimaryButton-ak17f2-1 dehHou jntNfS"
```
dcwither, line 11: I'll add the PR that either removes these, or makes them consistent

.github/dependabot.yml
anniehu4: yeah that, and I removed the block with `"/packages/styles"`

packages/components/package.json
anniehu4: they're both from `tailwindcss` -- you're totally right they should be dev dependencies though

packages/components/src/playground/button.tsx
anniehu4: yep, I see them in there! thanks for checking 👍

packages/components/src/playground/button.tsx
anniehu4: ooh ok, I did like that too. the `className` api might be a linaria-only thing but I can play a little

anniehu4: updated, and I think it's now closer to what I was doing with linaria? lmk if I interpreted wrong

dcwither (approved):

PR #231, opened 2020-12-15
feat: export global styles and fonts

anniehu4:
### Summary:
https://github.com/chanzuckerberg/lp-design-system/pull/227/files#r543564195

We'll need to export some global styles for `traject` to use, including Preflight (from tailwind) and Fonts (while we're currently sticking with Arimo, this may change in the future)

### Test Plan:
- tests pass
- `npm start` runs Storybook, confirm text is now in Arimo
- only Percy changes (against parent PR) should be from font family change


package.json
```diff
@@ -7,7 +7,7 @@
     "build": "NODE_ENV=production lerna run build",
     "lint": "npm run lint:styles && npm run lint:scripts",
     "lint:fix": "npm run lint:styles:fix && npm run lint:scripts:fix",
-    "lint:styles": "stylelint --ignore-path .gitignore packages",
+    "lint:styles": "stylelint --ignore-path .gitignore packages/components/**/*.{css,scss,js,jsx,ts,tsx}",
```
anniehu4, line 10: instead of continuing to add to `ignoreFiles`, figured this was more predictable

packages/components/src/styles/global.js
```diff
@@ -0,0 +1,32 @@
+import { GlobalStyles } from "twin.macro";
+import React from "react";
+import { createGlobalStyle } from "styled-components";
+
+const GlobalFonts = createGlobalStyle`
```
anniehu4, line 5: I was inspired by https://dev.to/alaskaa/how-to-import-a-web-font-into-your-react-app-with-styled-components-4-1dni

Please let me know if this is a bad idea!

package.json
dcwither: I think we should do `packages/**/*.{css,scss,js,jsx,ts,tsx}` in case we add other bundles - though this may not play nice with the new addition of `dist` into the project. We may have to explicitly exclude that one.

packages/components/src/styles/global.js
dcwither: I don't think it's a bad idea, but thinking about it more, we may want to add other ways to import it since these are gonna stay basically forever. Either that, or to include the css variables from tokens into it somehow.

dcwither (approved):

packages/components/src/styles/global.js
anniehu4: ah yeah exporting the css variables would be good

PR #232, opened 2020-12-15
Dcwither/styled css variables

dcwither:
### Summary:
Create a new format for style dictionary config, this allows us to pull in css variables into tailwind for easy theming and to get rid of the pesky `--tw-bg-opacity` and `--tw-text-opacity`

Some bonus tasks done in this PR:
- CI uses `npm ci --no-optional` to resolve node-gyp issues
- No more tokens file headers, this prevents unnecessary compilation annoyance
- rename `config.js` to `style-dictionary.config.js` and reference it properly for more happiness

### Test Plan:
Check the tokens build, `npm start`, CI

### DIFF
--- a/packages/components/.storybook/preview.tsx
+++ b/packages/components/.storybook/preview.tsx
@@ -1,8 +1,10 @@
+import "@chanzuckerberg/czedi-kit-tokens/css/variables.css";
+
 import * as React from "react";
 
 import { DocsContainer, DocsPage } from "@storybook/addon-docs/blocks";
-import EDSGlobalStyles from "../src/styles/global";
 
+import EDSGlobalStyles from "../src/styles/global";
 import { addDecorator } from "@storybook/react";
 import { addParameters } from "@storybook/react";
 import { withA11y } from "@storybook/addon-a11y";
--- a/packages/components/tailwind.config.js
+++ b/packages/components/tailwind.config.js
@@ -1,5 +1,4 @@
 const tokens = require("@chanzuckerberg/czedi-kit-tokens/json/variables-nested.json");
-
 module.exports = {
   theme: {
     colors: tokens.eds.color,
@@ -8,6 +7,7 @@ module.exports = {
     },
     fontSize: tokens.eds.size.font,
     lineHeight: tokens.eds.size["line-height"],
+    extend: {},
   },
   variants: {},
   plugins: [],
--- a/packages/components/tailwind.config.js
+++ b/packages/components/tailwind.config.js
@@ -1,4 +1,4 @@
-const tokens = require("@chanzuckerberg/czedi-kit-tokens/json/variables-nested.json");
+const tokens = require("@chanzuckerberg/czedi-kit-tokens/json/css-variables-nested.json");
 module.exports = {
   theme: {
     colors: tokens.eds.color,
--- a/packages/tokens/config.js
+++ b/packages/tokens/config.js
@@ -4,31 +4,35 @@ const EDSStyleDictionary = StyleDictionary.extend({
   platforms: {
     scss: {
       transforms: [...StyleDictionary.transformGroup.scss, "name/kebabCase"],
-      buildPath: "dist/scss/",
+      buildPath: "dist/",
       files: [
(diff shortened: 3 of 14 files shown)


packages/components/.storybook/preview.tsx
```diff
@@ -1,8 +1,10 @@
+import "@chanzuckerberg/czedi-kit-tokens/css/variables.css";
+
 import * as React from "react";
 
 import { DocsContainer, DocsPage } from "@storybook/addon-docs/blocks";
-import EDSGlobalStyles from "../src/styles/global";
 
+import EDSGlobalStyles from "../src/styles/global";
```
dcwither, line 7: I think this is probably a result of my auto formatting, I'll add a rule to our eslint so it applies for all of us.

packages/components/tailwind.config.js
```diff
@@ -8,6 +7,7 @@ module.exports = {
     },
     fontSize: tokens.eds.size.font,
     lineHeight: tokens.eds.size["line-height"],
+    extend: {},
```
dcwither, line 10: I think this was merge conflicts

packages/tokens/config.js
```diff
@@ -4,31 +4,35 @@ const EDSStyleDictionary = StyleDictionary.extend({
   platforms: {
     scss: {
       transforms: [...StyleDictionary.transformGroup.scss, "name/kebabCase"],
-      buildPath: "dist/scss/",
+      buildPath: "dist/",
       files: [
         {
-          destination: "_variables.scss",
+          destination: "scss/_variables.scss",
```
dcwither, line 10: So now the platforms build to arbitrary sub folders because the commonality between them is the transforms rather than the files

packages/tokens/config.js
```diff
@@ -62,4 +66,29 @@ EDSStyleDictionary.registerTransform({
...
+// copied from https://github.com/amzn/style-dictionary/blob/v3.0.0-rc.1/lib/common/formats.js#L96
+function minifyDictionary(obj) {
+  if (typeof obj !== "object" || Array.isArray(obj)) {
+    return obj;
+  }
+
+  var toRet = {};
+
+  if (obj.value) {
+    return `var(--${obj.name})`;
```
dcwither, line 78: this is the custom portion

packages/components/.storybook/preview.tsx
```diff
@@ -1,8 +1,10 @@
+import "@chanzuckerberg/czedi-kit-tokens/css/variables.css";
```
anniehu4, line 1: what is this doing? or is it setting up for next changes?

packages/tokens/config.js
```diff
@@ -62,4 +66,29 @@ EDSStyleDictionary.registerTransform({
   },
 });
 
+// copied from https://github.com/amzn/style-dictionary/blob/v3.0.0-rc.1/lib/common/formats.js#L96
+function minifyDictionary(obj) {
```
anniehu4, line 70: maybe name `minifyCSSDictionary` to distinguish?

anniehu4 (approved):

packages/tokens/config.js
ahuth: 🤔 

I'm not that familiar with how these tokens are built/transformed. Does this mean that we should expect more sub-directories in dist?

packages/components/.storybook/preview.tsx
dcwither: This adds the css variables to the root - in the long run we would want this tied in with the EDS Global Styles maybe

packages/tokens/config.js
dcwither: I'll go with `minifyDictionaryToCSSVars`

dcwither:
cc @anniehu4 on the additional changes

> Some bonus tasks done in this PR:
> 
> - CI uses npm ci --no-optional to resolve node-gyp issues
> - No more tokens file headers, this prevents unnecessary compilation annoyance
> - rename config.js to style-dictionary.config.js and reference it properly for more happiness

PR #243, opened 2020-12-16
Update package-lock

ahuth:
After running `npm install` from the package root, the package-lock.json is being updated with these changes.

Anyone else seeing that?


anniehu4:
weird, I don't see that 😮  isn't `@chanzuckerberg/czedi-kit-tokens` a dependency, and should be in the package-lock? But I guess the build passed so it shouldn't be??

ahuth:
Weird. Not sure if czedi-kit-tokens should be in the package-lock or not... 🤔 Maybe?

@anniehu4 can you confirm which versions of node and npm you're on when running `npm install`? If our versions match, I'll close this, and chalk it up to "dependency management is hard".

My versions are:
- Node: 14.5.0
- Npm: 6.14.5

anniehu4 (approved):
oooh ok after I removed `node_modules` and re-installed, I see this too. probably related to removing the `styles/` dependency. everything still works locally 🚢

ahuth:
Super weird. Thanks!

PR #244, opened 2020-12-16
Storybook 6 updates

ahuth:
This PR updates the stories for Storybook 6's component story format.

Before:

storybook old

After:

storybook new

Note that there's now a controls section! Unfortunately, actions are no longer working correctly (I think because we don't define `onClick` explicitly as a prop... maybe...). We can look into that some more once we get real components going.


packages/components/.storybook/main.js
```diff
@@ -3,10 +3,9 @@ const path = require("path");
 module.exports = {
   stories: ["../src/*.stories.mdx", "../src/**/*.stories.@(ts|tsx|mdx)"],
   addons: [
-    "@storybook/addon-actions",
     "@storybook/addon-links",
+    "@storybook/addon-essentials",
```
ahuth, line 7: addon-essentials includes actions, docs, controls, and some more.

packages/components/.storybook/preview.tsx
```diff
@@ -1,22 +1,11 @@
 import * as React from "react";
-
-import { DocsContainer, DocsPage } from "@storybook/addon-docs/blocks";
 import { GlobalStyles } from "twin.macro";
 
-import { addDecorator } from "@storybook/react";
-import { addParameters } from "@storybook/react";
-import { withA11y } from "@storybook/addon-a11y";
```
ahuth, line 8: No need to configure the docs or a11y addons here (as of v6)

packages/components/src/playground/button.spec.tsx
```diff
@@ -1,28 +1,33 @@
+import { primary, secondary } from "./button.stories";
```
ahuth, line 1: The import ordering we're enforcing here seems weird 😕

packages/components/src/playground/button.spec.tsx
```diff
@@ -1,28 +1,33 @@
...
+    expect(container.firstChild).toMatchInlineSnapshot(`
+      <button
+        class="button__ButtonComponent-ak17f2-0 VxUNz"
+      >
+        Secondary Button
+      </button>
     `);
   });
 
   it("passes accessibility checks", async () => {
```
ahuth, line 29: The secondary button doesn't actually appear to pass axe checks. But somehow this test still passes when I change it to secondary, instead of primary 😬

packages/components/src/playground/button.stories.tsx
```diff
@@ -1,30 +1,35 @@
 import * as React from "react";
-
 import Button from "./button";
-import { action } from "@storybook/addon-actions";
```
ahuth, line 4: Component story format is nice because we no longer need to import any storybook addons. Not importing storybook stuff is important for things like `getSnapshots` and `prepareStory` to work, because non-storybook environments such as Jest won't necessarily have these addons available.

packages/components/src/playground/button.stories.tsx
```diff
@@ -1,30 +1,35 @@
...
+import { Story } from "@storybook/react/types-6-0";
```
ahuth, line 3: Normally could write

```ts
import type { Story } from ...
```

But our eslint parsing doesn't seem to be able handle that 🤔

packages/components/src/playground/button.stories.tsx
```diff
@@ -1,30 +1,35 @@
...
 
-export const text = (): JSX.Element => (
-  <Button onClick={action("clicked")}>Hello Button</Button>
-);
+const Template: Story<Args> = (args) => <Button {...args} />;
+
+export const text = Template.bind(null);
+text.args = {
+  children: "Hello Button",
+};
```
ahuth, line 17: A downside I've noticed to component story format is we lose type safety. There's nothing here enforcing that we've actually provided all necessary args.

Haven't come up with a good solution for that, yet.

packages/components/src/playground/button.spec.tsx
anniehu4: hm yeah we're just using default https://eslint.org/docs/rules/sort-imports ... maybe should use the same rule you added to traject?

packages/components/src/playground/button.spec.tsx
anniehu4: eek o.o is this where we need axe-storybook

anniehu4 (commented):
awesome! 🎉  everything looks good but I did just land a change before seeing this that added a Typography story ... will that need to be refactored before landing? if not I could do that refactor instead

packages/components/src/playground/button.spec.tsx
ahuth: Oh, ha, I didn't realize that eslint had a built-in rule for this.

Maybe we should use the eslint-plugin-import/order rule, and be consistent between traject, frontend-libs, axe-storybook-testing, and this. (And then of course add it to Along, too).

ahuth:
Thanks!

> I did just land a change before seeing this that added a Typography story ... will that need to be refactored before landing? if not I could do that refactor instead

There are other conflicts anyway, so I'll update here.

packages/components/src/playground/button.spec.tsx
dcwither: yeah, that one has the `fix` I was planning on getting to that today, so no need to handle it yourself

packages/components/src/playground/button.spec.tsx
dcwither: Yeah, I think it's not high enough contrast, but it's not the real button anyway 😄

packages/components/src/playground/button.stories.tsx
dcwither: can we do `text.args: ButtonProps = ...`?

packages/components/src/playground/button.spec.tsx
```diff
@@ -1,38 +1,43 @@
...
 import { axe } from "jest-axe";
+import { prepareStory } from "@chanzuckerberg/story-utils";
 import { render } from "@testing-library/react";
-import { variants } from "./button.stories";
 
 describe("<Button />", () => {
-  it("renders the component", () => {
-    const { container } = render(variants());
+  it("renders primary buttons", () => {
+    const { container } = render(prepareStory(primary));
```
dcwither, line 8: probably because it's not published yet, but this could be reduced down to the `generateSnapshots`

dcwither (approved):

packages/components/src/playground/button.spec.tsx
lesliecodes: oh it's ready now! I can update it for lp-design-system

packages/components/src/playground/button.stories.tsx
ahuth: Unfortunately, that doesn't work

ahuth: We may have to pass the args through a function or something to get type safety

```ts
text.args = assignArgs<Args>({
  children: "Hello",
});
```

Or maybe pass the whole thing through a function

```ts
export const text = makeStory(template, {
  children: "Hello Button",
});
```

packages/components/src/playground/button.spec.tsx
ahuth: Cool 👍 . I'm going to leave that for a follow up, because I'm lazy 😸

packages/components/src/typography/typography.spec.tsx
```diff
@@ -8,25 +8,25 @@ describe("<Typography />", () => {
     expect(container.firstChild).toMatchInlineSnapshot(`
       <div>
         <p
-          class="typography__TypographyComponent-sc-8fjfkf-0 gwgBif"
+          class="typography__TypographyComponent-sc-8fjfkf-0 dJNOnb"
```
ahuth, line 11: Not sure why the classname would've changed

ahuth:
Damn, these snapshots...

packages/components/src/playground/button.stories.tsx
dcwither: thanks for flagging

dcwither (approved):
Thanks for updating the stories 🔥

ahuth:
Closing.

This is out of date, and I think we'll want to start over if we want to do this.

PR #245, opened 2020-12-16
Revert "Update package-lock"

ahuth:
Reverts chanzuckerberg/lp-design-system#243

Actually, now I think this is supposed to be present. But it's weird because:

- Sometimes running `npm install` from the top-level removes it...
- Sometimes running `npx lerna bootstrap` adds it back...

But I think your initial thought was right, @anniehu4 - since it's in the package.json for components/, it should be in the package-lock. Probably.

@dcwither do you have any insights as to what the heck is going on here?


anniehu4 (approved):
approving but clearly i dunno what's going on haha

ahuth:
Okay, nvm. I think we do still want these changes.

Turns out I was only seeing an issue when running `npm install` from the components package, not on lerna's bootstrap command. But bootstrap is the way to go, not an install from individual packages.

PR #247, opened 2020-12-17
test(snapshots): remove classes from snapshots

dcwither:
### Summary:
Pulled `https://github.com/chanzuckerberg/along/blob/main/client/src/test/removeStylesJestSerializer.ts` from along - we should consider moving this into https://github.com/chanzuckerberg/frontend-libs

### Test Plan:
`npm test`

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -35,7 +35,7 @@
       "extends": ["plugin:@typescript-eslint/recommended"]
     },
     {
-      "files": ["*.spec.ts", "*.spec.tsx"],
+      "files": ["*.spec.ts", "*.spec.tsx", "**/test/**"],
       "plugins": ["jest"],
       "extends": ["plugin:jest/recommended"],
       "env": {
--- a/packages/components/jest.config.js
+++ b/packages/components/jest.config.js
@@ -3,7 +3,7 @@
 
 module.exports = {
   clearMocks: true,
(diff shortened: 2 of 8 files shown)


dcwither:
https://github.com/chanzuckerberg/lp-design-system/pull/247/checks?check_run_id=1573354928#step:4:29 Noooo!

I think this has to do with changing dependencies and busting the dependencies cache

.eslintrc
```diff
@@ -35,7 +35,7 @@
       "extends": ["plugin:@typescript-eslint/recommended"]
     },
     {
-      "files": ["*.spec.ts", "*.spec.tsx"],
+      "files": ["*.spec.ts", "*.spec.tsx", "**/test/**"],
```
dcwither, line 38: test folder is supposed to be only for test utils

ahuth (approved):
Nice, LGTM

ahuth:
> we should consider moving this into https://github.com/chanzuckerberg/frontend-libs

Makes sense. I would like to look into getting rid of the jest-styled-components dependency when we do that.

anniehu4:
woot! task for tracking the followup: https://app.clubhouse.io/connections/story/62274/move-removestylesjestserializer-to-https-github-com-chanzuckerberg-frontend-libs

PR #248, opened 2020-12-18
build(scss): remove node-sass and any use of scss in components

dcwither:
### Summary:
No more scss, which most importantly means no more node-sass!

### Test Plan:
`npm start` + CI. If someone else can pull down to test it out (you may need to clear your storybook cache with `rm -rf packages/components/node_modules/.cache/storybook` (I should maybe add a note about this in our `README.md`.


packages/components/docs/token-table.tsx
```diff
@@ -11,12 +11,26 @@ type TokenTableProps = {
   renderExample: (name: string, value: string) => JSX.Element;
 };
 
+const StyledTable = styled.table`
```
dcwither, line 14: very open to naming suggestions - not a super fan of this name - I could also do a css block instead.

ahuth: Seems reasonable. Could also be something like `TokensTable`

ahuth (approved):
Hell yes 🤘

packages/components/docs/token-table.tsx
dcwither: The problem was the name overlap with `TokenTable` so I'd rather not have it be a deceptively similar name.

I'll merge for now, and we can discuss the naming conventions for this sort of thing later.

PR #249, opened 2020-12-21
fix(typography): render Arimo in storybook properly

anniehu4:
### Summary:
https://app.clubhouse.io/connections/story/33653/ds-typography-confirm-with-design-through-percy

Sean pointed out that the Storybook fonts didn't actually look like Arimo -- turns out I wasn't loading the font files properly :')

### Test Plan:
Confirm storybook type:
Screen Shot 2020-12-21 at 6 48 07 PM

Matches Figma:
Screen Shot 2020-12-21 at 6 50 00 PM

Especially look at the number 1, because that was clearly off before. Percy changes are expected

### DIFF
--- a/packages/components/src/styles/global.js
+++ b/packages/components/src/styles/global.js
@@ -1,13 +1,17 @@
 import { GlobalStyles } from "twin.macro";
 import React from "react";
+import arimoBoldWoff from "./fonts/arimo-bold.woff";
+import arimoBoldWoff2 from "./fonts/arimo-bold.woff2";
+import arimoRegularWoff from "./fonts/arimo-regular.woff";
+import arimoRegularWoff2 from "./fonts/arimo-regular.woff2";
 import { createGlobalStyle } from "styled-components";
 
 const GlobalFonts = createGlobalStyle`
   @font-face {
     font-family: 'Arimo';
     src:
-      url(fonts/arimo-regular.woff2) format('woff2'),
-      url(fonts/arimo-regular.woff) format('woff');
+      url(${arimoRegularWoff2}) format('woff2'),
+      url(${arimoRegularWoff}) format('woff');
     font-weight: 400;
     font-style: normal;
   }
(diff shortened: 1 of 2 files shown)


packages/components/src/styles/global.js
```diff
@@ -1,22 +1,26 @@
...
 import { createGlobalStyle } from "styled-components";
 
 const GlobalFonts = createGlobalStyle`
   @font-face {
     font-family: 'Arimo';
     src:
-      url(fonts/arimo-regular.woff2) format('woff2'),
-      url(fonts/arimo-regular.woff) format('woff');
+      url(${arimoRegularWoff2}) format('woff2'),
+      url(${arimoRegularWoff}) format('woff');
```
anniehu4, line 14: Follows https://stackoverflow.com/a/55242731

ahuth: Interesting 🤔 . What's the imported value of the woff files? Is it a URL, or just the name of the font?

ahuth (approved):
LGTM

packages/components/src/styles/global.js
anniehu4: great question, this is what I see when logging:
```
static/media/arimo-regular.f279a33d.woff
static/media/arimo-regular.286d6967.woff2
static/media/arimo-bold.9fd2270a.woff
static/media/arimo-bold.3d3529a4.woff2
```
so looks like a derived URL?

ahuth: Cool, okay, that makes sense. Thanks for looking into it!

packages/components/src/styles/global.js
```diff
@@ -1,22 +1,26 @@
 import { GlobalStyles } from "twin.macro";
 import React from "react";
+import arimoBoldWoff from "./fonts/arimo-bold.woff";
```
dcwither, line 3: I think for export purposes, we probably don't want this part in our components exports, but instead as a step in our README. 

That or we need to expect some level of loader for this file, maybe that's fine.

packages/components/src/styles/global.js
dcwither: Is that true for just the compiled version as well? `dist/styles/global.js`

dcwither (commented):
Just realized this is a `.js` file, should we make it a `.ts` file?

PR #251, opened 2021-01-05
refactor: import fonts from google and remove export

anniehu4:
### Summary:
see clubhouse story in comments for more context. this PR:
- imports from Google Fonts for our own development
- stops exporting fonts, instead adds import instructions to the README
- moves global.js -> global.tsx

### Test Plan:
all checks pass


anniehu4:
CH story (trying out the github integration): https://app.clubhouse.io/czi-edu/story/62454/ds-handle-font-imports-for-dependencies

packages/components/.storybook/preview.tsx
```diff
@@ -1,4 +1,5 @@
 import "@chanzuckerberg/czedi-kit-tokens/css/variables.css";
+import "../src/styles/fonts.css";
```
anniehu4, line 2: _Why is this in `src/styles/` instead of `.storybook/`?_

In case we change our minds later and want to export it, or import it directly into `typography.tsx`, or something. no strong feeling tbh

packages/components/src/styles/global.tsx
```diff
@@ -0,0 +1,6 @@
+import React, { ReactElement } from "react";
+import { GlobalStyles } from "twin.macro";
+
+const EDSGlobalStyles = (): ReactElement => <GlobalStyles />;
+
+export default EDSGlobalStyles;
```
anniehu4, line 6: _Why remove the Fonts export?_
- traject already imports Arimo
- I had trouble including the google import inside `createGlobalStyles` (https://github.com/styled-components/styled-components/pull/2997)

I vaguely feel like it's better to let the application decide how to handle importing the required font anyways? but it is an extra setup step

ahuth: Agreed that the application can/should import the font in the best way for it. Different apps may need to use different fonts, as well.

ahuth (approved):

dcwither (approved):
This makes sense to me, thanks for adding the detailed questions for additional context.

PR #261, opened 2021-01-12
feat(tokens): create legacy typography tokens

anniehu4:
### Summary:
EDS heading sizes ([Figma](https://www.figma.com/file/sKEhNrLEdexPeVVn58dkWx/SDL-(Summit-Design-Language)-Explorations?node-id=393%3A228)) H1-H3 are too large for anything in Traject. This adds "legacy" type tokens starting at the H4 size for easier migration.

This is the mapping:
- legacy H1 = EDS H4
- legacy H2/H3 = EDS H5
- legacy H4 matches the current traject H4 styling (and EDS button)
- legacy H5 matches the current traject H5 styling (and EDS caption)

### Test Plan:
- run Storybook, confirm sizes match designs
- all checks pass except for expected changes in Percy


packages/components/src/typography/typography.stories.tsx
```diff
@@ -7,45 +7,45 @@ export default {
...
+    <Typography size="h4">H4 Heading 4 14/24</Typography>
     <br />
     <br />
     <Typography bold size="h5">
-      H5 Section Sub Head 18/24
+      H5 Heading 5
     </Typography>
     <br />
-    <Typography size="h5">H5 Heading 5 18/24</Typography>
+    <Typography size="h5">H5 Heading 5 12/20</Typography>
```
anniehu4, line 44: note that only Heading sizes are changed -- body sizes are the same as EDS

packages/tokens/properties/size/font.json
```diff
@@ -1,4 +1,64 @@
...
+      "font": {
+        "h1": {
+          "value": "1.5rem",
+          "comment": "Font size for Heading 1 sized text"
+        },
+        "h2": {
+          "value": "1.125rem",
+          "comment": "Font size for Heading 2 sized text"
+        },
+        "h3": {
```
anniehu4, line 13: _Why only define tokens up to H3?_

The H4 and H5 sizes are actually smaller than the body size, so it felt appropriate to leverage the "sm" and "xs" sizes instead

packages/components/src/typography/typography.stories.tsx
```diff
@@ -7,45 +7,45 @@ export default {
   component: Typography,
 };
 
-export const SizeRamp = (): JSX.Element => (
+export const LegacySizeRamp = (): JSX.Element => (
```
ahuth, line 10: Is the idea that existing usages in Learning Platform can use the legacy ramp, and then be migrated to non-legacy over time?

ahuth (approved):

packages/components/src/typography/typography.stories.tsx
anniehu4: yeah exactly, it'll make this first migration easier. then there are a couple directions the non-legacy sizing can go, but we can figure it out later

PR #262, opened 2021-01-12
feat(typography): add neutral color option

anniehu4:
### Summary:
Many headings/body text in Traject are gray (specifically, `eds-color-neutral-500`), so we should provide that as a color option as well

Screen Shot 2021-01-12 at 1 56 56 PM

### Test Plan:
- all checks pass
- see Approved changes in percy -- only changes should be from adding "Neutral" examples


packages/components/src/typography/typography.tsx
```diff
@@ -46,6 +47,8 @@ function styleFromColor(color: TypographyColor): TwStyle {
       return tw`text-brand-700`;
     case "info":
       return tw`text-info-700`;
+    case "neutral":
+      return tw`text-neutral-500`;
```
ahuth, line 51: Unrelated to your changes in this PR (which LGTM), but since this is just a tailwind classname, why do we need `tw`?

Could we instead do something like

```jsx
return (
  <TypographyComponent
    className={classnames(styleFromColor(color), styleFromSize(size))}
  />
);
```

And have it work the same? It's not clear to me what we're getting out of the macro system.  I'm sure I'm just missing something, though!

---

Like I said, not related to your actual changes.

ahuth (approved):

packages/components/src/typography/typography.tsx
anniehu4: that's a good question -- from my reading, I would say the main benefit of the macro is no extra runtime code. from [twin.macro docs](https://github.com/ben-rogerson/twin.macro#features):
> Adds no size to your build - Twin converts classes you’ve used into css objects using Babel and then compiles away, leaving no runtime code

and there are some other nice things in there like catching mistypings. 

Otherwise, it seems (from [this article](https://www.freecodecamp.org/news/how-to-style-your-react-apps-with-less-code-using-tailwind-css-and-styled-components/#tailwind-css-and-styled-components))
> "It's perfectly fine to use the utility classes with the className attribute, but if you want to nest properties you have to use the twin.macro library"

in that case, we'd add a separate `tailwind.css` file that imports the necessary utilities. I think for the first reason, if we stick with `styled-components` we should definitely leverage the macro. If we switch to Linaria, I'm not as sold on the macro ...

cc @dcwither for other thoughts

PR #266, opened 2021-01-13
chore: update stories and tests to use Component Story Format

anniehu4:
### Summary:
Pulls heavily from https://github.com/chanzuckerberg/lp-design-system/pull/244 to update stories to Component Story Format 🙂 The only real change is this now uses `generateSnapshots` from story-utils also. 

This will enable automated a11y checks in https://github.com/chanzuckerberg/lp-design-system/pull/267!

### Test Plan:
- all checks pass
- See above referenced PR for before/after screenshots in Storybook, all stories should show up with a "Controls" section now

PR #267, opened 2021-01-13
test: add accessibility tests to CI

anniehu4:
### Summary:
Installs axe-storybook-testing to add automated a11y testing to CI. Depends on stories being in CSF (#266)

### Test Plan:
- Confirm accessibility tests on all stories ran in Checks -> build -> "Run Accessibility Tests"


packages/components/package.json
```diff
@@ -51,6 +53,7 @@
     "@babel/preset-env": "^7.12.7",
     "@babel/preset-react": "^7.12.7",
     "@babel/preset-typescript": "^7.12.7",
+    "@chanzuckerberg/axe-storybook-testing": "^2.0.0",
```
ahuth, line 56: We should add an explicit (dev) dependency on axe-core in this package.

That's because axe-core is a peer dep of axe-storybook-testing. This means axe-storybook-testing expects anyone using it to also have a dependency on axe-core.

Peer deps are overrused, in my opinion, but I still went with it for axe-storybook-testing 😬 .

packages/components/package.json
```diff
@@ -25,11 +25,13 @@
     "build:js": "babel src -d dist --extensions '.js,.jsx,.ts,.tsx' --ignore src/**/*.spec.ts,src/**/*.spec.tsx --ignore src/**/*.stories.ts,src/**/*.stories.tsx",
     "build:types": "tsc --project . --declaration --emitDeclarationOnly",
     "build:watch": "npm run build:js -- --watch",
-    "build:storybook": "build-storybook",
+    "build:storybook": "build-storybook -o storybook-static",
     "deploy:docs": "storybook-to-ghpages --ci",
     "prepack": "npm run build && cp ./package.json dist/",
     "snapshot": "percy-storybook --widths=320,1280",
     "storybook": "start-storybook -p 6006",
+    "storybook:axe": "npm run build:storybook && npm run storybook:axeOnly",
```
ahuth, line 33: Is jest-axe still being used? If not, should we remove it?

packages/components/package.json
anniehu4: ooh good catch, dunno why npm didn't warn me

packages/components/package.json
anniehu4: nope, will remove 👍

packages/components/package.json
anniehu4: still dunno what happened earlier, but apparently `story-utils` also needed a higher peer dep for testing-library/react below :-)

ahuth (approved):
LGTM 🤘 

Thanks for doing this!

PR #270, opened 2021-01-20
chore(release): publish v0.0.1-alpha.6

anniehu4:
### Summary:
[CH task](https://app.clubhouse.io/czi-edu/story/65420/ds-publish-typography-component)

Publishing typography component to start traject migration. Also snuck in the [tailwind config export](https://app.clubhouse.io/czi-edu/story/62334/ds-prep-tailwind-config-for-export)

### Test Plan:
Run `npx lerna run prepack`, confirm:
- packages/tokens/dist includes `json/css-variables-nested.json`
- packages/components/dist includes
Screen Shot 2021-01-20 at 5 49 07 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,75 +3,61 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
-## [0.0.1-alpha.5](https://github.com/chanzuckerberg/lp-design-system/compare/v0.0.1-alpha.4...v0.0.1-alpha.5) (2020-10-19)
-
+## [0.0.1-alpha.6](https://github.com/chanzuckerberg/lp-design-system/compare/v0.0.1-alpha.5...v0.0.1-alpha.6) (2021-01-20)
 
 ### Features
 
-* **tokens:** replace eds token numbers ([59b3552](https://github.com/chanzuckerberg/lp-design-system/commit/59b355274973098315ba0bd531a760c2a0d9499e))
+- **typography:** add Typography component ([36f5d5c](https://github.com/chanzuckerberg/lp-design-system/commit/36f5d5c9e87e3515f69391426157361c05d523dd))
+- **typography:** add legacy and eds font tokens ([ddbaf64](https://github.com/chanzuckerberg/lp-design-system/commit/ddbaf648273c6780a73ff1c66ab5d7bac31191a3))
+- **tokens:** add css variables export in json ([c86ac32](https://github.com/chanzuckerberg/lp-design-system/commit/c86ac32c3477cbb67028bcd7a887f1520928853e))
+- **styles:** export Tailwind global styles ([66313dc](https://github.com/chanzuckerberg/lp-design-system/commit/66313dca2cfaaaee6666ef30884b3807867084d1))
+- **button:** add button example with styled-components and twin.macro ([fa8069a](https://github.com/chanzuckerberg/lp-design-system/commit/fa8069ac2f94abe84863e0033488f23f20f6a253))
 
+## [0.0.1-alpha.5](https://github.com/chanzuckerberg/lp-design-system/compare/v0.0.1-alpha.4...v0.0.1-alpha.5) (2020-10-19)
 
+### Features
 
-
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -3,75 +3,61 @@
...
 
-## [0.0.1-alpha.5](https://github.com/chanzuckerberg/lp-design-system/compare/v0.0.1-alpha.4...v0.0.1-alpha.5) (2020-10-19)
-
+## [0.0.1-alpha.6](https://github.com/chanzuckerberg/lp-design-system/compare/v0.0.1-alpha.5...v0.0.1-alpha.6) (2021-01-20)
 
 ### Features
 
-* **tokens:** replace eds token numbers ([59b3552](https://github.com/chanzuckerberg/lp-design-system/commit/59b355274973098315ba0bd531a760c2a0d9499e))
-
+- **typography:** add Typography component ([36f5d5c](https://github.com/chanzuckerberg/lp-design-system/commit/36f5d5c9e87e3515f69391426157361c05d523dd))
```
anniehu4, line 10: I didn't include any of the build/test changes? but I am admittedly a publishing n00b

also idk why there are so many whitespace changes it just auto-formatted on save

packages/components/package.json
```diff
@@ -27,7 +27,7 @@
     "build:watch": "npm run build:js -- --watch",
     "build:storybook": "build-storybook -o storybook-static",
     "deploy:docs": "storybook-to-ghpages --ci",
-    "prepack": "npm run build && cp ./package.json dist/",
+    "prepack": "npm run build && cp ./package.json ./tailwind.config.js dist/",
```
anniehu4, line 30: this includes the `playground/` folder as well ... I didn't think it was a big deal yet but we could rm -rf it here or ignore it when building

packages/components/package-lock.json
```diff
@@ -5035,11 +5035,6 @@
         }
       }
     },
-    "@chanzuckerberg/czedi-kit-tokens": {
```
anniehu4, line 5038: didn't know if/how to update this but I figure it'll get installed correctly later

packages/components/package.json
anniehu4: oh also re: tailwind config, I tested
```
import {theme} from '@chanzuckerberg/components/tailwind.config';
```
in traject, and it was working. could also import the whole config object

CHANGELOG.md
ahuth: I'm good with the whitespace changes 👍 

Also, I think it's reasonable to not include build/test changes. Sometimes I include those kinds of things, but it's only really necessary to include user/developer-facing changes (from the standpoint of someone consuming these packages).

packages/components/package.json
ahuth: Regarding the import, I think it would be better to import the theme directly from `@chanzuckerberg/czedi-kit-components`, unless we need to export the whole config for other reasons.

Also, don't have to do it now, but I'd be pretty conservative about what ends up in the published package, and would probably rm the playground folder.

Could do that as part of a build script ran in the build step, isntead of the pack, like we did in https://github.com/chanzuckerberg/frontend-libs/blob/83af6282feefee6cacd3137aa867513a4c6a1f43/scripts/build.sh.

ahuth (approved):

packages/components/package.json
anniehu4: yeah good point, I'll rm playground

re: import, in my head I was imagining there's a chance we actually reference the file in some postcss or tailwind compilation step...we might not though, in which case I'd agree with you. I added this context to https://app.clubhouse.io/czi-edu/story/62334/ds-prep-tailwind-config-for-export so we don't forget

ahuth: That's true. We may have to reference that config file somewhere, at some point.

packages/components/package-lock.json
ahuth: Hmmm, weird. I'd expect Lerna to do whatever it needs to do correctly.

anniehu4: oh god I tried `lerna version` and it pushed a commit for me 😭 and also didn't update this correctly lol...sad

ahuth: Yeah, that's a downside to Lerna - when what you wanted to happen didn't happen, it's still going to push stuff (commits, tags, etc).

Annoyingly, there also isn't a `--dry-run` flag, which would be useful.

Issue #272, opened 2021-01-20
Documentation needed: common lerna commands


ahuth:
We could use a wiki or readme entry on developing in this monorepo. Specifically thinking it would be useful to document commands for

- installing dependencies in a package
- publishing
- publishing pre-release versions

PR #273, opened 2021-01-21
chore: update package-lock and publishing instructions

anniehu4:
### Summary:
I did some silly stuff while trying to publish version alpha.6, including manually changing package-lock.json 😅 

### Test Plan:
n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # lp-design-system
+
 ![Test CI](https://github.com/chanzuckerberg/lp-design-system/workflows/Test%20CI/badge.svg)
 
 This is the Learning Platform Design System or `@chanzuckerberg/czedi-kit` (working title).
@@ -32,7 +33,7 @@ npm run lint:fix
 
 ### Run Locally
 
-```
+```bash
 npm start
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -56,4 +57,7 @@ npx lerna create '@chanzuckerberg/czedi-kit-<package-name>' \
 
 ### Publishing
 
-We're not ready to publish yet
+**We are only publishing alpha pre-releases at this time**
+
+1. Confirm everything is green on CI
+2. Run `lerna publish ???`
```
anniehu4, line 63: what andrew suggested:
```
lerna publish --canary --preid alpha
```

what I actually did:
- find + replace all version numbers
- update CHANGELOGs
- manually create a tag + release
- panic???
- run `lerna publish from-package`

@dcwither as lerna expert, what do you recommend / what have you been doing?

anniehu4:
fyi @ahuth , didn't add you as a reviewer yet b/c I wanted Devin to confirm first

ahuth:
Related: #272

README.md
dcwither: "expert" lol.

```
lerna publish --canary --preid alpha
```

Should work, though it won't give a commit message that agrees with our CI/commit linter.

What I've been doing to make sure my publishes will contain what I expect:

```bash
git checkout master 
npx lerna version --no-push --conventional-commits --cd-version prerelease --message "chore(release): publish v0.0.1-alpha.<version>"
# confirm tags/commit/changelog etc
npx lerna publish from-git --no-push
git push origin --tags
git push origin master # not sure if this step is necessary
```

This is also why it would be great to get a publish script out.

The `--no-push` means that I don't have to delete remote stuff after a messed up publish.

anniehu4: oof okay thanks! will add all that + context

README.md
```diff
@@ -56,4 +57,34 @@ npx lerna create '@chanzuckerberg/czedi-kit-<package-name>' \
...
+
+and confirm that tags/commit/changelog etc. look correct
+
+4. Push commit and tags to remote:
+
+```bash
+git push origin --tags
+```
+
+5. Finally, run `git push origin master` (this may not be strictly necessary, but doesn't hurt)
```
ahuth, line 86: Does the `git push origin --tags` step push only tags, or also the commits that get created by lerna? If it only pushes tags, then I think this step will be necessary.

ahuth (approved):
LGTM

README.md
anniehu4: ah good call, it looks like it only pushes tags

PR #284, opened 2021-02-05
feat(typography): split into Heading and Text components

anniehu4:
### Summary:
[[ch119356]](https://app.clubhouse.io/czi-edu/story/119356/ds-split-typography-into-heading-and-text-components)

From [discussion](https://docs.google.com/document/d/1qlnMcX03slmFYsCspAOxo6lx2posZvhemJdkAzLzdlk/edit?ts=601b1f47#), there are a few benefits to this:
- heading and non-heading text are different semantic use cases
- leaves the door open for something like a HeadingContext in the future
- traject is already used to a Heading component
- it's a pretty common DS pattern

I also removed Typography from the playground button and capitalized the file names

### Test Plan:
- Confirm snapshots look the same as before (just split across 2 files)
- Confirm no button changes in Percy (only visual changes are from new Heading/Text file names)
- Confirm docs look good


packages/components/src/playground/button.tsx
```diff
@@ -20,13 +20,7 @@ const ButtonComponent = styled.button<BaseButtonProps>(({ variant }) => [
...
   const { children, ...rest } = props;
 
-  return (
-    <ButtonComponent {...rest}>
-      <Typography bold color="white" size="button">
-        {children}
-      </Typography>
-    </ButtonComponent>
-  );
+  return <ButtonComponent {...rest}>{children}</ButtonComponent>;
```
anniehu4, line 23: doesn't matter that much since it's a playground, but following up on a previous comment that this probably shouldn't use Typography

packages/components/src/components/Text/Text.tsx
```diff
@@ -102,15 +84,7 @@ const TypographyComponent = styled.span<BaseTypographyProps>(
...
 
-export type TypographyProps = {
-  /**
-   * This prop can be used to specify which size text element should
-   * actually be rendered, in the case that you want to render an element
-   * as one heading but style it as if it were another. If both an `as` prop
-   * and a `size` prop are passed, the `as` will be used to determine the element
-   * and the `size` will be used to determine the styling.
-   */
-  as?: TypographyElement;
```
anniehu4, line 113: I'm making the choice here that `Text` should always be rendered as `<p>` -- only Headings should use this prop

packages/components/src/components/Heading/Heading.tsx
```diff
@@ -0,0 +1,44 @@
...
+   * actually be rendered, in the case that you want to render an element
+   * as one heading but style it as if it were another. If both an `as` prop
+   * and a `size` prop are passed, the `as` will be used to determine the element
+   * and the `size` will be used to determine the styling.
+   */
+  as?: HeadingElement;
+};
+
+const HeadingComponent = styled(TypographyComponent).attrs({
+  bold: true,
```
anniehu4, line 19: Currently in traject, all headings are indeed bold. so I'm comfortable enforcing this, and we can revisit if there's a design need later

also this is how the component is rendered in DOM:
Screen Shot 2021-02-05 at 12 15 11 PM

packages/components/src/components/Heading/Heading.stories.tsx
```diff
@@ -45,36 +45,6 @@ Heading5.args = {
   children: "Heading 5 12/20",
 };
 
-export const Body = Template.bind(null);
```
anniehu4, line 48: all moved to `Text.stories.tsx`

packages/components/src/Text/Text.tsx
```diff
@@ -102,15 +84,7 @@ const TypographyComponent = styled.span<BaseTypographyProps>(
...
 
-export type TypographyProps = {
-  /**
-   * This prop can be used to specify which size text element should
-   * actually be rendered, in the case that you want to render an element
-   * as one heading but style it as if it were another. If both an `as` prop
-   * and a `size` prop are passed, the `as` will be used to determine the element
-   * and the `size` will be used to determine the styling.
-   */
-  as?: TypographyElement;
```
anniehu4, line 113: I'm making the choice here that `Text` should always be rendered as `<p>` -- only Headings should use this prop

packages/components/src/Heading/Heading.tsx
```diff
@@ -0,0 +1,44 @@
...
+   * This prop can be used to specify which size heading should
+   * actually be rendered, in the case that you want to render an element
+   * as one heading but style it as if it were another. If both an `as` prop
+   * and a `size` prop are passed, the `as` will be used to determine the element
+   * and the `size` will be used to determine the styling.
+   */
+  as?: HeadingElement;
+};
+
+const HeadingComponent = styled(TypographyComponent).attrs({
```
anniehu4, line 18: Currently in traject, all headings are indeed bold. so I'm comfortable enforcing this, and we can revisit if there's a design need later

also this is how the component is rendered in DOM:
Screen Shot 2021-02-05 at 12 15 11 PM

packages/components/src/Heading/Heading.stories.tsx
```diff
@@ -45,36 +45,6 @@ Heading5.args = {
   children: "Heading 5 12/20",
 };
 
-export const Body = Template.bind(null);
```
anniehu4, line 48: all moved to `Text.stories.tsx`

packages/components/src/Heading/Heading.tsx
```diff
@@ -0,0 +1,44 @@
...
+  bold: true,
+})``;
+
+function Heading(props: HeadingProps): JSX.Element {
```
anniehu4, line 22: also deciding to not include margin or className here just yet based on discussion, may revisit in future

packages/components/src/Heading/Heading.tsx
```diff
@@ -0,0 +1,44 @@
...
+   * as one heading but style it as if it were another. If both an `as` prop
+   * and a `size` prop are passed, the `as` will be used to determine the element
+   * and the `size` will be used to determine the styling.
+   */
+  as?: HeadingElement;
+};
+
+const HeadingComponent = styled(TypographyComponent).attrs({
+  bold: true,
+})``;
```
anniehu4, line 20: Currently in traject, all headings are indeed bold. so I'm comfortable enforcing this, and we can revisit if there's a design need later

also this is how the component is rendered in DOM:
Screen Shot 2021-02-05 at 12 15 11 PM

anniehu4: also this is how the component is rendered in DOM:
Screen Shot 2021-02-05 at 12 15 11 PM

versus directly using `TypographyComponent`:
Screen Shot 2021-02-05 at 1 19 28 PM

I think the extra className is ok? this follows [styled-components' guidance](https://styled-components.com/docs/faqs#when-to-use-attrs) that
> The rule of thumb is to use attrs when you want every instance of a styled component to have that prop, and pass props directly when every instance needs a different one

packages/components/src/Heading/Heading.tsx
```diff
@@ -0,0 +1,44 @@
+import { TextProps, TypographyComponent } from "../Text/Text";
+import React from "react";
+import { styled } from "twin.macro";
+
+type HeadingElement = "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
+
+export type HeadingProps = TextProps & {
```
ahuth, line 7: Are we only exporting HeadingProps for storybook? If so, that's fine, but we could also infer them in the story

```ts
// Heading.stories.tsx

type Args = React.ComponentProps<typeof Heading>
```

Or if we want to use HeadingProps for other reasons, that's fine, too.

packages/components/src/Text/Text.tsx
```diff
@@ -93,7 +75,7 @@ interface BaseTypographyProps {
   size: TypographySize;
 }
 
-const TypographyComponent = styled.span<BaseTypographyProps>(
+export const TypographyComponent = styled.p<BaseTypographyProps>(
   ({ bold, color, size }) => [
     tw`font-arimo`,
```
ahuth, line 80: Are we going to specify this on every component? Should have folks specify this on the `<body>` tag, instead?

packages/components/src/playground/button.tsx
```diff
@@ -12,6 +11,7 @@ export type ButtonProps = React.ButtonHTMLAttributes<HTMLButtonElement> &
 
 const ButtonComponent = styled.button<BaseButtonProps>(({ variant }) => [
   tw`py-2 px-4 rounded border-none`,
+  tw`font-arimo text-white font-bold text-sm leading-body`,
```
ahuth, line 14: Why are these in a separate item from `py-2 px-4 rounded border-none`? Are we grouping them logically?

packages/components/src/Text/Text.tsx
```diff
@@ -102,15 +84,7 @@ const TypographyComponent = styled.span<BaseTypographyProps>(
...
-export type TypographyProps = {
-  /**
-   * This prop can be used to specify which size text element should
-   * actually be rendered, in the case that you want to render an element
-   * as one heading but style it as if it were another. If both an `as` prop
-   * and a `size` prop are passed, the `as` will be used to determine the element
-   * and the `size` will be used to determine the styling.
-   */
-  as?: TypographyElement;
+export type TextProps = {
```
ahuth, line 87: Nit, can just call this `Props`. Especially if we don't need to export this for anything other than storybook (since we can infer the props in the stories and remove the export).

packages/components/src/Text/Text.tsx
```diff
@@ -127,16 +101,15 @@ export type TypographyProps = {
    */
   color?: TypographyColor;
   /**
-   * The size of the html element. If no `as` prop is provided, then
-   * the component uses this value to determine which html tag to render
+   * The size of the html element. If used as a Heading and no `as` prop is provided,
```
ahuth, line 104: Should we remove the docs here about using as a Heading component, since there is a Heading component now?

packages/components/src/Heading/Heading.tsx
ahuth: 🤔 Do we want headings to always be bold in the design system? Or are we going for maxmizing the ease of integrating into Learning Platform currently?

ahuth: Also, I wouldn't worry about the extra class.

packages/components/src/Heading/Heading.tsx
```diff
@@ -0,0 +1,44 @@
...
+    children,
+    color = "base",
+    size,
+    /**
+     * Components that wrap typography sometimes requires props such as event handlers
+     * to be passed down into the element. One example is the tooltip component.  It
+     * attaches a onHover and onFocus event to the element to determine when to
+     * trigger the overlay.
+     */
+    ...rest
```
ahuth, line 34: So this will allow passing `className`, which I'm cool with 😄 . But wanted to point that out

packages/components/src/Heading/Heading.tsx
ahuth, line 18: Are we getting much out of basing this off of the typography component? Wondering if it makes sense to isolate things between text and heading.

Or let me know if we're getting some useful stuff out of sharing code between them.

packages/components/src/Heading/Heading.tsx
ahuth, line 20:
```ts
})``;
```

🤔 This is annoying.

packages/components/src/Text/Text.tsx
```diff
@@ -93,7 +75,7 @@ interface BaseTypographyProps {
   size: TypographySize;
 }
 
-const TypographyComponent = styled.span<BaseTypographyProps>(
+export const TypographyComponent = styled.p<BaseTypographyProps>(
```
ahuth, line 78: Should Typography be it's own file?

packages/components/src/Text/Text.tsx
```diff
@@ -127,16 +101,15 @@ export type TypographyProps = {
...
-   * the component uses this value to determine which html tag to render
+   * The size of the html element. If used as a Heading and no `as` prop is provided,
+   * then the component uses this value to determine which html tag to render
    * (e.g. 'h1', 'h2', etc.)
    */
   size: TypographySize;
 };
 
-function Typography(props: TypographyProps): JSX.Element {
+function Text(props: TextProps): JSX.Element {
```
ahuth, line 111: I know this isn't part of the change here, but can/should we remove the `JSX.Element` return values from these components? I don't think they're giving us much, and TypeScript will happily infer the right thing (whether that's a string, an element, null, a number, etc).

packages/components/src/playground/button.tsx
ahuth: Oh yes, button should definitely not use typography (no need to nest a p in a button)

packages/components/src/Heading/Heading.tsx
anniehu4: ehhhh we kinda went back and forth...Sean originally wanted always bold, but then some LP designers wanted a non-bold option, but in LP they're all bold anyways...though since Along allows passing a weight maybe we should keep it flexible

packages/components/src/Heading/Heading.tsx
anniehu4: hah you're right 🤭

packages/components/src/Text/Text.tsx
anniehu4: yeah I thought about that, will do it in a follow-up

packages/components/src/Heading/Heading.tsx
anniehu4: it's probably over-DRYing. I can instead do:
- some util/typography file with helpers `styleFromColor` and `styleFromSize`
- Text and Heading create their own styled components, using those helpers

then the prop comments can be a bit clearer too

packages/components/src/playground/button.tsx
anniehu4: I was thinking first line is about shape/spacing and this line is about text. can leave comments

packages/components/src/Text/Text.tsx
anniehu4: hm yeah I thought I added it b/c TS complained earlier but I'll check

anniehu4: Screen Shot 2021-02-05 at 3 14 00 PM

ah it was ESLint that was unhappy. should we turn off this rule?

packages/components/src/playground/button.tsx
ahuth: Cool, that organization makes sense 👍

packages/components/src/Text/Text.tsx
ahuth: Oh, I see. It wants explicit types on module boundaries (such as an export).

My opinion would be to disable rule. But it's fine either way, and probably not worth worrying about on this PR.

anniehu4: okie i'll also do it in a follow-up

packages/components/src/Heading/Heading.tsx
anniehu4: I added the option back for now ~ flexibility ~

packages/components/src/Heading/Heading.tsx
```diff
@@ -0,0 +1,69 @@
...
+  size: TypographySize;
+};
+
+interface BaseHeadingProps {
+  bold: boolean;
+  color?: TypographyColor;
+  size: TypographySize;
+}
+
+const HeadingComponent = styled.span<BaseHeadingProps>(
```
anniehu4, line 50: de-coupled from `TextComponent`. we may want to customize the two differently in the future

packages/components/src/Heading/Heading.tsx
```diff
@@ -0,0 +1,81 @@
...
```
ahuth, line 50: Just realized that even though this says `styled.span`, it'll get correctly overriden by the `as`/`size` props.

Musing out loud here: I wonder if it would make sense to split the styles for the different heading levels.

```
const H1 = styled.h1<BaseHeadingProps>`
  font-size: 32px // or whatever it is
`;

const H2 = styled.h2<BaseHeadingProps>(...);
const H3 = styled.h3<BaseHeadingProps>(...);
// etc
```

Moving away from dynamic styles, and towards static styles where possible has a couple benefits
- Easier to read and understand, in my opinion. Especially as more of them get added.
- There's less runtime cost. And if we can get to all static for a component, there's no runtime cost.

Like I said, just thinking out loud, and things to think and talk about. Nothing to do for now.

anniehu4: thanks for the musing 👍 I thought about this previously, but I wasn't sure how much we could actually save since
- the element can still be changed by `as`
- there's still dynamic `color`/`bold` props

I hadn't considered readability though! and using static styles where possible is def a good general approach

PR #304, opened 2021-02-20
Prep package names

ahuth:
### Summary:

Rename the packages:
- czedi-kit-components -> eds-components
- czedi-kit-tokens -> eds-tokens

### Test Plan:

- n/a

### DIFF
--- a/packages/components/package-lock.json
+++ b/packages/components/package-lock.json
@@ -5035,11 +5035,6 @@
         }
       }
     },
-    "@chanzuckerberg/eds-tokens": {
-      "version": "0.0.1-alpha.6",
-      "resolved": "https://registry.npmjs.org/@chanzuckerberg/eds-tokens/-/eds-tokens-0.0.1-alpha.6.tgz",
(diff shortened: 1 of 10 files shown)


packages/components/package-lock.json
```diff
@@ -5035,11 +5035,6 @@
         }
       }
     },
-    "@chanzuckerberg/czedi-kit-tokens": {
-      "version": "0.0.1-alpha.6",
```
ahuth, line 5039: lerna does not seem to want this entry here (running a bootstrap deletes it). Also, I think it was out of sync with the actual version).

Not having it here makes sense to me, since the package-lock.json is not published, and is only needed for local development. Lerna is taking care of setting up components' node_modules to have tokens.

packages/components/package.json
```diff
@@ -1,13 +1,13 @@
...
   "version": "0.0.1-alpha.7",
-  "description": "React components for czedi-kit",
+  "description": "React components for the education design system",
   "keywords": [
     "design system",
     "components"
   ],
   "author": "Devin Witherspoon <dcwither@gmail.com>",
-  "homepage": "https://github.com/chanzuckerberg/edu-design-system/tree/main/packages/czedi-kit-components#readme",
+  "homepage": "https://github.com/chanzuckerberg/edu-design-system/tree/main/packages/components",
```
ahuth, line 10: As far as I can tell, the previous link here wasn't valid. Or maybe was, but is dead now.

PR #305, opened 2021-02-21
Update all deps

ahuth:
### Summary:

Noticed a bunch of dependbot alerts, and updated all dependencies. Looks like these are all dev deps, and our actual dependencies are already up to date.

### Test Plan:

- CI


anniehu4 (approved):

ahuth:
Thanks!

PR #307, opened 2021-02-22
chore: cleanup type tokens, default props

anniehu4:
### Summary:
few cleanup things:
- Changes H3 to 16px. this follows Sean's new [ramp](https://www.figma.com/file/Clqzd6LnZN1ryBQcH35zBx/EDS-Foundations-%22Clean-State%22--Alpha?node-id=1846%3A452), and currently H2 and H3 are the exact same size which doesn't make much sense anyways
- Removes some duplicate tokens so all tokens have unique values
- changes prop destructuring to show default values in storybook correctly:
Screen Shot 2021-02-22 at 3 27 21 PM

### Test Plan:
only visual change is from H3 getting smaller


.eslintrc
```diff
@@ -34,6 +34,7 @@
       "files": ["*.ts", "*.tsx"],
       "extends": ["plugin:@typescript-eslint/recommended"],
       "rules": {
+        "@typescript-eslint/explicit-function-return-type": "off",
```
anniehu4, line 37: for some reason I kept getting warnings that the component functions needed return types :(

packages/components/src/Heading/Heading.tsx
```diff
@@ -55,21 +55,19 @@ const HeadingComponent = styled.span<BaseHeadingProps>(
...
-  } = props;
-
+function Heading({
+  as,
+  bold = true,
+  children,
+  size,
+  /**
+   * Components that wrap typography sometimes require props such as
+   * event handlers, tabIndex, ref, etc. to be passed down into the element.
```
ahuth, line 65: We may want to remove reference to `ref` here, since I don't think it would work (without `forwardRef`)

ahuth (approved):

PR #338, opened 2021-02-26
Dcwither/button experiments

dcwither:
### Summary:
This is a barebones implementation of the button variants. I'll be expanding this to include size, as well as links, and I'll take a stab at passing another component as the `as` to test out how it'll work with `Link` and a button with tracking.

### Test Plan:
storybook rendering as well as `npm test` though I'm mostly updating the snapshots anyway.


dcwither:
> ## [size-limit](https://github.com/ai/size-limit) report
> Path	Size
> packages/components/dist/index.js, packages/components/dist/Heading/Heading.js, packages/components/dist/Heading/index.js, packages/components/dist/Text/Text.js, packages/components/dist/Text/index.js, packages/components/dist/styles/global.js, packages/components/dist/util/typography.js	0 B (-100%)
> packages/components/dist/index.js, packages/components/dist/Button/button.js, packages/components/dist/Heading/Heading.js, packages/components/dist/Heading/index.js, packages/components/dist/Text/Text.js, packages/components/dist/Text/index.js, packages/components/dist/styles/global.js, packages/components/dist/util/clickable.js, packages/components/dist/util/typography.js	6.16 KB (+100% 🔺)

Ok, so this type of size limit report isn't that helpful because it resets whenever we add a new file. I'll look into an alternative that involves the global export (not my ideal).

packages/components/twin.d.ts
```diff
@@ -1,17 +1,9 @@
 import "twin.macro";
+import "@types/styled-components";
```
dcwither, line 2: @anniehu4 I think this addresses the same issue as the piece I deleted below, but I want to check with you since you added it - what should I test to make sure it worked?

anniehu4: ah I just grabbed it from the [Twin guide](https://github.com/ben-rogerson/twin.examples/tree/master/react-styled-components#complete-the-typescript-setup). I think if there's no typescript error on the Heading component (or anywhere we pass `as` to twin), that means it's working

packages/components/src/Button/button.tsx
```diff
@@ -0,0 +1,32 @@
...
+type ButtonHTMLElementProps = React.ButtonHTMLAttributes<HTMLButtonElement>;
+
+export type ButtonProps = {
+  /**
+   * The button contents or label.
+   */
+  children: ReactNode;
+  type: ButtonHTMLElementProps["type"];
+  disabled: ButtonHTMLElementProps["disabled"];
+  as: "button" | React.ComponentType<any>;
```
anniehu4, line 13: should we allow `React.ComponentType<any>` 🤔

packages/components/src/Button/button.tsx
```diff
@@ -0,0 +1,32 @@
...
+  variant: "flat",
+  color: "brand",
+  disabled: false,
+  type: "button",
+  size: "small",
+};
+
+Button.defaultProps = defaultProps;
+
+export default Button;
```
anniehu4, line 32: nit: at some point, can we make these files `Button.*` instead of `button.*` 😅

dcwither:
Closing since we've since merged a version of this using css modules

dcwither:
Closing since we've since merged a version of this using css modules

PR #360, opened 2021-03-12
Dcwither/css modules

dcwither:
### Summary:
Migrating to css modules - this may involve some config changes in traject (untested). See https://github.com/chanzuckerberg/edu-design-system/issues/362

### Test Plan:
storybook + updated tests (buttons are broken right now)


packages/components/src/util/typography.module.css
```diff
@@ -0,0 +1,94 @@
+/* sizes */
```
anniehu4, line 1: future thought: maybe this folder should be named `common` or `base` or something instead of `util` since it's actually defining components & styles now?

packages/components/docs/token-table.tsx
```diff
@@ -1,6 +1,9 @@
 import * as React from "react";
 
-import { styled } from "twin.macro";
+import classNames from "classnames/bind";
+import styles from "./token-table.module.css";
+
+const cx = classNames.bind(styles);
```
anniehu4, line 6: future thought -- might be nice to write a little util combining these 2 lines into 1 import

packages/components/src/util/typography.tsx
```diff
@@ -0,0 +1,88 @@
...
+  | "alert"
+  | "base"
+  | "brand"
+  | "info"
+  | "neutral"
+  | "success"
+  | "warning"
+  | "white";
+
+export type TypographyProps<IComponent extends React.ElementType> = {
```
anniehu4, line 30: @ahuth not blocking but would appreciate your review on the typescript because I certainly know less than either of you

anniehu4 (approved):
lgtm besides removing `packages/components/storybook-output.txt`

.eslintrc
```diff
@@ -47,11 +47,12 @@
...
-      "files": ["config.js", "*.config.js"],
+      "files": ["config.js", "*.config.js", "main.js"],
       "env": {
         "browser": false,
         "node": true
       }
     }
-  ]
+  ],
+  "ignorePatterns": ["!.storybook"]
```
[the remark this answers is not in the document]
dcwither: it means don't ignore it - dot files apparently get ignored by default (has to do with `.git` files etc)

packages/components/docs/token-table.module.css
```diff
@@ -0,0 +1,14 @@
+
```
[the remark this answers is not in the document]
dcwither: I'll set up stylelint with prettier and it should take care of this for us (though I personally kind of like a leading newline) helps me see it more easily since it gives distance from the tabs above.

packages/components/docs/token-table.tsx
dcwither: Trying to think how that would be possible with the import syntax - we could write our own babel plugin. Is that what you mean?

packages/components/src/util/typography.tsx
```diff
@@ -0,0 +1,88 @@
...
+    </Component>
+  );
+}
+
+Typography.defaultProps = {
+  weight: null,
+  color: "base",
+};
+
+export default Typography;
```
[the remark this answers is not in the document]
dcwither: oh 🤦

packages/components/docs/token-table.tsx
anniehu4: oh I didn't think it through very much... does sound like more effort than it's worth just to avoid a lil bit of repeated syntax

packages/components/docs/token-table.tsx
```diff
@@ -11,23 +14,9 @@ type TokenTableProps = {
...
-
-  td {
-    font-family: monospace;
-  }
-`;
-
 const TokenTable = ({ tokens, renderExample }: TokenTableProps) => {
   return (
-    <StyledTable>
+    <table className={cx("token-table")}>
```
ahuth, line 19: Mentioned this in the RFC, but what's the benefit of using `bind` over directly using the class?

```suggestion
    <table className={styles['token-table']}>
```

The values of `styles` are class names, and I'm not aware of any downsides of using them as such.

packages/components/src/util/typography.module.css
```diff
@@ -0,0 +1,94 @@
...
+
+.typography,
+.typography--size-body {
+  @apply text-body leading-body;
```
ahuth, line 5: I asked @dcwither this before, but don't remember the answer 🤦 - can we not use the global Tailwind classes and get the performance / bundle size benefits of Tailwind?

To use those, this would look something like

```suggestion
  composes: text-body leading-body from global;
```

packages/components/src/util/typography.tsx
```diff
@@ -0,0 +1,88 @@
...
+  children,
+  color,
+  size,
+  weight,
+  ...rest
+}: TypographyProps<IComponent>) {
+  const Component = as;
+  return (
+    <Component
+      className={cx(`typography--size-${size}`, `typography--color-${color}`, {
```
ahuth, line 73: This is seriously the best thing about CSS modules (and also Linaria) - there's no runtime, dynamic stuff happening here. Just setting a different class in different situations.

packages/components/src/util/typography.module.css
```diff
@@ -0,0 +1,94 @@
...
+  @apply text-warning-700;
+}
+
+.typography--color-white {
+  @apply text-white;
+}
+
+/**
+ * font-weights
+ * allows override for text components, shouldn't be used by heading components
```
ahuth, line 85: This is a sign that typography and headings should be completely separate components and styles.

packages/components/src/util/typography.tsx
ahuth: Looks pretty good. My only suggestion is to drop the 'I' prefix from the type param. I don't believe it's needed (although let me know if it is), and we shouldn't get in the habit of adding it.

```suggestion
export type TypographyProps<Component extends React.ElementType> = {
```

or even

```suggestion
export type TypographyProps<T extends React.ElementType> = {
```

packages/components/src/util/typography.module.css
dcwither: The difficulty is that in an unknown number of cases bootstrap has its own utility classes which can conflict with these. It's definitely something to look into, but right now I'm worried about those overriding these styles in unpredictable ways.

packages/components/src/util/typography.tsx
dcwither: Definitely not needed, and probably an outdated pattern - I just remember `I` was sometimes used to signal interface and `T` was sometimes used to signal Type. Sorry I didn't see your review before merging, will look into this more after the button stuff.

packages/components/docs/token-table.tsx
dcwither: No benefit here, but for the more complex components it's a nice bit of syntactic sugar for concatenating an unknown number of classes together.

packages/components/src/util/typography.module.css
dcwither: hmm - I kinda hoped we could manage that with types preventing the override in heading. Same way they limit which elements can be used.

PR #363, opened 2021-03-16
Dcwither/button css modules

dcwither:
### Summary:

This adds a first draft of buttons to the css modules - I could put them back in the playground folder if we're worried about them being used.

### Test Plan:

- Storybook + percy screenshots capture the visual changes. 
- Snapshots capture that the classes are properly used

I could put some env checks around the state code, but that seems like overkill and it's helpful for testing in other cases.

Reduction in coverage appears to be only from the stories file since we're not including the grid in the snapshot tests.

### Open Questions:

How do you feel about the new stylelint rules/added classes to simulate the pseudoclasses?


.stylelintrc
```diff
@@ -18,7 +18,9 @@
       }
     ],
     "custom-property-empty-line-before": "never",
-    "no-eol-whitespace": null
+    "no-eol-whitespace": null,
+    "selector-nested-pattern": "^(&(:|\\.[\\w-]*--state-)(hover|focus|focus-within|disabled),?\\s*)+$",
```
dcwither, line 22: Sadly I can't add comments to json files (maybe that's a reason to make this a `stylelint.config.js`), but this basically only allows certain pseudoclasses (we may want more) and certain classes that communicate a simulated activity state to be nested.

Probably worthwhile to expand to all pseudoclasses and keep the `state-` the same.

packages/components/package.json
```diff
@@ -23,7 +23,7 @@
     "build": "npm run build:clean && npm run build:js && npm run build:styles && npm run build:types && rm -rf dist/playground",
     "build:clean": "rm -rf dist/",
     "build:js": "babel src -d dist --extensions '.js,.jsx,.ts,.tsx' --ignore src/**/*.spec.ts,src/**/*.spec.tsx --ignore src/**/*.stories.ts,src/**/*.stories.tsx",
-    "build:styles": "postcss src/**/*.css --dir dist/ --base src/",
+    "build:styles": "postcss \"src/**/*([a-zA-Z-_])?(.module).css\" --dir dist/ --base src/ --verbose",
```
dcwither, line 26: prevents `stories.module.css` files from getting put in dist.

packages/components/package.json
```diff
@@ -55,7 +55,7 @@
     "@babel/preset-react": "^7.12.13",
     "@babel/preset-typescript": "^7.12.17",
     "@chanzuckerberg/axe-storybook-testing": "^2.0.1",
-    "@chanzuckerberg/story-utils": "^1.2.1",
+    "@chanzuckerberg/story-utils": "^2.0.0",
```
dcwither, line 58: updates api for skipping rule

packages/components/src/Button/button.stories.tsx
```diff
@@ -0,0 +1,108 @@
...
+          ))}
+        </li>
+      ))}
+    </ul>
+  );
+};
+
+allVariants.parameters = {
+  snapshot: {
+    skip: true,
```
dcwither, line 106: this makes the snapshots way too bloated

packages/components/src/styles/tailwind-variables.css
```diff
@@ -0,0 +1,11 @@
...
+/* Copied from output of `@tailwind utilities` */
+:root {
+  --tw-shadow: 0 0 #0000;
+  --tw-ring-inset: var(--tw-empty,/*!*/ /*!*/);
+  --tw-ring-offset-width: 0;
+  --tw-ring-offset-color: #fff;
+  --tw-ring-color: rgba(59, 130, 246, 0.5);
+  --tw-ring-offset-shadow: 0 0 #0000;
+  --tw-ring-shadow: 0 0 #0000;
+}
```
dcwither, line 11: still need to add a stylelint prettier integration to fix this automatically

packages/tokens/properties/color/base.json
```diff
@@ -60,6 +60,7 @@
   },
   "eds": {
     "color": {
+      "transparent": { "value": "transparent" },
```
dcwither, line 63: I needed this in tailwind, but I should probably just add it to a config - no reason to make a variable for it.

dcwither:
> 226.72 KB (+98.41% 🔺)

😕  - I'll look into this tomorrow

package.json
```diff
@@ -51,6 +51,7 @@
     "@size-limit/preset-small-lib": "^4.9.2",
     "@typescript-eslint/eslint-plugin": "^4.15.1",
     "@typescript-eslint/parser": "^4.15.1",
+    "classnames": "^2.2.6",
```
anniehu4, line 54: oh, kinda surprised that last PR worked fine without this 👀

packages/components/src/styles/tailwind-variables.css
anniehu4: Why not include this in `global.css`?

packages/components/package.json
anniehu4: Would it be a lot of work to move the ButtonGrid out to a separate place now? Inspired by Brad Frost's repo he shared, something like:
```
.storybook
  - components/ButtonGrid
      - ButtonGrid.module.css
      - ButtonGrid.tsx
      - ButtonGrid.stories.tsx
```

I don't love the idea of having files that are only for storybook in `src/`

packages/components/src/Button/button.stories.tsx
```diff
@@ -0,0 +1,148 @@
...
+};
+
+export const link = Template.bind(null);
+link.args = {
+  children: "Link Button",
+  variant: "link",
+  color: "alert",
+};
+
+export const linkInBody = (args) => (
```
anniehu4, line 53: 🕺

packages/components/src/Button/button.stories.tsx
```diff
@@ -0,0 +1,148 @@
...
+    </ul>
+  );
+};
+
+allVariants.parameters = {
+  snapshot: {
+    skip: true,
+  },
+  axe: {
+    skip: true,
```
anniehu4, line 146: guessing we have to skip b/c of the disabled buttons, but it would be really cool to run this across all colors & hocus states 😮

packages/components/src/Button/button.tsx
```diff
@@ -0,0 +1,35 @@
...
+  /**
+   * The button contents or label.
+   */
+  children: ReactNode;
+  type: ButtonHTMLElementProps["type"];
+  disabled: ButtonHTMLElementProps["disabled"];
+  as: "button" | React.ComponentType<any>;
+  variant: ClickableProps<"button">["variant"];
+  color: ClickableProps<"button">["color"];
+  size: ClickableProps<"button">["size"];
```
anniehu4, line 16: nit: alphabetize

packages/components/src/util/clickable.module.css
```diff
@@ -0,0 +1,164 @@
...
+
+.button--size-medium {
+  @apply px-6 py-4 text-sm leading-snug;
+}
+
+.button--size-large {
+  @apply px-8 py-5 leading-7;
+
+  /* manual font size sinde there's no match in tokens */
+  font-size: 1.25rem;
```
anniehu4, line 23: not blocking, but in its final form this should definitely have a corresponding token size

packages/components/src/util/clickable.tsx
```diff
@@ -0,0 +1,58 @@
...
+  variant: "flat" | "outline" | "minimal" | "link";
+  /**
+   * The color of the element.
+   */
+  color: "alert" | "brand" | "neutral" | "success" | "warning";
+  /**
+   * The size of the element.
+   */
+  size: "small" | "medium" | "large";
+  className: string;
```
anniehu4, line 20: nit: alphabetize

packages/components/src/util/clickable.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <Component
+      className={cx(
+        `button`,
+        `button--variant-${variant}`,
+        `button--color-${color}`,
+        {
+          // For testing in storybook and percy
+          [`button--state-${state}`]: state,
+          [`button--size-${size}`]: variant !== "link",
```
anniehu4, line 46: ooh I like that we can check prop combos this way too

packages/components/src/util/clickable.tsx
```diff
@@ -0,0 +1,58 @@
...
+  size,
+  state,
+  variant,
+  ...rest
+}: ClickableProps<IComponent>) {
+  const Component = as;
+  return (
+    <Component
+      className={cx(
+        `button`,
```
anniehu4, line 40: would a normal link (rendered & styled as link) also go through this component do you think?

packages/components/package.json
dcwither: that's a good point - I'll explore it

packages/components/src/styles/tailwind-variables.css
dcwither: I could, it would just mean that we also have to copy/paste this for the app as well.

packages/components/src/util/clickable.tsx
dcwither: yes

packages/components/src/Button/button.stories.tsx
dcwither: that's a good point - I'll think about an alternate that is axe only

packages/components/package.json
dcwither: I'll move `tokens.stories` out as well

package.json
dcwither: almost definitely a dev dependency of something else. Would only work by fluke in another project.

packages/components/src/util/clickable.tsx
dcwither: Though this may be a reason to think of a name other than button - like `interactive` or `clickable` like it currently is. I like that clickable encompasses more than just button, but it feels too broad/meaningless honestly.

packages/components/.storybook/Button/button.stories.tsx
```diff
@@ -0,0 +1,82 @@
+import * as React from "react";
+
+import Clickable, { ClickableProps } from "../../src/util/clickable";
+
+import { ButtonProps } from "../../src/Button";
+import Heading from "../../src/Heading";
+import Text from "../../src/Text";
+import styles from "./button.module.css";
+
+export { default } from "../../src/Button/button.stories";
```
anniehu4, line 10: very cool how this keeps them all in the same storybook folder!

anniehu4 (approved):
yay this looks great! I think we can continue discussing the types & classNames syntax outside of this

packages/components/src/Button/button.tsx
```diff
@@ -0,0 +1,35 @@
...
+};
+
+function Button(props: ButtonProps) {
+  const { ...rest } = props;
```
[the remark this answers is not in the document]
anniehu4: were you still gonna change something here haha

packages/components/.storybook/Button/button.stories.tsx
dcwither: Right! I had a hunch it might, but I was really happy to be proven right 😄

PR #365, opened 2021-03-24
refactor(classnames): move to camelcase classnames

dcwither:
### Summary:

Let's see what the repo would look like with a different way of referencing css modules! Also since this style doesn't use bind, we can use [`clsx`](https://github.com/lukeed/clsx) which is smaller and more performant than `classnames`.

### Test Plan:

- `npm test`
- CI
- Percy (should have no differences from buttons)

One benefit that looks like a downside is we have a more accurate picture of our coverage.

### DIFF
--- a/packages/components/src/Button/__snapshots__/button.spec.tsx.snap
+++ b/packages/components/src/Button/__snapshots__/button.spec.tsx.snap
@@ -2,7 +2,7 @@
 
 exports[`<Button /> flat story renders snapshot 1`] = `
 <button
-  class="button button--variant-flat button--color-success button--size-medium"
+  class="button sizeMedium variantFlat colorSuccess"
   type="button"
 >
   ​
@@ -12,7 +12,7 @@ exports[`<Button /> flat story renders snapshot 1`] = `
 
 exports[`<Button /> link story renders snapshot 1`] = `
 <button
-  class="button button--variant-link button--color-alert"
+  class="button variantLink colorAlert"
   type="button"
 >
   ​
@@ -22,11 +22,11 @@ exports[`<Button /> link story renders snapshot 1`] = `
 
 exports[`<Button /> linkInBody story renders snapshot 1`] = `
 <p
-  class="typography--size-body typography--color-base"
+  class="typography sizeBody colorBase"
 >
   This text surrounds the 
   <button
-    class="button button--variant-link button--color-brand"
+    class="button variantLink colorBrand"
     type="button"
   >
     ​
@@ -38,11 +38,11 @@ exports[`<Button /> linkInBody story renders snapshot 1`] = `
 
 exports[`<Button /> linkInHeading story renders snapshot 1`] = `
 <p
-  class="typography--size-h1 typography--color-base"
+  class="typography sizeH1 colorBase"
 >
   This text surrounds the 
   <button
-    class="button button--variant-link button--color-brand"
+    class="button variantLink colorBrand"
     type="button"
   >
     ​
@@ -54,7 +54,7 @@ exports[`<Button /> linkInHeading story renders snapshot 1`] = `
 
 exports[`<Button /> outline story renders snapshot 1`] = `
 <button
-  class="button button--variant-outline button--color-brand button--size-medium"
+  class="button sizeMedium variantOutline colorBrand"
   type="button"
 >
   ​
--- a/packages/components/src/util/clickable.tsx
+++ b/packages/components/src/util/clickable.tsx
@@ -1,8 +1,6 @@
 import React from "react";
-import classNames from "classnames/bind";
+import clsx from "clsx";
 import styles from "./clickable.module.css";
-
-const cx = classNames.bind(styles);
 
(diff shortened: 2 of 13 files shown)


packages/components/src/Button/__snapshots__/button.spec.tsx.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`<Button /> flat story renders snapshot 1`] = `
 <button
-  class="button button--variant-flat button--color-success button--size-medium"
+  class="button sizeMedium variantFlat colorSuccess"
```
dierat, line 5: This is so much easier to read when inspecting elements in dev tools.

packages/components/src/util/clickable.tsx
```diff
@@ -20,7 +18,7 @@ export type ClickableProps<IComponent extends React.ElementType> = {
   /**
    * A hidden prop for visual testing
    */
-  state?: "inactive" | "hover" | "focus" | "active" | "disabled";
```
dierat, line 23: Could you help me understand the changes around "disabled" in this PR?

packages/components/src/util/typography.tsx
```diff
@@ -70,9 +68,32 @@ function Typography<IComponent extends React.ElementType>({
...
+        color === "base" && styles.colorBase,
+        color === "brand" && styles.colorBrand,
+        color === "info" && styles.colorInfo,
+        color === "neutral" && styles.colorNeutral,
+        color === "success" && styles.colorSuccess,
+        color === "warning" && styles.colorWarning,
+        color === "white" && styles.colorWhite,
+        // Weights
+        weight === "bold" && styles.weightBold,
+        weight === "normal" && styles.weightNormal,
```
dierat, line 95: Mmm too bad you have to write everything out like this. But the string interpolation is easier to mess up, and I don't think it throws an error if you do; the class just isn't applied.

packages/components/src/util/clickable.tsx
ahuth: I think the styles are now based off of the `:disabled` pseudo-class, instead of an explicit class.

ahuth (approved):
LGTM

I dig it, but also have a well-known and explicit bias for this pattern. So I'm good with whatever pattern works the best for our use cases.

packages/components/src/util/typography.tsx
dierat: I was referring to mistyping the class name which creates a legitimate string but it's not associated with an actual style (I did this yesterday 😅).

anniehu4: good points abt the downsides of string interpolation! I think this component may be the most cases to type out, and it's not terrible.

I wonder if it's worthwhile to break it up some (for readability) -- e.g. moving the logic of picking a className for size to a `sizeStyle` function? could do the same for button, though that one's not as bad

anniehu4 (approved):
thanks for showing how this looks! I am on board 👍

anniehu4:
a random musing -- if we think of stories as visual tests (which are captured thru percy), the coverage for `clickable.tsx` should actually be really high 🤔 does CodeCov only consider Jest tests?

ahuth:
> a random musing -- if we think of stories as visual tests (which are captured thru percy), the coverage for clickable.tsx should actually be really high 🤔 does CodeCov only consider Jest tests?

Good point. I believe CodeCov only considers jest tests. But if we write tests based off of stories, it should reflect in the %.

packages/components/src/util/typography.tsx
dcwither: I wanted to keep it in one place to see:
1. how extreme it would get
2. all the possible classes in one place
3. an extreme case we could share as a possible strategy to product engineers

> I think this component may be the most cases to type out, and it's not terrible.
I agree here, and I think this style, while not always beautiful, leads to really clear direction on what to do

dcwither:
> > a random musing -- if we think of stories as visual tests (which are captured thru percy), the coverage for clickable.tsx should actually be really high 🤔 does CodeCov only consider Jest tests?
> 
> Good point. I believe CodeCov only considers jest tests. But if we write tests based off of stories, it should reflect in the %.

Yup, the only thing was that the snapshot was unreasonably large, and we would probably end up updating it without looking at the changes. That plus axe tests failing on disabled buttons led me to stopping it for that.

Definitely something I want to revisit though. We could also explore [property based testing](https://github.com/dubzzz/fast-check#getting-started).

packages/components/src/util/typography.tsx
dcwither: > I don't think it throws an error if you do; the class just isn't applied.

I don't think this throws an error either, since it gets the following types:
```ts
declare module "*.module.css" {
  const content: Record<string, string>;
  export default content;
}
```

We could explore something like https://github.com/mrmckeb/typescript-plugin-css-modules to ensure the types work. There's also https://www.npmjs.com/package/css-modules-flow-types-cli on the flow side for the app.

PR #398, opened 2021-04-07
feat(typography): added spacing to typography components

dcwither:
### Summary:
- Added a spacing property to the text components, defaulting to `margin-bottom: 0`
- Moved away from `Component.defaultProps` format due to deprecation of `defaultProps` on function components

### Test Plan:
- Percy screenshot comparison (should be no change)
- Inspect confirming the margin present when spacing isn't `none`

This PR also flagged that there's an issue with our docgen when using `export default function Component...` the docgen doesn't capitalize the name of the component:
**Works:**
```
try {
    // @ts-ignore
    Button.displayName = "Button";
    // @ts-ignore
    Button.__docgenInfo = { "description": "", "displayName": "Button", "props": { "as": { "defaultValue": { value: "button" }, "description": "", "name": "as", "required": false, "type": { "name": "\"button\" | ComponentClass<any, any> | FunctionComponent<any>" } }, "children": { "defaultValue": null, "description": "The button contents or label.", "name": "children", "required": true, "type": { "name": "ReactNode" } }, "color": { "defaultValue": { value: "brand" }, "description": "", "name": "color", "required": false, "type": { "name": "\"alert\" | \"brand\" | \"neutral\" | \"success\"" } }, "disabled": { "defaultValue": { value: false }, "description": "", "name": "disabled", "required": false, "type": { "name": "boolean" } }, "size": { "defaultValue": { value: "medium" }, "description": "", "name": "size", "required": false, "type": { "name": "\"small\" | \"medium\" | \"large\"" } }, "type": { "defaultValue": { value: "button" }, "description": "", "name": "type", "required": false, "type": { "name": "\"button\" | \"submit\" | \"reset\"" } }, "variant": { "defaultValue": { value: "flat" }, "description": "", "name": "variant", "required": false, "type": { "name": "\"link\" | \"flat\" | \"outline\"" } } } };
    // @ts-ignore
    if (typeof STORYBOOK_REACT_CLASSES !== "undefined")
        // @ts-ignore
        STORYBOOK_REACT_CLASSES["src/Button/button.tsx#Button"] = { docgenInfo: Button.__docgenInfo, name: "Button", path: "src/Button/button.tsx#Button" };
}
catch (__react_docgen_typescript_loader_error) { }
```
**Doesn’t Work:**
```
try {
    // @ts-ignore
    button.displayName = "button";
    // @ts-ignore
    button.__docgenInfo = { "description": "", "displayName": "button", "props": { "as": { "defaultValue": { value: "button" }, "description": "", "name": "as", "required": false, "type": { "name": "\"button\" | ComponentClass<any, any> | FunctionComponent<any>" } }, "children": { "defaultValue": null, "description": "The button contents or label.", "name": "children", "required": true, "type": { "name": "ReactNode" } }, "color": { "defaultValue": { value: "brand" }, "description": "", "name": "color", "required": false, "type": { "name": "\"alert\" | \"brand\" | \"neutral\" | \"success\"" } }, "disabled": { "defaultValue": { value: false }, "description": "", "name": "disabled", "required": false, "type": { "name": "boolean" } }, "size": { "defaultValue": { value: "medium" }, "description": "", "name": "size", "required": false, "type": { "name": "\"small\" | \"medium\" | \"large\"" } }, "type": { "defaultValue": { value: "button" }, "description": "", "name": "type", "required": false, "type": { "name": "\"button\" | \"submit\" | \"reset\"" } }, "variant": { "defaultValue": { value: "flat" }, "description": "", "name": "variant", "required": false, "type": { "name": "\"link\" | \"flat\" | \"outline\"" } } } };
    // @ts-ignore
    if (typeof STORYBOOK_REACT_CLASSES !== "undefined")
        // @ts-ignore
        STORYBOOK_REACT_CLASSES["src/Button/button.tsx#button"] = { docgenInfo: button.__docgenInfo, name: "button", path: "src/Button/button.tsx#button" };
}
catch (__react_docgen_typescript_loader_error) { }
```


packages/components/src/util/typography.tsx
```diff
@@ -93,6 +100,11 @@ function Typography<IComponent extends React.ElementType>({
         // Weights
         weight === "bold" && styles.weightBold,
         weight === "normal" && styles.weightNormal,
+        // Spacing
+        spacing === "none" && styles.spacingNone,
```
dcwither, line 104: @anniehu4 Looking for recommendations on whether this is a good starting point for spacing or if there's a better resource for defaults.

anniehu4: I'd def recommend including `mb-1`, and I'm not suuuper sure there's a need for `mb-8` 🤔  given that, maybe we should also think of spacing in multiples of 8px

reasoning is most of traject's headings currently have `5px` margin bottom ([claro/_components/_app.scss](https://github.com/FB-PLP/traject/blob/cefa996b6c9593581f76364c4c6ad36fc20f4c8e/app/assets/stylesheets/claro/_components/_app.scss#L29)) and of the spacing variables in `_variables.scss`:
```
// Spacing
$spacing-none: 0;
$spacing-xxs: 2px;
$spacing-xs: 5px;
$spacing-sm: 10px;
$spacing-md: 15px;
$spacing-lg: 20px;
$spacing-lg-2: 25px;
$spacing-xl: 30px;
```

I only commonly see up to 20px being used

packages/components/src/util/typography.tsx
```diff
@@ -54,15 +56,20 @@ export type TypographyProps<IComponent extends React.ElementType> = {
    * we will add stricter enforcement -- e.g. enforcing a boldness
    * for each size or checking for mutually exclusive props.
    */
-  weight: "bold" | "normal" | null;
+  weight?: "bold" | "normal" | null;
+  /**
+   * Specifies the bottom-margin that should be applied to the typography element
+   */
+  spacing?: TypographyMargin;
```
anniehu4, line 63: could we actually handle this similarly to `weight`, where there's a default `mb-0` set on all typography, and extra styles / className only gets passed if it's explicitly set?

plop-templates/Component/Component.tsx.hbs
```diff
@@ -30,10 +30,6 @@ function {{pascalCase name}}({ children, align }: Props) {
     </div>
   );
 }
-
-{{pascalCase name}}.defaultProps = {
-  children: "Hello world =^_^=",
-  align: "left",
-};
```
anniehu4, line 37: move these into the function declaration above

dcwither: oops

packages/components/src/util/typography.tsx
dcwither: Oh no - I'm pretty sure our design docs mention our spacing as multiples of 8px, but maybe that's too optimistic 🤔

packages/components/src/util/typography.tsx
dcwither: was considering that, sounds good

dcwither (commented):
Thanks for the review @anniehu4, what do you think of the general naming `none/1x/2x...`?

anniehu4:
just to document it:
> what do you think of the general naming none/1x/2x...?

I like it! and I like it better than semantic naming for spacing like "xs" or "tight". We should make sure to clearly document what it's in multiples of though (which I'm assuming is being clarified as part of pilot also)

packages/components/src/util/typography.tsx
anniehu4: you know... it actually might be fine 🤔  maybe everything is too squished rn and 8px would actually look nicer

anniehu4: actually, you know what -- down to try it and revisit if we do need smaller 🙂  I trust sean's eye more than mine

dcwither:
> We should make sure to clearly document what it's in multiples of though (which I'm assuming is being clarified as part of pilot also)

I spoke with Sean, and I think we're gonna settle on each `x` being `8px/0.25rem`, and an initial `half` being `4px/0.25rem`.

anniehu4:
cool, sounds good!

packages/components/package.json
```diff
@@ -30,7 +30,7 @@
     "deploy:docs": "storybook-to-ghpages --ci",
     "prepack": "npm run build && cp ./package.json ./tailwind.config.js dist/",
     "snapshot": "percy-storybook --widths=320,1280",
-    "storybook": "start-storybook -p 6006",
+    "storybook": "start-storybook -p 6008",
```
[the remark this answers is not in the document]
anniehu4: why not `6007`? just asking b/c that's what I get prompted with whenever 6006 isn't available

packages/tokens/properties/size/font.json
```diff
@@ -104,6 +104,10 @@
           "value": "2rem",
           "comment": "Line height for Heading 4 sized text"
         },
+        "h5": {
```
anniehu4, line 107: oh, I excluded this intentionally because it was the exact same as the `body` value. was thinking that "tokens" should have unique values, otherwise it's a bit confusing which one to use

packages/components/src/util/typography.tsx
anniehu4, line 104: should this be removed now?

anniehu4 (approved):
few comments but nothing blocking

packages/components/src/util/typography.tsx
dcwither: I like the idea of having the ability to be explicit in the case that someone is using a condition, something like:

`spacedOut ? '1x' : 'none'` instead of `spacedOut ? '1x' : undefined`

packages/tokens/properties/size/font.json
dcwither: Yeah, I'm being a little inconsistent here, but I think having them line up, even if they're the same means we can
- Distinguish them more easily in config changes
- More consistently reference them in our tailwind config for `text-h5`

Since this is still an alpha change, I can revert if the discussion ends up changing.

packages/components/package.json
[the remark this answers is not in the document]
dcwither: I should probably go with a more arbitrary value, but basically because 6007 is the next one, when I was messing with it, it was also already taken, so just naively spacing it apart more.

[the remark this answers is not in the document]
anniehu4: oh lol 😅

packages/components/src/util/typography.tsx
anniehu4: oh as in there's no corresponding `.spacingNone` style in typography.module.css now

PR #409, opened 2021-04-12
Remove styled-components and rename util/ folder

anniehu4:
### Summary:
[ch128860](https://app.clubhouse.io/czi-edu/story/128860/cleanup-old-imports-after-css-modules-switch)

For uninstalls I ran:
```
// in packages/components
npm uninstall --save-dev @types/styled-components jest-styled-components
npm uninstall --save styled-components
// in root
npm uninstall --save-dev stylelint-config-styled-components
```

Otherwise, just renamed `src/util` to `src/common`

### Test Plan:
build passes


packages/components/.storybook/Button/button.stories.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from "react";
 
-import Clickable, { ClickableProps } from "../../src/util/clickable";
+import Clickable, { ClickableProps } from "../../src/common/clickable";
```
anniehu4, line 3: These are no longer just "utils" -- also considered "base" or "shared"

packages/components/src/styles/global.css
```diff
@@ -1,8 +1,7 @@
-
```
anniehu4, line 1: prettier changes -- think `stylelint-config-styled-components` was suppressing some of these rules

ahuth (approved):
LGTM 🤘

PR #410, opened 2021-04-13
docs(intellisense): added intellisense for css modules in typescript

dcwither:
### Summary:
Adds intellisense to our css modules to help avoid hard to find typos, or cascade issues from classname changes. Sadly this doesn't help type safety directly. There are alternatives like [typed-css-modules](https://www.npmjs.com/package/typed-css-modules) that do, but I couldn't find a good fit into our process for it.

### Test Plan:
image


packages/components/css-modules.d.ts
```diff
@@ -1,4 +1,4 @@
 declare module "*.module.css" {
-  const content: Record<string, string>;
-  export default content;
+  const classes: Record<string, string>;
+  export default classes;
```
anniehu4, line 3: curious, was this needed to work with the plugin?

anniehu4 (approved):

packages/components/css-modules.d.ts
dcwither: It wasn't, I just decided to go with something consistent with what was in their docs.

ahuth (commented):
Nice, LGTM

ahuth (approved):

PR #425, opened 2021-04-22
feat(typography): allow classname pass through on text components

dcwither:
### Summary:
Allow classname pass through for text. Although this will probably lead to some style overriding in our projects, there's too many valid use cases that we want to support to encode into the text and heading components.

e.g. line clamping, responsive margins, etc.

### Test Plan:
Added test for Text component with pass through. It felt unnecessary to add another story since this doesn't create an immediately visual change.

### DIFF
--- a/packages/components/src/Heading/Heading.tsx
+++ b/packages/components/src/Heading/Heading.tsx
@@ -14,6 +14,7 @@ type Props = {
    */
   as?: HeadingElement;
   children: TypographyProps<HeadingElement>["children"];
+  className: TypographyProps<HeadingElement>["className"];
   color?: TypographyProps<HeadingElement>["color"];
   size: TypographyProps<HeadingElement>["size"];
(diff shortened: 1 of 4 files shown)


packages/components/src/Heading/Heading.tsx
```diff
@@ -14,6 +14,7 @@ type Props = {
    */
   as?: HeadingElement;
   children: TypographyProps<HeadingElement>["children"];
+  className: TypographyProps<HeadingElement>["className"];
```
lesliecodes, line 17: should this be optional?

lesliecodes (approved):
one comment, otherwise LGTM

packages/components/src/Heading/Heading.tsx
dcwither: yes, but at the typography level (which is also missing), this will inherit that property. I'll add that, nice catch 👍

PR #450, opened 2021-06-02
feat(typography): allow color passthrough on text components

liza8bit:
### Summary:
TLDR: This is not a finished PR, just looking for feedback on the problem/possible solutions re: inheriting color for typography components

#### The Problem:
When working on the Banner EDS candidate (https://github.com/FB-PLP/traject/pull/6758), one of the suggestions was to pull out a `Banner.Title` component that ensures the `Heading` component styling stays consistent:

```
Banner.Title = function BannerTitle(props: {|
  children?: React$Node,
  as: string,
|}) {
  return (
    <Heading as={props.as} size="h3">
      {props.children}
    </Heading>
  );
};
```

and usage would look something like

```
<Banner>
  <Banner.Title> Title Content </Banner.Title>
  <Banner.Body> Body Content </Banner.Body> 
</Banner>
```

But Heading and Text have a default color value `base`, so the result is that the banner color was being overriden:
image

#### This PR:
This PR is an attempt to fix that by setting up Typography components to be able to "inherit" color, similar to SVGIcon (eg in the above screenshot, the icon gets the banner's color)

I say attempt because if the default is `currentColor` or `color: inherit`, and the parent doesn't have a color set, then it just uses black, which isn't actually what I want.

The ideal is "if the parent has a color, use that, but if not, use `base`", but I can't seem to figure out a way to do that. CSS variables can have a fallback, but the this doesn't seem to exist for `currentColor` or `inherit`. 

#### Other Options:
* Go back to having heading be a prop and the content being plain text, so we can set the color within the component
* When using `Banner.Title` and `Banner.Body`, you have to (redundantly) set the color scheme - pretty sure we dont want this, since it could lead to inconsistent looking banners
* Slightly wacky but we could [use cloneElement to inject the color elements into the children? ](https://stackoverflow.com/questions/50786382/how-to-safely-inject-props-into-react-children)(I can't tell if this is too 😬  or not) 

### Test Plan:
TBD

### DIFF
--- a/packages/components/src/common/typography.module.css
+++ b/packages/components/src/common/typography.module.css
@@ -59,6 +59,10 @@
   @apply text-info-700;
 }
 
+.colorInherit {
+  color: inherit;
+}
+
 .colorNeutral {
   @apply text-neutral-500;
 }
@@ -92,7 +96,7 @@
  * font-weights
  * allows override for text components, shouldn't be used by heading components
  **/
- 
+
 .weightNormal {
   @apply font-normal;
 }
--- a/packages/components/src/Text/Text.stories.tsx
+++ b/packages/components/src/Text/Text.stories.tsx
@@ -56,3 +56,15 @@ BodyColorInfoBold.args = {
   size: "body",
   weight: "bold",
 };
+
+export const TextColorInherit = (args) => (
+  <Text color="alert" size="body">
+    This text surrounds the <Text {...args} /> and shows it should inherit color
+    from the parent
+  </Text>
+);
+TextColorInherit.args = {
+  children: "Child Text",
+  color: "inherit",
+  size: "body",
+};
--- a/packages/components/src/Text/__snapshots__/Text.spec.tsx.snap
+++ b/packages/components/src/Text/__snapshots__/Text.spec.tsx.snap
@@ -47,3 +47,17 @@ exports[`<Text /> Overline story renders snapshot 1`] = `
   Overline 12/20
 </p>
 `;
+
+exports[`<Text /> TextColorInherit story renders snapshot 1`] = `
+<p
+  class="typography sizeBody colorAlert"
+>
+  This text surrounds the 
+  <p
+    class="typography sizeBody colorInherit"
+  >
+    Child Text
+  </p>
+   and shows it should inherit color from the parent
+</p>
+`;
--- a/packages/components/src/common/typography.tsx
+++ b/packages/components/src/common/typography.tsx
@@ -20,6 +20,7 @@ export type TypographyColor =
   | "base"
   | "brand"
   | "info"
+  | "inherit"
   | "neutral"
   | "success"
   | "warning"
@@ -99,6 +100,7 @@ function Typography<IComponent extends React.ElementType>({
         color === "base" && styles.colorBase,
         color === "brand" && styles.colorBrand,
         color === "info" && styles.colorInfo,
+        color === "inherit" && styles.colorInherit,
         color === "neutral" && styles.colorNeutral,
         color === "success" && styles.colorSuccess,
         color === "warning" && styles.colorWarning,


packages/components/src/common/typography.module.css
```diff
@@ -55,6 +55,10 @@
   @apply text-brand-700;
 }
 
+.colorCurrent {
+  color: currentColor;
```
liza8bit, line 59: `text-color` [should be a thing ](https://tailwindcss.com/docs/text-color#class-reference) but tailwind didnt seem to recognize it? 🤔

packages/components/src/Heading/Heading.stories.tsx
```diff
@@ -55,3 +55,10 @@ Heading4ColorNeutral.args = {
   color: "neutral",
   size: "h4",
 };
+
+export const Heading4ColorCurrent = Template.bind(null);
+Heading4ColorCurrent.args = {
+  children: "Current color Heading 4",
+  color: "currentColor",
```
liza8bit, line 62: this isnt really testing anything/not sure that we really need a story, but this was just so i could see a) that currentColor was being set correctly and b) you can see the color difference when you switch to color being `base` :(

liza8bit:
Another thought- maybe `base` could be left as the default, and then the cloneElement solution could be used to get the components to inherit the right color from the banner? (I'm thinking if just passing the color prop through cloneElement could still end up with a different color than what was intended in the context of the banner)

anniehu4 (commented):
Thank you for doing this! We had thought about allowing color&size inheritance but weren't sure without a use case yet

Re:
> The ideal is "if the parent has a color, use that, but if not, use base"

I wonder if this would work:
```
.typography {
  @apply text-font-base;
}

.colorInherit {
  color: inherit; // (I'd prefer inherit as I think it's more precise to this use case)
}
```
When the CSS module compiles, the `.typography` class styles get applied before the color styles (based on where they were declared in the file). see:
Screen Shot 2021-06-03 at 7 59 30 PM

Another potential approach is not setting a default value on the `color` prop. That seems fine b/c I'm struggling to think of cases where the parent wouldn't have a color (i.e. wouldn't text always at least inherit from the `<body>` color?)

anniehu4:
Separately, having `inherit` as the default feels like more intuitive behavior as a consumer. But we should double-check whether that's fine for the existing usages of `<Text>` in LP

anniehu4:
Separately, thinking about whether `"initial"` or `"base"` should be the default. I think the former is more intuitive behavior, but the latter forces more consistency / thoughtfulness of styling

liza8bit:
Hm yeah I could see an argument for either- in this case specifically, `base` being the default would still require manually setting `initial` on both `Banner.Title` and `Banner.Body`, so dev experience wise it would pretty similar to setting the same color on all of the Banner, Title, and Body- though you wouldn't have to know which color goes with which 🤔 

Setting the color later in the file doesn't seem to work; as long as `inherit` evaluates to a real color, it'll use that/override the base color, even if it just evaluates to black :') 
image
image


I updated https://github.com/FB-PLP/traject/pull/6758/files with the 'clone element' solution in the meantime (just trying to mock out what the final result would look like), if seeing that example is helpful :O

anniehu4:
> though you wouldn't have to know which color goes with which

Honestly, this is the main thing I was thinking about -- if the banner designs are consistent in color, we may as well try to make that the default behavior. The banner can always use color="inherit" so users don't have to set it

> as long as inherit evaluates to a real color, it'll use that/override the base color, even if it just evaluates to black :')

this miiight be something that seems worse in Storybook than it actually would be in an app 🤔  In `.storybook/preview-body.html`, if you add:
```
body {
  color: var(--eds-color-font-base);
}
```
that's what Storybook will inherit to instead of black. We can (and should) add similar styling to traject anyways. Though if the default value is still "base", people should be thoughtful about setting "inherit"

anniehu4:
> though you wouldn't have to know which color goes with which

Honestly, this is the main thing I was thinking about -- if the banner designs are consistent in color, we may as well try to make that the default behavior. The banner can always use color="inherit" so users don't have to set it

> as long as inherit evaluates to a real color, it'll use that/override the base color, even if it just evaluates to black :')

this miiight be something that seems worse in Storybook than it actually would be in an app 🤔  In `.storybook/preview-body.html`, if you add:
```
body {
  color: var(--eds-color-font-base);
}
```
that's what Storybook will inherit to instead of black. We can (and should) add similar styling to traject anyways. Though if the default value is still "base", people should be thoughtful about setting "inherit"

packages/components/src/common/typography.module.css
ahuth: Weird. Maybe our color customization clobbers text-transparent and text-color?

anniehu4: yeah, it only allows the EDS color tokens right now and 'transparent'. We can add currentColor though:
```
// in tailwind.config.js
module.exports = {
  theme: {
    colors: {
      current: 'currentColor',
      ...
```

liza8bit:
I went back to `base` as the default for now :o feels like the consistency/thoughtfulness would be good, also not sure if anyone is relying on that default for styling rn so didn't want to shake the boat too much 😅

anniehu4:
#452 should be landed today or tomorrow, and I think it'd be nice to include it in the next release since it's also typography related. If you need it sooner though lmk and I can do a release for this one

Issue #295, opened 2021-02-12
Heading and Text should forward ref


anniehu4:
## Request summary
Change Heading & Text to use `React.forwardRef`. 2 other options:
- accept a `ref` prop (a bit more annoying for end user b/c they have to type a unique prop name instead of just passing `ref`)
- use forwardRef with [`useImperativeHandle`](https://reactjs.org/docs/hooks-reference.html#useimperativehandle). @dcwither suggested this but I still don't fully understand it / whether it provides extra benefit for these type components

## Relevant component(s)
Heading, Text

## Example usage
Heading from traject:
```
<Heading
  headingRef={firstAction}
  onClick={() => setExpanded(prev => !prev)}
  size="h1"
  tabIndex={-1}
>
```

anniehu4:
cc @dcwither @ahuth

ahuth:
Woah, what's `useImperativeHandle` for? I'm not familiar with that one. Shouldn't be necessary to forward a ref to the component.

dcwither:
I don't think forwardRef is necessary for this particular component though I think you had mentioned a scenario where people were trying to `focus()` on the `Heading`s in modals etc. in which case we should explore how much they need that vs. a possible work around.

I think it's more necessary for things like buttons etc. since they will need imperative function access like `focus()`.

ahuth:
There are some use cases for focusing on a heading element.

For example, it's sort of ambiguous (as far as I know) whether modals should focus on the first focusable element or the heading element (although we tend to do the first focusable element in LP, which seems reasonable).

Also, LP is currently focusing on a heading in 1 case (that I know of) to work around limitations of the react-bootstrap button (we can't get a ref to the button in the version that we're on). So we may need this to have a smooth rollout.

I think it makes sense for heading (although maybe not with useImperitiveHandle. Still don't know what that does). And it could make sense to expose it for all "atom" components. Although that's a discussion for another time, perhaps.

anniehu4:
yeah, this is the PR for context: https://github.com/FB-PLP/traject/pull/5025/files. It does seem we could remove this specific case once there's a non-bootstrap button...though I can imagine (because of modals, etc.) this might come up again for Headings, so I think it's worth adding

dcwither:
[...]

PR #452, opened 2021-06-07
feat(typography): allow ref pass through on Text components

dierat:
### Summary:
Allow consumers of the `Text` componnent to pass down a `ref` to access the underlying DOM
element directly.

GitHub issue:
https://github.com/chanzuckerberg/edu-design-system/issues/295

Clubhouse story:
https://app.clubhouse.io/czi-edu/story/141247/add-forward-ref-to-heading-text-components

### Test Plan:
I added a unit test that uses a ref to focus on the text element produced by a <Text/>
element to verify the ref is correctly being passed down.

### DIFF
--- a/packages/components/src/Text/Text.spec.tsx
+++ b/packages/components/src/Text/Text.spec.tsx
@@ -24,7 +24,7 @@ describe("<Text />", () => {
     `);
   });
 
-  it("should pass the passthrough ref", async () => {
+  it("should pass the passthrough ref", () => {
     const textRef = React.createRef();
     render(
       <Text ref={textRef} size="body" tabIndex={-1}>
@@ -34,7 +34,7 @@ describe("<Text />", () => {
 
     textRef.current.focus();
 
-    const textElement = await screen.getByText("Some Text");
+    const textElement = screen.getByText("Some Text");
     expect(textElement).toEqual(document.activeElement);
   });
 });
--- a/packages/components/src/Text/Text.spec.tsx
+++ b/packages/components/src/Text/Text.spec.tsx
@@ -1,6 +1,6 @@
 import * as TextStoryFile from "./Text.stories";
 
-import React, { useEffect, useRef } from "react";
+import React, { useEffect } from "react";
 
 import { render, screen } from "@testing-library/react";
 import Text from "./Text";
@@ -25,22 +25,16 @@ describe("<Text />", () => {
   });
 
   it("should pass the passthrough ref", async () => {
-    const TestComponent = () => {
-      const textRef = useRef<HTMLParagraphElement>(null);
+    const textRef = React.createRef();
+    render(
+      <Text ref={textRef} size="body" tabIndex={-1}>
+        Some Text
+      </Text>,
+    );
 
-      useEffect(() => {
-        textRef.current && textRef.current.focus();
-      });
+    textRef.current.focus();
 
-      return (
-        <Text ref={textRef} size="body" className="passthrough" tabIndex={-1}>
-          Some Text
-        </Text>
-      );
-    };
-    render(<TestComponent />);
     const textElement = await screen.getByText("Some Text");
-
     expect(textElement).toEqual(document.activeElement);
   });
 });
--- a/packages/components/src/Text/Text.spec.tsx
+++ b/packages/components/src/Text/Text.spec.tsx
@@ -1,8 +1,8 @@
 import * as TextStoryFile from "./Text.stories";
 
-import { render, screen } from "@testing-library/react";
+import React, { useEffect, useRef } from "react";
 
-import React from "react";
+import { render, screen } from "@testing-library/react";
 import Text from "./Text";
 import { generateSnapshots } from "@chanzuckerberg/story-utils";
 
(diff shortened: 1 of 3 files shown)


packages/components/src/Text/Text.spec.tsx
```diff
@@ -23,4 +23,24 @@ describe("<Text />", () => {
...
+        textRef.current && textRef.current.focus();
+      });
+
+      return (
+        <Text ref={textRef} size="body" className="passthrough" tabIndex={-1}>
+          Some Text
+        </Text>
+      );
+    };
+    render(<TestComponent />);
```
ahuth, line 41: I think it would be simpler and clearer to test using `React.createRef` instead of `React.useRef` (both of which return the same data structure.

Would look something like

```jsx
const textRef = React.createRef();
render(<Test ref={textRef} {...the other props} />);

textRef.current.focus();

const textElement = await screen.getByText("Some Text");
expect(textElement).toEqual(document.activeElement);
```

It's less hoops to jump through, and it's easier to control when the element is focused.

packages/components/src/Text/Text.spec.tsx
```diff
@@ -23,4 +23,24 @@ describe("<Text />", () => {
       </p>
     `);
   });
+
+  it("should pass the passthrough ref", async () => {
```
ahuth, line 27: I'm a big fan of tests. Wondering if it's more trouble than it's worth in this case, though?

ahuth (approved):
LGTM

packages/components/src/Text/Text.tsx
```diff
@@ -17,22 +17,17 @@ type Props = {
   spacing?: TypographyProps<TextElement>["spacing"];
 };
 
-function Text({
-  as,
-  children,
-  /**
+const Text = forwardRef<HTMLElement, Props>(({ as, children, /**
```
anniehu4, line 20: Thanks! could you also add this to `Heading.tsx`?

packages/components/src/Text/Text.spec.tsx
anniehu4: +1. I think this suggestion makes the test lightweight enough & worth it

ahuth: Also just remembered there's a `toHaveFocus` matcher coming from jest-dom that we could use here.

packages/components/src/Text/Text.spec.tsx
```diff
@@ -23,4 +23,24 @@ describe("<Text />", () => {
...
+      });
+
+      return (
+        <Text ref={textRef} size="body" className="passthrough" tabIndex={-1}>
+          Some Text
+        </Text>
+      );
+    };
+    render(<TestComponent />);
+    const textElement = await screen.getByText("Some Text");
```
ahuth, line 42: `await` only works with `findBy*` queries. We should make this either

```suggestion
    const textElement = await screen.findByText("Some Text");
```

or

```suggestion
    const textElement = screen.getByText("Some Text");
```

We should also definitely get the [testing library eslint plugin](https://github.com/testing-library/eslint-plugin-testing-library) turned on sometime.

packages/components/src/Text/Text.spec.tsx
dierat: I mean, I can remove it, but I didn't like if this were somehow to break, we could be causing accessibility issues where we're trying to focus on a modal or whatever and no one notices. Is there any reason to not have it when it's already here?

packages/components/src/Text/Text.spec.tsx
dierat: That's so much better – thank you!

For the `toHaveFocus`, I'm hitting an error that said `expect(...).toHaveFocus is not a function`, which is why I used this roundabout method.

packages/components/src/Text/Text.spec.tsx
```diff
@@ -1,8 +1,8 @@
 import * as TextStoryFile from "./Text.stories";
 
-import { render, screen } from "@testing-library/react";
+import React, { useEffect } from "react";
```
anniehu4, line 3: huh, weird that this isn't an error 🤔

anniehu4 (approved):

packages/components/src/Text/Text.spec.tsx
dierat: Oops! Yeah, it woulda been nice if that had come up as an error; I didn't notice it.

dierat: Fixed here: https://github.com/chanzuckerberg/edu-design-system/pull/454

packages/components/src/Text/Text.spec.tsx
ahuth: Nope, no reason to remove it since it's already there.

To me, it's not always valuable things we technically should, but are mostly about wiring things up (such as passing a ref through forward ref to the component). But that's my 1.5 cents, and this is a reasonable test 👍. Especially since you find value in it.

packages/components/src/Text/Text.spec.tsx
ahuth: Added linter in #455

PR #455, opened 2021-06-09
Add eslint-plugin-testing-library

ahuth:
### Summary:

This PR adds the eslint plugin for testing library.

### Test Plan:

- CI
- Added an error intentionally and ran linters locally

### DIFF
--- a/packages/components/src/Text/Text.spec.tsx
+++ b/packages/components/src/Text/Text.spec.tsx
@@ -34,6 +34,7 @@ describe("<Text />", () => {
     textRef.current.focus();
 
     const textElement = screen.getByText("Some Text");
+    // eslint-disable-next-line testing-library/no-node-access
     expect(textElement).toEqual(document.activeElement);
   });
 });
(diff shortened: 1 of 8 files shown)


packages/components/src/Text/Text.spec.tsx
```diff
@@ -34,6 +34,7 @@ describe("<Text />", () => {
     textRef.current.focus();
 
     const textElement = screen.getByText("Some Text");
+    // eslint-disable-next-line testing-library/no-node-access
```
ahuth, line 37: The new eslint rules don't like directly accessing `document.activeElement`, it seems 🤔 

We could disable the rule. Or upgrade testing library and get access to `toHaveFocus`.

anniehu4: Or is the issue that we don't have `@testing-library/jest-dom` installed?

anniehu4 (approved):

packages/components/src/Text/Text.spec.tsx
ahuth: That's probably it. I assumed we had that installed. But if not, that's definitely the problem.

dierat (approved):

ahuth:
Thanks!

PR #486, opened 2021-06-23
chore: add display: flex to Heading

dierat:
### Summary:
We're currently migrating the headings in `traject` to use the eds `Heading` component. In the process, we found a bug where icons are being pushed to a separate line than the text. https://app.clubhouse.io/czi-edu/story/144708/icon-truncated-into-second-line-and-hover-on-it-tooltip-position-shows-at-wrong-place-while-navigating-to-mentee-profile

To fix this, we either need to give all `block` elements in headings `display: inline-block` or we add `display: flex` to the `Heading` styles. I figure that the latter makes more sense because we will, I think, always want `Heading` children to sit next to each other (unless they're wrapping due to space constraints).

### Test Plan:
Verify all the text in the `Heading` story `HeadingWithMultipleDivs` is on one line.
screenshot of HeadingWithMultipleDivs story

### DIFF
--- a/packages/components/src/Heading/Heading.module.css
+++ b/packages/components/src/Heading/Heading.module.css
@@ -1,0 +1,3 @@
+.heading {
+  @apply flex;
+}
--- a/packages/components/src/Heading/Heading.stories.tsx
+++ b/packages/components/src/Heading/Heading.stories.tsx
@@ -55,3 +55,13 @@ Heading4ColorNeutral.args = {
   color: "neutral",
   size: "h4",
 };
+
+export const HeadingWithMultipleDivs = Template.bind(null);
+HeadingWithMultipleDivs.args = {
+  size: "h3",
+  children: (
+    <>
+      <div>Heading in a Div</div>&nbsp;<div>Followed by Another Div</div>
+    </>
+  ),
+};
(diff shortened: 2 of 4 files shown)


dierat:
https://app.clubhouse.io/czi-edu/story/144708/icon-truncated-into-second-line-and-hover-on-it-tooltip-position-shows-at-wrong-place-while-navigating-to-mentee-profile

anniehu4:
hmm I'm not convinced that we always want this behavior in Heading 🤔  I think with the source bug, the issue is the styling is targeting the legacy heading className `.app-heading`:

Screen Shot 2021-06-24 at 2 48 54 PM

seems rare enough that it's more predictable to make callsites add the styling themselves if they need it?

dierat:
> hmm I'm not convinced that we always want this behavior in Heading 🤔 I think with the source bug, the issue is the styling is targeting the legacy heading className `.app-heading`:
> 
> Screen Shot 2021-06-24 at 2 48 54 PM
> 
> seems rare enough that it's more predictable to make callsites add the styling themselves if they need it?

I find it difficult to image a situation where we'd want a heading to wrap like this, but I'm also fine with putting it on the specific heading instance instead. 👍

dierat:
Alternate PR: https://github.com/FB-PLP/traject/pull/7109

anniehu4:
I think it could make sense to add to a higher-level "Header" or "SectionHeader"-like component, but I haven't seen it added to a heading / typography component before

dierat:
Fixed in https://github.com/FB-PLP/traject/pull/7109

PR #487, opened 2021-06-24
Add grading-yellow tokens and update storybook

anniehu4:
### Summary:
For the alpha UI kit, we're planning on keeping two of the "legacy" yellow colors, so we (Sean, Ashley, & I) settled on adding them as `eds-color-grading-yellow-` tokens. [See colors in Figma](https://www.figma.com/file/P9VPyI821gLq832tuU2WeY/UI-Kit-(Alpha)?node-id=2321%3A5169). 

I also removed the SCSS-style `$` since we're encouraging CSS modules

### Test Plan:
Confirmed new yellows are in Storybook, above the warning colors:
Screen Shot 2021-06-24 at 2 40 05 PM

Confirmed extra notes are there:
Screen Shot 2021-06-24 at 2 40 10 PM


packages/tokens/properties/color/base.json
```diff
@@ -1,63 +1,4 @@
 {
-  "legacy": {
```
anniehu4, line 2: I just moved these below the `eds` tokens, which explains some of the changes in `tokens/dist/`

packages/tokens/properties/color/base.json
```diff
@@ -116,6 +57,12 @@
         "600": { "value": "#574EFF" },
         "700": { "value": "#433BDE" }
       },
+      "grading": {
+        "yellow": {
+          "100": { "value": "#FDF1D0" },
+          "500": { "value": "#FFC55B" }
+        }
+      },
```
anniehu4, line 65: added these two

packages/components/.storybook/tokens.stories.mdx
```diff
@@ -12,28 +12,41 @@ import styles from "./tokens.module.css";
 ## EDS Colors
 
 export const mapEDSColorToTokenArray = (color) => {
-  return Object.keys(tokens.eds.color[color])
+  // Handle nested color names
+  const colorKeys = color.split("-");
+  const colorTokens = colorKeys.reduce((obj, key) => {
+    return obj[key];
+  }, tokens.eds.color);
+  return Object.keys(colorTokens)
```
dierat, line 20: Gonna be honest – it took me a while to figure out what you were doing in this reduce. I feel like there should be a pattern that's a little easier to reason about for this situation. At this point, it looks to me like you could just use the last value in the `colorKeys` array like
```
const colorToUse = colorKeys[colorKeys.length - 1];
return Object.keys(tokens.eds.color[colorToUse])
```
or something?

ahuth: Could also extract some functions or utils (not sure how that works with mdx)

```js
const colorTokens = getTokensForColor(color);
return Object.keys(colorTokens);
```

packages/components/.storybook/tokens.stories.mdx
```diff
@@ -12,28 +12,41 @@ import styles from "./tokens.module.css";
...
     .sort((shade1, shade2) => shade1.localeCompare(shade2))
     .map((shade) => ({
-      name: `$eds-color-${color}-${shade}`,
```
ahuth, line 18: Was the `$` just being added before because we previously wanted to document the scss variables, instead of the css variables?

anniehu4: yep!

packages/components/.storybook/tokens.stories.mdx
anniehu4: the tokens file (`variables-nested.json`) looks like:
```
"brand": {
    ...
    "700": "#433BDE"
},
"grading": {
    "yellow": {
        "100": "#FDF1D0",
        "500": "#FFC55B"
    }
 },
```

so the key `a-b` means we need to access `tokens[a][b]`. It's easier if I assume there's max 1 level of nesting that's going to happen, but thought it _might_ be useful to generalize.

Can def extract functions though (and maybe add comments? haven't tried that)

anniehu4: could be a hasty abstraction though, to handle just 1 extra level of nesting it could be
```
let colorTokens = tokens.eds.color[colorKeys[0]];
if colorKeys.length === 2 {
    colorTokens = colorTokens[colorKeys[1]]
}
```

ahuth (approved):

packages/components/.storybook/tokens.stories.mdx
```diff
@@ -11,29 +11,54 @@ import styles from "./tokens.module.css";
...
+
+export const getTokensForColor = (color) => {
+  const maybeNestedKeys = color.split("-");
+  // Each "-" separator corresponds to an extra level of nesting in variables-nested.json
+  // For example, the tokens for color "grading-yellow" are stored under
+  // "grading": {
+  //   "yellow": { ...tokens }
+  // }
+  // This will index in another level for each colorKey separated by "-"
+  // and return the innermost object with the color tokens
```
dierat, line 24: This helps, thanks!

dierat (approved):

anniehu4:
thanks all!

PR #493, opened 2021-07-06
feat(button): add warning color and display flex

dierat:
Clubhouse story: https://app.clubhouse.io/czi-edu/story/143864/buttons

### Summary:
This PR adds a couple of small updates to the `Button` component to prepare it for launch:
- re-adds the "warning" color (because we'll need it in the banners)
- adds `display: "inline-flex"` to the button (so icons can be used inside)
- adds 2 stories with a test icon to verify the `flex` change is working as expected
- adds 1 story with a `data-test-id` attribute to verify other important attributes not
  explicitly exposed will work with the component

### Test Plan:
`npm start` to run storybook,
and navigate to the `Button` stories.

Then:
- inspect the `With Data Test Id` button and verify the button element has a `data-test-id` with a value of "fake-test-id"
- verify the `Link With Icon` and `Outline With Icon` stories have the icon on the same line as the text
- verify the `All Variants` stories have a new `warning` color set that matches the pattern established by the other color variants
- verify that none of the other story outputs have been altered

### Screenshots
new-stories
warning-color

### DIFF
--- a/packages/components/src/common/clickable.tsx
+++ b/packages/components/src/common/clickable.tsx
@@ -10,7 +10,7 @@ export type ClickableProps<IComponent extends React.ElementType> = {
   /**
    * The color of the element.
    */
-  color: "alert" | "brand" | "neutral" | "success";
+  color: "alert" | "brand" | "neutral" | "success" | "warning";
   /**
    * The size of the element.
    */
@@ -54,8 +54,8 @@ function Clickable<IComponent extends React.ElementType>({
         color === "brand" && styles.colorBrand,
         color === "neutral" && styles.colorNeutral,
         color === "success" && styles.colorSuccess,
+        color === "warning" && styles.colorWarning,
         // Interactive States (for testing)
-        state === "inactive" && styles.stateInactive,
         state === "hover" && styles.stateHover,
         state === "focus" && styles.stateFocus,
         state === "active" && styles.stateActive,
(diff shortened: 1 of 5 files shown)


anniehu4:
yay! I'll take a real review a little later, but for now have a nit that it's easier to get context on feature commits if we also include the component name. something like `feat(button):` vs. `feat(typography):`

dierat:
> yay! I'll take a real review a little later, but for now have a nit that it's easier to get context on feature commits if we also include the component name. something like `feat(button):` vs. `feat(typography):`

Updated, thanks!

ahuth (approved):

packages/components/src/common/clickable.tsx
```diff
@@ -54,8 +54,8 @@ function Clickable<IComponent extends React.ElementType>({
         color === "brand" && styles.colorBrand,
         color === "neutral" && styles.colorNeutral,
         color === "success" && styles.colorSuccess,
+        color === "warning" && styles.colorWarning,
         // Interactive States (for testing)
-        state === "inactive" && styles.stateInactive,
```
anniehu4, line 58: intentional deletion?

anniehu4 (approved):

packages/components/src/common/clickable.tsx
dierat: Yes! Sorry, I meant to leave a comment here. I couldn't find any references to `stateInactive` outside of this line.

PR #494, opened 2021-07-06
chore(button): specify HTML attr styles, use className on Clickable

dierat:
### Summary:
This PR accomplishes 2 tasks:
- explicitly allow HTML button attributes on the `Button` props
- use the `className` prop if it's passed into the `Clickable` component

The `Clickable` component already has `className` in the types, and we have a test to verify that it's being applied to the element, but it's currently **removing** the default styles we're adding to the element. This change adds the `className`, if passed in, to the styles list so it exists so they are both applied.

Question: should the `className` be applied at the top or bottom of the stack? I put it at the top so our default styles win out, which I think makes sense for highly opinionated design system components.

### Test Plan:
It looks like type checking is currently not running on story files, so, to test this, I added a little test component to the bottom of `button.tsx` and ran `npx lerna run build:types` in my terminal.

This test button should not raise any flow errors:
```
export const TestButton = (
  <Button
    data-testid="fake-data-test-id"
    role="alert"
    aria-live="polite"
    tabIndex={-1} 
  >
    Test button
  </Button>
);
```

For testing `className`, verify that the new story renders like a normal button, has the fake class name, and also has the usual class names.

### DIFF
--- a/packages/components/src/Button/__snapshots__/button.spec.tsx.snap
+++ b/packages/components/src/Button/__snapshots__/button.spec.tsx.snap
@@ -103,10 +103,10 @@ exports[`<Button /> outlineWithIcon story renders snapshot 1`] = `
 exports[`<Button /> withDataTestId story renders snapshot 1`] = `
 <button
   class="button sizeMedium variantFlat colorAlert"
-  data-test-id="fake-test-id"
+  data-testid="fake-test-id"
   type="button"
 >
   ​
-  Button with data-test-id
+  Button with data-testid
 </button>
 `;
--- a/packages/components/src/Button/__snapshots__/button.spec.tsx.snap
+++ b/packages/components/src/Button/__snapshots__/button.spec.tsx.snap
@@ -110,3 +110,13 @@ exports[`<Button /> withDataTestId story renders snapshot 1`] = `
   Button with data-testid
 </button>
 `;
+
+exports[`<Button /> withFakeClassName story renders snapshot 1`] = `
+<button
+  class="fake-className button sizeMedium variantOutline colorWarning"
+  type="button"
+>
+  ​
+  With fake className
+</button>
+`;
--- a/packages/components/src/Button/button.tsx
+++ b/packages/components/src/Button/button.tsx
@@ -3,7 +3,7 @@ import React, { ReactNode } from "react";
 
 type ButtonHTMLElementProps = React.ButtonHTMLAttributes<HTMLButtonElement>;
 
-type ButtonPropsBase = ButtonHTMLElementProps & {
+export type ButtonProps = ButtonHTMLElementProps & {
   as?: "button" | React.ComponentType<any>;
   /**
    * The button contents or label.
@@ -13,8 +13,6 @@ type ButtonPropsBase = ButtonHTMLElementProps & {
   size?: ClickableProps<"button">["size"];
   variant?: ClickableProps<"button">["variant"];
 };
-
-export type ButtonProps = Omit<ButtonPropsBase, "style">;
 
 function Button({
   as = "button",
--- a/packages/components/src/Button/button.tsx
+++ b/packages/components/src/Button/button.tsx
@@ -3,18 +3,18 @@ import React, { ReactNode } from "react";
 
 type ButtonHTMLElementProps = React.ButtonHTMLAttributes<HTMLButtonElement>;
 
-export type ButtonProps = {
+type ButtonPropsBase = ButtonHTMLElementProps & {
   as?: "button" | React.ComponentType<any>;
   /**
    * The button contents or label.
    */
   children: ReactNode;
   color?: ClickableProps<"button">["color"];
-  disabled?: ButtonHTMLElementProps["disabled"];
   size?: ClickableProps<"button">["size"];
-  type?: ButtonHTMLElementProps["type"];
   variant?: ClickableProps<"button">["variant"];
 };
+
+export type ButtonProps = Omit<ButtonPropsBase, "style">;
 
 function Button({
   as = "button",
--- a/packages/components/src/Clickable/Clickable.tsx
+++ b/packages/components/src/Clickable/Clickable.tsx
@@ -37,12 +37,14 @@ function Clickable<IComponent extends React.ElementType>({
   size,
   state,
   variant,
+  className,
   ...rest
 }: ClickableProps<IComponent>) {
   const Component = as;
   return (
     <Component
       className={clsx(
+        className,
         styles.button,
         // Sizes
         variant !== "link" && [
--- a/packages/components/src/Button/button.stories.tsx
+++ b/packages/components/src/Button/button.stories.tsx
@@ -71,10 +71,10 @@ linkInHeading.args = {
 
 export const withDataTestId = Template.bind(null);
 withDataTestId.args = {
-  children: "Button with data-test-id",
+  children: "Button with data-testid",
   color: "alert",
   variant: "flat",
-  "data-test-id": "fake-test-id",
+  "data-testid": "fake-test-id",
 };
 
 export const linkWithIcon = Template.bind(null);
--- a/packages/components/src/Button/button.stories.tsx
+++ b/packages/components/src/Button/button.stories.tsx
@@ -90,3 +90,11 @@ outlineWithIcon.args = {
   color: "warning",
   variant: "outline",
 };
+
+export const withFakeClassName = Template.bind(null);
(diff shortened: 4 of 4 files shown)


packages/components/src/Button/__snapshots__/button.spec.tsx.snap
```diff
@@ -103,10 +103,21 @@ exports[`<Button /> outlineWithIcon story renders snapshot 1`] = `
 exports[`<Button /> withDataTestId story renders snapshot 1`] = `
 <button
   class="button sizeMedium variantFlat colorAlert"
-  data-test-id="fake-test-id"
```
dierat, line 106: I always make this typo.

packages/components/src/Button/button.tsx
```diff
@@ -3,19 +3,19 @@ import React, { ReactNode } from "react";
...
 
-export type ButtonProps = {
+type ButtonPropsBase = ButtonHTMLElementProps & {
   as?: "button" | React.ComponentType<any>;
   /**
    * The button contents or label.
    */
   children: ReactNode;
   color?: ClickableProps<"button">["color"];
-  disabled?: ButtonHTMLElementProps["disabled"];
```
dierat, line 13: I removed `disabled` and `type` from this list since we're now using everything in `ButtonHTMLElementProps`. Without combining `ButtonHTMLElementProps`, the other attributes we want (`data-testid`, `role`, `aria-live`, `tabIndex`) freak flow out.

packages/components/src/Button/__snapshots__/button.spec.tsx.snap
ahuth: We could use a lint rule for this. Helps align puppeteer and testing-library usage, as well (since puppeteer or the like don't really care what data-* attributes we use, but testing-library does).

packages/components/src/Button/button.tsx
```diff
@@ -3,19 +3,19 @@ import React, { ReactNode } from "react";
...
   size?: ClickableProps<"button">["size"];
-  type?: ButtonHTMLElementProps["type"];
   variant?: ClickableProps<"button">["variant"];
 };
 
+export type ButtonProps = Omit<ButtonPropsBase, "style">;
```
ahuth, line 17: We do we want to prevent people from using the style attribute? Is it to prevent inline styles? Or did we run into some issue in Learning Platform or something?

packages/components/src/Button/button.tsx
ahuth: > Without combining ButtonHTMLElementProps, the other attributes we want (data-testid, role, aria-live, tabIndex) freak flow out.

Do you mean that we can't add those attributes without also combining `ButtonHTMLElementProps`? That seems odd 🤔

ahuth:
Looks like we explicitly exclude story files from typescript in https://github.com/chanzuckerberg/edu-design-system/blob/2c7570ae088abd997f719d8ef1ab30c8912de287/packages/components/tsconfig.json#L21.

Not sure why that is, but I think we should remove that and start using TypeScript to check story and spec files.

packages/components/src/Button/button.tsx
dierat: This is the test output before this change when testing on a button that has `data-testid`, `role`, `aria-live`, and `tabIndex`:
```
src/Button/button.tsx(48,5): error TS2322: Type '{ children: string; "data-testid": string; role: string; "aria-live": string; tabIndex: number; }' is not assignable to type 'IntrinsicAttributes & Pick<ButtonPropsBase, "as" | "children" | "color" | "size" | "variant" | "disabled" | "type">'.
  Property 'role' does not exist on type 'IntrinsicAttributes & Pick<ButtonPropsBase, "as" | "children" | "color" | "size" | "variant" | "disabled" | "type">'.
  ```

File: packages/components/src/Button/button.tsx (hunk 3 above)
Comment (dierat) [reply packages/components/src/Button/button.tsx:17]: > Is it to prevent inline styles?

Yeah, so they have to use our props and/or css modules.

> Or did we run into some issue in Learning Platform or something?

No, this isn't in response to any existing code, just thinking through what situations could a developer get into that we would want to avoid.

I'm totally open to being convinced that `style` is valid and should be allowed through.

File: packages/components/src/Button/__snapshots__/button.spec.tsx.snap (hunk 1 above)
Comment (dierat) [reply packages/components/src/Button/__snapshots__/button.spec.tsx.snap:106]: Do you think it's common enough to make that useful? I only know that I'm constantly mixing it up, but I assumed it was just me.

File: packages/components/src/Button/button.tsx (hunk 3 above)
Comment (ahuth) [reply packages/components/src/Button/button.tsx:17]: I don't think there's generally a use case for inline styles, although sometimes it's handy when we need an escape hatch from static CSS (even though that's what we want 99.9999999999999999% of the time).

Personally I probably wouldn't worry about preventing it's use. But it doesn't seem actively harmful to do so. If we run into a case where we want it we could always bring it back.

File: packages/components/src/Button/button.tsx (hunk 2 above)
Comment (ahuth) [reply packages/components/src/Button/button.tsx:13]: Weird. I can't repro that, but was probably testing something different.

In any case, intersecting with `ButtonHTMLElementProps` here seems good 👍

**Review by ahuth — APPROVED** [review]

File: packages/components/src/Button/__snapshots__/button.spec.tsx.snap (hunk 1 above)
Comment (ahuth) [reply packages/components/src/Button/__snapshots__/button.spec.tsx.snap:106]: I think it's probably useful because of how easy the rule is to write (detect jsx attributes of `data-test-id`).

But the hard part may be figuring out where to put it 🤔 . Directly into this repo? In a new eslint plugin in https://github.com/chanzuckerberg/frontend-libs ?

Comment (ahuth) [reply packages/components/src/Button/__snapshots__/button.spec.tsx.snap:106]: Probably what we want is a new eslint-plugin-testing package in frontend-libs that contains a data-testid rule and Devin's [no-wrapper-snapshot](https://github.com/chanzuckerberg/along/blob/231e447b96b7e7839a0e178db351162138e03a4c/client/src/eslint/no-wrapper-snapshot.js) rule from Along.

File: packages/components/src/Button/button.tsx (hunk 3 above)
Comment (dierat) [reply packages/components/src/Button/button.tsx:17]: Looking at the `Button` component again, I'm suddenly realizing that `className` isn't a supported prop (it is on `Clickable`). Passing in `className` (even an empty string) removes all the internal styling from the element. So uh we should probably guard against that too (unless we do actually want to support it, in which case we should add it instead).

Thoughts @dcwither ?

Screen Shot 2021-07-08 at 7 17 49 AM

Comment (anniehu4) [reply packages/components/src/Button/button.tsx:17]: oo good catch. I think it's better to support it than not to encourage adoption. we can change `Clickable` to destructure `className`:
```
<Component
      className={clsx(
        className,
        styles.button,
        // ... the rest
```

File: packages/components/src/Button/__snapshots__/button.spec.tsx.snap (hunk 1 above)
Comment (anniehu4) [reply packages/components/src/Button/__snapshots__/button.spec.tsx.snap:106]: ^testing package would be cool!

File: packages/components/src/Button/button.tsx (hunk 3 above)
Comment (anniehu4) [reply packages/components/src/Button/button.tsx:17]: actually ... now I'm wondering, what's the point of omitting `style` if we're not omitting `className` (because the latter is much much more common for styling) 🤔  should we just allow both?

**Review by anniehu4 — APPROVED** [review]
Let's do one more followup (here or separately) to allow ` className`. I don't feel super strongly about omitting `style` vs. not because it's not common on LP (maybe it is more on Along?)

> Not sure why that is, but I think we should remove that and start using TypeScript to check story and spec files.

agree! https://app.clubhouse.io/czi-edu/story/147126/enable-type-checking-on-story-files just need to fix the existing errors first :')

File: packages/components/src/Clickable/Clickable.tsx [hunk 4]
```diff
@@ -37,12 +37,14 @@ function Clickable<IComponent extends React.ElementType>({
...
   state,
   variant,
+  className,
   ...rest
 }: ClickableProps<IComponent>) {
   const Component = as;
   return (
     <Component
       className={clsx(
+        className,
```
dierat, line 47: This is where the `className` was breaking. The prop is already being passed down properly using `...rest`, but it was being applied to `Clickable` *instead of* the internal styles.

dierat:
Update: I removed the `style` prop logic and fixed up `className` in `Clickable` so it's now being applied as expected.

anniehu4 (approved):
thanks!

PR #504, opened 2021-07-14
chore(buttons): make outline button background transparent

dierat:
### Summary:
I was pairing with Sean over the `Banner` component and he pointed out that
the outline buttons have a white background when he intended for them to be
transparent.

Currently, these buttons aren't being used anywhere, so I think this is a safe
change to make. Thoughts?

#### Buttons on dark background, before this change
b

#### Buttons on dark background, after this change
a

### Test Plan:
Run storybook,
navigate to the "Medium Variants on Dark Background" story,
verify the outline buttons have a transparent background (except for the hover state, which should be unchanged).

### DIFF
--- a/packages/components/src/Clickable/Clickable.module.css
+++ b/packages/components/src/Clickable/Clickable.module.css
@@ -147,14 +147,14 @@
 }
 
 .variantOutline {
-  background-color: var(--button-secondary-color);
+  background-color: transparent;
   border-color: var(--button-primary-color);
   color: var(--button-primary-color);
 
   
   &:focus,
   &.stateFocus {
-    background-color: var(--button-secondary-color);
+    background-color: transparent;
     color: var(--button-primary-color);
   }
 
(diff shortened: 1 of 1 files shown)


packages/components/src/Clickable/Clickable.module.css
```diff
@@ -147,14 +147,14 @@
 }
 
 .variantOutline {
-  background-color: var(--button-secondary-color);
+  background-color: transparent;
```
dierat, line 150: Could also just remove the style since it's not overriding anything. But it does make it clear though that the buttons intentionally have a transparent background.

anniehu4 (approved):
ahhh are product designers also aware these are supposed to be transparent? I know there's been some confusion b/w white vs. transparent vs. whatever background color happened to be in Figma before 😬

ahuth (approved):
LGTM

dierat:
> ahhh are product designers also aware these are supposed to be transparent? I know there's been some confusion b/w white vs. transparent vs. whatever background color happened to be in Figma before 😬

Tbh I'm not sure and I wouldn't be surprised if there's still some confusion here 😑 Actually I thought it was standard for them to be transparent, so I had the opposite issue on Along (where it was designed to have white backgrounds and I implemented them as transparent).

PR #506, opened 2021-07-19
chore(colors): update success and warning 500 tokens

dierat:
### Summary:
https://app.clubhouse.io/czi-edu/story/149360/update-success-and-warning-500-tokens

The icons on the success and warning banners don't currently have enough
color contrast, so we're going to bump the tokens they're using (level 500).

success: #00A66A (from #00A96C)
warning: #E06C00 (from #ED7200)

#### Before
before

#### After
after

### Test Plan:
Run storybook (`npm start`),
navigate to `/?path=/story/tokens--page`,
verify that `eds-success-500` has a value of `#00A66A` and `eds-warning-500` has a value of `#E06C00`.


anniehu4 (approved):
🚀

ahuth (approved):

PR #509, opened 2021-07-20
feat(banner): copy over SvgIcon to support Banner and Toast

dierat:
### Summary:
This code is pulled (with minimal changes) from https://github.com/FB-PLP/traject/tree/8d9f4dd5b612ca24c49b1d00d0ac89808baf887e/app/assets/javascripts/v2/core/EDSCandidates/SvgIcon

Clubhouse ticket: https://app.clubhouse.io/czi-edu/story/149303/banners-done

We want to move the `Banner` and `Toast` components into this repo, but they both rely on the `SvgIcon` for their icons. This PR brings over the `SvgIcon` to support the `Banner` and `Toast`, which will come in separate changes.

I'm only bringing over the icons they will actually use (`Close`, `CheckCircle`, `Forum`, `Notification`, `SentimentVeryDissatisfied`, and `Warning`), and I'm not going to export this component, so it can only be used within this repo. It sounds like we want to move over to a different icon library (?), so these icons are temporary and not intended to be used externally.

NOTE: There are still some typescript issues here that I need some help with – one is the `SvgIconProps` error and there are 3 `any`s floating around.

### Storybook screenshots
Screen Shot 2021-07-20 at 12 10 23 PM
Screen Shot 2021-07-20 at 12 10 30 PM
Screen Shot 2021-07-20 at 12 10 18 PM

### Test Plan:
Run storybook (`npm start`),
navigate to `/?path=/docs/svgicon-wip--small`,
verify the icons render at the size their stories suggest,
and verify 6 icons are present in the "all icons" story.


package-lock.json
```diff
@@ -19023,6 +19023,11 @@
         }
       }
     },
+    "utility-types": {
+      "version": "3.10.0",
+      "resolved": "https://registry.npmjs.org/utility-types/-/utility-types-3.10.0.tgz",
+      "integrity": "sha512-O11mqxmi7wMKCo6HKFt5AhO4BwY3VV68YU07tgxfz8zJTIxr4BpsezN49Ffwy9j3ZpwwJp4fkRwjRzq3uWE6Rg=="
+    },
```
dierat, line 19030: This gets us the `$Diff` flow type. The flot-to-ts conversion script relies on `utility-types` to support some existing flow code without having to completely rewrite it.

.vscode/settings.json
```diff
@@ -7,5 +7,7 @@
   ],
   "css.validate": false,
   "scss.validate": false,
-  "stylelint.enable": true
+  "stylelint.enable": true,
+  "typescript.validate.enable": true,
+  "javascript.validate.enable": true,
```
dierat, line 12: Turns on type validation in VSCode. See https://github.com/chanzuckerberg/edu-design-system/pull/514

packages/components/src/SvgIcon/SvgIcon.tsx
```diff
@@ -0,0 +1,106 @@
...
+      className,
+      ...rest
+    }: TestProps,
+    ref: any,
+  ) => {
+    // custom css variables are not in type CSSProperties, so we use this to
+    // convince typescript our variable is safe
+    const style: CSS.Properties = {
+      ["--svg-icon-size" as any]: size,
+    };
```
dierat, line 79: This gets us around a typescript error but creates a lint error. (ﾉ≧∇≦)ﾉ ﾐ ┸━┸

packages/components/src/SvgIcon/SvgIcon.tsx
```diff
@@ -0,0 +1,106 @@
...
+   * Accessible title that will be read out with screen readers.
+   */
+  title: string;
+};
+
+// Just for testing! The props being used should be SvgIconProps, but it's not passing
+// typescript right now.
+type TestProps = BaseProps & {
+  title?: string;
+};
```
dierat, line 60: I think this is the last typescript error. If you use `SvgIconProps` below instead of `TestProps` and run `npx lerna run build:types`, you get this error:
```
src/SvgIcon/utils/createSvgIcon.tsx(22,8): error TS2322: Type '{ children: ReactNode; color?: string | undefined; size?: string | undefined; className?: string | undefined; role: "img" | "presentation"; block?: boolean | undefined; ref: any; }' is not assignable to type '(IntrinsicAttributes & BaseProps & { role: "presentation"; } & RefAttributes<SVGElement>) | (IntrinsicAttributes & BaseProps & { ...; } & RefAttributes<...>)'.
  Property 'title' is missing in type '{ children: ReactNode; color?: string | undefined; size?: string | undefined; className?: string | undefined; role: "img" | "presentation"; block?: boolean | undefined; ref: any; }' but required in type '{ role: "img"; title: string; }'.
```

We're trying to say that title is required if the role is "img" and not present if the role is "presentation" and typescript is not buying it.

packages/components/src/SvgIcon/SvgIcon.tsx
```diff
@@ -0,0 +1,102 @@
...
+  title: string;
+};
+
+export type SvgIconProps = PresentationProps | ImgProps;
+
+interface SvgStyle extends React.CSSProperties {
+  "--svg-icon-size"?: string;
+}
+
+const SvgIcon = React.forwardRef<SVGElement, SvgIconProps>(
```
dierat, line 61: This is the one currently-failing typescript error.

Error output:
```
src/SvgIcon/utils/createSvgIcon.tsx(22,8): error TS2322: Type '{ children: ReactNode; color?: string | undefined; size?: string | undefined; className?: string | undefined; role: "img" | "presentation"; block?: boolean | undefined; ref: any; }' is not assignable to type '(IntrinsicAttributes & BaseProps & { role: "presentation"; } & RefAttributes<SVGElement>) | (IntrinsicAttributes & BaseProps & { ...; } & RefAttributes<...>)'.
  Property 'title' is missing in type '{ children: ReactNode; color?: string | undefined; size?: string | undefined; className?: string | undefined; role: "img" | "presentation"; block?: boolean | undefined; ref: any; }' but required in type '{ role: "img"; title: string; }'.
```

dierat: If I add `title?: string;` to `BaseProps`, I still get and error over the `role` values being different:
```
src/SvgIcon/utils/createSvgIcon.tsx(22,8): error TS2322: Type '{ children: ReactNode; title?: string | undefined; color?: string | undefined; size?: string | undefined; className?: string | undefined; role: "img" | "presentation"; block?: boolean | undefined; ref: any; }' is not assignable to type '(IntrinsicAttributes & BaseProps & { role: "presentation"; } & RefAttributes<SVGElement>) | (IntrinsicAttributes & BaseProps & { ...; } & RefAttributes<...>)'.
  Type '{ children: ReactNode; title?: string | undefined; color?: string | undefined; size?: string | undefined; className?: string | undefined; role: "img" | "presentation"; block?: boolean | undefined; ref: any; }' is not assignable to type '{ role: "img"; title: string; }'.
    Types of property 'role' are incompatible.
      Type '"img" | "presentation"' is not assignable to type '"img"'.
        Type '"presentation"' is not assignable to type '"img"'.
```

dierat: I feel like I have to have a wrapper that accepts either prop set and returns one of two components depending on the `role` value.

packages/components/src/SvgIcon/SvgIcon.tsx
```diff
@@ -0,0 +1,102 @@
...
+  (
+    {
+      block = false,
+      children,
+      color = "currentColor",
+      size,
+      className,
+      ...rest
+    }: SvgIconProps,
+    ref: any,
```
dierat, line 71: #1 of 3 `any` types in this PR.

packages/components/src/SvgIcon/utils/createSvgIcon.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from "react";
+import { $Diff } from "utility-types";
+import SvgIcon from "../SvgIcon";
+import type { SvgIconProps } from "../SvgIcon";
+
+type IconProps = $Diff<
+  SvgIconProps,
+  {
+    children: any;
```
dierat, line 9: Number 2 of 3 `any` types in this PR.

packages/components/src/SvgIcon/utils/createSvgIcon.tsx
```diff
@@ -0,0 +1,35 @@
...
+>;
+
+/**
+ * Generates SVG Icons from paths and fragments.
+ */
+export default function createSvgIcon(
+  path: React.ReactNode,
+  displayName: string,
+): React.FC<IconProps> {
+  const Component = (props: IconProps, ref: any) => {
```
dierat, line 20: Number 3 of 3 `any` types in this PR.

anniehu4:
Woo nice job getting rid of a bunch of them :') and thanks for leaving the inline comments. I'm gonna take a look locally to make sure any suggestions don't break more things

package.json
```diff
@@ -86,5 +86,8 @@
     "*.css": [
       "stylelint --fix"
     ]
+  },
+  "dependencies": {
+    "utility-types": "^3.10.0"
```
anniehu4, line 91: can this be a `devDependency`? ie. install with `npm install -D` instead

packages/components/src/SvgIcon/SvgIcon.tsx
anniehu4: I found [this comment](https://github.com/FB-PLP/traject/pull/5761#discussion_r595636765) from @ahuth from when he added this for MaterialIcon and was hoping it had more context around what we should do in Typescript 😅 

Would introducing a `switch` statement into `createSvgIcon` based on the `role` help?

anniehu4: So this works for me:
```
  const Component = (
    props: SvgIconProps,
    ref: React.ForwardedRef<SVGSVGElement>,
  ) => {
    // eslint-disable-next-line @typescript-eslint/no-unused-vars
    const { role: _, ...rest } = props;

    return props.role === "img" ? (
      <SvgIcon ref={ref} role={props.role} title={props.title} {...rest}>
        {path}
      </SvgIcon>
    ) : (
      <SvgIcon ref={ref} role={props.role} {...rest}>
        {path}
      </SvgIcon>
    );
  };
```

but using `IconProps` (after it goes through $Diff or Omit) doesn't work with the same code 🤔  other things that did not work (still threw the 'title' not a property error):
```
const {role, title} = props;
```

```
const {role} = props;

return role === "img" ? <SvgIcon role={role} title={props.title}>
```

packages/components/src/SvgIcon/utils/createSvgIcon.tsx
ahuth: Good call. `$Diff` confuses the heck out of me. Let's remove it, and any other Flow-isms.

packages/components/src/SvgIcon/SvgIcon.tsx
ahuth: I think this is hitting a limitation of `Pick`, which is how both [Omit is implemented](https://github.com/microsoft/TypeScript/blob/7c4c0a396203adb9ce5a989b61a04ac25657246e/src/lib/es5.d.ts#L1496)  and [$Diff is implemented](https://github.com/piotrwitek/utility-types/blob/df2502ef504c4ba8bd9de81a45baef112b7921d0/src/utility-types.ts#L50).

Haven't been able to find any official reference to this, but I believe it doesn't "distribute" over a union like we'd expect.

So we have something like

```ts
type IconProps = PresentationProps | ImgProps;
```

and we expect

```ts
Omit<IconProps, 'children'>
```

to essentially be equivalent to

```ts
Omit<PresentationProps, 'children'> | Omit<ImgProps, 'children'>
```

But that doesn't seem to be the case. Not exactly sure what it is equivalent to.

Might be related to any of these (although it's kind of hard to tell. I find these issues/posts hard to read)
- https://github.com/microsoft/TypeScript/issues/12959
- https://stackoverflow.com/questions/57103834/typescript-omit-a-property-from-all-interfaces-in-a-union-but-keep-the-union-s
- https://davidgomes.com/pick-omit-over-union-types-in-typescript/

ahuth: One fix would be to not try to omit `children`, and discard any value for it in `createSvgIcon`.

That would mean someone could pass `children` to an icon, which would promptly get ignored. Of course we'd like a type error there, but maybe it's unlikely to be an issue in this repo.

ahuth: Another idea: get rid of `createSvgIcon` completely and [implement icons like we do in Along](https://github.com/chanzuckerberg/along/blob/62990d05130e55cbf7af2a7c52814160ee594999/client/src/ui/Icons/PDF.tsx).

Basically looks something like

```ts
// SomeIcon.tsx

export default function SomeIcon(props: Props) {
  return (
    <SvgIcon {...props}>
      <Circle />
    </SvgIcon>
  );
}
```

Since there should only be a few icons in here, the `createSvgIcon` abstraction may be unnecessary.

packages/components/src/SvgIcon/utils/createSvgIcon.tsx
ahuth: Do we actually need to forward refs for our use case here?

packages/components/src/SvgIcon/SvgIcon.stories.tsx
```diff
@@ -0,0 +1,133 @@
...
+          ...Object.keys(Tokens).filter((key) => {
+            return key.startsWith("LegacyColor") || key.startsWith("EdsColor");
+          }),
+        ],
+      },
+    },
+  },
+};
+
+type Args = React.ElementProps<typeof SvgIcon> & {
```
ahuth, line 45: In TypeScript it's

```suggestion
type Args = React.ComponentProps<typeof SvgIcon> & {
```

ahuth, line 45: In TypeScript it's

```suggestion
type Args = React.ComponentProps<typeof SvgIcon> & {
```

packages/components/src/SvgIcon/utils/createSvgIcon.tsx
```diff
@@ -0,0 +1,35 @@
...
+    );
+  };
+
+  if (process.env.NODE_ENV !== "production") {
+    // Need to set `displayName` on the inner component for React.memo.
+    // React prior to 16.14 ignores `displayName` on the wrapper.
+    Component.displayName = `${displayName}Icon`;
+  }
+
+  return React.memo(React.forwardRef(Component));
```
ahuth, line 34: For our use case here do we need React.memo at all? Wondering if we can get rid of it (and the associated displayName stuff).

It's not clear that memoizing a leaf node like this matters too much.

(basically these icons are at the bottom of the React element tree. There isn't anything else to render other than an HTML element. So a memo/cache hit doesn't obviously save a bunch of rendering).

packages/components/src/SvgIcon/utils/createSvgIcon.tsx
anniehu4: I wonder if it was copied over from MaterialUI? I'm not sure the use cases for forwarding ref to an icon 🤔

packages/components/src/SvgIcon/SvgIcon.tsx
anniehu4: I think we do want to move more of the icons over though. We're just not yet because we're about to change the icon set

packages/components/src/SvgIcon/utils/createSvgIcon.tsx
anniehu4: @dcwither thoughts?

packages/components/src/SvgIcon/utils/createSvgIcon.tsx
ahuth: Use case would be if someone wants to get a handle to the actual SVG element itself.

Since we're not doing that here (as in, none of the components in this repo that I see are capturing a ref) we don't really need it.

dierat:
Thanks for all the suggestions! I put out an alternative PR that has a simplified version of this component since it's just going to be used to support the `Banner` and `Toast` for now (and they definitely don't need the ref or memo stuff): https://github.com/chanzuckerberg/edu-design-system/pull/516

dierat:
https://github.com/chanzuckerberg/edu-design-system/pull/516 merged; I'm closing this PR for now.

PR #514, opened 2021-07-21
chore: show ts errors inline in VSCode

anniehu4:
### Summary:
Typescript errors weren't showing up inline for me in VSCode 😢  I usually have validation set to false in my User workspace because of traject (sad), so this should turn it on for the EDS project always

### Test Plan:
Forced an error in a file and confirmed inline error shows up:
Screen Shot 2021-07-21 at 12 01 32 PM

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,5 +7,7 @@
   ],
   "css.validate": false,
   "scss.validate": false,
-  "stylelint.enable": true
+  "stylelint.enable": true,
+  "typescript.validate.enable": true,
+  "javascript.validate.enable": true,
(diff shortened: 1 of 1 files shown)


anniehu4:
@dcwither lmk if there's a reason we wouldn't want this

dierat (approved):

ahuth:
Seems perfectly reasonable.

I will mention that you can turn off TypeScript just for Traject (there's a button to disable the TypeScript built-in package just for the current workspace).

anniehu4:
oooooomg I hadn't even seen that page before 🤦

PR #516, opened 2021-07-22
feat(svgicon): copy over SvgIcon + 6 icons from traject

dierat:
### Summary:
Split off from https://github.com/chanzuckerberg/edu-design-system/pull/509

Part of https://app.clubhouse.io/czi-edu/story/149303/banners-done

This PR pulls over the `SVGIcon` component and 6 icons from `traject` to support the `Banner` and `Toast` components. The `SVGIcon` component is not being exported for use outside of this repo; it's only being used to get `Banner` and `Toast` out the door.

The version in `traject` has a utility method called `createSvgIcon` that has a bit of logic we don't think we need. Or, well, we definitely don't need it at this point. So this version of porting over the component does it "the Along way" and does not use this utility method. We can revisit the exact implementation later.

### Screenshots from storybook
Screen Shot 2021-07-22 at 4 18 07 PM
Screen Shot 2021-07-22 at 4 18 13 PM
Screen Shot 2021-07-22 at 4 18 21 PM

### Test Plan:
Run storybook (`npm start`),
navigate to `/?path=/docs/svgicon-wip--small`,
verify that in the first 4 stories the icons render as expected at the sizes indicated,
verify that in the 5th story the icons are inline with the text,
and verify that all 6 icons render normally in the 6th story.


packages/components/src/SVGIcon/Icons/SentimentVeryDissatisfied.tsx
```diff
@@ -0,0 +1,14 @@
...
+import React from "react";
+
+export default function SentimentVeryDissatisfied(props: IconProps) {
+  return (
+    <SVGIcon {...props}>
+      <>
+        <circle cx="15.5" cy="9.5" r="1.5" />
+        <circle cx="8.5" cy="9.5" r="1.5" />
+        <path d="M11.99 2C6.47 2 2 6.48 2 12s4.47 10 9.99 10C17.52 22 22 17.52 22 12S17.52 2 11.99 2zM12 20c-4.42 0-8-3.58-8-8s3.58-8 8-8 8 3.58 8 8-3.58 8-8 8zm0-6c-2.33 0-4.32 1.45-5.12 3.5h1.67c.69-1.19 1.97-2 3.45-2s2.75.81 3.45 2h1.67c-.8-2.05-2.79-3.5-5.12-3.5z" />
+      </>
```
dierat, line 11: This is an example of why we have to pass in `children` through the component instead of just getting path data and creating a `path` element ourselves – some icons will be composed of multiple vector elements that are children of the `svg` element.

packages/components/src/SVGIcon/SVGIcon.tsx
```diff
@@ -0,0 +1,111 @@
...
+    return (
+      <svg {...svgCommonProps}>
+        {props.title}
+        {children}
+      </svg>
+    );
+  } else {
+    return <svg {...svgCommonProps}>{children}</svg>;
+  }
+}
```
dierat, line 109: This is the last typescript error. I thought that by checking the `role` before trying to access `title` would get us past this, but I guess it doesn't?

```
src/SVGIcon/SVGIcon.tsx(102,16): error TS2339: Property 'title' does not exist on type 'SVGIconProps'.
  Property 'title' does not exist on type 'PresentationalIconProps'.
```

anniehu4: I think typescript can only parse it correctly if you check `props.role` instead of the already destructured `role` :') there's probably a technical explanation somewhere, but that's what I've found when testing

packages/components/src/SVGIcon/SVGIcon.tsx
```diff
@@ -0,0 +1,110 @@
...
+  title: string;
+}
+
+interface PresentationalIconProps extends IconPropsBase {
+  role: "presentation";
+}
+
+export type SVGIconProps = PresentationalIconProps | FunctionalIconProps;
+
+export type IconProps =
```
anniehu4, line 55: nice 👍

anniehu4 (approved):
approving pending TS fix

packages/components/src/SVGIcon/SVGIcon.tsx
dierat: That was it! 🎉 🎉 🎉

dierat:
Merging even though codecov is failing because the icons are all being tested using a loop, which codecov doesn't seem to be picking up.

PR #517, opened 2021-07-23
feat(banner): copy Banner component from traject

dierat:
### Summary:
Second part of https://app.clubhouse.io/czi-edu/story/149303/banners-done

The only code changes made from the `traject` version were related to transferring it from `flow` to `typescript`.

#### Screenshot from Storybook
Just adding one so you get the idea.
Banner storybook with the stories 'Warning Dismissable', 'Alert Dismissable', and 'Dismissable With Action', visible

#### Video from Storybook
To show all the other versions + hover effects.

https://user-images.githubusercontent.com/7761701/126793219-2480f1d9-2638-4106-9cc0-95bd9b9184eb.mp4

### Test Plan:
Run storybook (`npm start`),
navigate to `/?path=/story/banner--brand`,
verify the stories look and act exactly like the `traject` counterpart,
and especially make sure the layout is correct for the dismisssable with action version on desktop and mobile screens.


packages/components/src/Text/Text.tsx
```diff
@@ -10,7 +10,7 @@ type Props = {
    */
   as?: TextElement;
   children: TypographyProps<TextElement>["children"];
-  className: TypographyProps<TextElement>["className"];
```
dierat, line 13: Was hitting a type error because `className` was a required prop. I don't think we actually want it to be required?

packages/components/src/Banner/Banner.tsx
```diff
@@ -0,0 +1,196 @@
...
+};
+
+/**
+ * This should import a Text element type
+ */
+Banner.Body = function BannerBody(props: { children?: React.ReactNode }) {
+  return props.children ? (
+    <Text color={"inherit"} size="body">
+      {props.children}
+    </Text>
```
dierat, line 137: `size` is a required prop in Text. Should it be optional and just default to body?

packages/components/src/Heading/Heading.tsx
```diff
@@ -14,7 +14,7 @@ type Props = {
    */
   as?: HeadingElement;
   children: TypographyProps<HeadingElement>["children"];
-  className: TypographyProps<HeadingElement>["className"];
```
dierat, line 17: Was hitting a type error because `className` was a required prop. I don't think we actually want it to be required?

packages/components/tailwind.config.js
```diff
@@ -50,6 +50,16 @@ module.exports = {
...
+      padding: {
+        18: "4.5rem",
+      },
+      // Adding custom shadows based on numbers rather than sm/md/lg
+      boxShadow: {
+        0: "0px 0px 0px 1px var(--eds-color-neutral-200)",
+        1: "0px 0px 0px 1px var(--eds-color-neutral-200), 0px 2px 3px rgba(0, 0, 0, 0.02), 0px 4px 8px rgba(0, 0, 0, 0.08)",
+        2: "0px 0px 0px 1px var(--eds-color-neutral-200), 0px 10px 15px -3px rgba(0, 0, 0, 0.1), 0px 4px 6px -2px rgba(0, 0, 0, 0.05)",
+        3: "0px 0px 0px 1px var(--eds-color-neutral-200), 0px 16px 20px -5px rgba(0, 0, 0, 0.1), 0px 10px 10px -5px rgba(0, 0, 0, 0.04)",
+      },
```
dierat, line 62: We're using `shadow-1` in the `Banner` styling.

packages/components/src/Text/Text.tsx
anniehu4: agree 👍

packages/components/src/Banner/Banner.tsx
anniehu4: that sounds good to me

packages/components/tailwind.config.js
anniehu4: excited to remove these from the traject config :excellent:

anniehu4 (approved):

packages/components/src/Banner/Banner.tsx
dierat: Cool, I'll do that in a separate PR.

PR #519, opened 2021-07-23
feat(tag): copy Tag over from traject

dierat:
### Summary:
Clubhouse ticket: https://app.clubhouse.io/czi-edu/story/149292/tags-done

Minimal changes made just to move from flow to typescript and to satisfy the linter.

### Screenshots
Tag in storybook with the 'Default' story visible
Tag in storybook with the 'Color Variants', 'With Icon', and 'With Long Text And Icon' stories visible

### Test Plan:
Run storybook (`npm start`),
navigate to `/?path=/docs/tag--default`,
and verify the stories output is identical to the ones in `traject`.

### DIFF
--- a/packages/components/src/Tag/Tag.spec.tsx
+++ b/packages/components/src/Tag/Tag.spec.tsx
@@ -1,0 +1,8 @@
+// @flow
+
+import * as TagStoryFile from "./Tag.stories";
+import { generateSnapshots } from "@chanzuckerberg/story-utils";
+
+describe("<Tag />", () => {
+  generateSnapshots(TagStoryFile);
+});
--- a/packages/components/src/Tag/Tag.stories.tsx
+++ b/packages/components/src/Tag/Tag.stories.tsx
@@ -48,6 +48,17 @@ ColorVariants.args = {
   ...defaultArgs,
 };
 
+export const OutlineVariants: Story<Args> = (args) => (
+  <div className={styles.tagList}>
+    {colorOptions.map((color) => {
+      return <Tag key={color} {...args} color={color} variant="outline" />;
+    })}
+  </div>
+);
+OutlineVariants.args = {
+  ...defaultArgs,
+};
+
 export const WithIcon = Template.bind(null);
 WithIcon.args = {
   ...defaultArgs,
--- a/packages/components/src/Tag/Tag.stories.tsx
+++ b/packages/components/src/Tag/Tag.stories.tsx
@@ -1,0 +1,62 @@
+// @flow
+
+import * as React from "react";
+import Tag, { stylesByColor } from "./Tag";
+import type { Color } from "./Tag";
+import type { Story } from "@storybook/react";
+import WarningIcon from "../SVGIcon/Icons/Warning";
+import styles from "./Tag.stories.module.css";
+
+const colorOptions: Array<Color> = Object.keys(stylesByColor);
+
+export default {
+  title: "Tag",
+  component: Tag,
+
+  argTypes: {
+    color: {
+      control: {
+        type: "select",
+        options: colorOptions,
+      },
+    },
+  },
+};
+
+type Args = React.ElementProps<typeof Tag>;
+
+const Template: Story<Args> = (args) => <Tag {...args} />;
+
+const defaultArgs = {
+  children: "Tag text",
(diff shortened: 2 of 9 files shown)


packages/components/src/Tag/Tag.tsx
```diff
@@ -0,0 +1,69 @@
+// @flow
+
+import * as React from "react";
+import Text from "../Text";
+import { TypographyColor } from "../common/typography";
+import cx from "classnames";
+import styles from "./Tag.module.css";
+
+export type Color = Omit<
```
anniehu4, line 9: I think `Omit` only works on object types / interfaces, I'm seeing a weird type for `Color` after it:
Screen Shot 2021-07-23 at 2 25 50 PM

I'm also not finding any documentation around picking out of a union of strings like this ... it may be most straightforward to just explicitly type it here 😬

packages/components/src/Tag/Tag.tsx
```diff
@@ -0,0 +1,69 @@
...
+  TypographyColor,
+  "white" | "base" | "info" | "inherit"
+>;
+
+export const stylesByColor: { [K in Color]: string } = {
```
anniehu4, line 14: I think we no longer need the `{ |K in Color]: string }` annotation, typescript can now infer it (unlike flow)

anniehu4 (approved):

packages/components/src/Tag/Tag.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+export const ColorVariants: Story<Args> = (args) => (
+  <div className={styles.tagList}>
+    {colorOptions.map((color) => {
+      return <Tag key={color} {...args} color={color} />;
+    })}
+  </div>
+);
+ColorVariants.args = {
+  ...defaultArgs,
+};
```
anniehu4, line 49: oop just noticed -- could we add the OutlineVariants story as well? 
```
export const OutlineVariants: Story<Args> = args => (
  <div className={styles.tagList}>
    {colorOptions.map(color => {
      return <Tag key={color} {...args} color={color} variant="outline" />;
    })}
  </div>
);
OutlineVariants.args = {
  ...defaultArgs,
};
```

dierat: Oh yeah! I tried to get the recent changes but I must've missed some things. I'll double-check the css as well to make sure it's synced with the traject version.

packages/components/src/Tag/Tag.spec.tsx
```diff
@@ -0,0 +1,8 @@
+// @flow
```
ahuth, line 1: We need to remove the flow stuff when moving these files over

packages/components/src/Tag/Tag.stories.tsx
```diff
@@ -0,0 +1,73 @@
+// @flow
+
+import * as React from "react";
```
ahuth, line 3: In TypeScript there's no need to do the star import. Should be

```js
import React from "react";
```

packages/components/src/Tag/Tag.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+    color: {
+      control: {
+        type: "select",
+        options: colorOptions,
+      },
+    },
+  },
+};
+
+type Args = React.ElementProps<typeof Tag>;
```
ahuth, line 26: `React.ElementProps` doesn't exist, and this should be `React.ComponentProps`.

We need to be careful when moving Flow stuff over that we convert it, since apparently we don't have TypeScript being checked on CI 😬

packages/components/src/Tag/Tag.spec.tsx
anniehu4: ah good catch, I assumed we were running `flow-to-ts` on these which would remove them. One of our cleanup tasks for this period will be writing a script to do all this

packages/components/src/Tag/Tag.stories.tsx
anniehu4: We do have TS checking in the [build step](https://github.com/chanzuckerberg/edu-design-system/runs/3146904330), I think this one is from story files not being checked

PR #521, opened 2021-07-23
chore(banner): remove padding props and long stories

dierat:
### Summary:
We have 3 different levels of padding on the `Banner` component. The reasoning was that larger banners with lots of text need more padding or they look really cramped. But we actually don't want people to put lots of text into banners and are going to advice against it. So why would we support this version of the component?

This PR removes the padding props and also the stories that show the banner with lots of text. I verified with Sean that it's safe to remove these.

### Test Plan:
Verify Percy shows there are no changed stories (besides the ones removed).

### DIFF
--- a/packages/components/src/Banner/Banner.module.css
+++ b/packages/components/src/Banner/Banner.module.css
@@ -1,5 +1,5 @@
 .bannerDialog {
-  @apply w-full rounded border shadow-1;
+  @apply w-full p-8 rounded border shadow-1;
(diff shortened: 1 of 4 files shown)


dierat:
youre-killing-me-codecov

anniehu4 (approved):
makes sense 👍 

I'll look into the codecov config next week: https://app.clubhouse.io/czi-edu/story/150553/setup-codecov-yml

PR #524, opened 2021-07-26
chore: remove flow remnants

dierat:
### Summary:
Follow-up to https://github.com/chanzuckerberg/edu-design-system/pull/519

This PR changes 3 recurring issues that are left over after bringing in code that once used flow but is now typescript:
- remove `// @flow` from the top of files
- change `React.ElementProps` to `React.ComponentProps`
- change `import * as React from "react";` to `import React from "react";`

The last pattern existed in a bunch of files, not just ones ported over recently, so I just updated everything to be consistent.

### Test Plan:
- automated tests pass
- check out storybook (`npm start`) and verify there are no changes to the UI

PR #525, opened 2021-07-26
Various cleanups

ahuth:
[ch147126]

### Summary:

This PR
- Adds a script for running typescript locally. Can run `npm run types` at either the top level or packages/components.
- Adds a separate CI step for type checking. Technically this is duplicate work with the build step. But it could make what's going on more obvious.
- Enable type checking for stories and tests
- Allow Text and Heading components to receive `tabIndex`
- Ignore type errors in SvgIcon, for now. We'll need to figure out what the types should be there.
- ~~Use `React.ComponentProps` instead of `React.ElementProps`~~ Handled by #524
- ~~Remove the Flow pattern of `import * as React from "react"`, which isn't necessary here~~ Handled by #524
- Various type fixes
- Update package author to "CZI"

### Test Plan:

- CI

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -35,7 +35,8 @@
     "storybook:axeOnly": "axe-storybook --build-dir storybook-static",
     "start": "npm run storybook",
     "test": "jest",
-    "test:ci": "npm run test -- --ci --coverage && cat ./coverage/lcov.info | codecov"
+    "test:ci": "npm run test -- --ci --coverage && cat ./coverage/lcov.info | codecov",
+    "types": "tsc --noEmit"
   },
   "bugs": {
     "url": "https://github.com/chanzuckerberg/edu-design-system/issues"
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -6,7 +6,7 @@
     "design system",
     "components"
   ],
-  "author": "Devin Witherspoon <dcwither@gmail.com>",
+  "author": "CZI <edu-frontend-infra@chanzuckerberg.com>",
   "homepage": "https://github.com/chanzuckerberg/edu-design-system/tree/main/packages/components",
   "license": "MIT",
   "main": "dist/index.js",
--- a/packages/components/src/Button/button.stories.tsx
+++ b/packages/components/src/Button/button.stories.tsx
@@ -15,9 +15,7 @@ export default {
   },
 };
 
-type Args = React.ComponentProps<typeof Button> & {
-  "data-testid"?: string;
-};
+type Args = React.ComponentProps<typeof Button>;
 
 const Template: Story<Args> = (args) => <Button {...args} />;
 
--- a/packages/components/src/Button/button.stories.tsx
+++ b/packages/components/src/Button/button.stories.tsx
@@ -15,7 +15,10 @@ export default {
   },
 };
 
-type Args = React.ComponentProps<typeof Button>;
+type Args = React.ComponentProps<typeof Button> & {
+  "data-testid"?: string;
+};
+
 const Template: Story<Args> = (args) => <Button {...args} />;
 
 const heartIcon = (
@@ -44,7 +47,7 @@ link.args = {
   variant: "link",
 };
 
-export const linkInBody = (args) => (
+export const linkInBody: Story<Args> = (args) => (
   <Text size="body">
     This text surrounds the <Button {...args} /> and shows that the link should
     adhere to its appearance
@@ -56,7 +59,7 @@ linkInBody.args = {
   variant: "link",
 };
 
-export const linkInHeading = (args) => (
+export const linkInHeading: Story<Args> = (args) => (
   <Text size="h1">
     This text surrounds the <Button {...args} /> and shows that the link should
     adhere to its appearance
--- a/packages/components/src/Heading/Heading.spec.tsx
+++ b/packages/components/src/Heading/Heading.spec.tsx
@@ -9,14 +9,15 @@ describe("<Heading />", () => {
   generateSnapshots(HeadingStoryFile);
 
   it("should pass through ref", () => {
-    const headingRef = React.createRef();
+    const headingRef = React.createRef<HTMLHeadingElement>();
     render(
       <Heading ref={headingRef} size="h1" tabIndex={-1}>
         Some Heading
       </Heading>,
     );
 
-    headingRef.current.focus();
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    headingRef.current!.focus();
 
     const headingElement = screen.getByText("Some Heading");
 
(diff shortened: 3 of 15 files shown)


packages/components/.storybook/Button/button.stories.tsx
```diff
@@ -1,9 +1,7 @@
-import * as React from "react";
-
 import Clickable, { ClickableProps } from "../../src/Clickable";
-
 import { ButtonProps } from "../../src/Button";
 import Heading from "../../src/Heading";
+import React from "react";
```
ahuth, line 4: This import ordering isn't quite what I expected. Something to look into later.

packages/components/src/Button/button.stories.tsx
```diff
@@ -16,7 +15,10 @@ export default {
   },
 };
 
-type Args = React.ComponentProps<typeof Button>;
+type Args = React.ComponentProps<typeof Button> & {
+  "data-testid"?: string;
```
ahuth, line 19: One of the stories is providing a `data-testid`

packages/components/src/Heading/Heading.spec.tsx
```diff
@@ -9,14 +9,15 @@ describe("<Heading />", () => {
...
+    const headingRef = React.createRef<HTMLHeadingElement>();
     render(
       <Heading ref={headingRef} size="h1" tabIndex={-1}>
         Some Heading
       </Heading>,
     );
 
-    headingRef.current.focus();
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    headingRef.current!.focus();
```
ahuth, line 20: It's kind of subtle, but notice the non-null assertion - `!`.

I think in tests it's perfectly reasonable to use them, since the test would fail if `headingRef.current` happens to be missing.

packages/components/src/Heading/Heading.tsx
```diff
@@ -17,8 +17,9 @@ type Props = {
   className?: TypographyProps<HeadingElement>["className"];
   color?: TypographyProps<HeadingElement>["color"];
   size: TypographyProps<HeadingElement>["size"];
-  weight?: TypographyProps<HeadingElement>["weight"];
   spacing?: TypographyProps<HeadingElement>["spacing"];
+  tabIndex?: number;
```
ahuth, line 21: As the comment in this file points out, we could add `React.HTMLAttributes<HTMLHeadingElement>` to the Props type to get all the attributes someone could pass in.

That makes sense to me, but I opted here for the smaller change.

packages/components/src/SVGIcon/SVGIcon.stories.tsx
```diff
@@ -49,34 +49,42 @@ type Args = React.ComponentProps<typeof SVGIcon> & {
...
 
 const defaultArgs = {
   icon: Object.keys(allIcons)[0],
-  role: "presentation",
+  role: "presentation" as const,
 };
 
 export const Small = Template.bind(null);
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore todo: figure out SVGIcon types
```
ahuth, line 63: I think the SvgIcon types will take some investigation to figure out what's up and how we should write them. Opted to punt on that for from this PR.

packages/components/tsconfig.json
```diff
@@ -11,12 +11,13 @@
     "moduleResolution": "node",
     "noUnusedLocals": true,
     "outDir": "dist/",
+    "resolveJsonModule": true,
```
ahuth, line 14: SvgIcon imports `@chanzuckerberg/eds-tokens/json/variables.json`, and TIL that we need to tell TypeScript to handle that.

anniehu4:
> Use React.ComponentProps instead of React.ElementProps
Remove the Flow pattern of import * as React from "react", which isn't necessary here

You might need to re-merge once https://github.com/chanzuckerberg/edu-design-system/pull/524 goes in

ahuth:
> You might need to re-merge once #524 goes in

Thanks for the heads up. I'll remove the conflicting commits, and let #525 handle the imports and ElementProps.

packages/components/package.json
```diff
@@ -6,7 +6,7 @@
     "design system",
     "components"
   ],
-  "author": "Devin Witherspoon <dcwither@gmail.com>",
+  "author": "CZI <edu-frontend-infra@chanzuckerberg.com>",
```
anniehu4, line 9: good call 👍

packages/components/src/Button/button.stories.tsx
anniehu4: should we add this to the Button component itself (will this be an issue for consumers of Button that try to use `data-testid`)?

packages/components/src/Heading/Heading.tsx
anniehu4: sg, I added a task for it: https://app.clubhouse.io/czi-edu/story/66248/use-react-htmlattributes-htmlheadingelement-in-heading-props

packages/components/src/Tag/Tag.stories.tsx
```diff
@@ -29,7 +31,7 @@ const Template: Story<Args> = (args) => <Tag {...args} />;
 
 const defaultArgs = {
   children: "Tag text",
-  color: "warning",
+  color: "warning" as const,
```
anniehu4, line 34: for learning -- what is `as const` doing?

anniehu4 (commented):
thanks so much, this is great!

packages/components/src/Button/button.stories.tsx
ahuth: We definitely could, and probably should.

Here I just directly fixed the type failure (in the story), but will switch it over to the component.

packages/components/src/Tag/Tag.stories.tsx
ahuth: Great question!

The `color` prop is a union of specific string values: https://github.com/chanzuckerberg/edu-design-system/blob/9c66b28dccc71fc4693578909bbbfbc45b3104a0/packages/components/src/Tag/Tag.tsx#L8-L14

and doesn't accept arbitrary strings.

In TS (and Flow) when defining an object like

```ts
const defaultArgs = {
  color: 'warning',
};
```

the type ends up being

```ts
{
  color: string
}
```

By adding the `as const` we're telling TypeScript to instead interpret values as specific values, and it ends up being

```ts
{
  color: "warning"
}
```

which we can pass to the component.

ahuth: Can use `as const` with many types, such as

- `const a = "hi" as const; // => type is "hi", not string`
- `const b = false as const; // => type is false, not boolean`

or objects such as this example.

dierat:
https://github.com/chanzuckerberg/edu-design-system/pull/524 has been merged 👍

packages/components/src/Button/button.stories.tsx
ahuth: Done

dierat (approved):

ahuth:
Thanks!

PR #528, opened 2021-07-27
Some more cleanups

ahuth:
### Summary:

This PR does some more cleanups
- Remove the unused Storybook "links" addon
- Fix some warning messages due to a couple imports during builds
- Upgrade Storybook (and related dependencies)
- Upgrade both typescript entries
- Upgrade testing-library
- Upgrade Jest
- Upgrade Node for local dev (will require running `nodenv install` after merging)

### Test Plan:

- CI
- Tests and stuff


package.json
```diff
@@ -69,7 +69,7 @@
     "size-limit": "^4.9.2",
     "stylelint": "^13.11.0",
     "stylelint-config-standard": "^20.0.0",
-    "typescript": "^4.1.5"
+    "typescript": "^4.3.5"
```
ahuth, line 72: Kind of annoying, although not that big a deal, to have to specify typescript in 2 places.

I think it's because the top level needs it for linting, and packages/components needs it for type checking (which I added, so it's my own damn fault).

We could maybe set things up so it's only needed at the top level. This is also the kind of yarn workspaces handles nicely, since deps are only installed at the top level.

But not worth worrying about too much right now.

packages/components/src/Banner/__snapshots__/Banner.spec.tsx.snap
```diff
@@ -67,6 +67,7 @@ exports[`<Banner /> AlertDismissable story renders snapshot 1`] = `
       fill="currentColor"
       height="28px"
       role="img"
+      style="--svg-icon-size: 28px;"
```
ahuth, line 70: Seems that Jest 24 wasn't capturing these in the snapshots. But I think it's correct to include them.

packages/components/src/Button/index.ts
```diff
@@ -1 +1,2 @@
-export { default, ButtonProps } from "./button";
+export { default } from "./button";
+export type { ButtonProps } from "./button";
```
ahuth, line 2: Was seeing a warning when building for Storybook

```
WARNING in ./src/Button/index.ts 1:0-48
"export 'ButtonProps' was not found in './button'
```

Not sure why exporting both at the same time would be problematic.

Maybe TypeScript or Babel wasn't removing the type export as expected, and separating them fixes the warning because we can annotate that ButtonProps is a type (and it's easier to tell that it needs to be removed). That's just a guess, though.

.node-version
```diff
@@ -1 +1 @@
-14.5.0
+14.17.3
```
ahuth, line 1: This is the version currently used in Along, and one day we'll use it in Learning Platform.

dierat (approved):
Thanks for cleaning all this up! 👍

anniehu4 (approved):
niice

ahuth:
Thanks!

package.json
anniehu4: 👍  tracking: https://app.clubhouse.io/czi-edu/story/151292/only-import-typescript-at-the-top-level

PR #532, opened 2021-07-27
chore: use clsx instead of cx

dierat:
### Summary:
I ported over several components that are using `cx` (`classnames`), but the rest of the components are using `clsx`. This PR just updates all usage of `cx` to `clsx`.

I added this to the list of things we might want to automatically update when porting over a component: https://app.clubhouse.io/czi-edu/story/150855/write-migration-script-for-components-from-traject-eds

### Test Plan:
Spot check that there are no changes in styling in storybook.


ahuth (approved):
LGTM

anniehu4 (approved):
nice, good call. wonder if we could totally uninstall `classnames` to help enforce it 🤔

PR #533, opened 2021-07-27
Migrate to Yarn v3

ahuth:
### Summary:

This PR tries out using [Yarn workspaces](https://yarnpkg.com/features/workspaces) to simplify working in the repo. It would still use Lerna for publishing.

Using Yarn workspaces would simplify working with the repo a little bit.

There wouldn't be any lerna commands to remember to bootstrap deps, or to add dependencies to a package. There would only be `yarn install`.

- Just downloaded the repo and need to set it up? Just run `yarn install`
- Want to install dependency in a package? `cd` into it and run `yarn install <dependency name>`, as you normally would.

No `npx lerna --scope=somewhere do the thing damnit` commands to learn and remember. No more running `npm install <dep name>` in a package and getting the Lerna symlinks messed up.

Implications:
- Only a single lock file
- https://github.com/lerna/lerna/issues/2631 unfortunately means we can't publish from the dist/ directories currently. I think this is okay, and simplifies things a bit. But it is a slight change in how the packages would be used.

### Test Plan:

- [x] CI
- [ ] Publish alpha versions of the packages


.gitattributes
```diff
@@ -0,0 +1,8 @@
+# Ensure that all files have line feeds for line endings, so that they're consistent regardless of
+# which platform a change was developed on.
+* text=auto eol=lf
```
ahuth, line 3: This just means people can use the repo on Windows without inadvertently changing all the line endings. I just threw it in here for good measure, but it's not related to Yarn or workspaces.

.gitattributes
```diff
@@ -0,0 +1,8 @@
...
+
+# Identify Yarn 2 releases and plugins as binary content.
+# https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored
+/.yarn/releases/** binary
+/.yarn/plugins/** binary
```
ahuth, line 8: Interestingly, the version of Yarn (and any plugins) are actually JS files that get committed to the repo. This config tells git to treat them as binary data, so we don't end up with large diffs anytime we update them.

.yarnrc.yml
```diff
@@ -0,0 +1,7 @@
+nodeLinker: node-modules
+
+plugins:
+  - path: .yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
```
ahuth, line 4: A lot of Yarn functionality is in plugins. Added the workspace-tools one, which lets us to stuff like `yarn workspaces foreach run test`

packages/components/package.json
```diff
@@ -10,32 +10,28 @@
...
-  "publishConfig": {
-    "access": "public",
-    "directory": "dist/",
-    "registry": "https://registry.npmjs.org/"
-  },
   "repository": {
     "type": "git",
     "url": "git+https://github.com/chanzuckerberg/edu-design-system.git"
   },
+  "files": ["dist", "tailwind.config.js"],
```
ahuth, line 17: Since this is now publishing the whole directory as the package, this tells npm to only actually publish dist/ and tailwind.config.js.

npm will also include the package.json and README.md.

packages/components/package.json
```diff
@@ -10,32 +10,28 @@
...
   "scripts": {
-    "build": "npm run build:clean && npm run build:js && npm run build:styles && npm run build:types && rm -rf dist/playground",
+    "build": "yarn run build:clean && yarn run build:js && yarn run build:styles && yarn run build:types && rm -rf dist/playground",
     "build:clean": "rm -rf dist/",
-    "build:js": "babel src -d dist --extensions '.js,.jsx,.ts,.tsx' --ignore src/**/*.spec.ts,src/**/*.spec.tsx --ignore src/**/*.stories.ts,src/**/*.stories.tsx",
+    "build:js": "babel src -d dist --extensions '.js,.jsx,.ts,.tsx' --ignore 'src/**/*.spec.ts','src/**/*.spec.tsx' --ignore 'src/**/*.stories.ts','src/**/*.stories.tsx'",
```
ahuth, line 21: Not sure why, but I needed to add some quotes around the paths to make this work right.

packages/components/src/SVGIcon/SVGIcon.stories.tsx
```diff
@@ -3,7 +3,7 @@ import React from "react";
 import SVGIcon from "./SVGIcon";
 import type { Story } from "@storybook/react";
 import Text from "../Text";
-import Tokens from "@chanzuckerberg/eds-tokens/json/variables.json";
+import Tokens from "@chanzuckerberg/eds-tokens/dist/json/variables.json";
```
ahuth, line 6: This is an implication of publishing the entire package (instead of just dist/), and is a result of https://github.com/lerna/lerna/issues/2631.

We'd need to go update anywhere this is used in Traject. I think it's fine, and simplifies the build process a little bit. 

But we could alternatively output the built files to `packages/tokens/` instead of `packages/tokens/dist`, since they are committed to the repo anyway.

---

Note that this same issue doesn't affect eds-components, since we can still tell npm that dist/index.js is the "main" file.

packages/components/src/Heading/Heading.tsx
```diff
@@ -22,16 +22,26 @@ type Props = {
...
-   * Components that wrap typography sometimes require props such as
-   * event handlers, tabIndex, etc. to be passed down into the element.
-   * TODO: add better typing or documentation for optional props,
-   * e.g. React.HTMLAttributes<HTMLHeadingElement>
-   */ ...rest }: Props, ref) => (
-  <Typography as={as || size} size={size} ref={ref} {...rest}>
-    {children}
-  </Typography>
-));
+const Heading = forwardRef<HTMLElement, Props>(
```
ahuth, line 25: Eslint/Prettier fixed up this file (and a few others). No idea why this either didn't happen with NPM, or is happening differently.

.github/workflows/test.yml
```diff
@@ -8,25 +8,24 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 14.x]
+        node-version: [12.x, 14.x]
```
ahuth, line 11: Haven't found docs on this, but it doesn't look like Yarn v2/v3 supports Node 10.

Seems okay for the design system, since all the apps we support are on Node 14.

dcwither:
> unfortunately means we can't publish from the dist/ directories currently. I think this is okay, and simplifies things a bit. But it is a slight change in how the packages would be used

This is too bad, but maybe this would be softened by making the project side effects false (or side effects *.css if need be). Then we can enable tree shaking in traject.

packages/components/package.json
```diff
@@ -93,6 +93,5 @@
     "storybook-css-modules-preset": "^1.1.1",
     "tailwindcss": "^2.0.3",
     "typescript": "^4.3.5"
-  },
-  "gitHead": "8680dac1fa3c685e2a35f80d9ceb157f5f5343bd"
```
ahuth, line 97: https://github.com/lerna/lerna/issues/1880 indicates that `gitHead` is something Lerna should remove after publishing, and isn't intended to be committed.

.github/workflows/test.yml
```diff
@@ -41,26 +40,25 @@ jobs:
...
+        uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node-version }}
+          cache: 'yarn'
 
       - name: Install Dependencies ⬆️
-        run: |
-          npm ci --no-optional
-          npx lerna bootstrap
+        run: yarn install --immutable
```
ahuth, line 49: `--immutable` tells yarn to blow up if there would be any changes made to the yarn.lock file. This would happen if the package.json somehow got out of sync with the yarn.lock. The previous `npm ci` command would also do that (among other things).

.yarnrc.yml
```diff
@@ -0,0 +1,9 @@
...
+    spec: "@yarnpkg/plugin-workspace-tools"
+  - path: .yarn/plugins/@yarnpkg/plugin-interactive-tools.cjs
```
ahuth, line 6: The interactive-tools plugin lets you run `yarn upgrade-interactive` to open an interactive CLI for upgrading packages. It'll show you which packages have updates available, and lets you select which ones you want to update. I use it all the time in the other repos (axe-storybook-testing, frontend-libs, along).

packages/components/src/Heading/Heading.tsx
ahuth: The [prettier online playground agrees with the changes made in this PR](https://prettier.io/playground/#N4Igxg9gdgLgprEAuc0DOMAEAJOBDAEwEsoBzTAXkwDMIAnAdzzoICU5qAKT4TPNADSYwACyIAbAnQRC0RAF5whAegBUqgDpRMO1ZgDCEALYAHaAhhpMMEXiwM6eE9YCeZ0o5MiXmNMbgwREZwVtIAjgCuRNKYJnQQJlZoEaJ8aFo6mHpwAG4WmLZQBOJwdILWeABGAJJFcAAeQgFgAHTWEJiVcLH8aHAEmAQQDNokMB023XAlwbAtGbqYACoA8gAiK0h8BANdMPB0riYk5PSDEGARszB2RNA0ZwmB0HjisfGJAgtZmHAtpG12HgwDAWtglgBZAAyAEF9nQiJUIvA0AAecHQ3CEE4AURmFgAfN9VMpMC1ydIMJgAL5bAAKH3K0moAEpKATMJwFqilm4IB4nN40hRgPxMAAfcW+BRwanSxQiuSKOXMkXMuXAcktSkwalE7SZYCiCRSBDU7nKXnuTzefUslkAbhAAhATzuUDQyFAzHiDDpzAQnpQryYLk9LsqjjAAGsAgBlEzAk7IGB0CJKEBwIxdHb9KF4MgRPCkOAAMXoRjsgTIyBAeGREGdIBEMCM4gA6mIUYmwHA44GiIEcoOXLWwGhwyASH06DAGcXK8hqK8+i6AFZoeoAISjsZgcbwwShJDgS5XGY39TjJxKAEUIhB4GfxKuQImyqVa5UqtMm3Exu2RAEDYyAABwAAwunEEB9O2ni1nEISlHkTaRI+cAMgkQZ1mgAC0UBwP0-RNuEUTSPOpCLkgy4vhmfRGEQKZpnRN5wPe6HPq+NyVIBwEiMgABMLqpngEgnIYRhUZmaAAKxNhEfRLFUQY0a+OTprUBAWHGYAIiYMAwkUcYwC4JScbK1JAA).

packages/components/package.json
```diff
@@ -10,33 +10,32 @@
...
-    "registry": "https://registry.npmjs.org/"
-  },
   "repository": {
     "type": "git",
     "url": "git+https://github.com/chanzuckerberg/edu-design-system.git"
   },
+  "files": [
+    "dist",
+    "tailwind.config.js"
+  ],
```
ahuth, line 20: This tells npm to only publish files from `dist` and the `tailwind.config.js` file. Everything else (such as src/) will be ignored, except for some files that are always included in the published package (package.json, readme, changelog, etc).

ahuth: An alternative would be a .npmignore file listing the src/, .storybook/, coverage/, and storybook-static/ directories.

package.json
```diff
@@ -33,20 +35,8 @@
...
-  "keywords": [
-    "CZI",
-    "Learning Platform"
-  ],
   "author": "CZI",
   "license": "MIT",
-  "bugs": {
-    "url": "https://github.com/chanzuckerberg/edu-design-system/issues"
-  },
-  "homepage": "https://github.com/chanzuckerberg/edu-design-system#readme",
```
ahuth, line 49: Since this top-level package.json isn't for a published package, we don't need the repo, keywords, bugs, homepage, etc. info.

ahuth:
@chanzuckerberg/edu-frontend-infra what do you think? Does this change make sense, or seem reasonable? Do you hate it? Do you love it? Do you not care one way or the other? Let me know your thoughts.

dierat:
> @chanzuckerberg/edu-frontend-infra what do you think? Does this change make sense, or seem reasonable? Do you hate it? Do you love it? Do you not care one way or the other? Let me know your thoughts.

I'm ambivalent, but it does seem a little simpler, so 🤷

.eslintrc
```diff
@@ -6,10 +6,10 @@
   "parser": "@typescript-eslint/parser",
   "plugins": ["react-hooks", "testing-library"],
   "extends": [
+    "eslint:recommended",
     "plugin:react/recommended",
-    "plugin:prettier/recommended",
     "plugin:testing-library/react",
-    "eslint:recommended"
+    "plugin:prettier/recommended"
```
ahuth, line 12: I don't think it was affecting anything, bu https://github.com/prettier/eslint-plugin-prettier#recommended-configuration says prettier/recommended should be the last extension.

packages/components/src/Heading/Heading.tsx
ahuth: Actually, I bet when installing the deps through yarn we ended up with an updated version of Prettier, and they may have tweaked their formatting.

anniehu4: that would make sense 🤷‍♀️

anniehu4 (commented):
left some clarifying Qs, but overall I like this change! the workspace commands make more sense, and looks like it works pretty well with lerna 👍 

I feel ambivalent about importing `tokens` from "dist/", those get imported pretty rarely. I'm more wary about having to import from "dist/" for components until we have tree shaking set up 🤔 Mainly thinking about Icons, which we're recommending to import directly from their file, and they're pretty commonly used

I guess one alternative is outputting directly into packages/components, adding everything else to `.npmignore`, then cleaning up after so they're not committed? Is that too messy 😬 or is it possible to hold this change until we do have tree-shaking?

package.json
```diff
@@ -1,23 +1,26 @@
...
     "lint:scripts": "eslint --ignore-path .gitignore --ext=js,jsx,ts,tsx .",
-    "lint:scripts:fix": "npm run lint:scripts -- --fix",
-    "postinstall": "lerna bootstrap",
+    "lint:scripts:fix": "yarn run lint:scripts --fix",
     "prettier": "prettier --write \"**/*.{js,jsx,ts,tsx}\"",
     "publish-releases": "lerna publish from-git --no-push",
-    "start": "lerna run build --ignore=@chanzuckerberg/eds-components && lerna run start --parallel",
-    "test": "lerna run test",
-    "types": "lerna run types"
+    "start": "echo 'Do we need this?'",
```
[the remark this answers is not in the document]
ahuth: Ah, I've been cd-ing into packages/components and running `npm run start`. I missed that this was also running `start` in that package.

Will add it back.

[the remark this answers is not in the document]
ahuth: Done

package.json
```diff
@@ -1,23 +1,26 @@
 {
-  "name": "@chanzuckerberg/czedi-kit",
-  "version": "0.0.0",
+  "name": "edu-design-system",
+  "private": true,
```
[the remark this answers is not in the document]
ahuth: That's exactly right. We wouldn't want to do that anyway, and this prevents us from accidentally doing so.

package.json
```diff
@@ -1,23 +1,28 @@
...
     "lint:scripts": "eslint --ignore-path .gitignore --ext=js,jsx,ts,tsx .",
-    "lint:scripts:fix": "npm run lint:scripts -- --fix",
-    "postinstall": "lerna bootstrap",
+    "lint:scripts:fix": "yarn run lint:scripts --fix",
     "prettier": "prettier --write \"**/*.{js,jsx,ts,tsx}\"",
     "publish-releases": "lerna publish from-git --no-push",
-    "start": "lerna run build --ignore=@chanzuckerberg/eds-components && lerna run start --parallel",
-    "test": "lerna run test",
-    "types": "lerna run types"
+    "publish-releases:alpha": "yarn publish-releases --dist-tag alpha",
```
ahuth, line 22: Added scripts for publishing alpha versions.

Normally with Lerna you'd use one command to do this - `lerna publish --canary`.

But we have it split into the two separate phases (create-releases/version and publish-releases/publish), which helps us verify everything is correct before pushing.

That works great for us (and I like it), but it does mean we'll want to make sure each phase's command is correct. Hopefully these scripts help us do that.

packages/components/package.json
```diff
@@ -11,33 +11,32 @@
   "license": "MIT",
   "main": "dist/index.js",
   "sideEffects": false,
-  "publishConfig": {
-    "access": "public",
-    "directory": "dist/",
-    "registry": "https://registry.npmjs.org/"
```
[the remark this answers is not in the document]
ahuth: That is the default for registry, but it actually looks scoped packages (which these are with the `@chanzuckerberg/`) have a default of 'restricted' - https://docs.npmjs.com/cli/v7/using-npm/config#access

My test publish published them publicly, and I'm not sure why. Maybe because paid npm plans are going away?

Will add this back to explicitly override the documented default value. Good catch 👍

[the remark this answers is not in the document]
ahuth: Done

package.json
```diff
@@ -55,7 +48,6 @@
     "@typescript-eslint/eslint-plugin": "^4.15.1",
     "@typescript-eslint/parser": "^4.15.1",
     "classnames": "^2.2.6",
-    "codecov": "^3.8.1",
```
ahuth, line 58: Had to move codecov to the package.json that references it. Otherwise Yarn couldn't find its binary (binaries still get installed in the packages that install them, not the top level necessarily).

ahuth:
Thanks @anniehu4!

> is it possible to hold this change until we do have tree-shaking?

Yes, this is definitely possible. There's no rush on this. Other than my illogical desire to merge any PR ASAP :mashing-button-emoji:

> I'm more wary about having to import from "dist/" for components until we have tree shaking set up 🤔 Mainly thinking about Icons, which we're recommending to import directly from their file, and they're pretty commonly used

Are you thinking that for icons it's too annoying to have to do

```diff
- import FooIcon from '@chanzuckerberg/eds-components/icons/FooIcon';
+ import FooIcon from '@chanzuckerberg/eds-components/dist/icons/FooIcon';
```

?

Also, is anyone using icons from the eds-components package, yet? I thought that LP was only using icons defined in LP at this point in time?

> I guess one alternative is outputting directly into packages/components, adding everything else to .npmignore, then cleaning up after so they're not committed? Is that too messy

Yeah, that would work. Or there may be other strategies that involve moving some files around during publishing (such as copying the package.json and README files into dist/ and publishing from there or something).

I don't think those are really worth doing, though, since adding the `dist/` to an import path seems like a pretty small price to pay. It may be better to accept that and keep the build simpler.

But that's my 2 cents. Let me know what you think

ahuth:
https://github.com/FB-PLP/traject/pull/7846 updates these packages in Learning Platform for this branch, and shows the dist/ implications.

anniehu4:
> Also, is anyone using icons from the eds-components package, yet? I thought that LP was only using icons defined in LP at this point in time?

They're not, yet! But all the new Rounded icons from the UI kit are here, and I was planning to remove SvgIcon from LP in favor of these

anniehu4:
> Are you thinking that for icons it's too annoying to have to do

```suggestion
- import FooIcon from '@chanzuckerberg/eds-components/icons/Foo';
+ import FooIcon from '@chanzuckerberg/eds-components/dist/icons/Foo';
```

that was my original instinct, but upon reflection I think it's not such a big deal. Even with tree shaking, I don't necessarily think people should do 
```
import {FooIcon} from '@chanzuckerberg/eds-components'
```

If anything, it'd make more sense to actually create a separate Icons package then.

packages/components/package.json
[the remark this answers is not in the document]
anniehu4: iinteresting, thanks for checking!

package.json
anniehu4: thanks!

ahuth:
Creating a separate icons package makes sense to me.

For the import paths, we could change from `dist/` to `lib/`

```diff
- import FooIcon from '@chanzuckerberg/eds-components/dist/icons/FooIcon';
+ import FooIcon from '@chanzuckerberg/eds-components/lib/icons/FooIcon';
```

Maybe that would seem less annoying/weird?

anniehu4 (approved):
okay I'm down! I do think migrating to Yarn is worthwhile in the long run.

**last thought on `dist/`:**
> Or there may be other strategies that involve moving some files around during publishing (such as copying the package.json and README files into dist/ and publishing from there or something).

If this is easy (looking at the workaround from [this comment](https://github.com/lerna/lerna/issues/2631#issuecomment-675217008)), it's not that many files to move around. I think the slight complexity added to our build (which only we have to deal with) is worth it to make the experience nicer for everyone else.

Otherwise, it's not a huge deal, we can just communicate the change clearly.

anniehu4:
Just saw your last comment!

> For the import paths, we could change from dist/ to lib/

This does sound better to me 👍

ahuth:
Thanks!

Merging, even though the 10.x tests have not reported as finished. That's because they've been deleted... hoping that once this is in master github won't think it needs to wait for those tests.

PR #537, opened 2021-07-28
chore(banner): split BannerIcon into separate file

dierat:
### Summary:
Clubhouse ticket: https://app.clubhouse.io/czi-edu/story/149303/banners-done

The `Toast` is currently reaching into the `Banner` to get its `Icon` component since that part is so similar between the two components. This PR splits it out into a separate component.

Question: Should shared components that are not exported from `components` be in a different section so it's clearer that although this component is in storybook, it's not available for external use?

### Screenshot
Screen Shot 2021-07-29 at 4 44 33 PM

### Test Plan:
Run storybook (`npm start`),
navigate to `/?path=/story/common-notificationicon--color-variants`,
verify the 5 icons are rendering as expected with the right colors and titles.

Also verify there are no changes registered in Percy for the `Banner` and `Toast` components.


packages/components/src/Banner/__snapshots__/Banner.spec.tsx.snap
```diff
@@ -5,7 +5,7 @@ exports[`<Banner /> Alert story renders snapshot 1`] = `
   class="bannerDialog horizontal colorAlert"
 >
   <div
-    class="icon iconAlert"
```
dierat, line 8: Mm I need to take another look at the `icon` style..

anniehu4:
> Question: Should shared components that are not exported from components be in a different section so it's clearer that although this component is in storybook, it's not available for external use?

I think that makes sense -- we could put this in `common` (like where `Clickable`) used to be

packages/components/src/Banner/Banner.module.css
```diff
@@ -12,10 +12,6 @@
   --svg-icon-size-default: 2.5rem;
 }
 
-.icon {
-  @apply justify-self-center;
-}
-
```
dierat, line 18: This styling doesn't seem to have been doing anything so it wasn't moved over to the new component.

packages/components/src/Banner/__snapshots__/Banner.spec.tsx.snap
dierat: We can do without the `icon` style; it wasn't doing anything.

packages/components/src/Banner/index.ts
```diff
@@ -1 +1,2 @@
-export { default, BannerIcon } from "./Banner";
+export { default } from "./Banner";
+export { BannerColor } from "./Banner";
```
anniehu4, line 2: nit: `export type { BannerColor }`

packages/components/src/common/Notifications/NotificationIcon/NotificationIcon.tsx
```diff
@@ -0,0 +1,61 @@
...
+  alert: {
+    icon: DangerousRoundedIcon,
+    style: styles.iconAlert,
+    title: "alert",
+  },
+};
+
+/**
+ * An icon that appears in notifcation components (Banner and Toast).
+ * This icon communicates the tone of the notification and is through its image,
```
anniehu4, line 48: nit:
```suggestion
 * This icon communicates the tone of the notification through its image,
```

packages/components/src/common/Notifications/NotificationIcon/NotificationIcon.tsx
```diff
@@ -0,0 +1,61 @@
...
+ * color, and title text.
+ */
+const NotificationIcon = ({ variant }: Props) => {
```
anniehu4, line 51: curious why you switched to `variant` instead of `color`?

packages/components/src/common/Notifications/NotificationIcon/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from "./NotificationIcon";
```
anniehu4, line 1: this is pretty nitty, but I think it'd be fine if all the `NotificationIcon` files were directly under `Notifications/` instead of in another subfolder. I realize you probably used `plop` which did this automatically

anniehu4 (approved):

packages/components/src/common/Notifications/NotificationIcon/index.ts
anniehu4: nvm!! just saw your DismissButton PR so this makes more sense

packages/components/src/common/Notifications/NotificationIcon/NotificationIcon.tsx
dierat: It already kinda bugs me that `Banner` and `Toast` have a prop called `color` that also controls which icon is used, but it felt especially weird to use `color` in the situation where it primarily controls which icon is returned. But I don't care that much and am fine with changing it to `color` if consistency is more important.

anniehu4: mmm yep makes sense

PR #538, opened 2021-07-28
chore: fix SvgIcon story types

anniehu4:
### Summary:

Removes the `ts-ignore`s from SVGIcon.stories.tsx. Next PR will add `createSvgIcon` and a license statement

### Test Plan:
- tsc passes
- confirmed that passing something like `role: "test"` still
Screen Shot 2021-07-28 at 4 52 27 PM
 fails type checking

### DIFF
--- a/packages/components/src/SvgIcon/SvgIcon.stories.tsx
+++ b/packages/components/src/SvgIcon/SvgIcon.stories.tsx
@@ -54,37 +54,29 @@ const Template: Story<Args> = ({ icon, color, ...rest }) => {
 };
 
 const defaultArgs = {
-  icon: Object.keys(allIcons)[0],
+  icon: Object.keys(allIcons)[0] as keyof typeof allIcons,
   role: "presentation" as const,
 };
 
 export const Small = Template.bind(null);
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore todo: figure out SvgIcon types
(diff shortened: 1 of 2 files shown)


packages/components/src/SVGIcon/SVGIcon.stories.tsx
```diff
@@ -54,37 +54,29 @@ const Template: Story<Args> = ({ icon, color, ...rest }) => {
 };
 
 const defaultArgs = {
-  icon: Object.keys(allIcons)[0],
+  icon: Object.keys(allIcons)[0] as keyof typeof allIcons,
```
anniehu4, line 57: for some reason, this was actually the source of the TS errors 🤔 it was unhappy that icon was type `string`

packages/components/src/SVGIcon/SVGIcon.tsx
```diff
@@ -23,12 +23,6 @@ interface IconPropsBase {
    * @chanzuckerberg/eds-tokens/json/variables.json
    */
   color?: string;
-  /**
-   * The role of the icon.
-   * "img" requires a title prop also being passed in
-   * "presentation" should not have a title prop because it will be hidden from screenreaders
-   */
-  role: "img" | "presentation";
```
anniehu4, line 31: Typescript can infer this from the below union

dierat (approved):

PR #539, opened 2021-07-28
chore: update SvgIcon stories to only use color tokens file

anniehu4:
### Summary:

There was a TODO comment in the story file to use "fixed exports of js colors", instead of filtering through all tokens. I hit a "missing declaration file" issue when trying to import it directly, so I also added a `.ts` version of the file (more detail in inline comment)

### Test Plan:
confirm color control still works in storybook:
svg icon colors


packages/components/src/SVGIcon/SVGIcon.stories.tsx
```diff
@@ -1,9 +1,9 @@
+import * as ColorTokens from "@chanzuckerberg/eds-tokens/ts/colors";
```
anniehu4, line 1: This is exactly the same as `js/colors.js`, but avoids the error
```
Could not find a declaration file for module 'module-name'. '/path/to/module-name.js' implicitly has an 'any' type
```

I think we need to keep the `js` one around still, for flexible use

dierat (approved):

packages/tokens/style-dictionary.config.js
```diff
@@ -51,6 +51,18 @@ const EDSStyleDictionary = StyleDictionary.extend({
             },
           },
         },
+        {
+          format: "javascript/es6",
+          destination: "ts/colors.ts",
```
ahuth, line 56: Can we avoid needing the duplicate TS file by setting `checkJs: true` in the tsconfig?

ahuth (approved):

packages/tokens/style-dictionary.config.js
anniehu4: I tried turning on `allowJs` & `checkJs` but still saw the error 😬 

Screen Shot 2021-07-29 at 12 30 59 PM

anniehu4:
codecov :')

PR #540, opened 2021-07-28
Update SvgIcon for the latest axe-core

ahuth:
### Summary:

This PR
- Updates axe-core to 4.3.2
- Uses `aria-hidden` instead of `presentation="role"` for presentational/decorative icons (will explain more below)
- Replaces the `role` prop with ~~type~~ `purpose` (will explain more below)

### Wat, why?

Grab a cup of coffee (or your flavored beverage of choice) and let me tell you the story of why I made all these changes 📖 

#### Chapter 1 - More upgrades, more problems

All I wanted to do was update axe-core (213e146069bb8e04ba0ff836cac4d6b8e1875cd6), but that caused some axe testing failures. Looks like [axe-core's allowlist of elements that can have `role="presentation"` does not include `<svg>`](https://github.com/dequelabs/axe-core/blob/2777dbc20c028d88a906e49a3cd7032b482f0988/lib/commons/aria/index.js#L1462-L1488).

I can't find any documentation anywhere that svg elements shouldn't have that role. So I don't really agree that it should be forbidden.

#### Chapter 2 - Embrace the tools

But mdn has some interesting docs on [when to use aria-hidden vs presentation="role"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute#deciding_between_aria-hiddentrue_rolepresentation_and_rolenone), and our use case for icons seems closer to aria-hidden to me (completely remove from the accessibility tree).

So I changed SvgIcon to use aria-hidden when they're presentational/decorative (2e7650f371dc384a5a43900bbad415d7df376ab3).

#### Chapter 3 - Words have meaning, I think

An interesting ramification of that is now the `role` prop is decoupled from what actually happens on the HTML svg element. Meaning, `role="presentation"` would no longer result in the svg element literally having that role.

So I got to thinking (always dangerous) why not embrace that decoupling? And I renamed the prop from `role` to `purpose` (thank you Diedra for the prop name idea), and changed its possible values to "informative" and "decorative" (f226bd501006c9fa657cbae03766ca8858ccc3cd).

Those words may still be kind of fuzzy, but they could better indicate the purpose of them. An "informative" icon has information to communicate. A decorative one is for decoration - it's extraneous.

If you've made it this far, you can tell I'm way over-analyzing this. But over-analyzing is half the fun.

### Test Plan:

- CI
- Local storybook testing


packages/components/src/Banner/Banner.tsx
```diff
@@ -84,7 +84,7 @@ export const BannerIcon = ({ color }: { color: Color }) => {
 
   return (
     <div className={clsx(styles.icon, iconAssets.style)}>
-      <iconAssets.icon role="img" title={iconAssets.title} />
+      <iconAssets.icon type="informative" title={iconAssets.title} />
```
ahuth, line 87: More info in the PR description. If you think the switch from role (img|presentation) to type (informative|decorative) is more confusing or unnecessary, I'll switch it back.

ahuth:
Don't know why codecov thinks there's -0.04% less test coverage.

packages/components/src/Button/button.stories.tsx
```diff
@@ -79,14 +74,22 @@ withDataTestId.args = {
 
 export const linkWithIcon = Template.bind(null);
 linkWithIcon.args = {
-  children: <>Link with icon {heartIcon}</>,
+  children: (
+    <>
+      Link with icon <CheckCircle type="decorative" />
+    </>
+  ),
```
dierat, line 81: Thanks for updating this! That heart icon looked seriously fugly 😅

packages/components/src/SvgIcon/SvgIcon.tsx
```diff
@@ -89,15 +100,19 @@ function SvgIcon(props: SvgIconProps) {
     xmlns: "http://www.w3.org/2000/svg",
   };
 
-  if (props.role === "img") {
+  if (props.type === "informative") {
     return (
-      <svg {...svgCommonProps}>
+      <svg {...svgCommonProps} role="img">
         <title>{props.title}</title>
```
dierat, line 106: I kinda wonder if we should use an `aria-label` and drop the role + title for informative icons for consistency. It feels a little weird to me that one version of the svgs has a role and the other doesn't. 🤷

dierat (approved):
Weird that it doesn't allow `role="presentation"` on svgs, but I agree that `aria-hidden` makes more sense anyway.

Re: the API change, I think it would be a little misleading to have consumers pass in a `role` that won't actually result in a `role` attribute on the element, but I also feel like `type` is really vague. Maybe something like "purpose"? I also like the idea of switching from "img" to "informative" and from "presentation" to "decorative" because I think that will make it easier for people to reason about which category any particular icon falls into.

packages/components/src/SvgIcon/SvgIcon.tsx
anniehu4: One benefit of `<title>` (which our auditors request sometimes) is it'll appear when hovering on the icon, in case it's not clear what an icon means. I do generally like aria-label better, but that could be a reason to keep this

anniehu4 (approved):
also like "informative" & "decorative"! I'm actually not against keeping the prop named `"role"`, unless you're worried it'll cause confusion around the "role" prop for accessibility (I'd guess people don't think too much about that one)

ahuth:
Good point that `type` is really vague. Both `role` and `purpose` make a lot of sense.

I'll change it to `purpose` to deconflict with the HTML attribute. I don't think a conflict there is really a big problem, but for some reason "purpose" seems to go better with "informative"/"decorative" to me 🤔

ahuth:
Thanks! Merging (not sure what's up with codecov)

ahuth:
FYI, since https://github.com/dequelabs/axe-core/pull/3124 axe will now allow svg elements to have a role of presentation or img. I think our current implementation is still fine, or maybe even better, though.

PR #543, opened 2021-07-30
chore(banner): pull out color styles into shared styles module

dierat:
### Summary:
Clubhouse ticket: https://app.clubhouse.io/czi-edu/story/145432/create-shared-utilities-across-toast-banner

The color styles for the `Banner` and `Toast` components are mostly duplicated. Right now there are only 2 color on the `Toast`, but I think we're going to add more one day? So this will also save a little work in the future if that happens.

### Test Plan:
Verify there are no Percy changes,
and verify the hover colors are unchanged on the dismiss button in the `Banner` and `Toast` stories.


dierat:
https://github.com/chanzuckerberg/edu-design-system/pull/543

dierat:
app.clubhouse.io/czi-edu/story/145432/create-shared-utilities-across-toast-banner

packages/components/src/Toast/Toast.tsx
```diff
@@ -45,8 +46,8 @@ export default function Toast({
       className={clsx(
         className,
         styles.toastDialog,
-        color === "success" && styles.colorSuccess,
-        color === "alert" && styles.colorAlert,
+        color === "success" && colorStyles.colorSuccess,
+        color === "alert" && colorStyles.colorAlert,
```
dierat, line 50: I could also write a little utility to unduplicate these lines, but it's only two lines, so I'm gonna skip it for now.

packages/components/src/Toast/Toast.module.css
```diff
@@ -2,6 +2,8 @@
   @apply w-full max-w-100 pl-4 pr-2 grid gap-2 items-center;
   @apply border border-l-4 rounded shadow-1;
 
+  border-color: var(--banner-border-light-color);
+  border-left-color: var(--banner-border-dark-color);
```
dierat, line 6: The shared styles are setting variables, not actual border colors, because  the `Banner` border are different on desktop and mobile, so I had to update the `Toast` to use the variables.

packages/components/src/common/Notification.module.css
```diff
@@ -0,0 +1,34 @@
+.colorBrand {
+  @apply text-brand-700 bg-brand-100;
+  --banner-border-light-color: var(--eds-color-brand-200);
```
anniehu4, line 3: I think we could remove the `banner-` prefix for these. it's only kind of confusing in the Toast styles though, so not a biggie

anniehu4 (approved):

anniehu4:
thinking about folder structure -- might make sense to have

- common
 -- Notification
   ---- Notification.module.css
   ---- NotificationIcon.tsx

?

dierat:
> thinking about folder structure -- might make sense to have
> 
> * common
>   -- Notification
>   ---- Notification.module.css
>   ---- NotificationIcon.tsx
> 
> ?

Sounds good! I briefly thought about that too but wasn't sure if I was overthinking it.

packages/components/src/Banner/index.ts
```diff
@@ -1 +1,2 @@
-export { default, BannerIcon } from "./Banner";
+export { default } from "./Banner";
+export { BannerColor } from "./Banner";
```
ahuth, line 2: Should this be

```suggestion
export type { BannerColor } from "./Banner";
```

to help babel figure out it can be omitted from the JS?

dierat: Oh good call! Looks like while this PR was out, that was fixed and then moved + renamed, so there's nothing to fix here, but thanks for taking a look and pointing that out 👍

PR #545, opened 2021-07-30
feat(icons): add DangerousRounded for notifications

anniehu4:
### Summary:
[ch149303]

Adds Dangerous rounded icon to use as the Banner & Toast "alert" icon. I added it to a "custom" folder since it's not in https://material-ui.com/components/material-icons/, so I had to grab the raw SVG from https://fonts.google.com/icons?selected=Material+Icons+Round:campaign&icon.query=dangerous.

### Test Plan:
Confirmed allIcons story is updated:
Screen Shot 2021-07-30 at 10 35 58 AM

Banner "alert" variant:
Screen Shot 2021-07-30 at 10 36 18 AM

Toast "alert" variant (**note**: alignment looks a little off though?):
Screen Shot 2021-07-30 at 10 36 41 AM


packages/components/src/Banner/Banner.tsx
```diff
@@ -2,10 +2,10 @@ import Heading, { HeadingElement } from "../Heading";
 import Button from "../Button";
 import CheckCircleIcon from "../Icons/CheckCircle";
 import CloseIcon from "../Icons/Close";
+import DangerousRoundedIcon from "../Icons/custom/DangerousRounded";
```
anniehu4, line 5: keeping these named `*Rounded` in case we change our minds and need a mix of multiple sets in the future or something 🤓

dierat:
> note: alignment looks a little off though?

Looks like the text isn't quite vertically centered in `Toast` because the icon is taller than the line height. Adding `align-items: center;` on the `content` styles fixes this.

dierat (approved):

PR #546, opened 2021-07-30
build: remove all story and spec related files from dist/

anniehu4:
### Summary:
It was kind of annoying that `components/dist/` contained type declarations for story & spec files still:
Screen Shot 2021-07-30 at 3 54 39 PM

and the flow script was trying to parse those files also:
Screen Shot 2021-07-30 at 3 54 55 PM

This PR:
- skips story & spec files during flowgen (for speed reasons)
- removes the `playground/` folder (we can just mark stories as WIP)
- removes all story & spec filepaths **at the end of the build**

### Test Plan:
Confirmed that each component only has `index.js` (+ types), `[Component].js` (+ types), and `[Component.module.css]` files within `dist/`
Screen Shot 2021-07-29 at 7 35 41 PM


packages/components/package.json
```diff
@@ -20,9 +20,10 @@
     "url": "git+https://github.com/chanzuckerberg/edu-design-system.git"
   },
   "scripts": {
-    "build": "npm run build:clean && npm run build:js && npm run build:styles && npm run build:types && rm -rf dist/playground",
+    "build": "npm run build:clean && npm run build:js && npm run build:styles && npm run build:types && npm run build:finish",
     "build:clean": "rm -rf dist/",
-    "build:js": "babel src -d dist --extensions '.js,.jsx,.ts,.tsx' --ignore src/**/*.spec.ts,src/**/*.spec.tsx --ignore src/**/*.stories.ts,src/**/*.stories.tsx",
```
anniehu4, line 25: Adding single quotes would correctly ignore these for `babel`:
```
--ignore 'src/**/*.spec.ts','src/**/*.spec.tsx','src/**/*.stories.ts','src/**/*.stories.tsx'
```

but:
- `tsc` was still generating `.d.ts` files 😬 
- postcss was still including the `.stories.module.css` files 😬 

Instead of adding this custom ignore logic to every step (and partly because the babel glob matching is already bad), it felt cleaner to just remove everything with `.spec.` or `.stories.` at the end. A side benefit is we don't need to worry about where `.stories.module.css` files are included (and can move the button ones in `.storybook/Button` back)

packages/components/generateFlowTypes.sh
```diff
@@ -1,7 +1,12 @@
 #!/bin/sh
 
 for i in $(find dist/* -name "*.d.ts"); do
-    [ -f "$i" ] || break
+    if  [ ! -f "$i" ] || \
+        [[ "$i" == *".stories."* ]] || \
+        [[ "$i" == *".spec."* ]]
```
anniehu4, line 6: this isn't technically necessary for the `dist/` folder, but the `flowgen` script is a little slow so it's nice to skip these if we can

packages/components/package.json
ahuth: I like it 👍

ahuth (approved):
LGTM 🤘

PR #550, opened 2021-08-02
Scripts for easy package publishing

ahuth:
### Summary:

This PR adds scripts for publishing, to make it easier. Now to publish you'd

- Run `npm run create-releases`
- Check that everything looks right
- Run `npm run publish-releases`
- Run `git push origin head --follow-tags`

### Test Plan:

- Test publish

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "main": "index.js",
   "scripts": {
     "build": "NODE_ENV=production lerna run build",
+    "create-releases": "lerna version --no-push --conventional-commits --message 'chore(release): %s'",
     "lint": "npm run lint:styles && npm run lint:scripts",
     "lint:fix": "npm run lint:styles:fix && npm run lint:scripts:fix",
     "lint:styles": "stylelint --ignore-path .gitignore packages/**/*.{css,js,jsx,ts,tsx}",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -5,6 +5,7 @@
   "main": "index.js",
   "scripts": {
     "build": "NODE_ENV=production lerna run build",
+    "create-releases": "npx lerna version --no-push --conventional-commits --message 'chore(release): %s'",
```
ahuth, line 8: Couple changes to this script.

First, (if it wasn't before) this will now open an interactive prompt to verify the versions
Screen Shot 2021-08-02 at 3 25 19 PM

Second, notice the `--message 'chore(release): %s'`. Lerna will automatically fill in the version number at `%s`, so we don't have to specify the commit message.

dierat: Does this only work when bumping the patch version (not minor/major)?

package.json
```diff
@@ -5,6 +5,7 @@
   "main": "index.js",
   "scripts": {
     "build": "NODE_ENV=production lerna run build",
+    "create-releases": "lerna version --no-push --conventional-commits --message 'chore(release): %s'",
```
ahuth, line 8: Couple changes to this script from what's in the README:

1. (if it wasn't before) this will now open an interactive prompt to verify the versions
Screen Shot 2021-08-02 at 3 25 19 PM

2. Notice the `--message 'chore(release): %s'`. Lerna will automatically fill in the version number at `%s`, so we don't have to specify the commit message.

ahuth, line 8: @dierat, looks like I accidentally buried your question here:

> Does this only work when bumping the patch version (not minor/major)?

No, this on works on any version bump type, as determined by the conventional commit messages (fix, feat, etc)

dierat: Cool, thanks!

dierat (approved):

ahuth:
Thanks!

dcwither:
Love this change! Thanks for operationalizing my flimsy process 😆

ahuth:
Thanks! Your process wasn't flimsy, it was an MVP

PR #551, opened 2021-08-03
Export SvgIcon from the components package

ahuth:
### Summary:

Allow `<SvgIcon>` to be used from the components package. That way we can look at using it in Learning Platform and Along.

### Test Plan:

- CI

### DIFF
--- a/packages/components/src/SvgIcon/index.tsx
+++ b/packages/components/src/SvgIcon/index.tsx
@@ -1,1 +1,2 @@
 export { default } from "./SvgIcon";
+export type { IconProps } from "./SvgIcon";
--- a/packages/components/src/index.ts
+++ b/packages/components/src/index.ts
@@ -2,6 +2,8 @@ export { default as Banner } from "./Banner";
 export { default as Button } from "./Button";
 export { default as Clickable } from "./Clickable";
 export { default as Heading } from "./Heading";
+export { default as SvgIcon } from "./SvgIcon";
+export type { IconProps } from "./SvgIcon";
 export { default as Tag } from "./Tag";
 export { default as Text } from "./Text";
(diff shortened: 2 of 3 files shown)


packages/components/src/index.ts
```diff
@@ -2,6 +2,7 @@ export { default as Banner } from "./Banner";
 export { default as Button } from "./Button";
 export { default as Clickable } from "./Clickable";
 export { default as Heading } from "./Heading";
+export { default as SvgIcon, IconProps } from "./SvgIcon";
```
ahuth, line 5: Random thought - if we get enough icons, one day we could think about creating a logical separation between icons and eds components, and create a `@chanzuckerberg/eds-icons` package.

packages/components/src/SvgIcon/index.tsx
```diff
@@ -1 +1 @@
-export { default } from "./SvgIcon";
+export { default, IconProps } from "./SvgIcon";
```
anniehu4, line 1: does this work fine now, or should we still export Props on a separate line?

packages/components/src/index.ts
anniehu4: 💡 !

anniehu4 (approved):

packages/components/src/SvgIcon/index.tsx
ahuth: Good call, I forgot there was an issue. Everything is green, and I don't see TypeScript or Babel complaining about anything.

But when I look at the dist directory, there is an export for `_SvgIcon.IconProps`. I'll move this to its own line to prevent that from showing up in the js file at all.

dierat (approved):

ahuth:
Thanks!

PR #553, opened 2021-08-03
Add option ClassName attribute to Tag

CMWesterfield16:
### Summary:
Add the ability to pass a unique className to the new Tag component

### Test Plan:
No regressions

### DIFF
--- a/packages/components/src/Tag/Tag.tsx
+++ b/packages/components/src/Tag/Tag.tsx
@@ -31,6 +31,10 @@ type Props = {
    */
   children: React.ReactNode;
   /**
+   * Additional case-specific styling
+   */
+  className?: string;
+  /**
    * The tag icon (shouldn't provide color or size since those are determined
    * by the color prop).
    */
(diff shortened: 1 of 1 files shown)


packages/components/src/Tag/Tag.tsx
```diff
@@ -44,14 +48,15 @@ type Props = {
...
-function Tag({ color, children, icon, variant = "flat" }: Props) {
+function Tag({ color, children, className, icon, variant = "flat" }: Props) {
   return (
     <Text
       as="span"
       className={clsx(
         styles.tag,
         color && stylesByColor[color],
         variant === "outline" && styles.outline,
+        className,
```
anniehu4, line 59: nit: could you move to line 56 so it's first in the list? (just for consistency)

anniehu4 (approved):
thanks!!

anniehu4:
also, the lint error means the commit message needs to be prefixed with `feat:` or `fix:` (we're using the [Conventional Commits standard](https://www.conventionalcommits.org/en/v1.0.0/#summary))

PR #554, opened 2021-08-03
No side effects

ahuth:
### Summary:

Tell Webpack that eds-components has no side effects, to help it remove unused files more effectively when someone imports from index.js.

https://webpack.js.org/guides/tree-shaking/\#mark-the-file-as-side-effect-free

### Test Plan:

- CI

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -10,6 +10,7 @@
   "homepage": "https://github.com/chanzuckerberg/edu-design-system/tree/main/packages/components",
   "license": "MIT",
   "main": "dist/index.js",
+  "sideEffects": false,
   "publishConfig": {
     "access": "public",
     "directory": "dist/",


packages/components/package.json
```diff
@@ -10,6 +10,7 @@
   "homepage": "https://github.com/chanzuckerberg/edu-design-system/tree/main/packages/components",
   "license": "MIT",
   "main": "dist/index.js",
+  "sideEffects": false,
```
ahuth, line 13: Here we're basically promising to Webpack that there are no side effects.

Examples of side effects would be a module changing something outside of itself

```js
window.foo = 666;
```

or providing a CSS file for consumers to import

```js
// In Learning Platform or Along or something...

import '@chanzuckerberg/eds-components/some-styles.css';
```

By telling Webpack that there are no side effects, it'll remove unused  files from something like

```js
import {Heading} from '@chanzuckerberg/eds-components';
```

dcwither: I think it also promises there's no shared "global to the module" values like

```js
let count = 0

function increment1() {
  count += 1;
}

function increment2() {
  count += 2;
}
```

dcwither (approved):
Shouldn't be any issues for this lib

packages/components/package.json
ahuth: My understanding was shared globals **_within_** a module are fine, but shared stuff **_between_** modules is not.

Could be wrong about that. Either way, good call out.

ahuth:
Thanks!

ahuth:
Thanks for looking into that, and we should do more investigation.

> “Ensure no compilers transform your ES2015 module syntax into CommonJS modules (this is the default behavior of the popular Babel preset @babel/preset-env - see the documentation for more details)”

We're currently compiling to common js, so we'd need to change to output ES modules. Definitely worthy trying and seeing if we run into any issues.

> Add `/*#__PURE__*/` annotation to any components as needed

I don't think we really need to manually specify which statements are "pure". Especially since we've added the `sideEffects` annotation for modules. And it looks like Babel is actually adding `#__PURE__` on it's own in some cases already.

PR #561, opened 2021-08-05
Storybook cleanup to match UI kit

anniehu4:
### Summary:

Various cleanup things:
- moves button grid stories from `.storybook` into rest of the button stories
- only displays "alert" and "brand" colors for the `flat` buttons, per [UI kit recommendations](https://www.figma.com/file/YPTOEVRuas5ytXQEcwm4FK/UI-Kit-Sticker-Sheet-(Alpha)?node-id=226%3A1100)
- adds & renames button stories to better match UI kit (and so we can use the stories directly in Zeroheight instead of Figma assets)
- makes Toast stories dismissable by default
- fixes some controls

### Test Plan:
See Percy for visual changes


packages/components/src/Banner/Banner.stories.tsx
```diff
@@ -7,6 +7,14 @@ import type { Story } from "@storybook/react";
 export default {
   title: "Banner",
   component: Banner,
+  argTypes: {
+    elevation: {
+      control: {
+        type: "radio",
+        options: [0, 1],
```
anniehu4, line 14: Storybook is having trouble when an argument has value `0` 🤔  this fixes the explicit control, but the elevation arg is still not passed properly when the story is embedded: https://chanzuckerberg.github.io/edu-design-system/iframe.html?id=banner--elevation-0&args=&viewMode=story

packages/components/src/Button/button.stories.tsx
```diff
@@ -13,30 +33,48 @@ export default {
         type: "text",
       },
     },
+    color: {
```
anniehu4, line 36: for some reason these weren't showing up 😞  it's weird b/c the other Button props are declared similarly, but show up correctly

packages/components/src/Banner/Banner.stories.tsx
ahuth: Interesting. I noticed an issue with enums, such as

```ts
enum Cats {
  Anakin;
  Prosecco;
}
```

Maybe this is the same issue you found, since the default values will be `0` and `1`.

packages/components/src/Button/button.stories.tsx
```diff
@@ -1,8 +1,28 @@
...
+  "warning",
+];
+const variants: Array<ButtonProps["variant"]> = ["flat", "outline", "link"];
+const states: Array<ClickableProps<"button">["state"] | "disabled"> = [
+  "inactive",
+  "hover",
+  "focus",
+  "active",
+  "disabled",
+];
```
ahuth, line 25: Can we get rid of the explicit typing with something like

```ts
const states = [
  // ... items
] as const;
```

My thought process is this should result in a "tuple" type, where typescript understands what specific value each item in the array is.

As we iterate that array, we should end up with a type that TS lets us pass to the component

```ts
states.map((state) => {
  // TypeScript should understand here that state has the type of
  // 'inactive' | 'hover' | 'focus' | etc... , which we pass to `state`.
  return <SomeComponent state={state} />;
});
```

[Can see something like this working in the TS playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAysCGwIGcoF4oHIBmB7XmUAPlgEbwBOhJm5AXpgNwBQz2ArgHYDGwAlrk5QkyYAApRiCAC5YCEQEooAbwC+rboNFRR7bNnRQA2jnyYANGUoWrDALpR4qTZ1Etmu-QDoAtvDBiYsAAFnycAOZKaAB8KsxQCcIo4iFhkSyqCoxAA).

Hoping that works for all of these consts.

packages/components/src/Button/button.stories.tsx
```diff
@@ -101,3 +139,79 @@ withFakeClassName.args = {
...
+            {states.map((state) => (
+              <tr key={state}>
+                <th scope="row">
+                  <Text size="body" color={textColor}>
+                    {state}
+                  </Text>
+                </th>
+                {colors.map((color) => (
+                  <td key={color} className={styles["color"]}>
+                    <Clickable
```
ahuth, line 179: I think I asked this before, but why use Clickable instead of a button?

I may not understand what Clickable is, but I thought it was the base for button and links, but folks should import one of those.

packages/components/src/Button/button.stories.tsx
```diff
@@ -101,3 +139,79 @@ withFakeClassName.args = {
...
+  variants.map((variant) => {
+    // For now, the UI kit only includes alert & brand "flat" buttons
+    const colors = variant === "flat" ? ["alert", "brand"] : allColors;
+
+    return (
+      <React.Fragment key={variant}>
+        <Heading size="h2" color={textColor}>
+          {variant} - {size}
+        </Heading>
+        <table className={styles["variant"]}>
```
ahuth, line 168: Really doesn't matter, but I find it easier to type

```suggestion
        <table className={styles.variant}>
```

packages/components/src/Toast/Toast.stories.tsx
```diff
@@ -13,6 +13,7 @@ const Template: Story<Args> = (args: Args) => <Toast {...args} />;
 
 const defaultArgs = {
   children: "You've got toast!",
+  onDismiss: () => console.log("dismissed!"),
```
ahuth, line 16: Should we leave this off and let Storybook supply it's own handler that'll work with the actions addon?

That's a little indirect/hidden, but it can also be kind of nice to tie in to Storybook.

packages/components/src/Toast/Toast.stories.tsx
```diff
@@ -27,9 +28,9 @@ Alert.args = {
...
 };
 
-export const Dismissable = Template.bind(null);
-Dismissable.args = {
+export const NotDismissable = Template.bind(null);
+NotDismissable.args = {
   ...defaultArgs,
   color: "success",
-  onDismiss: () => console.log("dismissed!"),
+  onDismiss: undefined,
```
ahuth, line 35: If we do tie into the Storybook actions addon with the default args, I think we'd unfortunately have to make this `null`.

Otherwise the actions addon thinks we haven't passed anything and it should pass its own handler.

packages/components/src/Toast/Toast.stories.tsx
anniehu4: oh! yep, I'll try it again. honestly would be fine getting rid of the `NotDismissable` story anyways

packages/components/src/Button/button.stories.tsx
anniehu4: oops I didn't even notice that's what was there earlier. idk why this wouldn't be a button either, will update

packages/components/src/Button/button.stories.tsx
anniehu4: will try!

packages/components/src/Button/button.stories.tsx
anniehu4: ah I see -- it's because the `state` prop isn't allowed on `Button`, but we added it to `Clickable` just for storybook purposes. I wanted to restructure Clickable as a `createClickable` factory at some point, but for now I left a comment here

packages/components/src/Toast/__snapshots__/Toast.spec.tsx.snap
```diff
@@ -69,29 +69,6 @@ exports[`<Toast /> Dismissable story renders snapshot 1`] = `
       You've got toast!
     </p>
   </div>
-  <button
-    class="close button variantLink colorSuccess"
-    type="button"
-  >
```
anniehu4, line 75: side effect of the actions addOn is `onDismiss` isn't in the prepared story for these snapshots. I think it's fine since we have Percy testing here anyways

packages/components/src/Toast/Toast.stories.tsx
```diff
@@ -27,9 +28,12 @@ Alert.args = {
...
-Dismissable.args = {
+export const NotDismissable = Template.bind(null);
+NotDismissable.args = {
   ...defaultArgs,
   color: "success",
-  onDismiss: () => console.log("dismissed!"),
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore onDismiss is not nullable, but this is needed to remove the arg from
+  // storybook's actions addon
+  onDismiss: null,
```
dierat, line 38: Nice!

dierat (approved):

anniehu4:
I thiiink the `test (10.x)` step hanging is expected because this PR was opened before the yarn changes

ahuth:
Re: test (10.x) hanging, I'm actually seeing that on branches that came after the yarn changes 😞.

I'm going to open a support ticket with github about that.

PR #566, opened 2021-08-06
Interactive releases

ahuth:
### Summary:

Added a `create-releases:interactive` command that opens a CLI prompt asking you what version you want to create. This is useful if you know you want to publish a specific version, instead of following the conventional commits.

Screen Shot 2021-08-06 at 9 54 22 AM

For example, until we hit v1.0 we're going to follow
- Bump the minor version (0.y.0) when introducing a breaking change
- Bump the patch version (0.0.x) when introducing backwards compatible changes

Conventional commits won't let us do that, since it follows semver.

This is mostly to make it easier on my so I don't forget.

### Test Plan:

- Local testing


package.json
```diff
@@ -10,8 +10,9 @@
     "build": "NODE_ENV=production yarn workspaces foreach --parallel --topological run build",
     "build:clean": "yarn workspaces foreach --exclude eds-tokens run build:clean",
     "create-component": "plop",
-    "create-releases": "lerna version --no-push --conventional-commits --message 'chore(release): %s'",
+    "create-releases": "yarn create-releases:interactive --conventional-commits",
     "create-releases:alpha": "yarn create-releases --preid alpha prerelease",
+    "create-releases:interactive": "lerna version --no-push --message 'chore(release): %s'",
```
anniehu4, line 15: ooh that prompting is awesome! will this still update the `CHANGELOG` files automatically?

ahuth: Doh, I don't think it will. Good catch.

I'll settle instead of manually running things like
- `yarn create-releases minor`
- `yarn create-releases patch`

when needed.

PR #573, opened 2021-08-06
Misc Cleanups

ahuth:
### Summary:

This PR makes a couple miscellaneous cleanups

- ~~Uses [workspace ranges](https://yarnpkg.com/features/workspaces/#workspace-ranges-workspace) to define dependencies between packages to work around https://github.com/lerna/lerna/issues/1171~~
- Removes unnecessary duplicate commitlint deps
- Removes unused classnames dep
- Removes some instances of `lerna run` that I missed in #533.

### Test Plan:

- CI
- [x] Docs deploy
- [x] Tested bad commit messages
- [ ] ~~Alpha version publish~~ This didn't work, so I'll need to figure something else out. Removing the workspace ranges stuff


packages/components/package.json
```diff
@@ -49,7 +49,7 @@
     "react": ">= 16.8.0"
   },
   "dependencies": {
-    "@chanzuckerberg/eds-tokens": "^0.1.2",
+    "@chanzuckerberg/eds-tokens": "workspace:*",
```
ahuth, line 52: Instead of specifying a version, we tell Yarn to only match the version in the monorepo.

Doing this doesn't directly solve https://github.com/lerna/lerna/issues/1171, but works around it because we no longer need the yarn.lock file to be updated when publishing.

Also note that https://github.com/lerna/lerna/issues/2564 means this change will cause any `lerna run` calls to fail. That should be fine, because we can run anything through `yarn workspaces foreach`.

I will test that publishing still works.

packages/components/package.json
```diff
@@ -61,9 +61,6 @@
     "@babel/preset-typescript": "^7.12.17",
     "@chanzuckerberg/axe-storybook-testing": "^3.0.2",
     "@chanzuckerberg/story-utils": "^2.0.0",
-    "@commitlint/cli": "^8.3.5",
-    "@commitlint/config-conventional": "^8.3.4",
-    "@commitlint/prompt-cli": "^8.3.5",
```
ahuth, line 66: No need to have v8 here and v11 at the top level.

anniehu4 (approved):
https://app.clubhouse.io/czi-edu/story/132355/remove-classnames-import-in-favor-of-using-clsx-from-eds

anniehu4:
https://app.clubhouse.io/czi-edu/story/132355/remove-classnames-import-in-favor-of-using-clsx-from-eds

PR #575, opened 2021-08-06
Ensure the yarn.lock file gets updated when publishing

ahuth:
Hat tip @anniehu4 for the idea here

### Summary:

Currently Lerna doesn't update yarn.lock files when dependencies between monorepo packages increment. See https://github.com/lerna/lerna/issues/1171.

To work around that, we manually install deps during publishing so that the lock file gets updated.

Another workaround would be to [use Yarn workspace ranges to specify dependencies between package](https://yarnpkg.com/features/workspaces/#workspace-ranges-workspace), so that there's no need to reference specific version numbers.

Unfortunately, Lerna also doesn't support that, either - https://github.com/lerna/lerna/issues/2564.

### Test Plan:

- [x] Create a test release (didn't publish it)

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,8 @@
     "publish-releases:alpha": "yarn publish-releases --dist-tag alpha",
     "start": "yarn workspaces foreach run start",
     "test": "yarn workspaces foreach run test",
-    "types": "yarn workspaces foreach run types"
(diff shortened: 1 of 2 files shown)


anniehu4 (approved):
thanks!

PR #578, opened 2021-08-09
chore: use eslint-plugin-import/order

anniehu4:
### Summary:
[ch151882]

our current lint rule doesn't auto-fix import order issues, which is hellaaa annoying. this switches to use the rule that `traject` already uses.

Note: one alternative was [eslint-plugin-simple-import-sort](https://github.com/lydell/eslint-plugin-simple-import-sort/#how-is-this-rule-different-from-importorder), which is mainly nice b/c it can also sort exports. Decided against it though because
- it's only maintained by one person, lol
- it's probably good to have more predictable behavior w/ traject

### Test Plan:
- ran `yarn run lint:scripts:fix` and verified all import order issues were fixed
- build passes

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,7 +4,7 @@
     "browser": true
   },
   "parser": "@typescript-eslint/parser",
-  "plugins": ["react-hooks", "testing-library"],
+  "plugins": ["import", "react-hooks", "testing-library"],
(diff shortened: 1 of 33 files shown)


ahuth (approved):
LGTM, thanks for this

PR #580, opened 2021-08-10
Misc cleanups 2

ahuth:
### Summary:

This PR
- Fixes the eds-tokens build script, which I broke in #533 
- Improves the output of the build and start commands

### Test Plan:

- CI

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "packages/*"
   ],
   "scripts": {
-    "build": "NODE_ENV=production yarn workspaces foreach --parallel --topological run build",
+    "build": "NODE_ENV=production yarn workspaces foreach --interlaced --parallel --topological --verbose run build",
     "build:clean": "yarn workspaces foreach --exclude eds-tokens run build:clean",
     "create-component": "plop",
     "create-releases": "lerna version --no-push --conventional-commits --message 'chore(release): %s'",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -21,7 +21,7 @@
     "prettier": "prettier --write \"**/*.{js,jsx,ts,tsx}\"",
     "publish-releases": "lerna publish from-git --no-push",
     "publish-releases:alpha": "yarn publish-releases --dist-tag alpha",
-    "start": "yarn workspaces foreach run start",
+    "start": "yarn workspaces foreach --interlaced --parallel --verbose run start",
```
ahuth, line 24: - `--interlaced` prints output as it comes in from the test commands, instead of buffering and printing all at once
- `--parallel` runs the commands in parallel
- `--verbose` prefixes each line of output with the name of the package that generated it

packages/tokens/package.json
```diff
@@ -21,7 +21,7 @@
     "build": "style-dictionary build --config=./style-dictionary.config.js",
     "build:clean": "rm -rf lib/",
     "prepack": "yarn run build",
-    "start": "watch 'yarn run build' ./properties ./node_modules"
+    "start": "watch 'yarn run build' ./properties"
```
ahuth, line 24: There's no longer a node_modules directory that's local to the eds-tokens package, so this command was broken.

I believe most folks would restart any running processes when installing dependencies anyway, so we don't need to watch the node_modules directory.

ahuth:
I accepted the Percy diff: https://percy.io/0fae89a3/Edu-Design-System/builds/12013820/changed/676404644

Although it's kind of odd. Not sure why one of the banner stories would've been blank before (on main).

package.json
anniehu4: checking for understanding -- this doesn't need `--topological` because we're watching the tokens build in that `start` script?

anniehu4 (approved):

package.json
ahuth: Not exactly. Or, well, that's sort of it.

(For posterity) `--topological` makes the scripts run in dependency order. So first eds-tokens would run, then eds-components (since it has a dependency on eds-tokens).

That's great for the build scripts, since they run for a finite amount of time and then exit.

But `start` doesn't exit (until you ctrl-c it). So the eds-tokens start script would never exit, so the eds-components one never starts.

Hope that sort of made sense.

ahuth:
Thanks!

Still no word on the Node 10 test job...

PR #582, opened 2021-08-10
Allow buttons to take a ref

ahuth:
### Summary:

@pomfy113 and I were working on focus management for https://app.clubhouse.io/czi-edu/story/153582/screen-reader-ssp-adoption-setup-improvements-nvda-announces-extra-text-blank-when-press-enter-key-on-x-button-besides-the, and realized we can't use React refs (to manage focus) with the react-bootstrap buttons. 

React-bootstrap's buttons have refs, but get you an instance of the React class, not the element.

Decided it was a good opportunity to switch over to the EDS buttons, but realized that they don't take refs. This PR adds ref support to buttons.

### Test Plan:

- CI

### DIFF
--- a/packages/components/src/Button/button.tsx
+++ b/packages/components/src/Button/button.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode } from "react";
+import React, { ReactNode, forwardRef } from "react";
 import Clickable, { ClickableProps } from "../Clickable";
 
 type ButtonHTMLElementProps = React.ButtonHTMLAttributes<HTMLButtonElement>;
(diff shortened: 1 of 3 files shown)


pomfy113 (approved):

packages/components/src/Clickable/Clickable.tsx
```diff
@@ -30,17 +30,19 @@ export type ClickableProps<IComponent extends React.ElementType> = {
...
-  size,
-  state,
-  variant,
-  className,
-  ...rest
-}: ClickableProps<IComponent>) {
-  const Component = as;
+const Clickable = React.forwardRef(function <
+  IComponent extends React.ElementType,
+>(props: ClickableProps<IComponent>, ref: React.ForwardedRef<HTMLElement>) {
```
ahuth, line 35: Odd formatting by prettier (which comes with the territory of using it sometimes).

Chose to use a regular function instead of arrow here so that prettier would keep the indentation of the function body as-is, so I don't have to touch every line of this component.

packages/components/src/Button/button.tsx
```diff
@@ -15,15 +15,16 @@ export type ButtonProps = ButtonHTMLElementProps & {
...
-  ...rest
-}: ButtonProps) {
+const Button = forwardRef<HTMLButtonElement, ButtonProps>((props, ref) => {
+  const {
+    as = "button",
+    variant = "flat",
+    color = "brand",
+    disabled = false,
+    type = "button",
+    size = "medium",
```
anniehu4, line 25: does storybook still show these default values correctly? I remember a while ago we had to destructure the other way to preserve that

anniehu4 (commented):
thanks! one Q about keeping the prop destructuring where it is for storybook

packages/components/src/Button/button.tsx
ahuth: That's a nope. Will change

ahuth: Good catch

anniehu4 (approved):
thanks!

ahuth:
Thanks!

PR #594, opened 2021-08-16
Remove dependabot automatic PRs

ahuth:
### Summary:

This PR removes automatic dependabot PRs. We will still get notifications for security vulnerabilities.

I haven't found the automatically generated PRs to be that useful. And I didn't update dependabot's config when migrating to Yarn in #533, so it's not actually producing working PRs at the moment (of course we could fix that).

### Test Plan:

- CI


anniehu4:
> We will still get notifications for security vulnerabilities.

Through email? if we've been getting them so far, I haven't noticed 😅 

> I haven't found the automatically generated PRs to be that useful

I agree, though I'll admit I'm new to keeping up dependency management. Is there a regular schedule that we should set to update these, if there's no dependabot? (I don't want you to have to manually do it every single time because nobody else remembers)

Alternatively, could we set a lower frequency for dependabot checking and do a better job of actually landing them? Just thinking through what's most sustainable, curious what's worked best for you & @dcwither in the past

ahuth:
> Through email? if we've been getting them so far, I haven't noticed 😅

No, I haven't seen any email notifications for security vulnerabilities. Was referring to seeing the number of vulnerabilities in the "security" tab, like this (from LP, since this repo doesn't have any vulnerabilities currently):

Screen Shot 2021-08-16 at 10 38 45 AM

These dependabot updates are separate from security updates though (maybe a superset?). We could turn on dependabot update PRs specifically for security vulnerabilities, if we wanted to.

---

We could go with a lower frequency, or generally try to be better at landing the PRs. Would take updating this config to specify Yarn.

PR #598, opened 2021-08-19
feat: update neutral button colors

anniehu4:
### Summary:
[ch156491]

Noticed in the [UI Kit](https://www.figma.com/file/YPTOEVRuas5ytXQEcwm4FK/UI-Kit-Sticker-Sheet-(Core)?node-id=226%3A1100) that the Tertiary button is actually using `neutral-500` instead of `-600`. The darkness difference is noticeable enough that a product team was hesitant to use the new EDS button.

Note: the UI kit uses `-500` for the hover & active states as well, which is inconsistent with the other button color implementations. I'm choosing to keep consistency for now (b/c this update already brings it a lot closer to those designs), and can clarify more w/ Sean

### Test Plan:
- see Storybook changes
- confirmed color contrast rules in Storybook still all passed (on the "All Variants" story)

### DIFF
--- a/packages/components/src/Clickable/Clickable.module.css
+++ b/packages/components/src/Clickable/Clickable.module.css
@@ -68,9 +68,9 @@
     --button-primary-color: var(--eds-color-alert-300);
   }
 }
-  
+
 .colorNeutral {
-  --button-primary-color: var(--eds-color-neutral-600);
+  --button-primary-color: var(--eds-color-neutral-500);
   --button-secondary-color: var(--eds-color-white);
   --button-tertiary-color: var(--eds-color-neutral-100);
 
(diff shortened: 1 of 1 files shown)


packages/components/src/Clickable/Clickable.module.css
```diff
@@ -68,31 +68,31 @@
...
   &:focus,
   &.stateFocus {
-    --button-primary-color: var(--eds-color-neutral-700);
+    --button-primary-color: var(--eds-color-neutral-600);
   }
 
   &:active,
   &.stateActive {
     /* TODO: Replace with neutral active color */
-    --button-primary-color: var(--eds-color-neutral-700);
```
dierat, line 87: Can we leave this one as 700? The active state is supposed to be darker than the focus state, but it's currently the same. Dropping the other states down a level actually allows us to have the same value increase between these two states as the other hues.

dierat (approved):

packages/components/src/Clickable/Clickable.module.css
anniehu4: yeah I wavered on this one, but makes sense 👍

PR #617, opened 2021-08-27
feat(link): add Link component

dierat:
### Summary:

Our `Button` component has styles for both buttons and links, but there is no actual "Link" component. This adds that component. It's basically just a copy of `Button` but the defaults are set to make it look like a link (even though you can totally make it look like a button).

I was getting a little hung up on figuring out how to best consolidate the stories between `Button` and `Link`, so I decided to defer that work and just copy over all the stories for now.

We have also discussed changing `Button` to always create a `button` element and make this `Link` component always create an `a` element, but I'm deferring that as well. One thing I just now noticed is that, based on the type for the `as` prop, is that `Button` doesn't seem to actually be able to create an `a` element, only `button` or some other component. So maybe we can leave the `as` as-is.

### Screenshot
Link component in storybook, scrolled to the 'all variants' story, inspecting one of the buttons with Chrome devtools to show that it's using an anchor tag

### Test Plan:
Navigate to `http://localhost:6008/?path=/docs/link--primary`,
verify the `Link` stories are just like the `Button` stories except the button text says "Link" and it's creating an `a` element instead of a `button`.


packages/components/src/Link/Link.stories.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+export const TertiarySmall = Template.bind(null);
+TertiarySmall.args = {
+  ...defaultArgs,
+  color: "neutral",
+  size: "small",
+  variant: "outline",
+};
+
+export const LinkButton = Template.bind(null);
```
anniehu4, line 95: Would suggest:
- Removing all the stories above this (they're more design guidance for when to use what style, I added them to make Zeroheight match the UI Kit)
- renaming "Link Button" to "Link" throughout

packages/components/src/Link/Link.stories.tsx
```diff
@@ -0,0 +1,242 @@
...
+type Args = React.ComponentProps<typeof Link>;
+
+const Template: Story<Args> = (args) => <Link {...args} />;
+
+// TODO: Consolidate with button stories.
+
+const defaultArgs = {
+  children: "Link",
+  variant: "flat" as const,
+  size: "medium" as const,
```
anniehu4, line 45:
```suggestion
  variant: "link" as const,
```

packages/components/src/Link/Link.tsx
```diff
@@ -0,0 +1,31 @@
...
+  "disabled"
+>;
+
+export type LinkProps = LinkHTMLElementProps & {
+  as?: "a" | React.ComponentType<any>;
+  /**
+   * The link contents or label.
+   */
+  children: ReactNode;
+  color?: ClickableProps<"button">["color"];
```
anniehu4, line 15: should this be `"a"` instead of `"button"`?

packages/components/src/Link/Link.stories.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+export const mediumVariantsOnDarkBackground = () =>
+  renderSize("medium", "white", "Link");
+
+mediumVariantsOnDarkBackground.parameters = {
+  ...gridParameters,
+  backgrounds: {
+    default: "dark",
+  },
+};
```
anniehu4, line 242: Think you could replace this with

```
import * as ButtonStoryFile from "../Button/button.stories";

export const allVariants = ButtonStoryFile.allVariants;
allVariants.parameters = gridParameters;

export const mediumVariantsOnDarkBackground =
  ButtonStoryFile.mediumVariantsOnDarkBackground;
mediumVariantsOnDarkBackground.parameters = {
  ...gridParameters,
  backgrounds: {
    default: "dark",
  },
};
```

packages/tokens/lib/json/css-variables-nested.json
```diff
@@ -179,4 +179,4 @@
       }
     }
   }
-}
+}
```
anniehu4, line 182: ugh I really want to find a way to stop this

anniehu4: OH actually, this is great, idk why we added the newlines in the first place

packages/components/src/Link/Link.stories.tsx
dierat: Would you suggest this change for the `Button` component stories as well?

packages/components/src/Link/Link.stories.tsx
dierat: The only thing is the `Button` component is using a `button` tag in the grid and it would be really nice to have this one use an anchor tag for more thorough testing.

packages/components/src/Link/Link.tsx
dierat: Good call, thanks!

packages/components/src/Link/Link.stories.tsx
dierat: Oh and the other difference is the `disabled` row isn't applicable for this component since you can't disable links.

packages/components/src/Button/button.tsx
```diff
@@ -15,6 +15,13 @@ export type ButtonProps = ButtonHTMLElementProps & {
   variant?: ClickableProps<"button">["variant"];
 };
 
+/**
+ * Component for making buttons that do not navigate the user to another page.
+ *
+ * This component is called Button because it should be used to making button elements;
+ * however, it can be styled to look like a link. In terms of the look and feel of the
+ * component in the UI, the Button and Link components are exactly the same.
+ */
```
dierat, line 24: Added a docstring to the `Button` component that's a mirror image of the one on the `Link`.

packages/components/src/Link/Link.stories.tsx
anniehu4: oh, right 👍

packages/components/src/Link/Link.tsx
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Component for making links that navigate to a different page that is outside of the
+ * product (ie help desk or external resource).
+ *
+ * This component is called Link because it should be used to making link elements;
+ * however, it can be styled to look like a button. In terms of the look and feel of the
+ * component in the UI, the Link and Button components are exactly the same.
+ */
+const Link = forwardRef<HTMLAnchorElement, LinkProps>(
+  ({ as = "a", variant = "link", color = "brand", ...rest }, ref) => {
```
anniehu4, line 30: looking at the "Link: Button Variant With Icon" story, maybe this should also pass `size = "medium"` by default? if variant is "link", that prop will be ignored, but otherwise people might expect that default

packages/components/src/Link/Link.stories.tsx
anniehu4: showing the grids first makes sense 👍  re: `Button` component stories, maybe we can show just one link variant there and rename it `linkVariant`

anniehu4 (approved):
last comment about `size` prop but otherwise lgtm!

packages/components/src/Button/button.stories.tsx
```diff
@@ -37,6 +37,8 @@ type Args = React.ComponentProps<typeof Button>;
 
 const Template: Story<Args> = (args) => <Button {...args} />;
 
+// TODO: Consolidate with Link stories.
```
ahuth, line 40: Since Link and Button are separate components, should we skip consolidating them? It seems reasonable to me to keep their stories completely separate.

packages/components/src/Link/Link.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { ReactNode, forwardRef } from "react";
+import Clickable, { ClickableProps } from "../Clickable";
+
+type LinkHTMLElementProps = Omit<
+  React.AnchorHTMLAttributes<HTMLAnchorElement>,
+  "disabled"
+>;
+
+export type LinkProps = LinkHTMLElementProps & {
+  as?: "a" | React.ComponentType<any>;
```
ahuth, line 10: Since Link is for `<a>` and Button is for `<button>`, can we get rid of the `as` prop?

I can't think of too many scenarios where we'd want to allow arbitrary components. Maybe a specific Bootstrap or React-router link?

packages/components/src/Link/Link.tsx
```diff
@@ -0,0 +1,39 @@
...
+  children: ReactNode;
+  color?: ClickableProps<"a">["color"];
+  "data-testid"?: string;
+  size?: ClickableProps<"a">["size"];
+  variant?: ClickableProps<"a">["variant"];
+};
+
+/**
+ * Component for making links that navigate to a different page that is outside of the
+ * product (ie help desk or external resource).
```
ahuth, line 23: Couldn't a link also be to a location within the product?

packages/components/src/Link/Link.spec.tsx
```diff
@@ -0,0 +1,19 @@
...
+import * as LinkStoryFile from "./Link.stories";
+
+describe("<Link />", () => {
+  generateSnapshots(LinkStoryFile);
+
+  it("forwards refs", () => {
+    const ref = React.createRef<HTMLAnchorElement>();
+    render(prepareStory(LinkStoryFile.StandardLink, { ref }));
+
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
ahuth, line 13: 🤔 Maybe we should disable this lint rule for all test files

packages/components/src/Link/Link.stories.tsx
```diff
@@ -0,0 +1,184 @@
...
+          <tbody>
+            {states.map((state) => (
+              <tr key={state}>
+                <th scope="row">
+                  <Text size="body" color={textColor}>
+                    {state}
+                  </Text>
+                </th>
+                {colors.map((color) => (
+                  <td key={color} className={styles.color}>
```
ahuth, line 108: Would be nice to directly use Tailwind classes, instead of having to create a CSS file.

Could be interesting at some point to experiment with integrating Tailwind (ideally in [JIT mode](https://tailwindcss.com/docs/just-in-time-mode)) into our Storybook.

ahuth (commented):
❤️ Code looks perfect. Left a couple questions

packages/components/src/Link/Link.tsx
dierat: The react-router link is definitely the main usage I'm expecting here. I think we're going to need some way to apply the button/link designs to other components and elements.

@anniehu4 and I have talked a bit about whether we should drop the `as` from the `Button` and `Link` components and have people use `Clickable` instead (but maybe with a different name) so the `Link` and `Button` components' purpose is clearer since they both just apply the same styles anyway. Thoughts?

packages/components/src/Link/Link.tsx
dierat: You probably know better than I do since I haven't worked in LP much. My interpretation is that the react router should be used for in-product navigation. I wouldn't want people to think that hey we should be using the new design system link component and miss the react router pattern when we're not actually trying to change that by adding the new component.

packages/components/src/Button/button.stories.tsx
dierat: We definitely don't need to (and it would be less work so ayyyy 👍 👍 ), but I figure it makes sense since they can create the exact same styles and the separation is more related to their underlying element than the appearance. I have a general concern about devs thinking you use `Link` if it looks like a link and `Button` if it looks like a button.

packages/components/src/Link/Link.stories.tsx
dierat: Good idea. I think I'll take another look at the stories in a separate PR so we can get the initial component out first.

packages/components/src/Button/button.stories.tsx
ahuth: >  I have a general concern about devs thinking you use Link if it looks like a link and Button if it looks like a button.

Legit concern 👍 . We'll have to educate and document (which you've already done) that you use Link to get a link, and use Button to get a button, regardless of styling.

To me, consolidating stories seems like it would increase confusion, not decrease. So I'm thinking separate makes the most sense.

Plus the shared styles via Clickable is more of an implementation detail, not inherent to the API, I think.

packages/components/src/Link/Link.tsx
ahuth: Seems reasonable. Expose Clickable for arbitrary components (which if I understand correctly is exactly why it exists). And ensure Link and Button get you exactly those things.

packages/components/src/Link/Link.tsx
ahuth: > My interpretation is that the react router should be used for in-product navigation

I think this is generally correct currently.

My thinking is this is an aspect of LP right now, but not something that the design system should really be worried about. The design system shouldn't really have LP-isms baked into it (within reason).

But at the end of the day, we're talking about semantics in a comment 😄, so either way is fine.

Maybe we could handle both sides by making the comment something like

> Component for making anchor tags.
>
> Note: if using a routing component like react-router, you'll probably need to pass in its Link component
> via the `as` prop.

ahuth (approved):

packages/components/src/Link/Link.tsx
dierat: I like that, thanks! 👍

packages/components/src/Button/button.stories.tsx
dierat: I'll go ahead and remove the todo comments 👍

dierat:
Going to merge this now; we can talk more about the `as` prop and stories in follow-up PRs.

PR #622, opened 2021-08-31
chore(banner): update api to use title and message instead of children

dierat:
### Summary:
Clubhouse ticket: https://app.clubhouse.io/czi-edu/story/157238/update-banner-api

The `Banner` component is pretty complex, and we're playing around with the API to hopefully be easier to reason about. We're concerned that it may be confusing to pass in the `action` as a prop while the heading and body are the `children`.

We want the `action` to be a prop to control its position (left side when the banner is laid out horizontally; centered when in mobile or card format) and the gab between it and the text. Right now, we're just assuming any `children` passed in is intended to be the text (heading and body) that appears above the `action`.

This PR removes the `children` prop and replaces it with one or more props for the text to be passed in.

#### First pass
https://github.com/chanzuckerberg/edu-design-system/pull/622/commits/448801c5897dbbc1177a8e0289c7253f8ad728a5

The suggested API in the clubhouse ticket looks like this:
```
<Banner
  action={<button>brew</button>}
  title={<h2>Coffee</h2>}
  message="is amazing"
/>
```
I'm a little concerned about the `title` usage here because I think it's confusing to usually expect devs to use components like `Heading` and `Text` for text except in this instance. So I was thinking that we wouldn't style `title` or wrap it in a `Heading` or anything so devs can pass in a `Heading` component or `Banner.Title` themselves.

But in that version, `title` and `message` don't really need to be passed in as separate props because we're not doing anything special to either of them.

#### Second pass
https://github.com/chanzuckerberg/edu-design-system/pull/622/commits/6ad905185de33bc40a465dba92b08a39ccd21a12

In the second commit, I combined `title` and `message` into one prop, `textContent`. So the only difference between this version and the current API is the `children` prop has another name and the contents needs to be passed in as a regular prop. If that makes sense.

### Test Plan:
There should be no UI changes picked up in the Percy test.

### DIFF
--- a/packages/components/src/Banner/Banner.stories.tsx
+++ b/packages/components/src/Banner/Banner.stories.tsx
@@ -30,7 +30,7 @@ const Template: Story<Args> = (args) => {
       textContent={
         <>
           <Banner.Title as="h1">{heading}</Banner.Title>{" "}
-          <Banner.Body>{content}</Banner.Body>
+          <Banner.Message>{content}</Banner.Message>
         </>
       }
     />
@@ -148,7 +148,7 @@ export const DismissableBelowContent = () => (
       textContent={
         <>
           <Banner.Title as="h1">{dialogArgs.heading}</Banner.Title>{" "}
-          <Banner.Body>{dialogArgs.content}</Banner.Body>
+          <Banner.Message>{dialogArgs.content}</Banner.Message>
         </>
       }
     />
--- a/packages/components/src/Banner/Banner.stories.tsx
+++ b/packages/components/src/Banner/Banner.stories.tsx
@@ -27,8 +27,12 @@ const Template: Story<Args> = (args) => {
   return (
     <Banner
       {...restArgs}
-      title={<Banner.Title as="h1">{heading}</Banner.Title>}
-      message={<Banner.Body>{content}</Banner.Body>}
+      textContent={
+        <>
+          <Banner.Title as="h1">{heading}</Banner.Title>{" "}
+          <Banner.Body>{content}</Banner.Body>
+        </>
+      }
     />
   );
 };
@@ -141,8 +145,12 @@ export const DismissableBelowContent = () => (
     </Heading>
     <Banner
       onDismiss={() => console.log("dismissed!")}
-      title={<Banner.Title as="h1">{dialogArgs.heading}</Banner.Title>}
-      message={<Banner.Body>{dialogArgs.content}</Banner.Body>}
+      textContent={
+        <>
+          <Banner.Title as="h1">{dialogArgs.heading}</Banner.Title>{" "}
+          <Banner.Body>{dialogArgs.content}</Banner.Body>
+        </>
+      }
     />
   </>
 );
--- a/packages/components/src/Banner/Banner.stories.tsx
+++ b/packages/components/src/Banner/Banner.stories.tsx
@@ -25,10 +25,11 @@ type Args = React.ComponentProps<typeof Banner> & {
 const Template: Story<Args> = (args) => {
   const { content, heading, ...restArgs } = args;
   return (
-    <Banner {...restArgs}>
-      <Banner.Title as="h1">{heading}</Banner.Title>
-      <Banner.Body>{content}</Banner.Body>
-    </Banner>
+    <Banner
+      {...restArgs}
+      title={<Banner.Title as="h1">{heading}</Banner.Title>}
+      message={<Banner.Body>{content}</Banner.Body>}
+    />
   );
 };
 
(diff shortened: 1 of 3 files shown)


packages/components/src/Banner/Banner.tsx
```diff
@@ -19,9 +19,11 @@ type Props = {
    */
   color?: NotificationVariant;
   /**
-   * The contents of the banner in addition to the icon
+   * The text content of the banner.
+   *
+   * Please note that this can contain text links, but block buttons should be passed in via the `action` prop.
    */
-  children: React.ReactNode;
+  textContent: React.ReactNode;
```
ahuth, line 26: I like your initial API in 448801c5897dbbc1177a8e0289c7253f8ad728a5

```jsx
<Banner
  {...restArgs}
  title={<Banner.Title as="h1">{heading}</Banner.Title>}
  message={<Banner.Body>{content}</Banner.Body>}
/>
```

although maybe we don't need to manually provide Banner.Body

```jsx
<Banner
  {...restArgs}
  title={<Banner.Title as="h1">{heading}</Banner.Title>}
  message={content}
/>
```

The PR description mentions why you changed it and went with this API

> But in that version, title and message don't really need to be passed in as separate props because we're not doing anything special to either of them.

One thing that is nice about keeping the separate `title` and `message` props (even if we're not doing anything special with either) is the logical separation. It may be more obvious what kind of content does and does not go in there.

Once we have a single prop

```jsx
textContent={
  <>
    <Banner.Title as="h1">{heading}</Banner.Title>{" "}
    <Banner.Body>{content}</Banner.Body>
  </>
}
```

It might be a little more natural to start putting other stuff (like buttons) in there too

```jsx
textContent={
  <>
    <Banner.Title as="h1">{heading}</Banner.Title>{" "}
    <Banner.Body>{content}</Banner.Body>
    <Button>click me!</Button>
  </>
}
```

But it's a small thing, so 🤷

ahuth (approved):
LGTM

packages/components/src/Banner/Banner.tsx
anniehu4: I agree ... I wonder if we could stick with the `title` and `message` props and get rid of `Banner.Title` and `Banner.Body`? Similar to what Andrew suggested:
```
<Banner {...restArgs} title={heading} message={content} />
```

Then internally we still render
```
// introduce a new headerAs prop as well
<Heading as={props.headerAs || "h2"} color={"inherit"} size="h3">
  {props.title}
</Heading>

<Text color={"inherit"} size="body">
  {props.message}
</Text>
```

Thoughts?

ahuth: Agreed that it would be nice to get rid of the Banner.Title and Banner.Body components (could still have internal components that wrap what people pass in, as Annie suggested).

Not sure if there are any ramifications of that or not, though.

ahuth:
Side note: tweaking the component API to avoid mistakes reminds me of [Poka-yoke](https://en.wikipedia.org/wiki/Poka-yoke) (not sure how to pronounce. Japanese for "mistake-proof") from manufacturing.

dierat:
> [Poka-yoke](https://en.wikipedia.org/wiki/Poka-yoke) (not sure how to pronounce

Cool word!

poh-KAH-yoh-keh https://translate.google.com/?sl=ja&tl=en&text=%E3%83%9D%E3%82%AB%E3%83%A8%E3%82%B1&op=translate

packages/components/src/Banner/Banner.stories.tsx
```diff
@@ -25,10 +25,15 @@ type Args = React.ComponentProps<typeof Banner> & {
...
-    <Banner {...restArgs}>
-      <Banner.Title as="h1">{heading}</Banner.Title>
-      <Banner.Body>{content}</Banner.Body>
-    </Banner>
+    <Banner
+      {...restArgs}
+      textContent={
+        <>
+          <Banner.Title as="h1">{heading}</Banner.Title>{" "}
+          <Banner.Body>{content}</Banner.Body>
```
anniehu4, line 33: I wonder if it'd help also to rename this to `Banner.Message`? "Body" does seem to communicate there can be all sorts of things in there

anniehu4 (approved):
let's try it! one plus of this is it's pretty close to the previous api (just move children into `textContent`)

packages/components/src/Banner/Banner.tsx
anniehu4: I think `message` would need to accept a ReactNode type anyways because of the link (and maybe both should allow it to be more flexible), so that case would still be fine. but I see how `textContent` makes more sense there

dierat: I'm also fine with the idea of using `title` and `message` even if they don't make any styling changes to the content just to make it clearer what we expect you to pass in.

dierat: I guess I'm just talking about the first commit https://github.com/chanzuckerberg/edu-design-system/commit/448801c5897dbbc1177a8e0289c7253f8ad728a5, but with renaming `Banner.Body` per @anniehu4 's other comment https://github.com/chanzuckerberg/edu-design-system/pull/622#discussion_r706338844

anniehu4: I have mild concern that people will be confused by having to pass `title={<Banner.Title>...</Banner.Title>}` and it'll lead to more custom styling 🤔  so I'd prefer one `textContent` prop over that

dierat: Ah, I see. Then let's go with `textContent` for now and see how it goes.

PR #623, opened 2021-09-02
fix: set line-height for NotificationIcon

anniehu4:
### Summary:

@lesliecodes noticed that the Toast icons had some extra vertical space at the top, so they weren't quite center aligned:
Screen Shot 2021-09-02 at 5 41 19 PM

what I found weird was the `svg` itself was size `20 x 20` as expected, but then the bounding `div` was suddenly `20 x 24`. Seems that `div` was inheriting a default `line-height` all the way from the `html` element (which for storybook happens to be 1.5, aka 24px).

~To be safe, this also sets SvgIcon's line height to match its `width` and `height`.~
This sets a `line-height` on that div, matching SvgIcon's width/height properties.

Alternate approaches:
- (if the `0.77px` diff is too annoying) setting `flex items-center` -- but that would only apply to the Toast icons, which felt more brittle
- removing the outer `div` entirely -- this broke the Vertical Banner variants, but we could apply some `width: 100%` styling to fix those

### Test Plan:
- Toast icon is now properly centered in a `20 x 20` box (note for some reason this adds an extra `0.77px`?? very perplexing)
Screen Shot 2021-09-02 at 6 20 01 PM

- No other stories change


lesliecodes:
Strange that Percy didn't catch any changes 🤔

anniehu4:
yeah that's bc the first change didn't actually work 😅 2nd try

anniehu4:
welp looks like percy can pick up a `0.77px` difference :') I'm gonna leave this for now, but if anyone picks it up or likes one of the other approaches better feel free to edit while I'm out

anniehu4:
FYI @dierat

dierat (approved):

anniehu4:
weird extra 0.77px of vertical space is now gone:

Screen Shot 2021-09-08 at 5 55 44 PM

PR #627, opened 2021-09-08
feat: rename clickable component

dierat:
### Summary:

We have a component called `Clickable`. It's a base component that can apply our button and link styling to any tag or component (preferably a button or link). My concern is that `Clickable` describes what kind of element it's intended to be used with (clickable, interactive elements), but it doesn't communicate what it does to that element (styles it). The question posed by this PR is: **Should we rename this component and, if so, to what?**

~This PR changes the name of `Clickable` to `ClickableStyleWrapper`, which came up in a conversation with @anniehu4 at some point. But I'm not really sure if that's the best name either. So I'm open for suggestions (including the suggestion to leave it as-is if we think `Clickable` is sufficiently descriptive).~

This PR changes the name of `Clickable` to `ClickableStyle` in response to the comments.

### Test Plan:
Automated tests should be sufficient for this change.


packages/components/src/Link/Link.tsx
```diff
@@ -34,7 +36,7 @@ const Link = forwardRef<HTMLAnchorElement, LinkProps>(
     ref,
   ) => {
     return (
-      <Clickable
+      <ClickableStyleWrapper
```
dierat, line 39: Note to self: the dosstring will need to be updated if https://github.com/chanzuckerberg/edu-design-system/pull/628 merges first.

packages/components/src/Button/button.stories.tsx
```diff
@@ -189,8 +189,8 @@ const renderSize = (
                 {colors.map((color) => (
                   <td key={color} className={styles.color}>
                     {/* To pass the "state" prop (only used for demonstration in storybook),
-                    we must use Clickable instead of Button */}
-                    <Clickable
+                    we must use ClickableStyleWrapper instead of Button */}
+                    <ClickableStyleWrapper
```
anniehu4, line 193: > Should we rename this component

definite yes from me -- I think you pointed out the confusion of the current name well

> and, if so, to what?

I'm ok with `ClickableStyleWrapper` because it's at least more descriptive. Other thoughts:
- `ClickableStyle` without the "Wrapper"
- `WithClickableStyle` to be more HOC-like

packages/components/src/ClickableStyleWrapper/ClickableStyleWrapper.stories.tsx
```diff
@@ -0,0 +1,21 @@
...
+  title: "ClickableStyleWrapper",
+  component: ClickableStyleWrapper,
+};
+
+type Args = React.ComponentProps<typeof ClickableStyleWrapper>;
+
+const Template: Story<Args> = (args) => <ClickableStyleWrapper {...args} />;
+
+export const ClickableStyleWrapperAsAnATag = Template.bind(null);
+ClickableStyleWrapperAsAnATag.args = {
```
anniehu4, line 15: I'm not sure it's useful to surface visual stories, since this is now more of an implementation detail 🤔  Instead we could make this `.stories.mdx` so it's more of a documentation page on usage. MDX files can still render code  to show examples and such

packages/components/src/Button/button.stories.tsx
ahuth: ClickableStyle is nice.

Can't think of any other good names. The naming challenge may be a sign that it's a weird abstraction. But of course it's useful to have for the react router links.

packages/components/src/ClickableStyleWrapper/ClickableStyleWrapper.stories.tsx
dierat: I feel like it's still useful to have this component in storybook so when devs go to use it with react router (or whatever), they can see how it works and play around with it.

I'll make a ticket for taking another look at the button, link, and clickable stories generally to figure out what we think is the best way to present this set of components based on who's using them and how.

dierat: Feel free to add more suggestions and clarify the ones I listed so far: https://app.clubhouse.io/czi-edu/story/160018/reorganize-button-link-and-clickablestyle-stories

packages/components/src/Button/button.stories.tsx
```diff
@@ -1,6 +1,6 @@
 import { Story } from "@storybook/react/types-6-0";
 import React from "react";
-import Clickable from "../Clickable";
+import ClickableStyle from "../ClickableStyle";
```
ahuth, line 3: Another name idea just occurred to me: Actionable

...because buttons and links are the main way that action happens on a web site? 🤷

ahuth (approved):
LGTM

anniehu4 (approved):

packages/components/src/Button/button.stories.tsx
dierat: How about `ActionableStyle`?! 😅 I think I'll stick with `ClickableStyle` for now, but I'm happy to revisit this if people think it's weird or if we decide we like something else better.

PR #628, opened 2021-09-08
feat(link + button): remove 'as' prop

dierat:
### Summary:

Our `Button` and `Link` components currently have an `as` prop that lets users pass in a component to be styled like a button or link. These wrap the `Clickable` component, which does the same thing and is available in `traject`, so we don't actually need `Button` and `Link` components to have an `as` prop at all. I think that it may also be clearer when to which components if the `Button` always returns a `button`, `Link` always returns an `a`, and `Clickable` is the only one that can return anything.

This PR updates the `Button` and `Link` components to no longer accept an `as` prop. Thoughts?

### Test Plan:
Automated tests should be sufficient for this change.

### DIFF
--- a/packages/components/src/Link/Link.tsx
+++ b/packages/components/src/Link/Link.tsx
@@ -7,7 +7,6 @@ type LinkHTMLElementProps = Omit<
 >;
 
 export type LinkProps = LinkHTMLElementProps & {
-  as?: "a" | React.ComponentType<any>;
   /**
    * The link contents or label.
    */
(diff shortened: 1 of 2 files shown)


packages/components/src/Link/Link.tsx
```diff
@@ -25,18 +24,15 @@ export type LinkProps = LinkHTMLElementProps & {
  * however, it can be styled to look like a button. In terms of the look and feel of the
  * component in the UI, the Link and Button components are exactly the same.
  *
- * Note: when using a routing component like react-router, you'll probably want to pass
- * in its Link component via the `as` prop.
+ * Note: when using a routing component like react-router, you'll probably want to
+ * use Clickable and pass in the routing component via the `as` prop.
```
dierat, line 28: Note to self: this line will need to be updated if https://github.com/chanzuckerberg/edu-design-system/pull/627 merges first.

anniehu4 (approved):
👍  agree this makes it clearer we're distinguishing these components by _behavior_, not appearance

ahuth (approved):

PR #637, opened 2021-09-14
chore(link, banner): export Banner props and Link from index

dierat:
### Summary:
When updating the `eds/components` package in `traject`, I ran into 2 issues (one big; one small):
- we can't import the `Link` component because it's not being exported from `index.ts`
- we can't reuse the `Banner` props in `NuxMegaphone` because it's not being exported at all

Relevant PR: https://github.com/FB-PLP/traject/pull/8370

### Test Plan:
Verify `npx lerna run build:types` completes with no errors or warnings.

### DIFF
--- a/packages/components/src/Banner/Banner.tsx
+++ b/packages/components/src/Banner/Banner.tsx
@@ -9,7 +9,7 @@ import NotificationIcon, {
 } from "../common/Notifications/NotificationIcon";
 import styles from "./Banner.module.css";
 
-type BannerProps = {
+export type BannerProps = {
   /**
    * Additional class names passed in for styling
    */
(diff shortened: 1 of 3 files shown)


packages/components/src/Banner/Banner.tsx
```diff
@@ -9,7 +9,7 @@ import NotificationIcon, {
 } from "../common/Notifications/NotificationIcon";
 import styles from "./Banner.module.css";
 
-type Props = {
+type BannerProps = {
```
anniehu4, line 12:
```suggestion
export type BannerProps = {
```

we should also add to `Banner/index.ts`:
```
export type { BannerProps } from "./Banner";
```

that way we can import from `lib/Banner` instead of needing `lib/Banner/Banner`

anniehu4: One way you could test is making sure within `index.ts` you can do
```
import {BannerProps} from "./Banner";
```

ahuth (approved):
LGTM

PR #643, opened 2021-09-20
feat(clickable-style): add icon support

dierat:
### Summary:
Shortcut ticket: https://app.shortcut.com/czi-edu/story/160030/icon-button
Figma: https://www.figma.com/file/YPTOEVRuas5ytXQEcwm4FK/UI-Kit-Sticker-Sheet-Core?node-id=226%3A1100

This PR adds support for icons in `ClickableStyle`, `Button`, and `Link`.

We have been talking about an "icon button" as if it were a separate component, but it didn't make sense to me to make a separate component since these buttons will still need all of the things the existing buttons do (ie the same variants and colors) and would need to be able to produce a `button` or anchor element as needed.

The weird thing about adding these icons is the size. We want the icons to be larger than the text, but that causes the button height to increase. And we want there to be no extra space to the left or right of the icon path, but the icon viewboxes are square even if the icon path is not.

To try to mitigate these issues, I:
- set a height on the buttons instead of vertical padding
- added a prop to easily add margin to the left and right of the icon to offset the extra space

The downside of the height is that the buttons will not wrap correctly if needed. Ideally, they would never wrap, but there may be places where a button is wrapping because there isn't enough space for it to sit normally, and this change will cause that bug to look even worse. I don't have a solution for this, but I wanted to mention that so we're aware of the ramifications of this change.

### Screenshot
ClickableStyle icon stories in storybook

### Test Plan:
`yarn start`,
navigate to `http://localhost:6008/?path=/docs/clickablestyle--clickable-style-with-icon-on-left-side`,
verify the icon is showing up on the left with -4px horizontal margin and 8px right padding in the "Clickable Style With Icon On Left Side" story,
and verify the icon is showing up on the right with -4px horizontal margin and 8px left padding in the "Clickable Style With Icon On Left Side" story,
and verify the buttons are the same size as before.


packages/components/src/ClickableStyle/ClickableStyle.stories.tsx
```diff
@@ -17,5 +19,72 @@ ClickableStyleAsAnATag.args = {
...
+  as: "a",
+  color: "brand",
+  variant: "flat",
+  size: "medium",
+  children: "ClickableStyle",
+  icon: {
+    iconElement: <ArrowBackRoundedIcon purpose="decorative" />,
+    iconHorizontalMargin: "-4px",
+  },
+};
```
dierat, line 36: Kinda wanna add another grid to the button stories so we can see the icons on the left, right, and along on all the variants. I put these here because I didn't want to get too bogged down in thinking about the button stories' organization.

packages/components/src/Link/Link.stories.tsx
```diff
@@ -151,22 +151,20 @@ mediumVariantsOnDarkBackground.parameters = {
...
 linkWithIcon.args = {
   ...defaultArgs,
-  children: (
-    <>
-      Link <CheckCircleRoundedIcon purpose="decorative" />
-    </>
-  ),
+  children: "Link",
+  icon: {
+    iconElement: <CheckCircleRoundedIcon purpose="decorative" />,
```
anniehu4, line 156: side note -- I wonder if it'd make sense for icons to default to `purpose: 'decorative'`

packages/components/src/ClickableStyle/ClickableStyle.tsx
```diff
@@ -23,6 +23,33 @@ export type ClickableStyleProps<IComponent extends React.ElementType> = {
    * The style of the element.
    */
   variant: "flat" | "outline" | "link";
+  /**
+   * Optional icon that can be positioned to the left or right of the other button contents.
+   *
+   * Default position is 'left'. Use the iconPlacement prop to move it to the right side.
+   */
+  icon?: {
```
anniehu4, line 31: Could it make sense to split these, so there's an `icon` prop that just takes the element, then `iconProps` that has `placement` and `horizontalMargin` fields? Thinking that most of the time, people won't need to pass the extra props

packages/components/src/ClickableStyle/ClickableStyle.tsx
```diff
@@ -23,6 +23,33 @@ export type ClickableStyleProps<IComponent extends React.ElementType> = {
...
+     * If the button has no other contents, this prop will be ignored.
+     */
+    iconPlacement?: "left" | "right";
+    /*
+     * Margin to be added to icon's left and right sides.
+     *
+     * Icons typically have a little extra space between the path and the edge of the viewbox.
+     * Negative left and right margin can be used to remove this extra space.
+     */
+    iconHorizontalMargin?: any;
```
anniehu4, line 51: could this be type `string`?

packages/components/src/ClickableStyle/ClickableStyle.tsx
```diff
@@ -40,10 +67,43 @@ const ClickableStyle = React.forwardRef(
...
+
+      const {
+        iconElement,
+        iconPlacement = "left",
+        iconHorizontalMargin = 0,
+      } = icon;
+
+      if (
+        (side === "left" && iconPlacement !== "left") ||
+        (side === "right" && iconPlacement !== "right")
```
anniehu4, line 88: having the 2 different side props is confusing to me 🤔   might be cleaner to destructure `iconPlacement` beforehand and then do
```
iconPlacement === "left" && renderIcon(styles.iconLeft)
iconPlacement === "right" && renderIcon(styles.iconRight)
```

anniehu4 (commented):
> The downside of the height is that the buttons will not wrap correctly if needed

Could you check if there are any long buttons in traject currently that might be affected?

packages/components/src/ClickableStyle/ClickableStyle.tsx
ahuth, line 31: (You may have already answered this elsewhere)

Why do we need to add this, when we can already pass in an icon as `children`?

```jsx
<Button>
  <FooIcon>
  Hello!
</Button>
```

Is it because of the height/sizing issue mentioned in the PR description?

Is it definitely worth the added props (that will inevitably end up being insufficient for what someone needs to do)?

packages/components/src/ClickableStyle/ClickableStyle.tsx
ahuth: Or `string | 0`

packages/components/src/Link/Link.stories.tsx
ahuth: I think the idea was to force people to always choose, and think about which purpose they're selecting.

packages/components/src/ClickableStyle/ClickableStyle.tsx
anniehu4: they can, but there's some consistent styling that's needed every time. for example, this is what it looks like to have an icon in children:
Screen Shot 2021-09-21 at 11 13 29 AM

Diedra & I discussed that when we add tailwind utility classes to traject, this would provide less value because you could easily just add this bit of styling through the className. But for now, I think it's worth us making the most common case easier.

packages/components/src/Link/Link.stories.tsx
dierat: Oh, I'll update the story that has just an icon so it follows that pattern, in case anyone uses it as a guide.

packages/components/src/ClickableStyle/ClickableStyle.tsx
dierat: @ahuth do you think we need to accommodate a value of `0`? It's 0 by default, right?

packages/components/src/ClickableStyle/ClickableStyle.tsx
dierat: One thing to consider with making a separate component is devs will need to pass in the element again. I'm also not sure what we would call it since it could produce either a button or an anchor element (or a react router link; who knows). At this point, I'm mostly concerned that making a separate component would be confusing.

dierat: For why we're adding this at all and not asking people to pass in the icon as a child, the only added value (I think) is the spacing consistency.

The spacing argument is harder to make because of the square viewbox issue – I don't know what kind of spacing variance we would see if devs add their own spacing vs using this component and possibly adjusting for the extra space (but probably not). Right now, I actually feel like the benefit of the `iconHorizontalMargin ` is that the component API calls out the issue and suggests a solution. The solution itself (adding negative margin via a style attribute) is not interesting, but I think people are more likely to even notice there's a problem and see the solution because of this prop.

We're also setting the size and color of the svg, but that can be set on the button styles even if we don't know for sure if there's an icon in there.

dierat: Oh, I just thought of another benefit, which is that we can discourage people from using the "plain" (formerly "text") variant without an icon. They can get around it by just passing the icon in via the children as usual, so you can still get around our little fence if you want, but at least we're making people go around.

dierat: My argument against adding this is – is it annoying for people to learn this API? I don't think this is a common pattern, so people will be learning it from scratch and reminding each other in PRs to use it instead of passing the icon in via the children. Are the benefits of this change worth that time and effort?

packages/components/src/ClickableStyle/ClickableStyle.tsx
dierat: Oh duh, I'm setting it to 0 🤦

packages/components/src/ClickableStyle/ClickableStyle.tsx
anniehu4: Well I agree with everything that's been said 😂  But most of the benefits are imo not dealbreakers:
 - as long as designers follow our guidance for the "plain" variant, it should be fine not checking that in the code
 - spacing variance from devs styling it themselves doesn't bother me so much, unless it's bothering designers

I'm starting to waver on how much benefit the spacing consistency gives over the cost of a more complicated API. Is it possible to land the smaller changes (setting size & color of svg on the button styles) independent of the API change? I'd like to think it through some more / maybe get the opinion of a product dev (so we're not inventing pains that don't exist). and if it's easier we could all hop in a discussion circle haha

ahuth: Always happy to hop in a discussion circle.

There's probably no 100% right or wrong answer here. I'm mostly thinking of component API and what kinds of things do and don't belong in a component.

But you're both closer to the actual work/needs than I am. We can always go with what meets those needs (always a good place to start 😸) and iterate/learn from there.

packages/components/src/ClickableStyle/ClickableStyle.tsx
```diff
@@ -40,10 +68,31 @@ const ClickableStyle = React.forwardRef(
...
       variant,
       className,
+      icon,
+      iconProps,
       ...rest
     }: ClickableStyleProps<IComponent>,
     ref: React.ForwardedRef<HTMLElement>,
   ) => {
+    const renderIcon = (placement: IconPlacement) => {
+      const horizontalMargin = (iconProps && iconProps.horizontalMargin) || 0;
```
ahuth, line 78: Not necessary, but I think we have access to optional chaining

```suggestion
      const horizontalMargin = iconProps?.horizontalMargin ?? 0;
```

ahuth (approved):
LGTM

packages/components/src/ClickableStyle/ClickableStyle.tsx
dierat: I'm totally fine with just landing the css changes so icons assume the expected height and color. I'll do that in a separate PR so we can easily reuse this one if we for some reason change our minds later.

I'll also add some spacing to the existing icon button stories and add stories for just an icon so people can see from storybook how we are expecting icons to be used in buttons.

Is there further guidance we should leave for people who may be expecting an icon button component/variant? Probably in zeroheight?

dierat:
Continuing this work in https://github.com/chanzuckerberg/edu-design-system/pull/650

PR #644, opened 2021-09-20
chore(clickable-style): reduce horizontal padding from 20px to 16px

dierat:
### Summary:
I noticed that the horizontal padding for buttons in figma is 14px (not counting the border), but they have 20px in the code. I asked Sean about it, and he said he actually prefers 16px. So he's going to make the padding 2px higher in figma and I'm making it 4px lower in the code.

#### Before
button all variants story in storybook before this change

#### After
button all variants story in storybook before this change

### Test Plan:
`yarn start`,
and verify the small and medium buttons have 1 rem horizontal padding.

### DIFF
--- a/packages/components/src/ClickableStyle/ClickableStyle.module.css
+++ b/packages/components/src/ClickableStyle/ClickableStyle.module.css
@@ -11,11 +11,11 @@
 
 /* Component Sizes */
 .sizeSmall {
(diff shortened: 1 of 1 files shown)


anniehu4 (approved):
16 is nice b/c it's more consistent with our 8px grid too 👍

ahuth (approved):

PR #645, opened 2021-09-21
feat(clickable-style): add text variant

dierat:
### Summary:
Shortcut ticket: https://app.shortcut.com/czi-edu/story/160030/icon-button
Design: https://www.figma.com/file/P9VPyI821gLq832tuU2WeY/UI-Kit-Master-Core?node-id=3072%3A11792

There is a style of button in the UI kit referred to as `TextButton`. This is floating text that has the same padding as the flat and outline buttons but doesn't have a background color in its default state. There was a bit of discussion around what this button is and how are we distinguishing that it's interactive in a way besides color since it doesn't have an underline.

I talked it out with Sean and I think I understand this variant now – it is only intended to be used with an icon. (I still have reservations about this from an accessibility point of view because there are also times when non-interactive text is paired with an icon, but that's pretty rare and this is better than just having text floating around with 0 indication.) In the UI sticker sheet, it's currently shown with no icon by default, but Sean said he'll remove that (or update them to all have icons maybe?) to make that clearer.

Now that we're on the same page there, I'm fine with adding the `TextButton` style. This PR adds it as a variant, called "text". I'm putting this out before going deeper to verify that we're all on board with this approach.

I have one follow-up task to address after this (left a TODO comment):
- We should probably make the types restrictive here so it's not possible to create a "text" button with no icon.

I also just noticed that the disabled states for all the buttons in the core UI kit design are light neutral. The code is following the sticker sheet, which uses a light version of the button color (ie success buttons are a light green when disabled). So, please ignore that discrepancy between this change and the design. https://app.shortcut.com/czi-edu/story/163311/button-link-disabled-states-should-be-gray

#### Design
Screen Shot 2021-09-20 at 8 50 05 PM

#### Storybook
Screen Shot 2021-09-20 at 8 46 29 PM

### Test Plan:
`yarn start`,
verify there's now a "text" variant,
verify it looks like the design (except the disabled state is not gray),
hover over the disabled buttons,
and verify the background color doesn't change.

### DIFF
--- a/packages/components/src/Button/button.stories.tsx
+++ b/packages/components/src/Button/button.stories.tsx
@@ -9,7 +9,7 @@ import styles from "./button.stories.module.css";
 
 const sizes = ["small", "medium", "large"] as const;
 const allColors = ["alert", "brand", "neutral", "success", "warning"] as const;
-const variants = ["flat", "outline", "link"] as const;
+const variants = ["flat", "outline", "link", "plain"] as const;
 const states = ["inactive", "hover", "focus", "active", "disabled"] as const;
 
 // For now, the UI kit only includes alert & brand "flat" buttons
@@ -115,6 +115,19 @@ linkInHeading.args = {
   color: "brand",
 };
 
+export const plainSmall = Template.bind(null);
+plainSmall.args = {
+  children: (
+    <>
+      Button{" "}
+      <AddRoundedIcon className={styles.iconButton} purpose="decorative" />
+    </>
+  ),
+  color: "brand",
+  size: "small",
+  variant: "plain",
+};
+
 export const withDataTestId = Template.bind(null);
 withDataTestId.args = {
   children: "Button with data-testid",
@@ -179,6 +192,10 @@ const renderSize = (
 ) =>
   variants.map((variant) => {
     const colors = variant === "flat" ? flatColors : allColors;
+    const icon =
+      variant === "plain" ? (
+        <AddRoundedIcon className={styles.iconButton} purpose="decorative" />
+      ) : null;
 
     return (
       <React.Fragment key={variant}>
@@ -207,6 +224,7 @@ const renderSize = (
                       disabled={state === "disabled"}
                     >
                       {children}
+                      {icon}
                     </ClickableStyle>
                   </td>
                 ))}
--- a/packages/components/src/Button/button.stories.tsx
+++ b/packages/components/src/Button/button.stories.tsx
@@ -160,10 +160,10 @@ outlineWithIcon.args = {
   variant: "outline",
 };
 
-export const flatWithOnlyIcon = Template.bind(null);
-flatWithOnlyIcon.args = {
+export const plainWithOnlyIcon = Template.bind(null);
+plainWithOnlyIcon.args = {
   children: <AddRoundedIcon purpose="informative" title="add" />,
-  variant: "flat",
+  variant: "plain",
 };
 
 export const withFakeClassName = Template.bind(null);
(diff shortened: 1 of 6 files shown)


packages/components/src/ClickableStyle/ClickableStyle.module.css
```diff
@@ -172,6 +177,35 @@
...
+  &:hover,
+  &.stateHover {
+    background-color: var(--tertiary-color);
+    color: var(--primary-color);
+  }
+
+  &:active,
+  &.stateActive {
+    /* TODO: primary is being set to 700 on click; should be 500 here */
+    background-color: var(--primary-color);
```
anniehu4, line 199: isn't it actually `-800` now 😬   If these colors are only used in the text button, I'm tempted to call
- `tertiary-color` -> `text-hocus-bg-color`
- add a `text-active-bg-color` for the `-500` variants

it is an annoying large mix of colors, but at least it's clearer then not every variant needs that many

packages/components/src/ClickableStyle/ClickableStyle.tsx
```diff
@@ -22,7 +22,7 @@ export type ClickableStyleProps<IComponent extends React.ElementType> = {
   /**
    * The style of the element.
    */
-  variant: "flat" | "outline" | "link";
+  variant: "flat" | "outline" | "text" | "link";
```
anniehu4, line 25: > We should probably make the types restrictive here so it's not possible to create a "text" button with no icon.

Agree -- let's either make a TODO here or be sure to land after the icon change.

I'm also wondering if `"plain"` might be more descriptive than `"text"`? these variant names are already different from the UI Kit groupings

packages/components/src/Button/button.stories.tsx
```diff
@@ -85,6 +85,14 @@ TertiarySmall.args = {
   variant: "outline",
 };
 
+export const TextSmall = Template.bind(null);
```
anniehu4, line 88: > Regardless of the type setup, we should probably put an icon in this variant's stories so it is always presented as being used with an icon.

Yes, let's 😀

anniehu4 (commented):
> I also just noticed that the disabled states for all the buttons in the core UI kit design are light neutral.

Do we need to change this in code, or does the UI kit need to be updated?

dierat:
> > I also just noticed that the disabled states for all the buttons in the core UI kit design are light neutral.
> 
> Do we need to change this in code, or does the UI kit need to be updated?

I'm still not sure but I think the code needs updating. I'm talking to Sean async about it.

packages/components/src/ClickableStyle/ClickableStyle.tsx
dierat: Sounds good – I found "text" really confusing before because links are also just text. "Plain" makes sense because links are underlined whereas this variant has no decoration (aside from the icon but any of the buttons can have an icon).

dierat:
> > > I also just noticed that the disabled states for all the buttons in the core UI kit design are light neutral.
> > 
> > 
> > Do we need to change this in code, or does the UI kit need to be updated?
> 
> I'm still not sure but I think the code needs updating. I'm talking to Sean async about it.

https://app.shortcut.com/czi-edu/story/163311/button-link-disabled-states-should-be-gray

packages/components/src/ClickableStyle/ClickableStyle.module.css
dierat: Good idea! I replaced `tertiary` with `plain-background-color` and just tracked the background color there since, you're right, it's not being reused by any of the other variants.

> isn't it actually -800 now 😬 

yes and the click color is 600 so I was wrong all around lol

anniehu4 (approved):
nice, thanks for the updates!

PR #651, opened 2021-09-22
Add Chromatic for Visual Regression Testing

jinlee93:
### Summary:
- [Chromatic](https://www.chromatic.com/docs/) is used to for visual regression testing of Storybook stories
- Chromatic hosts Storybook
- Set up to run every push, like Percy

### Test Plan:
- Team use and discussion
- 
image

image


.github/workflows/test.yml
```diff
@@ -62,6 +62,30 @@ jobs:
         env:
           PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
 
+  chromatic-deployment:
```
jinlee93, line 65: Made chromatic it's own job. Should I make it a a step instead?

.github/workflows/test.yml
```diff
@@ -62,6 +62,30 @@ jobs:
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # 👈 Required to retrieve git history
+      - name: Install dependencies
+        run: yarn install --immutable
+      - name: Build Storybook
+        run: yarn workspace @chanzuckerberg/eds-components run build:storybook
```
jinlee93, line 80: In the Percy job, we were using `yarn workspaces foreach run build:storybook` but Storybook only exists in `components`, so went with this instead. Are we expecting more packages under this monorepo and should I make it the `foreach` instead?

packages/components/package.json
```diff
@@ -31,6 +31,7 @@
     "build:types": "tsc --project . --declaration --emitDeclarationOnly",
     "build:watch": "yarn run build:js --watch",
     "build:storybook": "build-storybook -o storybook-static",
+    "chromatic": "chromatic --project-token=CHROMATIC_PROJECT_TOKEN --skip 'dependabot/**' --only-changed",
```
jinlee93, line 34: the `--only-changed` enables the [Turbosnap](https://www.chromatic.com/docs/turbosnap) feature where Chromatic "identifies component files that change, then intelligently builds and snapshots only the stories associated with those components ... However, we don’t recommend using TurboSnap immediately when starting out with Chromatic since the configuration is more complicated and can lead to difficult to debug scenarios or UI changes being missed. Instead, become familiar with Chromatic’s out of the box behavior and once your project has been running smoothly for some time consider trying out TurboSnap."
Should we use TurboSnap from the get go? It was working when I tried it out

lesliecodes:
The [UI Review](https://www.chromatic.com/pullrequest?appId=61313967cde49b003ae2a860&number=651&view=activity) is pretty cool. At first I wondered if we want designers to approve, but it looks like you can only add people who have github accounts so they would all need to make one.

packages/components/package.json
lesliecodes: Let's establish a baseline first. We're still on the free plan so no worries about going over the limit right now. This also helps us know the upper bound of our usage.

.github/workflows/test.yml
lesliecodes: This is right call. I think Percy was approved at tier 3, where as chromatic was approved at tier 2 since it has access to git history.

jinlee93:
> The [UI Review](https://www.chromatic.com/pullrequest?appId=61313967cde49b003ae2a860&number=651&view=activity) is pretty cool. At first I wondered if we want designers to approve, but it looks like you can only add people who have github accounts so they would all need to make one.

Think we can add designers and other users without repo access like so: 
image

.github/workflows/test.yml
ahuth: We can always update later, so this seems fine.

lesliecodes:
Percy and Chromatic have the same # of stories (which is good 😂 )
image

Also, Chromatic has a much better UI for grouping stories.

.github/workflows/test.yml
```diff
@@ -62,6 +62,32 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # 👈 Required to retrieve git history
+      - name: Install dependencies
+        run: yarn install --immutable
+      - name: Build Storybook 📚
+        run: yarn workspace @chanzuckerberg/eds-components run build:storybook
+      - name: Run Accessibility Tests 💛
+        run: yarn workspace @chanzuckerberg/eds-components run storybook:axeOnly
```
ahuth, line 82: Is this copy pasta and can be removed?

ahuth (approved):
LGTM

The UI is nice. And it's cool that you can go from a UI test directly to a storybook for the component!

packages/components/package.json
jinlee93: Got it, thanks for the advice, removed TurboSnap for now

.github/workflows/test.yml
jinlee93: Was intentionally added so that if we were to remove Percy, we can just remove the code block and would make more sense in terms of commit history, but makes sense to remove since it is double testing. What is your advice?

packages/components/package.json
```diff
@@ -31,6 +31,7 @@
     "build:types": "tsc --project . --declaration --emitDeclarationOnly",
     "build:watch": "yarn run build:js --watch",
     "build:storybook": "build-storybook -o storybook-static",
+    "chromatic": "chromatic --project-token=CHROMATIC_PROJECT_TOKEN --skip 'dependabot/**'",
```
anniehu4, line 34: Is this hooked up to the GH action automagically? or is it for local use?

ahuth: Automagically. Pretty sure Chromatic's CI integration expects a `chromatic` script to be present.

PR #675, opened 2021-10-20
Enable Tailwind utility and component classes

jinlee93:
### Summary:
- Enables [utility](https://tailwindcss.com/docs/utility-first) and [component](https://unpkg.com/tailwindcss@2.2.17/dist/components.css) tailwind classes
- As per conversation with Annie: 
> some of the paths forward are
> 1. prefix everything
> 2. scope tailwind
> 3. live with conflicts and if need higher specificity than the conflict, use scope and use @apply with the class
> After digging some in the `bootstrap-sass` code, we think the chance of conflict is actually pretty low. Relevant files:

>    - https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_utilities.scss
>    - Component classnames like .text- (https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_type.scss#L90)

> Next steps with this approach:

> - Bring in utilities directive to traject. We could add it directly to tailwind.base.css OR repurpose tailwind-sr-only.css and rename it tailwind.utilities.css to handle all utility-related things
> - QA to check for other glaring conflicts
> - Create guidance about using utility classes. Includes guidance on known existing conflicts (.invisible, .hidden) &
> - what to do if someone finds a conflict we didn't know about
> - Add https://www.npmjs.com/package/eslint-plugin-tailwindcss to lint utility classes. (Note: we could use this to ban bootstrap utilities & conflicting classes with the no-custom-classname rule, but that may be too aggressive given all the custom classnames that currently exist)

Decided to go with path 3. of living with the occasional conflict since there shouldn't be many.
My approach to dealing with conflicts without `!important` is to use the `@apply` directive to scope styling into higher specificity, or using custom classes. 
For `!important` using conflicts, such as `.hidden`, can use `@apply ...classes !important` or use custom classes
Examples and details are given in the components [readme](./packages/components/README.md)

### Test Plan:
CI tests, especially visual regression tests pass


jinlee93:
> ## [size-limit](https://github.com/ai/size-limit) report
> Path	Size
> components	56.05 KB (0%)
> styles	244.2 KB (+98.4% 🔺)

🤔  addition of TW utility classes def does increase the size

packages/components/tailwind.config.js
```diff
@@ -76,4 +76,5 @@ module.exports = {
     transitionProperty: ["motion-safe", "motion-reduce"],
   },
   plugins: [],
+  purge: ["./src/**/*.tsx"],
```
anniehu4, line 79: maybe we should make this
```suggestion
  purge: ["./src/**/*.{js,jsx,ts,tsx}"],
```
to be most safe & so it doesn't need to be overriden in traject

jinlee93: Thanks for the suggestion, committed as suggested

jinlee93:
Test pr: https://github.com/chanzuckerberg/edu-design-system/pull/678

packages/components/src/styles/global.css
```diff
@@ -12,4 +12,20 @@
...
+ * This injects Tailwind's base styles and any base styles registered by
+ * plugins.
+ */
 @tailwind base;
+
+/**
+ * This injects Tailwind's component classes and any component classes
+ * registered by plugins.
+ */
+@tailwind components;
```
ahuth, line 25: My read of https://tailwindcss.com/docs/extracting-components#extracting-component-classes-with-apply is that this is where the `@apply`-ed stuff goes.

But I thought we were already getting those in the compiled css? 🤔 

Is it possible we're now duplicating anything from `@apply`?

packages/components/src/styles/global.css
```diff
@@ -12,4 +12,20 @@
...
+
+/**
+ * This injects Tailwind's utility classes and any utility classes registered
+ * by plugins.
+ */
+@tailwind utilities;
```
ahuth, line 31: We may want to put the `@tailwind` stuff in another file. Something like `src/styles/tailwind.css`.

That way an App already using Tailwind (like Traject is/will?) can use it's own Tailwind stuff instead of importing 2 of them (one from here and one from the app)

packages/components/src/styles/global.css
anniehu4: I think Jin noted that the `components.css` class contains really few things: https://unpkg.com/tailwindcss@2.2.17/dist/components.css. My read of that section was that you can use `@layer` to control specificity, but it doesn't overlap with the content of `@tailwind components`

jinlee93: And if the component classes are not being used, they should be purged out

packages/components/src/styles/global.css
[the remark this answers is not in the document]
anniehu4: That makes sense to me! It's essentially what we're doing already (since `global.css` is only used in the storybook file), but we can be more explicit by only `purge`-ing story files. I'm less worried about the size since according to Tailwind,
> When removing unused styles with Tailwind, it’s very hard to end up with more than 10kb of compressed CSS

packages/components/src/styles/global.css
ahuth: Yeah you're right. Looks like `@apply` stuff gets inlined right into the .module.css file

packages/components/README.md
```diff
@@ -99,3 +99,32 @@ declare var x: AbstractComponent<Props, HTMLElement>;
...
+Use the [docs](https://tailwindcss.com/docs) to search for appropriate classes.
+
+**Conflicts may arise rarely with Bootstrap classes, some that we know of:**
+  - Most annoying, `.hidden`
+    - Both Tailwind and Bootstrap have the same styling for `.hidden` but Bootstrap applies the  `!important ` property, which can be annoying when trying to utilize Tailwind breakpoints, e.g.:
+    ```html
+    <div class="hidden md:block lg:inline">
+    ```
+    - Where the expectation is `display: none ` for breakpoints smaller than 768px, `display: block `for 768px - 1023px, and `display: inline ` for >= 1024px. However since Bootstrap applies the `!important` property, all screen sizes show `display: none`.
+    - This can be circumvented with the `@apply` directive or using custom CSS with regular media queries to scope styling into a different class name.
```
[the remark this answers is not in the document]
jinlee93: Done, thanks!

jinlee93:
EDS storybook:
- `display: none` workaround: https://614b969675171d003ac6f98a-xclfqqspzi.chromatic.com/?path=/story/admin-components-studentcoursemodificationspanel--normal&globals=measureEnabled:false
- `text-` align https://614b969675171d003ac6f98a-xclfqqspzi.chromatic.com/?path=/story/core-components-pagetitle--standard&globals=measureEnabled:false
- `.invisible .lg:visible' https://614b969675171d003ac6f98a-xclfqqspzi.chromatic.com/?path=/story/core-components-megaphone--info&globals=measureEnabled:false
- '.visible .lg:invisible' https://614b969675171d003ac6f98a-xclfqqspzi.chromatic.com/?path=/story/core-components-megaphone--success&globals=measureEnabled:false

PR #678, opened 2021-10-21
test(tailwind): tailwind utility classes

jinlee93:
Test pr to show utility classes working

### DIFF
--- a/packages/components/src/Banner/Banner.stories.tsx
+++ b/packages/components/src/Banner/Banner.stories.tsx
@@ -52,6 +52,7 @@ const dialogArgs = {
 
 export const Brand = Template.bind(null);
 Brand.args = {
+  className: "visible md:invisible",
   ...dialogArgs,
 };
 
--- a/packages/components/src/Heading/Heading.stories.tsx
+++ b/packages/components/src/Heading/Heading.stories.tsx
@@ -15,12 +15,14 @@ export const Heading1 = Template.bind(null);
 Heading1.args = {
   size: "h1",
   children: "Heading 1 24/32",
+  className: "text-right",
 };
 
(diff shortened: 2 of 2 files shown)


jinlee93:
> Test pr to show utility classes working

When running storybook locally, can modify className in the story also

packages/components/src/Banner/Banner.stories.tsx
```diff
@@ -52,6 +52,7 @@ const dialogArgs = {
 
 export const Brand = Template.bind(null);
 Brand.args = {
+  className: "visible md:invisible",
```
jinlee93, line 55: to show `.visible` and breakpoints working even tho `.visible` exists in both Bootstrap and TW

packages/components/src/Heading/Heading.stories.tsx
```diff
@@ -15,12 +15,14 @@ export const Heading1 = Template.bind(null);
 Heading1.args = {
   size: "h1",
   children: "Heading 1 24/32",
+  className: "text-right",
```
jinlee93, line 18: to show `.text-` alignment workign even tho existing in both Bootstrap and TW

packages/components/src/Banner/Banner.stories.tsx
anniehu4: Nice! There's no Bootstrap in this repo though, so it wouldn't be a problem here :') we should test this in `traject` though, where Bootstrap is imported

anniehu4 (commented):
wooot thanks!

packages/components/src/Banner/Banner.stories.tsx
jinlee93: oooop right

PR #692, opened 2021-10-28
test(button + link): add 'RecommendedVariants' story

dierat:
### Summary:
Shortcut ticket: https://app.shortcut.com/czi-edu/story/160018/reorganize-button-link-and-clickablestyle-stories 
Figma: https://www.figma.com/file/P9VPyI821gLq832tuU2WeY/UI-Kit-Master-(Core)?node-id=3072%3A11792

Sorry this is a big PR, but at least it only affects story files?

Design requested that we add a story for the `Button` and `Link` components that matches the figma design. So here I'm adding a story called `RecommendedVariants` that copies the structure in the core UI kit as closely as possible. However, I'm not adding all the states (hover, focus, etc) to this story. My idea is that we instead keep the grid showing all of the possible variants, and that's where we show all the states. So the `AllVariants` grid is more for regression testing via Chromatic/Percy. I don't think we really need to show all the states in the `RecommendedVariants` story because that one is more for designers and devs to see the main offering of buttons. I don't expect they'll really care what the hover state looks like when picking out which button to use in their feature. This also means we can leave the snapshot tests on because it's not going to be too ridiculously big.

My reason for putting all these test cases in a single story, instead of having a bunch of individual ones, is it can go in zeroheight as one chunk. But I'm open to revisiting that if people feel it's better to have individual stories.

This makes most of the other stories redundant, so I also cleared a bunch of them out. I'm leaving one standard story for each component at the top for people to play with, and the `Link` storybook page is also keeping the two stories where we show the `Link` component in a block of text.

I also did a little other cleanup, like renaming stories, while I was in here.

### Test Plan:
Verify the `RecommendedVariants` story for the `Button` and `Link` components matches the figma design.

### Video of Link stories after this change
https://user-images.githubusercontent.com/7761701/140103260-7fb7c165-d26e-448a-856c-0650262e4b95.mov

### DIFF
--- a/packages/components/.storybook/clickableStyleUtils.stories.tsx
+++ b/packages/components/.storybook/clickableStyleUtils.stories.tsx
@@ -15,44 +15,58 @@ export const getStandardSet = (
   variant: ClickableStyleProps<"button">["variant"],
   color: ClickableStyleProps<"button">["color"] = "brand",
 ) => (
-  <ul className="grid gap-y-4">
-    <Heading size="h3">Size Large</Heading>
+  <ul>
     <li>
-      <Component color={color} variant={variant}>
-        {componentName}
-      </Component>
-    </li>
-    <li>
-      <Component color={color} variant={variant}>
-        <ArrowBackRoundedIcon
-          className={styles.arrowBackIcon}
-          purpose="decorative"
-        />{" "}
-        {componentName}
-      </Component>
-    </li>
-    <li className="mb-4">
-      <Component color={color} variant={variant}>
-        {componentName}
-        <ArrowForwardRoundedIcon
-          className={styles.arrowForwardIcon}
-          purpose="decorative"
-        />
-      </Component>
+      <Heading size="h3" spacing="2x">
+        Size Large
+      </Heading>
+      <ul className="grid gap-y-4">
+        <li>
+          <Component color={color} variant={variant}>
+            {componentName}
+          </Component>
+        </li>
+        <li>
+          <Component color={color} variant={variant}>
+            <ArrowBackRoundedIcon
+              className={styles.arrowBackIcon}
+              purpose="decorative"
+            />{" "}
+            {componentName}
+          </Component>
+        </li>
+        <li className="mb-4">
+          <Component color={color} variant={variant}>
+            {componentName}
+            <ArrowForwardRoundedIcon
+              className={styles.arrowForwardIcon}
+              purpose="decorative"
+            />
+          </Component>
+        </li>
+      </ul>
     </li>
 
-    <Heading size="h3">Size Medium</Heading>
-    <li className="mb-4">
-      <Component color={color} variant={variant} size="medium">
-        {componentName}
-      </Component>
+    <li>
+      <Heading size="h3" spacing="2x">
+        Size Medium
+      </Heading>
+      <div className="mb-4">
+        <Component color={color} variant={variant} size="medium">
+          {componentName}
+        </Component>
+      </div>
     </li>
 
-    <Heading size="h3">Size Small</Heading>
-    <li className="mb-4">
-      <Component color={color} variant={variant} size="small">
-        {componentName}
-      </Component>
+    <li>
+      <Heading size="h3" spacing="2x">
+        Size Small
+      </Heading>
+      <div className="mb-4">
+        <Component color={color} variant={variant} size="small">
+          {componentName}
+        </Component>
+      </div>
     </li>
   </ul>
 );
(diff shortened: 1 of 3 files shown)


packages/components/tailwind.config.js
```diff
@@ -5,7 +5,10 @@ module.exports = {
   /*The main value in TW utility classes is in Storybook stories.
     We avoid using them in component styles to reduce chance of conflict with other libraries.
     Please configure downstream Tailwind config purge to include app files if necessary*/
-  purge: ["./src/**/*.stories.{js,jsx,ts,tsx}"],
+  purge: [
+    "./src/**/*.stories.{js,jsx,ts,tsx}",
+    "./.storybook/*.stories.{js,jsx,ts,tsx}",
+  ],
```
ahuth, line 11: Could this be condensed to

```suggestion
  purge: ["**/*.stories.{js,jsx,ts,tsx}"],
```

?

packages/components/src/Button/button.stories.tsx
```diff
@@ -41,123 +42,10 @@ const Template: Story<Args> = (args) => <Button {...args} />;
...
-  color: "warning",
-  variant: "outline",
-};
-
-export const plainWithOnlyIcon = Template.bind(null);
-plainWithOnlyIcon.args = {
-  children: <AddRoundedIcon purpose="informative" title="add" />,
-  variant: "plain",
-};
+export const RecommendedVariants = () =>
```
ahuth, line 47: Are we basically trading all the individual stories for 1 or 2 bigger ones that show more examples together?

Edit: found where you said:

> My reason for putting all these test cases in a single story, instead of having a bunch of individual ones, is it can go in zeroheight as one chunk. But I'm open to revisiting that if people feel it's better to have individual stories.

I'm not sure what the right answer is. Seems like there's tradeoffs either way.

One big one loses searching for or linking to a specific story. But makes it easier load into Zeroheight it seems.

ahuth: One possibility could be to do both. Individual stories could also be re-used in a larger story.

Not sure if that is too much work or not, though.

dierat: Do you mean like have x individual stories and then one repeater that houses all of them?

dierat: I'm personally not too opinionated about which direction we choose to go in. Do you have thoughts @anniehu4 (since you're been owning the zeroheight changes)?

packages/components/tailwind.config.js
dierat: Sadly, no, that doesn't seem to work 🤔

packages/components/src/Button/button.stories.tsx
ahuth: Yeah, meant X individual stories, and then the big ones for zeroheight that re-use them.

So like

```jsx
export const Modelo = () => {...}

export const Pacifico = ( ) => {...}

export const AllTheCerveza = () => (
  <ul>
    <li>{Modelo()}</li>
    <li>{Pacifico()}</li>
  </ul>
);
```

Just an idea. Not sure if this makes the problem worse or better

anniehu4: Maybe we could split so each Heading level in the recommended variants is its own story? then
- RecommendedVariants could compose all of them like Andrew suggested
- we could rename that first story "Default"

packages/components/.storybook/buttonUtils.stories.module.css
```diff
@@ -0,0 +1,24 @@
+.recommendedVariantSet,
+.recommendedVariantSet h2,
+.recommendedVariantSet button,
+.recommendedVariantSet a {
+  @apply mb-4;
```
anniehu4, line 5: nit: is it easier to put this in a `grid`? was thinking it might be nice to fill the horizontal space more anyways

packages/components/.storybook/buttonUtils.stories.module.css
```diff
@@ -0,0 +1,24 @@
...
+}
+
+.arrowBackIcon {
+  margin-left: -0.4em;
+  margin-right: 0.5em;
+}
```
anniehu4, line 11: you could try using [arbitrary TW values](https://tailwindcss.com/docs/just-in-time-mode#arbitrary-value-support) now 😱 
```
<ArrowBackRoundedIcon
  className="ml-[-0.4em] mr-[0.5em]"
/>
```

packages/components/tailwind.config.js
anniehu4: maybe `"./**/*.stories.{js,jsx,ts,tsx}"`?

anniehu4 (commented):
gah sorry I'm so late to reviewing these

packages/components/.storybook/buttonUtils.stories.tsx
```diff
@@ -0,0 +1,170 @@
+import React from "react";
```
anniehu4, line 1: nit: maybe we should call this `clickableStyleUtils` instead of `buttonUtils`

packages/components/src/Button/button.stories.tsx
dierat: Sounds good! 👍

packages/components/.storybook/buttonUtils.stories.module.css
dierat: Woah cool!

packages/components/.storybook/buttonUtils.stories.module.css
dierat: >was thinking it might be nice to fill the horizontal space more anyways

Do you mean you'd like to see them laid out in rows so they take up less vertical space?

packages/components/.storybook/buttonUtils.stories.module.css
dierat: I decided to leave those 2 styles in the module because it's annoying to tinker with the values when it's copied across different elements, but I updated them to use tw. Thanks for the hot tip!

packages/components/tailwind.config.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
   /*The main value in TW utility classes is in Storybook stories.
     We avoid using them in component styles to reduce chance of conflict with other libraries.
     Please configure downstream Tailwind config purge to include app files if necessary*/
-  purge: ["./src/**/*.stories.{js,jsx,ts,tsx}"],
+  purge: ["./src/**/*.stories.{js,jsx,ts,tsx}", "./.storybook/*"],
```
dierat, line 8: We talked about this separately but posting here for posterity – I couldn't get any path that didn't specify `.storybook` to work, and @anniehu4 suggested we just point to the whole directory instead of just specifying the `.stories` files.

packages/components/.storybook/clickableStyleUtils.stories.tsx
```diff
@@ -0,0 +1,208 @@
...
+      <Component color={color} variant={variant}>
+        {componentName}
+        <ArrowForwardRoundedIcon
+          className={styles.arrowForwardIcon}
+          purpose="decorative"
+        />
+      </Component>
+    </li>
+    <li>
+      <Component color={color} variant={variant} size="medium">
```
anniehu4, line 43: oop I didn't notice before -- let's include a label above the ones that are size "medium" or "small", because it's kind of subtle

packages/components/src/Link/Link.stories.tsx
```diff
@@ -46,11 +53,29 @@ const defaultArgs = {
...
+
+export const PlainRecommendedVariants = () =>
+  getPlainRecommendedVariants(Link, "Link");
+
+export const LinkRecommendedVariants = () =>
+  getLinkRecommendedVariants(Link, "Link");
+
+export const DestructiveRecommendedVariants = () =>
+  getDestructiveRecommendedVariants(Link, "Link");
+
```
anniehu4, line 78: let's keep `linkInBody` & `linkInHeading` above these, think it's important to illustrate that Link inherits its size

anniehu4 (approved):

packages/components/.storybook/buttonUtils.stories.module.css
anniehu4: yeah, but now that they're separate stories I don't mind it

packages/components/.storybook/clickableStyleUtils.stories.tsx
dierat: Added!

packages/components/src/Link/Link.stories.tsx
dierat: Moved – thanks!

PR #694, opened 2021-10-29
test(clickable-style): change story from using a tag to label tag

dierat:
### Summary:
Shortcut ticket: https://app.shortcut.com/czi-edu/story/160018/reorganize-button-link-and-clickablestyle-stories

We have one beautiful little story for the `ClickableStyle` component: the component with an `a` tag. That's pretty unnecessary now that we have the `Link` component, since that uses an `a` tag.

In this PR, I've updated the story to use a `label` tag instead. However, I'm kind of wondering if we need that either. Does this component need to have stories when it's so well-covered by the `Button` and `Link` components?

### Test Plan:
Verify the story has been updated in storybook.


anniehu4 (commented):
> Does this component need to have stories when it's so well-covered by the Button and Link components?

I also don't think so! We could fully convert this to a docs page (like `typography.stories.mdx`) instead, or just talk about it in Zeroheight

dierat:
An alternate direction was implemented in https://github.com/chanzuckerberg/edu-design-system/pull/708

PR #695, opened 2021-10-29
Prune and update dependencies

ahuth:
### Summary:

Removals:
- @storybook/addons (likely mistakenly added when bringing in storybook)
- core-js (remnant of @emotion from 0ded83306a14013dcbfb30421a725a31444177c1)

Updates:
- axe-storybook-testing
- typescript
- jest (and related)
- babel (and related)
- eslint (and related)
- storybook
- testing library
- autoprefixer
- chromatic
- postcss and postcss-cli

### Test Plan:

- CI
- Ran a local build
- Ran storybook locally


packages/components/package.json
```diff
@@ -54,45 +54,43 @@
...
-    "@chanzuckerberg/axe-storybook-testing": "^4.1.2",
+    "@chanzuckerberg/axe-storybook-testing": "^4.1.3",
     "@chanzuckerberg/story-utils": "^2.1.0",
     "@percy/storybook": "^3.3.1",
-    "@storybook/addon-a11y": "^6.3.8",
-    "@storybook/addon-essentials": "^6.3.8",
+    "@storybook/addon-a11y": "^6.3.12",
+    "@storybook/addon-essentials": "^6.3.12",
     "@storybook/addon-postcss": "^2.0.0",
-    "@storybook/addons": "^6.3.8",
```
ahuth, line 69: @storybook/addon seems to be used when creating addons, not when actually using storybook

ahuth:
Err wat, guess this broke storybook

packages/components/package.json
```diff
@@ -54,45 +54,44 @@
...
-    "axe-core": "^4.3.2",
-    "babel-jest": "^27.0.6",
-    "babel-loader": "^8.2.2",
-    "chromatic": "^5.10.1",
+    "autoprefixer": "^10.4.0",
+    "axe-core": "^4.3.5",
+    "babel-jest": "^27.3.1",
+    "babel-loader": "^8.2.3",
+    "chromatic": "^6.0.5",
+    "classnames": "^2.3.1",
```
ahuth, line 81: We're using classnames in a story, so it's in dev deps here. If we ever want to use it in an actual component, we'll need to move it to regular deps.

ahuth: Wait, just realized we're using clsx also... will change this and the story to that

packages/components/docs/token-table.tsx
```diff
@@ -1,9 +1,6 @@
-import classNames from "classnames/bind";
```
ahuth, line 1: `classnames` was being brought in through a storybook package, but with these updates it's no longer in the dependency tree.

The `no-extraneous-dependencies` should've stopped us from bringing this in without having an explicit dependency on it. Maybe that rule isn't turned on by default with the imports plugin? 🤔

ahuth:
This is green again, and should be ready to go in @chanzuckerberg/edu-frontend-infra

package.json
```diff
@@ -50,16 +50,16 @@
     "@commitlint/config-conventional": "^11.0.0",
     "@commitlint/prompt-cli": "^11.0.0",
     "@size-limit/preset-small-lib": "^4.12.0",
-    "@typescript-eslint/eslint-plugin": "^4.29.1",
-    "@typescript-eslint/parser": "^4.29.1",
-    "eslint": "^7.32.0",
+    "@typescript-eslint/eslint-plugin": "^5.2.0",
+    "@typescript-eslint/parser": "^5.2.0",
```
dierat, line 54: 5.3.0 was just published this morning.

dierat (approved):

packages/components/docs/token-table.tsx
ahuth: Added the rule in #697

package.json
ahuth: 👍 thanks

PR #696, opened 2021-10-29
RFC - Replace babel build system with tsc

ahuth:
### Summary:

Proposal: replace babel with the typescript compiler.

This greatly simplifies the build system, at the cost of the kinds of things we can build. Babel is much more extensible and capable. But TypeScript's `tsc` does everything we need it to for the design system.

### Test Plan:

- CI
- Local build
- Local storybook
- Local tests
- [x] Publish a pre-release version of the package (https://github.com/FB-PLP/traject/pull/9039)


packages/components/tsconfig.json
```diff
@@ -16,7 +17,7 @@
     "skipLibCheck": true,
     "strict": true,
     "suppressImplicitAnyIndexErrors": true,
-    "target": "es5"
+    "target": "es6"
```
ahuth, line 20: Our Babel build was targeting the last 2 versions of each browser, which doesn't meet Edu's browser support policy.

Unfortunately tsc doesn't have the ability to target specific browsers, so we give up that ability.

es6/es2015 is more broad than we need, but I think it's reasonable for the design system to output it.

packages/components/package.json
```diff
@@ -23,12 +23,11 @@
     "tailwind.config.js"
   ],
   "scripts": {
-    "build": "yarn run build:clean && yarn run build:js && yarn run build:styles && yarn run build:types && yarn run build:finish",
+    "build": "yarn run build:clean && yarn run build:js && yarn run build:styles && yarn run build:finish",
     "build:clean": "rm -rf lib/",
     "build:finish": "rm lib/**/*.spec.* lib/**/*.stories.*",
-    "build:js": "babel src -d lib --extensions '.js,.jsx,.ts,.tsx'",
+    "build:js": "tsc",
```
ahuth, line 29: I first looked at using esbuild for compilation, which is totally doable. But we would still need to use `tsc` to get the type definitions, so we may as well also use `tsc` to compile to js.

ahuth:
Interesting that codecov thinks we gained 5% of coverage with this 🤔 

And that size-limit thinks we gained 5% of output size 🤔 🤔 (this one makes some sense, since we've changed the build target to es6)

packages/components/tsconfig.json
jinlee93: LP browser support for FF is "firefox >= 52" but according to https://caniuse.com/es6, partial support for ff v52, not sure how accurate that is tho

ahuth: Good catch. The .browserslist in LP is actually out of date. Sometime we should update it to what's at https://go.czi.team/edu-browser-support.

For Firefox we actually want to support >= 86

anniehu4:
> Interesting that codecov thinks we gained 5% of coverage with this 🤔

lol codecov is ever mysterious

packages/components/package.json
anniehu4: few things I want to check:
1. input / output directories are now fully controlled by `tsconfig.json`?
2. are declaration files still being output into `lib/`?
3. should we consider using the `--build` or `--project` flag?

anniehu4 (commented):
> This greatly simplifies the build system, at the cost of the kinds of things we can build

definitely looks simpler! left a few clarifying Qs and was also wondering if you could give a quick overview of the cost -- i.e. at a high level, what things could we no longer do?

packages/components/package.json
ahuth: Good questions.

> input / output directories are now fully controlled by tsconfig.json?

That's right, the in and out directories are not fully controlled by tsconfig.json (which was already configured to do so. Probably just to match the babel config).

> are declaration files still being output into lib/?

Yep, declaration files are still being output. And we're now getting sourcemaps output, too. This technically makes the package larger, but they're pretty small and may be useful for debugging in LP.

> should we consider using the --build or --project flag?

We don't need `--project`, since it's for specifying a specific tsconfig.json, but it'll find the one in the package's root directory no problem.

We also don't need `--build` as currently written. That's for use with multiple typescript projects which we're not making use of right now. Could think about it if we added more packages.

https://github.com/chanzuckerberg/frontend-libs does use `--build`, and there are some nice and annoying things about it.

ahuth:
>  if you could give a quick overview of the cost -- i.e. at a high level, what things could we no longer do?

Sure. 2 things come to mind
- We can no longer target specific browsers. The best we can do is target specific JS versions (i.e. es5, es6, es2016, es2017, so on and so forth)
- We can no longer bring in arbitrary babel plugins to transform arbitrary stuff. Want Sass? It's possible with Babel. Want Flow, Elm, random proposed syntaxes? All possible with Babel. But not with `tsc`

I think for our use case sticking to what the `tsc` compiler makes possible is reasonable. Especially since we're sticking with TypeScript and React for the foreseeable future.

anniehu4 (approved):
Cool, thanks!

ahuth:
Thanks!

I'm going to merge. If we ever want to change to something different in the future (esbuild, back to Babel, typescript project references) it should be pretty straightforward.

PR #697, opened 2021-11-01
Turn on import/no-extraneous-dependencies

ahuth:
### Summary:

Turn on the import/no-extraneous-dependencies rule, which will force us to
- Only import packages that we have explicit dependencies on. That way our code doesn't accidentally work in our context, but doesn't break in others.
- Only import dev dependencies in specific "dev" files/paths

### Test Plan:

- `yarn lint`


dierat (approved):
Makes sense to me 👍

ahuth:
Thanks!

PR #704, opened 2021-11-02
CI tweaks

ahuth:
### Summary:

This PR
- Removes the unnecessary build matrices from the docs and size-limit jobs.
- Only runs tests on Node 14, instead of 14 and 12. We only have apps running on Node 14 at this point.
- Rename `chromatic-deployment` to `storybook`

### Test Plan:

- CI

### DIFF
--- a/.github/workflows/size-limit.yml
+++ b/.github/workflows/size-limit.yml
@@ -6,11 +6,6 @@ on:
 jobs:
   size:
     runs-on: ubuntu-latest
-
-    strategy:
-      matrix:
-        node-version: [12.x]
-
     env:
       CI_JOB_NUMBER: 1
     steps:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x]
+        node-version: [14.x]
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,7 +51,7 @@ jobs:
       - name: Build Dist 🔧
         run: yarn run build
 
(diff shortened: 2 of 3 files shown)


.github/workflows/size-limit.yml
```diff
@@ -6,11 +6,6 @@ on:
 jobs:
   size:
     runs-on: ubuntu-latest
-
-    strategy:
-      matrix:
-        node-version: [12.x]
```
ahuth, line 12: The node version wasn't even being used in this job. So this is probably copy pasta.

.github/workflows/test.yml
```diff
@@ -51,7 +51,7 @@ jobs:
       - name: Build Dist 🔧
         run: yarn run build
 
-  chromatic-deployment:
+  storybook:
```
ahuth, line 54: Figured this job could be everything that needs storybook to be built (a11y tests, chromatic)

.github/workflows/test.yml
```diff
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x]
+        node-version: [14.x]
```
ahuth, line 11: No point running tests on Node 12 since we no longer have any Node 12 apps.

I did decide to leave the matrix, so we can easily add more (such as Node 16) when the time comes.

.github/workflows/test.yml
jinlee93: Yes, appreciate the simplicity

ahuth:
Ah damnit, we're running into the issue we've hit before when removing jobs... it takes Github a while to figure out that it doesn't need to wait on that job anymore.

jinlee93 (approved):
LGTM, thanks for cleanup

ahuth:
Thanks!

PR #705, opened 2021-11-02
Use manual commitlint

jinlee93:
### Summary:
Tweaks original @commitlint/config-conventional config
- removed subject rules "Start Case" and "Sentence case"
- added "nit" type
### Test Plan:
- Lint ci passes

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,1 +1,131 @@
-module.exports = { extends: ["@commitlint/config-conventional"] };
+module.exports = {
+  rules: {
+    "body-leading-blank": [1, "always"],
+    "body-max-line-length": [2, "always", 100],
+    "footer-leading-blank": [1, "always"],
+    "footer-max-line-length": [2, "always", 100],
+    "header-max-length": [2, "always", 100],
+    "subject-case": [2, "never", ["pascal-case", "upper-case"]],
+    "subject-empty": [2, "never"],
+    "subject-full-stop": [2, "never", "."],
+    "type-case": [2, "always", "lower-case"],
+    "type-empty": [2, "never"],
+    "type-enum": [
+      2,
+      "always",
+      [
+        "build",
+        "chore",
+        "ci",
+        "docs",
+        "feat",
+        "fix",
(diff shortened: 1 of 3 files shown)


commitlint.config.js
```diff
@@ -1 +1,131 @@
-module.exports = { extends: ["@commitlint/config-conventional"] };
+module.exports = {
```
ahuth, line 1: Would it be easier to leave config-conventional, and only override the rules we want to tweak?

commitlint.config.js
```diff
@@ -1 +1,131 @@
...
+  rules: {
+    "body-leading-blank": [1, "always"],
+    "body-max-line-length": [2, "always", 100],
```
ahuth, line 4: Can we remove body and footer max length checks?

commitlint.config.js
```diff
@@ -1 +1,131 @@
...
+        "ci",
+        "docs",
+        "feat",
+        "fix",
+        "perf",
+        "refactor",
+        "revert",
+        "style",
+        "test",
+        "nit",
```
ahuth, line 28: Does lerna's `--conventional-commits` understand the what it should do with `nit`?

jinlee93: Oop didn't know about that making commitlint less configurable, will be closing PR

jinlee93:
lerna's `--conventional-commits` flag makes commitilnt less configurable

commitlint.config.js
anniehu4: man I'm so late to this lol -- is the idea that `--conventional-commits` totally bypasses our own config? I'm starting to think we should look into moving away from lerna next half. Either way, would be nice to be able to automatically build the `CHANGELOG` while using our own config

ahuth: Lerna seems abandoned, so moving away from it would be nice.

Looks like from https://yarnpkg.com/features/release-workflow Yarn will get its own release workflow at some point (experimental now). Haven't seen any conventional commit or auto-changelog integrations, though.

PR #706, opened 2021-11-02
Turn on turbosnap for EDS components

jinlee93:
### Summary:
Turns on Turbosnap for EDS Components
- Manually run script command in github action rather than importing workflow
- Accessibility tests run on the Storybook built by Chromatic
### Test Plan:
- Chromatic stays working
- CI (esp accessibility) tests run and pass


packages/components/package.json
```diff
@@ -30,7 +30,7 @@
     "build:styles": "postcss \"src/**/*.css\" --dir lib/ --base src/ --verbose",
     "build:watch": "yarn run build:js --watch",
     "build:storybook": "build-storybook -o storybook-static",
-    "chromatic": "chromatic --project-token=CHROMATIC_PROJECT_TOKEN --skip 'dependabot/**'",
+    "chromatic": "chromatic --only-changed",
```
ahuth, line 33: Interestingly, on the CI run chromatic actually ignored `--only-changed` because

> Could not retrieve changed files since baseline commit(s).
This typically happens after rebasing.

Something to keep an eye on after merging, to make sure it's actually working as expected.

.github/workflows/test.yml
```diff
@@ -65,17 +65,10 @@ jobs:
           fetch-depth: 0 # 👈 Required to retrieve git history
       - name: Install dependencies
         run: yarn install --immutable
-      - name: Build Storybook 📚
-        run: yarn workspace @chanzuckerberg/eds-components run build:storybook
+      - name: Build Storybook and Publish to Chromatic
```
ahuth, line 68: Can you leave a comment here that we're using Chromatic to build Storybook because that's necessary to use `--only-changed`, and link to the Chromatic issue where they discussed that?

ahuth:
I don't think commitlint is going to be happy with the merge and revert commits 😐

jinlee93:
> I don't think commitlint is going to be happy with the merge and revert commits 😐

I ran the commits msgs against commitlint and no errors

packages/components/package.json
jinlee93: Might be because Turbosnap wasn't on for the main branch, but yes it is something to keep an eye out for

.github/workflows/test.yml
jinlee93: Added comments as suggested, lmk if it's too verbose

packages/components/package.json
jinlee93: Latest commit didn't have that message but rather
> Failed to retrieve dependent story files
🤔

ahuth (approved):
LGTM

PR #708, opened 2021-11-03
test(clickable-style): remove redundant test

dierat:
### Summary:
Part of https://app.shortcut.com/czi-edu/story/160018/reorganize-button-link-and-clickablestyle-stories

Proposed as an alternative to https://github.com/chanzuckerberg/edu-design-system/pull/694

The `ClickableStyle` component currently has a lot of test coverage due to downstream components (`Button`, `Link`, and further down there's the `CloseButton` and places where buttons are used inside other components). I generally prefer for each component to have its own tests, but in this case, it does feel like a bit overkill. I think we should just remove it.

### Test Plan:
The remaining tests pass and the remaining stories load fine.


ahuth (approved):
🤘

PR #709, opened 2021-11-03
docs(button + link + clickable-style): explain differences between components in docstrings

dierat:
### Summary:
I wanted to update the `Button`, `Link`, and `ClickableStyle` docstrings to all explain the differences between these components a little better.

### Test Plan:
Check out `Button` and `Link` in storybook,
verify the docstrings are being rendered nicely.

#### Button docstring in storybook
Button docstring in storybook

#### Link docstring in storybook
Link docstring in storybook


packages/components/src/Link/Link.tsx
```diff
@@ -20,12 +20,15 @@ export type LinkProps = LinkHTMLElementProps & {
 /**
  * Component for making anchor tags.
  *
- * This component is called Link because it should be used to making link elements;
- * however, it can be styled to look like a button. In terms of the look and feel of the
- * component in the UI, the Link and Button components are exactly the same.
+ * This component is called Link because it should be used to make link elements;
```
ahuth, line 23: There's some pedantic nerd deep inside me somewhere that wants to update this to

```suggestion
 * This component is called Link because it should be used to make anchor elements;
```

But it doesn't really matter... 🤔  No, doesn't really matter

ahuth (approved):
LGTM. Thanks for the updates!

packages/components/src/Link/Link.tsx
dierat: No, I think that's a good point – I can see "link element" being confusing. I'll change it to `<a>` to be consistent with the rest of the text. Thanks!

dierat:
Thanks again!

PR #710, opened 2021-11-03
feat(button + link): rename 'plain' variant to 'text'

dierat:
### Summary:
When I paired with design on the newest button styling, we had a conversation about the naming of the "plain" variant. The designs continue to refer to this variant as "icon", which design prefers, but I'm not in favor of (because any of the buttons can have an icon; this variant just happens to require an icon). They said if we can't do "icon", how about "text"? I believe that we on the code side also discussed the name of this thing at length, and it was suggested that "text" feels too vague when we also have the "link" variant. Design is confident that the other designers will not confused "text" and "link" because "link" is such a structural concept in web design. So we compromised on "text".

This PR updates references to the "plain" variant to now call it "text".

### Test Plan:
Verify that there are no visual changes in the Button and Link storybook pages except that the "plain" variant is now referred to as "text".

### DIFF
--- a/packages/components/.storybook/clickableStyleUtils.stories.tsx
+++ b/packages/components/.storybook/clickableStyleUtils.stories.tsx
@@ -73,7 +73,7 @@ export const getStandardSet = (
   </ul>
 );
 
-export const getPlainRecommendedVariants = (
+export const getTextRecommendedVariants = (
   Component: typeof Button | typeof Link,
   componentName: "Button" | "Link",
 ) => (
@@ -84,7 +84,7 @@ export const getPlainRecommendedVariants = (
       </Heading>
       <ul className="grid gap-y-4">
         <li>
-          <Component variant="plain">
+          <Component variant="text">
             <ArrowBackRoundedIcon
               className={styles.arrowBackIcon}
               purpose="decorative"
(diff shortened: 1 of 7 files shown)


anniehu4 (commented):
Part of why I liked `plain` -- the UI Kit variants are already different from the code variants. i.e. in code, we don't call things "primary", "secondary", etc. So it feels totally fine to have slightly different naming in the UI Kit vs. code. And I think "plain" is more communicative as a code variant.

Thoughts? Was there confusion from product teams?

anniehu4:
We could rename the story from Plain -> Text though

dierat:
> Part of why I liked `plain` -- the UI Kit variants are already different from the `variant` prop in code, i.e. in code, we don't call things "primary", "secondary", etc. So it feels totally fine to have slightly different naming in the UI Kit vs. code. And I think "plain" is more communicative as a code variant.

That's true; "flat" and "outline" don't appear in the UI kit anywhere, but they just feel right for code variants.
 
> Was there confusion from product teams?

No, I was just looking at the UI kit and thinking about how we could make it more consistent with storybook.

> We could rename the story from Plain -> Text though, to be closer to the UI Kit

dierat:
I don't think we're going to pursue this direction.

dierat:
> We could rename the story from Plain -> Text though, to be closer to the UI Kit

After having thought about it more, I'm afraid that might make things more confusing. I think I'm just going to leave it as-is. If we get feedback that storybook is confusing because of this, we can revisit it. But I don't actually think it's going to be much of an issue. Designers will probably stick to figma mostly anyway.

PR #712, opened 2021-11-05
Storybook v6.4

ahuth:
### Summary:

This PR
- Updates Storybook to 6.4
- ~~Updates the stories to [component story format v3](https://storybook.js.org/blog/component-story-format-3-0/)~~ (will move to a followup PR)
- ~~Uses Storybook's auto-generated component titles (instead of specifying our own)~~ (nvm, might look at this in a followup)
- Update to test utils that work with Storybook 6.4
- Use @storybook/testing-react instead of our own custom helpers

### Test Plan:

- CI
- Storybook works
- axe-storybook-testing works
- Tests work


packages/components/.storybook/main.js
```diff
@@ -1,9 +1,5 @@
 module.exports = {
-  stories: [
-    "**/*.stories.mdx",
-    "../src/**/*.stories.@(ts|tsx|mdx)",
-    "**/*.stories.@(ts|tsx|mdx)",
-  ],
+  stories: ["../src", "../docs"],
```
ahuth, line 2: Couldn't find a way to get Storybook 6.4 to look at stories in its own .storybook/ directory. So I moved the stories there into the docs/ directory.

Not saying it's not possible, but I couldn't find an obvious way to do it.

packages/components/package.json
```diff
@@ -52,13 +52,12 @@
...
   "devDependencies": {
-    "@chanzuckerberg/axe-storybook-testing": "^4.1.3",
+    "@chanzuckerberg/axe-storybook-testing": "^5.0.0-alpha.1",
     "@chanzuckerberg/story-utils": "^2.1.0",
-    "@storybook/addon-a11y": "^6.3.12",
-    "@storybook/addon-essentials": "^6.3.12",
+    "@storybook/addon-a11y": "^6.4.0-beta.28",
+    "@storybook/addon-essentials": "^6.4.0-beta.28",
     "@storybook/addon-postcss": "^2.0.0",
-    "@storybook/preset-typescript": "^3.0.0",
```
ahuth, line 60: Not used

packages/components/src/Banner/Banner.stories.tsx
```diff
@@ -5,7 +5,6 @@ import Heading from "../Heading";
 import Banner from "./Banner";
 
 export default {
-  title: "Banner",
```
ahuth, line 8: Storybook can now generate titles for us, based on the directory structure.

We can still provide our own if we want, though.

packages/components/src/Button/button.spec.tsx
```diff
@@ -1,31 +1,30 @@
-import { generateSnapshots, prepareStory } from "@chanzuckerberg/story-utils";
+import { generateSnapshots } from "@chanzuckerberg/story-utils";
+import { composeStories } from "@storybook/testing-react";
 import { render, screen } from "@testing-library/react";
 import React from "react";
 import Button from "./button";
 import * as ButtonStoryFile from "./button.stories";
 
+const { Default } = composeStories(ButtonStoryFile);
```
ahuth, line 8: Storybook now has a `@storybook/testing-react` for use with testing library. I've retired our own `prepareStory`, in lieu of this.

packages/components/src/Button/button.spec.tsx
```diff
@@ -1,31 +1,30 @@
...
-      </Button>,
+      <Default className="exampleClassName" data-testid="example-class-name" />,
     );
     expect(screen.getByTestId("example-class-name")).toMatchSnapshot();
   });
 
   it("forwards refs", () => {
     const ref = React.createRef<HTMLButtonElement>();
-    render(prepareStory(ButtonStoryFile.Default, { ref }));
+    render(<Button ref={ref}>Hi</Button>);
```
ahuth, line 27: Unfortunately, @storybook/testing-react doesn't let us pass refs to `React.forwardRef` components.

This kind of makes sense, because we're rendering the **_story_** (which isn't a forwardRef), not the **_component_** (which is).

But still kind of annoying. Working around for now by rendering Button instead of a story.

packages/components/src/Link/Link.stories.tsx
```diff
@@ -39,7 +38,7 @@ const defaultArgs = {
   children: "Link",
   variant: "link" as const,
   color: "brand" as const,
-  href: "",
+  href: "www.example.com",
```
ahuth, line 41: Randomly gave the link examples an href. Looked wrong having an empty string 🧐

dierat: I like to use an empty string because you can click the component to see how it looks when you interact with it without being sent to another page.

ahuth: Makes sense

ahuth: Changing it back. I did notice that the browser will still redirect to another page. With an empty href (or #) I still get redirect to the current story's documentation page... which is weird.

Not sure if there's some browser thing I don't understand, or if Storybook is now intercepting link clicks or something (maybe that's likely).

dierat: Oh yeah, it does. Maybe I should put `event.preventDefault()` on them..

packages/components/src/Button/__snapshots__/button.spec.tsx.snap
```diff
@@ -1070,6 +1070,6 @@ exports[`<Button /> passes test ids down properly 1`] = `
   data-testid="example-test-id"
   type="button"
 >
-  Button with data-testid
+  Button
```
ahuth, line 1073: These snapshot tests are now using stories, instead of directly rendering the component.

ahuth:
@chanzuckerberg/edu-frontend-infra this is ready for review and merge

dierat (approved):

packages/components/src/Link/Link.spec.tsx
```diff
@@ -1,34 +1,39 @@
...
+    render(<Default data-testid="example-test-id" />);
     expect(screen.getByTestId("example-test-id")).toMatchSnapshot();
   });
 
   it("passes class names down properly", () => {
     render(
-      <Link className="exampleClassName" data-testid="example-class-name">
-        Link with example className
-      </Link>,
+      <Default className="exampleClassName" data-testid="example-class-name" />,
```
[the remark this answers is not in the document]
ahuth: Ah, didn't even notice. That's a totally valid value 😸

ahuth:
Thanks!

packages/components/.storybook/main.js
anniehu4: `docs/` kinda makes more sense anyways 👍

PR #713, opened 2021-11-05
build: stop TSC building the .storybook folder

anniehu4:
### Summary:
As of https://github.com/chanzuckerberg/edu-design-system/pull/692, the build output structure changed to:
```
- lib
    - .storybook
    - Banner
        - Banner.module.css
    -  src
        - Banner (etc)
            - Banner.js
```
i.e. the CSS files were processed correctly, but JS files were being additionally nested into an `src/` folder. This breaks the build in traject.

The issue is that `.storybook` files were being imported into files in `src`, so typescript was also compiling them, then outputting the folders in a way to match the parent directory. The `exclude` option doesn't work because ([reference](https://www.typescriptlang.org/tsconfig#exclude))
> A file specified by exclude can still become part of your codebase due to an **import statement in your code**, a types inclusion, a /// <reference directive, or being specified in the files list. It is not a mechanism that prevents a file from being included in the codebase - it simply changes what the include setting finds.

Two options:
1. (this PR) create a `src/storyUtils` folder for common story utils, which gets removed during the build. In the future, we'll need to make sure things inside `.storybook` are only used inside that folder.
2. Revert to `babel` compilation. I couldn't find a way to configure TS to typecheck story files without also building them, but @ahuth lmk if there is one

### Test Plan:
`yarn build`, confirm structure is as expected again:
Screen Shot 2021-11-05 at 9 57 56 AM


dierat:
Maybe also remove the `.storybook` line from the `tailwind.config` `purge` list: https://github.com/chanzuckerberg/edu-design-system/blob/main/packages/components/tailwind.config.js#L8

dierat (approved):
Thank you so much! ❤️

anniehu4:
> Maybe also remove the .storybook line from the tailwind.config purge list

I thought about this, but it's fine for a self-contained (i.e. not imported into `src/`) story to exist in `.storybook`, and we may want to style it. Can revisit if it's confusing though

ahuth (approved):
LGTM thanks for fixing

If at any point you feel like it's better to go back to Babel, let me know and we can make it happen pretty easily.

anniehu4:
Thanks! Landing now to unblock, will add lint rule soon as a followup

PR #716, opened 2021-11-09
build: restrict importing from .storybook, docs into src

anniehu4:
### Summary:
[ch172777]

### Test Plan:
forced an error:
Screen Shot 2021-11-08 at 5 09 13 PM

confirmed importing from something like `docs-editor/` doesn't trigger the error

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -69,6 +69,25 @@
         "browser": false,
         "node": true
       }
+    },
+    {
+      "files": ["packages/components/src/**/*.stories.{ts,tsx}"],
+      "rules": {
+        "no-restricted-imports": [
+          "error", {
+            "patterns": [
+              {
+                "group": [".storybook"],
+                "message": "Do not import files from `.storybook/` into `src`. If needed, move files into `src/storyUtils` instead."
+              },
+              {
+                "group": ["**/docs/*"],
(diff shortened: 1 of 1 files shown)


.eslintrc
```diff
@@ -69,6 +69,25 @@
...
       }
+    },
+    {
+      "files": ["packages/components/src/**/*.stories.{ts,tsx}"],
+      "rules": {
+        "no-restricted-imports": [
+          "error", {
+            "patterns": [
+              {
+                "group": [".storybook"],
```
anniehu4, line 80: using the path-matching pattern didn't work, so I needed to use the [object form](https://github.com/eslint/eslint/blob/04e91b615743cbb25cd9103e7329b056bd1572ad/docs/rules/no-restricted-imports.md#options). I'm guessing that means the former matches from the beginning of the path, which doesn't work for relative imports

.eslintrc
```diff
@@ -69,6 +69,25 @@
...
+                "message": "Do not import files from `.storybook/` into `src`. If needed, move files into `src/storyUtils` instead."
+              },
+              {
+                "group": ["**/docs/*"],
```
anniehu4, line 84: To prevent false positives from something like `docs-editor/`. Ideally this could be an even more specific path, but was tough to catch relative imports then

dierat (approved):

ahuth (approved):

PR #720, opened 2021-11-16
chore(clickable-style): remove empty space in children

dierat:
### Summary:
We have an empty space in the `ClickableStyle`, `Button`, and `Link` children. The comment says it's there to ensure the height of the button. We added `height` to the button styles a while back, to make sure icons don't expand the buttons and make them too tall, so we don't really need this anymore.

Strangely, the buttons are slightly *taller* after this change. According to the styles, the height is the same, but it's rendering very slightly taller. It's pretty subtle on an individual button.

before and after snapshots of the button stories in chromatic

### Test Plan:
Verify there are no significant changes in the chromatic snapshots.

### DIFF
--- a/packages/components/src/Banner/__snapshots__/Banner.spec.tsx.snap
+++ b/packages/components/src/Banner/__snapshots__/Banner.spec.tsx.snap
@@ -56,7 +56,6 @@ exports[`<Banner /> AlertDismissable story renders snapshot 1`] = `
     class="close dismiss button variantLink colorAlert"
     type="button"
   >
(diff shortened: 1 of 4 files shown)


ahuth (approved):
LGTM

Wow, it's like literally 1px taller (in Chrome, anyway). Super weird.

anniehu4 (approved):
wonder if the buttons are rendering taller or if the space between them is rendering slightly differently now 👀 since the 1st button in the column looks the same. either way think this is fine

PR #721, opened 2021-11-16
chore(close-button): add hover color, aria-label prop, and default prop vals

dierat:
### Summary:
The `CloseButton` component was only really intended to be used in the `Banner` and `Toast` components, but I'm using the `CloseButton` in the `Modal` component in `traject`, and I noticed a few things I wanted to change:
- having to add your own hover color is annoying and not intuitive
- you can't add a custom aria label
- you can't pass other attributes like `data-testid` down to the button
- there are no default `size` and `color` values (I like the be able to plug the component in, verify it works with the default values, and then customize it to meet my needs)

This PR attempts to address these issues.

I'm also wondering if we should pull the `CloseButton` out of `common` or not.

### Test Plan:
Verify the close button hover state is unchanged in the `Banner` and `Toast` components in storybook,
and verify no changes are registered in the chromatic snapshots (except to the sizes are different in the `CloseButton` stories).

### DIFF
--- a/packages/components/src/Banner/Banner.tsx
+++ b/packages/components/src/Banner/Banner.tsx
@@ -107,7 +107,7 @@ export default function Banner({
           className={styles.dismiss}
           color={color}
           onClose={onDismiss}
-          size="28px"
+          size="1.75rem"
         />
       )}
 
--- a/packages/components/src/Toast/Toast.tsx
+++ b/packages/components/src/Toast/Toast.tsx
@@ -57,9 +57,7 @@ export default function Toast({
           {children}
         </Text>
       </div>
-      {onDismiss && (
-        <CloseButton color={color} onClose={onDismiss} size="32px" />
-      )}
+      {onDismiss && <CloseButton color={color} onClose={onDismiss} />}
     </div>
   );
 }
--- a/packages/components/src/common/CloseButton/CloseButton.module.css
+++ b/packages/components/src/common/CloseButton/CloseButton.module.css
@@ -2,7 +2,7 @@
   @apply w-12 h-12 justify-center rounded-full;
 }
 
-/* The hover transition is inherited from the Button component */
+/* The hover transition is inherited from the ClickableStyle component */
 .button:hover {
   @apply text-white;
 
--- a/packages/components/src/common/CloseButton/CloseButton.module.css
+++ b/packages/components/src/common/CloseButton/CloseButton.module.css
@@ -10,26 +10,21 @@
 }
 
 .colorBrand {
-  @apply text-brand-700;
   --close-hover-color: var(--eds-color-brand-600);
 }
 
 .colorNeutral {
-  @apply text-neutral-700;
   --close-hover-color: var(--eds-color-neutral-600);
 }
 
 .colorSuccess {
-  @apply text-success-700;
   --close-hover-color: var(--eds-color-success-600);
 }
 
 .colorWarning {
-  @apply text-warning-700;
   --close-hover-color: var(--eds-color-warning-600);
 }
 
 .colorAlert {
-  @apply text-alert-700;
   --close-hover-color: var(--eds-color-alert-600);
 }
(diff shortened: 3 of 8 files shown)


packages/components/src/Banner/Banner.tsx
```diff
@@ -107,7 +107,7 @@ export default function Banner({
           className={styles.dismiss}
           color={color}
           onClose={onDismiss}
-          size="28px"
+          size="1.75rem"
```
dierat, line 110: I changed the px values to rem, just because.

packages/components/src/Toast/Toast.tsx
```diff
@@ -57,9 +57,7 @@ export default function Toast({
           {children}
         </Text>
       </div>
-      {onDismiss && (
-        <CloseButton color={color} onClose={onDismiss} size="32px" />
```
dierat, line 61: 2rem is now the default value for `size`, so we don't need to pass it in here anymore.

packages/components/src/common/CloseButton/CloseButton.stories.tsx
```diff
@@ -29,11 +23,23 @@ Default.args = {
 export const SmallerSize = Template.bind(null);
 SmallerSize.args = {
   ...defaultArgs,
-  size: "28px",
+  size: "1rem",
```
dierat, line 26: I made reduced this value just so it's more obvious that it's smaller than the other stories.

packages/components/src/common/CloseButton/CloseButton.tsx
```diff
@@ -2,25 +2,27 @@ import clsx from "clsx";
 import React from "react";
 import Button from "../../Button";
 import CloseRoundedIcon from "../../Icons/CloseRounded";
-import { NotificationVariant } from "../Notifications/NotificationIcon";
```
dierat, line 5: It felt weird to me that this button is so closely tied to the notification components now that it's being used somewhere else as well.

packages/components/src/common/CloseButton/CloseButton.tsx
```diff
@@ -2,25 +2,30 @@ import clsx from "clsx";
...
    * @chanzuckerberg/eds-tokens/json/variables.json
    */
-  size: string;
+  size?: string;
+  /**
+   * Custom aria-label. If undefined, "close" will be used.
+   */
+  "aria-label"?: string;
+  onClose: () => void;
+  className?: string;
```
dierat, line 28: I moved these to the bottom because I want the docs pages to have the props people will most want to play around with at the top of the list.

packages/components/src/common/CloseButton/CloseButton.module.css
```diff
@@ -1,11 +1,30 @@
...
+  --close-hover-color: var(--eds-color-success-600);
+}
+
+.colorWarning {
+  --close-hover-color: var(--eds-color-warning-600);
+}
+
+.colorAlert {
+  --close-hover-color: var(--eds-color-alert-600);
+}
```
dierat, line 30: Moved from `common/Notifications/Notification.module.css`.

packages/components/src/common/Notifications/Notification.module.css
```diff
@@ -2,33 +2,28 @@
...
   --border-light-color: var(--eds-color-warning-200);
   --border-dark-color: var(--eds-color-warning-500);
-  --close-hover-color: var(--eds-color-warning-600);
 }
 
 .colorAlert {
   @apply text-alert-700 bg-alert-100;
   --border-light-color: var(--eds-color-alert-200);
   --border-dark-color: var(--eds-color-alert-500);
-  --close-hover-color: var(--eds-color-alert-600);
```
dierat, line 33: Moved to `common/CloseButton/CloseButton.module.css`.

ahuth (approved):
LGTM

anniehu4 (approved):

PR #733, opened 2021-11-30
feat(colors): update grading-yellow-500 token

anniehu4:
### Summary:
[ch34359]

Sean and I are deciding for now that the `grading-yellow-500` token should match the shade of yellow that's currently being used in LP grading (`#F79D2D`). Motivation:
- this will quickly systematize that color
- we're finding it difficult to uniformly convert to the current token
- during Fix-Hack-Learn, the designers are going to take a bigger look at redesigning based on the new brand. so token shifts may happen anyways in the future, systematizing will at least make that easier

### Test Plan:
There are only 3 places currently using this token in `traject`, and they're all:
- related to R/Y/G (so this improves consistency)
- used as a background circle (this new shade is darker, so won't worsen color contrast)


ahuth (approved):
LGTM 🤘

PR #734, opened 2021-11-30
docs(banner): fix display names for subcomponents

dierat:
### Summary:
If you look at the `Banner` code snippets in storybook, it shows the subcomponents being called `BannerTitle` and `BannerMessage`. This is confusing because when you use them in your code, they're `Banner.Title` and `Banner.Message`. The component names in that snippet are coming from the `displayName` of the subcomponent, which are defined in `Banner` when they're first defined. We're taking these components and setting them to `Banner.Title` and `Banner.Message`, but that's not their names under the hood.

This PR adds display names for the `Banner` and its subcomponents so their names appear correctly in the code snippets on the docs page.

Alternatively, we could just update the docstring on the `Banner` with some example code that does use the right component names. https://github.com/chanzuckerberg/edu-design-system/pull/735

#### Before
Banner in storybook, before this change

#### After
Banner in storybook, after this change

### Test Plan:
Checkout the `Banner` in storybook,
look at the `code` feature on any of the stories,
and see that the example code says `Banner.Title` instead of `BannerTitle` and `Banner.Message` instead of `BannerMessage`.

Also verify Chromatic doesn't pick up any visual changes.

### DIFF
--- a/packages/components/src/Banner/Banner.tsx
+++ b/packages/components/src/Banner/Banner.tsx
@@ -113,27 +113,37 @@ export default function Banner({
     </article>
   );
 }
+Banner.displayName = "Banner";
 
+type TitleProps = {
+  children?: React.ReactNode;
+  as: HeadingElement;
+};
 /**
  * This should import a Heading element type
  */
-Banner.Title = (props: { children?: React.ReactNode; as: HeadingElement }) => {
+const BannerTitle: React.FC<TitleProps> = (props: TitleProps) => {
   return props.children ? (
     <Heading as={props.as} color={"inherit"} size="h3">
       {props.children}
     </Heading>
   ) : null;
 };
+Banner.Title = BannerTitle;
 Banner.Title.displayName = "Banner.Title";
 
+type MessageProps = {
+  children?: React.ReactNode;
+};
 /**
  * This should import a Text element type
  */
-Banner.Message = (props: { children?: React.ReactNode }) => {
+const BannerMessage: React.FC<MessageProps> = (props: MessageProps) => {
   return props.children ? (
     <Text color={"inherit"} size="body">
       {props.children}
     </Text>
   ) : null;
 };
+Banner.Message = BannerMessage;
 Banner.Message.displayName = "Banner.Message";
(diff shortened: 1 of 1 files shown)


packages/components/src/Banner/Banner.tsx
```diff
@@ -138,3 +113,27 @@ export default function Banner({
...
+ * This should import a Heading element type
+ */
+Banner.Title = (props: { children?: React.ReactNode; as: HeadingElement }) => {
+  return props.children ? (
+    <Heading as={props.as} color={"inherit"} size="h3">
+      {props.children}
+    </Heading>
+  ) : null;
+};
+Banner.Title.displayName = "Banner.Title";
```
anniehu4, line 127: I played around a little bit locally and the root issue is `displayName` is only defined on component types. I don't understand why TS insists on returning type `Element`, when `Heading` itself is correctly typed as a component though 😭 

Two ways I found of working around it are:
1) Casting as `React.FC` (aka `FunctionalComponent` type):
```
type TitleProps = {
  children?: React.ReactNode;
  as: HeadingElement;
};
/**
 * This should import a Heading element type
 */
const BannerTitle: React.FC<TitleProps> = (
  props: TitleProps,
): React.ReactElement | null => {
  return props.children ? (
    <Heading as={props.as} color={"inherit"} size="h3">
      {props.children}
    </Heading>
  ) : null;
};
Banner.Title = BannerTitle;
Banner.Title.displayName = "Banner.Title";
```
Unfortunately, it's extra verbose because I couldn't get directly casting `Banner.Title: React.FC<TitleProps>` to work

2) Just silencing the TS error 😬 
```suggestion
// @ts-expect-error TS thinks this is not a component type but it is 
Banner.Title.displayName = "Banner.Title";
```

or 3) as you suggested, updating the docstring (also don't know why the docstring would disappear??)

anniehu4: for some reason, setting
```
Banner.displayName = "Banner";
```
_before_ setting the other displayNames gets the docstring to show up for me 🧐

dierat: I updated the PR to cast the types to `React.FC` – thank you so much for figuring that out @anniehu4 ❤️ ❤️ ❤️

anniehu4 (approved):
lovely! thanks for taking this on 🙂

PR #735, opened 2021-11-30
docs(banner): add example to docstring

dierat:
#### Updated summary:
This PR updates the `Banner` docstring to have a code example. It was originally for clarifying something confusing about the code snippets under the stories, but that was fixed in Alternative to https://github.com/chanzuckerberg/edu-design-system/pull/734

#### Original summary:
Alternative to https://github.com/chanzuckerberg/edu-design-system/pull/734

The code snippets for the `Banner` stories in storybook are a little misleading because they show the subcomponents as being called `BannerTitle` and `BannerMessage`, but when you use them in the code, you refer to them as `Banner.Title` and `Banner.Message`.

This PR updates the `Banner` docstring to have a code example that uses the right names and also points out that the code snippets are wrong.

#### Before
Banner component in storybook, before this change

#### After
Banner component in storybook, after this change

### Test Plan:
Check out the `Banner` in storybook and verify the new docstring is present and makes sense.

### DIFF
--- a/packages/components/src/Banner/Banner.tsx
+++ b/packages/components/src/Banner/Banner.tsx
@@ -49,6 +49,22 @@ export type BannerProps = {
 
 /**
  * A banner used to provide and highlight information to a user or ask for a decision or action.
+ *
+ * ```tsx
+ * <Banner
+ *   onDismiss={handleDismiss}
+ *   textContent={
+ *     <>
+ *       <Banner.Title>{bannerTitle}</Banner.Title>
+ *       <Banner.Message>{bannerMessage}</Banner.Message>
+ *     </>
+ *   }
+ * />
+ * ```
+ *
+ * Please note the code snippets under each Banner story on storybook are misleading
+ * because they use <BannerTitle> and <BannerMessage> when they are actually called
+ * <Banner.Title> and <Banner.Message>.
  */
 export default function Banner({
   className,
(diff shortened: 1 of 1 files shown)


packages/components/src/Banner/Banner.tsx
```diff
@@ -74,6 +74,22 @@ Banner.Message = function BannerMessage(props: { children?: React.ReactNode }) {
...
+ *       <Banner.Title>{bannerTitle}</Banner.Title>
+ *       <Banner.Message>{bannerMessage}</Banner.Message>
+ *     </>
+ *   }
+ * />
+ * ```
+ *
+ * Please note the code snippets under each Banner story on storybook are misleading
+ * because they use <BannerTitle> and <BannerMessage> when they are actually called
+ * <Banner.Title> and <Banner.Message>.
```
anniehu4, line 92: could you try
```suggestion
 * because they use `<BannerTitle>` and `<BannerMessage>` when they are actually called
 * `<Banner.Title>` and `<Banner.Message>`.
```
? rendering weird in the screenshot

anniehu4 (approved):

packages/components/src/Banner/Banner.tsx
dierat: Yes, thanks!

anniehu4 (approved):

PR #737, opened 2021-12-01
Component story format v3

ahuth:
### Summary:

Convert stories to [component story format (CSF) v3](https://storybook.js.org/blog/component-story-format-3-0/).

### Test Plan:

- CI

### DIFF
--- a/packages/components/src/Banner/Banner.stories.tsx
+++ b/packages/components/src/Banner/Banner.stories.tsx
@@ -1,4 +1,4 @@
-import type { Story } from "@storybook/react";
+import type { StoryObj } from "@storybook/react";
 import React from "react";
 import Button from "../Button";
 import Heading from "../Heading";
@@ -15,26 +15,17 @@ export default {
       },
     },
   },
+  args: {
+    content:
+      "Summit Learning has a full-time team dedicated to constantly improving our curriculum. To see the updates, click into the course.",
+    heading:
+      "New curriculum updates are available for one or more of your courses.",
+  },
 };
 
 type Args = React.ComponentProps<typeof Banner> & {
   content: string;
   heading: string;
-};
-
-const Template: Story<Args> = (args) => {
-  const { content, heading, ...restArgs } = args;
-  return (
-    <Banner
-      {...restArgs}
-      textContent={
-        <>
-          <Banner.Title as="h1">{heading}</Banner.Title>{" "}
-          <Banner.Message>{content}</Banner.Message>
-        </>
-      }
-    />
-  );
 };
 
 const action = (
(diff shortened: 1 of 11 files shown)


packages/components/src/Banner/Banner.stories.tsx
```diff
@@ -15,186 +15,203 @@ export default {
       },
     },
   },
+  args: {
+    content:
+      "Summit Learning has a full-time team dedicated to constantly improving our curriculum. To see the updates, click into the course.",
+    heading:
+      "New curriculum updates are available for one or more of your courses.",
+  },
```
ahuth, line 23: Top level args are inherited by each story.

We could always do this, but we used to lint against it due to limitations of `prepareStory`.

packages/components/src/Button/button.stories.tsx
```diff
@@ -69,17 +68,21 @@ const gridParameters = {
   },
 };
 
-export const AllVariants = () => getAllVariantsWithStates("button", "Button");
-AllVariants.parameters = gridParameters;
+export const AllVariants = {
```
ahuth, line 71: Generally I think we should explicitly type stories - i.e.

```ts
export const AllVariants: StoryObj<Args> = {
```

But for these stories that don't use or set args, it doesn't really matter.

packages/components/src/Heading/Heading.stories.tsx
```diff
@@ -9,48 +9,53 @@ export default {
...
-
-export const Heading1 = Template.bind(null);
-Heading1.args = {
-  size: "h1",
-  children: "Heading 1 24/32",
+export const Heading1: StoryObj<Args> = {
+  args: {
+    size: "h1",
+    children: "Heading 1 24/32",
+  },
```
ahuth, line 16: Notice that none of the stories actually define a `render` function...

For any story that does nothing but apply `args` to the component - i.e.

```tsx
(args) => <Heading {...args />;
```

we can omit it and Storybook will do that for us.

packages/components/src/SvgIcon/SvgIcon.stories.tsx
```diff
@@ -41,119 +41,139 @@ type Args = React.ComponentProps<typeof SvgIcon> & {
...
 
-const Template: Story<Args> = ({ icon, color, ...rest }) => {
-  const Icon = allIcons[icon];
-  const computedColor = color && ColorTokens[color];
-  return <Icon {...rest} color={computedColor} />;
-};
-
-const defaultArgs = {
-  icon: Object.keys(allIcons)[0] as keyof typeof allIcons,
-  role: "decorative" as const,
```
ahuth, line 52: Notice that `role` here isn't a valid prop (anymore), but TS failed to catch it.

Looks like object stories get us better types, and TS would not catch this.

packages/components/src/Button/button.stories.tsx
anniehu4: Could you explain why we should explicitly type every story?

anniehu4 (approved):
niice love how much more condensed it is

packages/components/src/Button/button.stories.tsx
ahuth: Mostly so it forces us to get the `args` right (if we are using any).

Also I think `composesStories` requires them to be annotated to infer types in tests.

ahuth: Maybe we should think about a lint rule for that 🤔

ahuth:
Thanks!

PR #738, opened 2021-12-01
feat(icons): add InfoRounded

anniehu4:
### Summary:
Needed for some tooltip designs.

### Test Plan:
Screen Shot 2021-12-01 at 5 46 13 PM

it's a bit different from the stickersheet design but I see that the "master" component was deleted so need to check with Sean if that's still the intended look 🤔  easy to change if so
Screen Shot 2021-12-01 at 5 50 40 PM


dierat (approved):
Nice 👍

PR #742, opened 2021-12-03
Update stuff

ahuth:
### Summary:

This PR:
- Fixes plop component generation
- Update everything, basically

### Test Plan:

- CI


packages/components/src/index.ts
```diff
@@ -1,11 +1,11 @@
...
 export { default as ClickableStyle } from "./ClickableStyle";
 export { default as Heading } from "./Heading";
-export { default as SvgIcon } from "./SvgIcon";
-export type { IconProps } from "./SvgIcon";
 export { default as Link } from "./Link";
+export { default as SvgIcon } from "./SvgIcon";
 export { default as Tag } from "./Tag";
 export { default as Text } from "./Text";
 export { default as Toast } from "./Toast";
+export type { IconProps } from "./SvgIcon";
```
ahuth, line 11: Plop alphabetized these. I left the changes

plopfile.js
```diff
@@ -50,13 +50,11 @@ module.exports = function (plop) {
         const indexFile = `${plop.getDestBasePath()}/packages/components/src/index.ts`;
 
         if (fs.existsSync(indexFile)) {
-          const nameRegex = /default as ([^\s]*)/;
```
ahuth, line 53: So I think plop is broken on `main` (but haven't tested that). Likely broke when we started exporting types, as I don't think this regex handles that.

Not related to the updates, but I noticed this when trying to update Plop.

plopfile.js
```diff
@@ -50,13 +50,11 @@ module.exports = function (plop) {
...
-          const sorted = fs
-            .readFileSync(indexFile, "utf8")
-            .split(";\n")
+          const lines = fs.readFileSync(indexFile, "utf8").split(";\n");
+          const sorted = lines
             .sort((a, b) => {
-              const aName = a.match(nameRegex)[1];
-              const bName = b.match(nameRegex)[1];
```
ahuth, line 59: 🤔 Why [1] and not [0]?

.stylelintrc
```diff
@@ -16,8 +17,10 @@
         ]
       }
     ],
+    "color-function-notation": null,
```
ahuth, line 20: Latest stylelint-config-standard wants us to write

```css
rgba(1 2 3 0.5)
```

instead of

```css
rgba(1, 2, 3, 0.5)
```

Didn't seem worth worrying about too much, so I turned off the rule. There's only 1 violation, though, so it would be easy to turn back on if we want.

.stylelintrc
```diff
@@ -16,8 +17,10 @@
...
     "custom-property-empty-line-before": "never",
     "no-eol-whitespace": null,
+    "selector-class-pattern": null,
```
ahuth, line 23: Stylelint now doesn't like `.fooBar` selectors (it wants `.foo-bar`).

I figured we know what we're doing and don't need this rule. But if we want it we could supply a different regex to match whatever pattern we want.

.stylelintrc
```diff
@@ -3,6 +3,7 @@
     "stylelint-config-standard"
   ],
   "rules": {
+    "alpha-value-notation": null,
```
ahuth, line 6: stylelint-config-standard doesn't like alpha values of `0.5` anymore. It wants `50%` for some reason.

Figured it doesn't really matter to us, and turned off the rule.

package.json
```diff
@@ -17,7 +17,7 @@
     "generateFlowTypes": "./packages/components/generateFlowTypes.sh",
     "lint": "yarn run lint:styles && yarn run lint:scripts",
     "lint:fix": "yarn run lint:styles:fix && yarn run lint:scripts:fix",
-    "lint:styles": "stylelint --ignore-path .gitignore packages/**/*.{css,js,jsx,ts,tsx}",
```
ahuth, line 20: There's no reason to stylelint the JS files, right?

On the latest version it completely explodes on them.

PR #743, opened 2021-12-03
Fix plops

ahuth:
### Summary:

Fix plopping.

### Test Plan:

- Plopped

### DIFF
--- a/packages/components/src/index.ts
+++ b/packages/components/src/index.ts
@@ -1,11 +1,11 @@
 export { default as Banner } from "./Banner";
-export type { BannerProps } from "./Banner";
 export { default as Button } from "./Button";
 export { default as ClickableStyle } from "./ClickableStyle";
 export { default as Heading } from "./Heading";
-export { default as SvgIcon } from "./SvgIcon";
-export type { IconProps } from "./SvgIcon";
 export { default as Link } from "./Link";
+export { default as SvgIcon } from "./SvgIcon";
 export { default as Tag } from "./Tag";
 export { default as Text } from "./Text";
 export { default as Toast } from "./Toast";
+export type { BannerProps } from "./Banner";
+export type { IconProps } from "./SvgIcon";
--- a/plopfile.js
+++ b/plopfile.js
@@ -40,7 +40,7 @@ module.exports = function (plop) {
         separator: "",
         path: "packages/components/src/index.ts",
         template:
-          'export { default as {{pascalCase name}} } from "./{{pascalCase name}}"',
+          'export { default as {{pascalCase name}} } from "./{{pascalCase name}}";',
       },
       // From https://github.com/bradfrost/czi-vanilla-storybook
       function sortIndex() {
@@ -50,17 +50,15 @@ module.exports = function (plop) {
         const indexFile = `${plop.getDestBasePath()}/packages/components/src/index.ts`;
 
         if (fs.existsSync(indexFile)) {
-          const nameRegex = /default as ([^\s]*)/;
-          const sorted = fs
-            .readFileSync(indexFile, "utf8")
-            .split(";\n")
-            .sort((a, b) => {
-              const aName = a.match(nameRegex)[1];
-              const bName = b.match(nameRegex)[1];
-              return aName.localeCompare(bName);
-            })
-            .join(";\n");
-          fs.writeFileSync(indexFile, sorted + ";\n");
+          // Split the index file into lines.
+          const lines = fs.readFileSync(indexFile, "utf8").split("\n");
+
+          // Sort the lines.
+          const sorted = lines.sort((a, b) => a.localeCompare(b)).join("\n");
+
+          // Write the sorted lines back to the file.
+          fs.writeFileSync(indexFile, sorted + "\n");
+
           return `index.ts lines sorted`;
         }
       },


packages/components/src/index.ts
```diff
@@ -1,11 +1,11 @@
...
 export { default as ClickableStyle } from "./ClickableStyle";
 export { default as Heading } from "./Heading";
-export { default as SvgIcon } from "./SvgIcon";
-export type { IconProps } from "./SvgIcon";
 export { default as Link } from "./Link";
+export { default as SvgIcon } from "./SvgIcon";
 export { default as Tag } from "./Tag";
 export { default as Text } from "./Text";
 export { default as Toast } from "./Toast";
+export type { IconProps } from "./SvgIcon";
```
ahuth, line 11: Our plop alphabetizes this file. Even though I didn't keep around any test components, I kept the alphabetization.

plopfile.js
```diff
@@ -50,13 +50,11 @@ module.exports = function (plop) {
         const indexFile = `${plop.getDestBasePath()}/packages/components/src/index.ts`;
 
         if (fs.existsSync(indexFile)) {
-          const nameRegex = /default as ([^\s]*)/;
```
ahuth, line 53: I think plop was broken on `main` (although I didn't test that). Noticed while doing some upgrades that it's broken. Probably broke when we started exporting types from index.ts.

plopfile.js
```diff
@@ -50,13 +50,11 @@ module.exports = function (plop) {
...
-          const sorted = fs
-            .readFileSync(indexFile, "utf8")
-            .split(";\n")
+          const lines = fs.readFileSync(indexFile, "utf8").split(";\n");
+          const sorted = lines
             .sort((a, b) => {
-              const aName = a.match(nameRegex)[1];
-              const bName = b.match(nameRegex)[1];
```
ahuth, line 59: Why [1] and not [0]? 🤔

anniehu4: `match` returns [full match, ...capturing group(s)] right?

plopfile.js
```diff
@@ -67,3 +65,37 @@ module.exports = function (plop) {
...
+  // Get the exported name. We have to handle cases like `export { default as Foo }` and
+  // `export type { Foo }`. To match these cases, this is currently looking for:
+  //
+  // 1. Opening curly bracket, for either `export {` or `export type {`.
+  // 2. Any characters immediately after the opening bracket. Mostly to match `default as`.
+  // 3. Capture the name of the thing that's exported.
+  // 4. Closing curly bracket.
+  //
+  //                            1 2   3    4
+  //                            ┃ ┃ ┏━┻━┓  ┃
```
anniehu4, line 90: 😮  how did you do that

plopfile.js
```diff
@@ -67,3 +65,37 @@ module.exports = function (plop) {
...
+/**
+ * Extract an exported name from an export statement.
+ *
+ * @example
+ * getNameFromExport('export { default as FooBear } from "./FooBear";');
+ * // => 'FooBear'
+ *
+ * @example
+ * getNameFromExport('export type { FooBat } from "./FooBat";');
+ * // => 'FooBat'
```
anniehu4, line 78: I wonder if, based on the `IconProps` example, it's better (and easier) to order based on file name? I think the only reason we want the sorting is to maintain some readability, and I don't actually care how the type vs. default export is ordered for a specific file as long as they're grouped

plopfile.js
ahuth: You're probably right. Would explain why `export type { foo }` went to the bottom of the list.

plopfile.js
ahuth: That sounds way better. And would better match how import statements should be ordered.

plopfile.js
ahuth: Unicode box drawing characters: https://en.wikipedia.org/wiki/Box-drawing_character

plopfile.js
ahuth: Wish the eslint imports plugin could order these for us. Looks like it doesn't

plopfile.js
```diff
@@ -50,17 +50,15 @@ module.exports = function (plop) {
...
-              const bName = b.match(nameRegex)[1];
-              return aName.localeCompare(bName);
-            })
-            .join(";\n");
-          fs.writeFileSync(indexFile, sorted + ";\n");
+          // Split the index file into lines.
+          const lines = fs.readFileSync(indexFile, "utf8").split("\n");
+
+          // Sort each line.
+          const sorted = lines.sort((a, b) => a.localeCompare(b)).join("\n");
```
ahuth, line 57: Instead of extracting the exported name or path, I decided to sort by the entire line. End result is we get 2 groups: the components and the types

```ts
export a from 'a';
export b from 'b';
export c from 'c';
export type b from 'b';
export type c from 'c';
```

I think that's fine for our purposes.

plopfile.js
```diff
@@ -50,17 +50,15 @@ module.exports = function (plop) {
...
-            .readFileSync(indexFile, "utf8")
-            .split(";\n")
-            .sort((a, b) => {
-              const aName = a.match(nameRegex)[1];
-              const bName = b.match(nameRegex)[1];
-              return aName.localeCompare(bName);
-            })
-            .join(";\n");
-          fs.writeFileSync(indexFile, sorted + ";\n");
+          // Split the index file into lines.
```
dierat, line 53: This section is so much simpler now ❤️

dierat (approved):
I am ambivalent about how things are sorted, but the code change is ✨

plopfile.js
anniehu4: I'm good with this! is consistent & predictable 👍 thanks for noticing & fixing this issue

anniehu4 (approved):

PR #746, opened 2021-12-07
chore(link): add fake href, stop navigating away in stories

dierat:
### Summary:
We have some `Link` stories that have an `href` that's just an empty string. When you click on those links, you're navigated to a different page that just has that story, which makes it kind of annoying to test the click interaction on these things.

This PR upgrades the `href` value to an actual example URL and also adds an `onClick` handler to stop the page navigation.

Please note that this only affects the top 3 stories (default, in body text, in a heading) and the big grids at the bottom (all variants) because I'm only updating the stories that already had an `href`. The other stories would be a little more work and repetition, and I don't think we need this for every single story. A couple is enough.

### Test Plan:
Navigate to `http://localhost:6008/?path=/docs/link--default`,
test clicking all the links in the stories,
and verify the top 3 and the big grids at bottom show a pointer cursor and focus the link on click but don't navigate you to another page.

### DIFF
--- a/packages/components/src/Link/Link.stories.tsx
+++ b/packages/components/src/Link/Link.stories.tsx
@@ -33,7 +33,7 @@ export default {
     children: "Link",
     variant: "link" as const,
     color: "brand" as const,
-    href: "www.example.com",
+    href: "https://go.czi.team/eds",
     onClick: (event: React.MouseEvent<HTMLElement>) => {
       // Allows the user to click the links for testing without being navigated away.
       event.preventDefault();
--- a/packages/components/src/Link/Link.stories.tsx
+++ b/packages/components/src/Link/Link.stories.tsx
@@ -33,7 +33,11 @@ export default {
     children: "Link",
     variant: "link" as const,
     color: "brand" as const,
-    href: "",
+    href: "www.example.com",
+    onClick: (event: React.MouseEvent<HTMLElement>) => {
+      // Allows the user to click the links for testing without being navigated away.
+      event.preventDefault();
+    },
   },
 };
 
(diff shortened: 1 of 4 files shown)


packages/components/src/storyUtils/clickableStyleUtils.tsx
```diff
@@ -298,7 +298,11 @@ const getVariantWithStates = (
                     variant={variant}
                     state={state}
                     disabled={state === "disabled"}
-                    href={tag === "a" ? "" : undefined}
+                    href={tag === "a" ? "www.example.com" : undefined}
+                    onClick={(event: React.MouseEvent<HTMLElement>) => {
```
ahuth, line 302: Can TypeScript figure out the type of `event` without the explicit annotation here?

```suggestion
                    onClick={(event) => {
```

ahuth (approved):
LGTM

packages/components/src/Link/Link.stories.tsx
```diff
@@ -33,7 +33,11 @@ export default {
     children: "Link",
     variant: "link" as const,
     color: "brand" as const,
-    href: "",
+    href: "www.example.com",
```
anniehu4, line 36: nit 🤓 
```suggestion
    href: "https://go.czi.team/eds",
```

anniehu4 (approved):
ty!!

packages/components/src/storyUtils/clickableStyleUtils.tsx
dierat: Guess not; I get
```
➤ YN0000: src/storyUtils/clickableStyleUtils.tsx(302,31): error TS7006: Parameter 'event' implicitly has an 'any' type.
```

anniehu4:
subliminal messaging lmao

PR #752, opened 2021-12-15
chore(clickable-style): give plain button hover style a border

dierat:
### Summary:
Ticket: https://app.shortcut.com/czi-edu/story/176514/revisit-hover-background-color-on-plain-variant

There's been some back and forth about the "plain" `Button` variant's hover style. The button has a very light background color, which looks great on white, but it kind of disappears into the light gray background that is commonly used in LP.

example of plain variant button on a gray background with the hover style visible

I talked to Sean and we decided to add a 2px border on the hover style (so it kind of turns into an "outline" variant button but it still has the light background color). I think the hope is that one day we may no longer have these gray backgrounds, but I don't think that day is going to come anytime soon. If it does, we can consider dropping the outline.

### Test Plan:
Navigate to `http://localhost:6008/?path=/story/button--all-variants`,
verify the hover and focus states are updated to have a border that matches the text color,
verify the other states are unchanged,
and specifically make sure that the disabled and active states don't now have a border that's a different color than the background.

#### Before
all variants story in storybook, before this change

#### After
all variants story in storybook, after this change

### DIFF
--- a/packages/components/src/ClickableStyle/ClickableStyle.module.css
+++ b/packages/components/src/ClickableStyle/ClickableStyle.module.css
@@ -196,7 +196,7 @@
 
 .variantPlain {
   color: var(--link-color);
-  border: var(--plain-background-color);
+  border-color: var(--plain-background-color);
   background-color: var(--plain-background-color);
 
   &:hover,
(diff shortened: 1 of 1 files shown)


dierat:
Side note  – I'm also wondering if it would be more useful to have "All variants on **gray** background" instead of "All variants on **dark** background" because 1) the purpose of those stories was just to visually communicate the opacity (because outline buttons have a transparent background) and 2) we don't actually have very dark backgrounds in LP as far as I know.

packages/components/src/ClickableStyle/ClickableStyle.module.css
```diff
@@ -196,22 +196,27 @@
 
 .variantPlain {
   color: var(--link-color);
-  border: var(--plain-background-color);
```
dierat, line 199: Invalid CSS.

ahuth (approved):
LGTM 🤘

anniehu4 (approved):

PR #753, opened 2021-12-15
chore(button, link): make dark background stories gray instead

dierat:
### Summary:
The `Button` and `Link` components both have a story that shows all of the variants of the component in one size (large) on a dark background. The reason we added these stories was to make the opacity of the buttons clear – specifically it's obvious when looking at these buttons that the outline variant has a transparent background (which there was some confusion around before). I used the standard "dark" background that comes with storybook because that accomplished the goal.

What I would like to do with this PR is change those stories to use the same gray that LP uses. This way, the stories will still communicate the opacity of the buttons but it will also help us see how well the buttons are working on this gray, which is very common in LP. I think the hope is that one day those gray backgrounds will be removed, but, for now, we have to live with it and these buttons are going to be living on it.

I also added a dark background that uses one of the EDS colors because I imagine it will still come in handy, but now it'll be consistent with the design system.

### Test Plan:
Navigate to `http://localhost:6008/?path=/story/button--large-variants-on-gray-background&globals=backgrounds.value:transparent` and `http://localhost:6008/?path=/story/link--large-variants-on-gray-background&globals=backgrounds.value:transparent`,
verify the background is now LP gray with dark text headings and labels,
verify you can still pick a dark background from the toolbar,
and verify all the other stories still have white backgrounds.

#### Before
The Link component's large variants on dark background, before this change

#### After
The Link component's large variants on gray background, after this change

### DIFF
--- a/packages/components/.storybook/preview.tsx
+++ b/packages/components/.storybook/preview.tsx
@@ -12,3 +12,18 @@ export const decorators = [
     </div>
   ),
 ];
+
+export const parameters = {
+  backgrounds: {
+    values: [
+      {
+        name: "gray",
+        value: "#f3f3f3",
+      },
+      {
+        name: "dark",
+        value: "#21272D", // eds-color-neutral-700
+      },
+    ],
+  },
+};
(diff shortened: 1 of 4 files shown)


ahuth (approved):
LGTM

PR #755, opened 2021-12-17
feat: add Checkbox component

anniehu4:
### Summary:
[ch164178]

Copies over Checkbox component from traject EDSCandidates. Mostly unchanged except for removing Flow-isms and traject style overriding.

### Test Plan:
- Storybook should match traject stories
- Published a [0.4.10-alpha release](https://www.npmjs.com/package/@chanzuckerberg/eds-components/v/0.4.10-alpha.0) and tested in traject


packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,217 @@
...
+ * Design system Checkbox.
+ */
+const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
+  (props, ref) => {
+    // All remaining props are passed to the `input` element
+    const { className, id, label, size, ...rest } = props;
+
+    if (
+      process.env.NODE_ENV !== "production" &&
+      !label &&
```
jeremiah-clothier, line 194: This would be a more thorough check for label text. Do we change the logic for checking a label to something more like that in case the developer passes a component without text/number as `label`?

https://github.com/FB-PLP/traject/blob/2a29bc21ae73061891bcc044761f98c20659bd03/app/assets/javascripts/v2/core/EDSCandidates/button.tsx#L200-L213

packages/components/src/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+      onChange={handleChange}
+      {...defaultArgs}
+      {...args}
+    />
+  );
+}
+
+export const Default: StoryObj<Args> = {};
+
+export const Checked = {
```
jeremiah-clothier, line 49: nit: could update all the `export const ...` types to `StoryObj<Args>` and get help typing from VSCode

jeremiah-clothier:
Not sure chromatic sends notifications: Looks like this story needs updating so that the checkbox appears on the left. https://www.chromatic.com/test?appId=61313967cde49b003ae2a860&id=61ddfb11677db1003a1ce4fd

packages/components/jest.setup.ts
```diff
@@ -1 +1,13 @@
...
+beforeEach(() => {
+  jest.spyOn(console, "error").mockImplementation(() => null);
+});
+
+// Restore all sandbox spies/mocks and clean up fixture DOM after _each_ test
+afterEach(() => {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore jest.spyOn adds this functionallity
+  console.error.mockRestore();
+});
```
ahuth, line 13: I forget why we need these. Can remove them?

packages/components/src/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const [checked, setChecked] = React.useState(true);
+  const handleChange = () => {
+    setChecked(!checked);
+  };
+
+  return (
+    <Checkbox
+      checked={checked}
+      onChange={handleChange}
+      {...defaultArgs}
```
ahuth, line 41: Would it be better to pass `defaultArgs` as `args` on the stories using CheckedExample, instead of directly passing to the component here? That way all args are handled the same?

packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,217 @@
...
+      style={svgStyle}
+    >
+      {children}
+    </label>
+  );
+};
+
+/**
+ * Design system Checkbox.
+ */
```
ahuth, line 186:
```suggestion
/**
 * Checkbox control indicating if something is on or off.
 * 
 * Requires either a visible label or an accessible name.
 */
```

packages/components/src/Checkbox/Checkbox.tsx
anniehu4: that's a good instinct! IIRC a lot of that logic was added specifically to handle icons awhile back. Now that Andrew's added a11y checking at the Icon level (i.e. informative icons must have text titles), I think a lighter level of checking here is fine

packages/components/src/Checkbox/Checkbox.stories.tsx
anniehu4: makes sense 😁

packages/components/jest.setup.ts
anniehu4: we test throwing an error in `Checkbox.spec.tsx`, so without this, Jest surfaces the console error in `yarn test` and it's kinda messy
Screen Shot 2022-01-12 at 6 32 58 PM

I should cleanup the comments some though

anniehu4: hmm ok here's the test for reference:
```
  test("throws an error if no label or aria-label", () => {
    expect(() => {
      render(<Checkbox />);
    }).toThrow(/must provide a visible label or aria-label/);
  });
```

The test passes regardless of mocking `console.error` or not, so I was referencing https://github.com/facebook/jest/pull/5267#issuecomment-356605468.

I'm down to make all console error calls throw for consistency w/ the other repos though! I figured we might test for thrown errors in more components so it made sense to handle it globally

packages/components/src/Checkbox/Checkbox.stories.tsx
anniehu4: Updated for all stories except the 3 that use custom renders

packages/components/src/Checkbox/Checkbox.module.css
```diff
@@ -0,0 +1,76 @@
...
+
+  /* Hide the input element visually.
+  This ensures the checkbox is still "physically" present so that all users,
+  especially on touch screen readers, still interact with the real checkbox element
+  where it would naturally be present. */
+  opacity: 0;
+}
+
+.checkboxInput + svg {
+  @apply text-neutral-500 align-top inline-block;
```
anniehu4, line 47: Added `inline-block` to match SvgIcon styles & to override [Tailwind Preflight's ](https://tailwindcss.com/docs/preflight#images-are-block-level) `display: block` styling.

confirmed custom positioning story is as expected:
Screen Shot 2022-01-14 at 2 58 55 PM

packages/components/src/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,122 @@
...
+const defaultArgs = {
+  disabled: false,
+  label: "Checkbox",
+};
+
+/**
+ * Controlled example to make checked stories interactive.
+ * We name this "Checkbox" for Storybook's code documentation
+ */
+function Checkbox({ checked: defaultChecked = false, ...rest }: Args) {
```
anniehu4, line 14: Made this the default component for Storybook b/c I liked what @jinlee93 did in the [Toggle](https://github.com/FB-PLP/traject/blob/master/app/assets/javascripts/v2/core/EDSCandidates/Toggle/Toggle.stories.tsx#L9)

packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,211 @@
+import clsx from "clsx";
+import React from "react";
+import { useUID } from "react-uid";
+import styles from "./Checkbox.module.css";
+
+// Allow all native HTML input attributes,
+// except for ones where we override type or strictness
+type CheckboxHTMLElementProps = Omit<
```
anniehu4, line 8: thanks to @jeremiah-clothier for reminding me that this is probs more maintainable going forward. Note that we do lose some explicit documentation (i.e. `disabled`, `aria-label`, and `readOnly` props are no longer explicitly here). I think this is an okay trade-off because there inevitably may be more HTML attributes that people need (e.g. `readOnly` was added because of a React warning we saw), but open to push-back.

Another thing to note is Storybook is having issues parsing the props in either case 😬 

Now, it only shows props in `defaultArgs` or `argTypes` and doesn't load descriptions:


Previously, it loaded more props but didn't include `size` and marked `id` as required (which isn't true for the overall component):


I'll make a ticket to dig into this more, maybe we're doing something wrong with the docgen config

packages/components/src/Checkbox/Checkbox.stories.tsx
anniehu4: Another thing to note is Storybook is having issues parsing the props in either case 😬 

Now, it only shows props in `defaultArgs` or `argTypes` and doesn't load descriptions:


Previously, it loaded more props but didn't include `size` and marked `id` as required (which isn't true for the overall component):


I'll make a ticket to dig into this more, maybe we're doing something wrong with the docgen config

packages/components/src/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+  },
+  argTypes: {
+    // For some reason Storybook does not infer all props correctly;
+    // we manually include the most relevant controls here.
+    checked: {
+      description: `Whether checkbox is checked. Defaults to false.
+        "indeterminate" can be used when a checkbox visually represents
+        a list of checkboxes that is "partially" checked.`,
+      control: "radio",
+      options: [true, false, "indeterminate"],
```
jeremiah-clothier, line 35: The `checked` controls aren't doing anything for me in the storybook. They do not update the state of the story. I think the story component may need to be updated because the checked prop is only a `defaultChecked` prop.

https://61313967cde49b003ae2a860-tshkmlvpue.chromatic.com/?path=/story/checkbox--default

anniehu4: oh no, weird, it works locally 😭 will investigate

anniehu4: gah something weird is happening with the state assignment:
Screen Shot 2022-01-18 at 1 24 18 PM

I'm gonna timebox this and just go back to the traject structure of stories

packages/components/src/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const handleChange = () => {
+    setChecked(!checked);
+  };
+
+  return <Checkbox checked={checked} onChange={handleChange} {...args} />;
+}
+
+export const Default: StoryObj<Args> = {};
+
+export const Checked: StoryObj<Args> = {
```
anniehu4, line 46: went back to the old (in traject) structure, confirmed Docs page & Controls panel work:
Kapture 2022-01-18 at 10 56 10

jeremiah-clothier (approved):

packages/components/src/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,120 @@
...
+const defaultArgs = {
+  disabled: false,
+  label: "Checkbox",
+};
+
+export default {
+  title: "Checkbox",
+  component: Checkbox,
+  args: defaultArgs,
+  argTypes: {
```
anniehu4, line 14: docs page:
Screen Shot 2022-01-18 at 10 42 27 AM

see https://github.com/chanzuckerberg/edu-design-system/pull/755#discussion_r785194198

packages/components/jest.setup.ts
ahuth: Oh, okay. I didn't fully understand what was happening here, @anniehu4.

This is testing that an error was actually thrown, and for some reason that still gets `console.error`-ed. And to keep non-actionable stuff like that from being printed, we suppress the console.error.

I'd personally mock as necessary, and if it starts being every damn component we could move to global. But if being global could make sense, as well.

packages/components/src/Checkbox/Checkbox.stories.tsx
ahuth: I also wonder if react-docgen-typescript expects to find types/descriptions associated with the component itself, and not passed through an intermediary.

packages/components/jest.setup.ts
```diff
@@ -1 +1,19 @@
...
+ */
+function throwOnConsoleError() {
+  console.error = function (...args: unknown[]) {
+    const [firstArg, ...restArgs] = args;
+
+    if (firstArg instanceof Error) {
+      throw firstArg;
+    }
+
+    throw new Error(format(firstArg, ...restArgs));
```
ahuth, line 15: Nice thing about this is it would've prevented the nested `<p>` tag console log we had for the longest time.

packages/components/src/Checkbox/Checkbox.spec.tsx
```diff
@@ -0,0 +1,21 @@
...
+import Checkbox from "./Checkbox";
+import * as CheckboxStoryFile from "./Checkbox.stories";
+
+describe("<Checkbox />", () => {
+  generateSnapshots(CheckboxStoryFile);
+
+  test("throws an error if no label or aria-label", () => {
+    expect(() => {
+      render(<Checkbox />);
+    }).toThrow(/must provide a visible label or aria-label/);
```
ahuth, line 13: Is this still annoyingly printing this error message when the test runs? Could spyOn `console.error` if it is.

packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,211 @@
...
+  label?: React.ReactNode;
+  /**
+   * Size of the checkbox label.
+   */
+  size?: LabelProps["size"];
+};
+
+interface SvgStyle extends React.CSSProperties {
+  "--checkbox-svg-size": string;
+}
```
ahuth, line 69: This reminds me, I wonder if we should export a type definition with all css variables we use. That way LP can consumers can supply values for these in style attributes with max autocompletion and type safety.

ahuth (approved):

packages/components/src/Checkbox/Checkbox.spec.tsx
anniehu4: nope! throwing the error fixed it 👍

packages/components/src/Checkbox/Checkbox.tsx
anniehu4: ooh I like it. tracked: https://app.shortcut.com/czi-edu/story/180821/export-type-definition-from-eds-with-all-css-variables-used

PR #759, opened 2021-12-19
chore(close-button): add forwardRef, test file

dierat:
### Summary:
Part of https://app.shortcut.com/czi-edu/story/176519/modal-follow-ups

While working on the `Modal` component in LP, I've been thinking about how multi-step modals will need to manually set focus on the first interactive element in the modal when the user moves to a different step. We set focus using refs, but the `CloseButton` doesn't actually support forwarding refs atm, so this PR adds that functionality.

I wanted to add a test to verify that the ref was properly being forwarded, and I realized I never added a spec file for this component (???), so I added that while I was in here.

Next up: I'm going to move the `CloseButton` out of `/common` since we now expect it will be used in LP outside of EDS components.

### Test Plan:
Verify automated tests pass.

### DIFF
--- a/packages/components/src/common/CloseButton/CloseButton.tsx
+++ b/packages/components/src/common/CloseButton/CloseButton.tsx
@@ -1,5 +1,5 @@
 import clsx from "clsx";
-import React from "react";
+import React, { forwardRef } from "react";
 import Button from "../../Button";
 import CloseRoundedIcon from "../../Icons/CloseRounded";
 import styles from "./CloseButton.module.css";
@@ -10,7 +10,7 @@ type CloseButtonProps = {
   /**
    * The color theme of the icon. Also affects the hover state.
    */
-  color: Variant;
+  color?: Variant;
   /**
    * Size of the icon. Does not affect actual button size. The button is larger than the
    * icon to ensure the hit box is large enough (for accessibility).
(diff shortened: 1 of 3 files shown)


packages/components/src/common/CloseButton/CloseButton.tsx
```diff
@@ -10,7 +10,7 @@ type CloseButtonProps = {
   /**
    * The color theme of the icon. Also affects the hover state.
    */
-  color: Variant;
+  color?: Variant;
```
dierat, line 13: I noticed that we have a default `color` prop value, but the prop is still required. Made it optional.

packages/components/src/common/CloseButton/CloseButton.tsx
```diff
@@ -31,32 +31,39 @@ type CloseButtonProps = {
...
+      onClick={onClose}
+      variant="link"
+      ref={ref}
+      {...rest}
+    >
+      <CloseRoundedIcon purpose="informative" size={size} title={ariaLabel} />
+    </Button>
+  ),
 );
+CloseButton.displayName = "CloseButton";
```
dierat, line 67: We have to manually add a `displayName` when adding `forwardRef`. 🤷

ahuth: Is that a lint rule making us do this? 🤔 

An alternative may be to define the forward ref like:

```js
const CloseButton = forwardRef<HTMLButtonElement, Props>(function CloseButton(props, ref) {
  // ...
});
```

Basically give the function a name that React can pull directly from.

Doesn't really matter either way. Both require duplicating the CloseButton name. Thought just occured to me (and https://reactjs.org/docs/forwarding-refs.html#displaying-a-custom-name-in-devtools indicates it would work)

ahuth (approved):
LGTM

PR #760, opened 2021-12-19
chore(close-button): move out of /common

dierat:
### Summary:
We now expect that the `CloseButton` will be used outside of EDS components, so I'm moving it out of the `/common` directory, which is intended to contain reusable internal code.

The reason I expect this to be used outside of EDS components is because multi-step components will need to manually add their own `CloseButton` so they can set a `ref` on it that they can then set focus on. An alternate method could be to add a `closeButtonRef` prop to the `Modal`. We didn't discuss this option, but I think that would also get the job done?

In any case, the `CloseButton` is now being used more than we initially anticipated, so I think moving it into the main section makes sense.

### Test Plan:
Verify the `CloseButton` has moved in storybook but otherwise behaves the same.


ahuth (approved):
LGTM

anniehu4 (commented):
If we want to be able to import with `import {CloseButton} from '@chanzuckerberg/eds-components'`, we should also re-export from `src/index.ts`

PR #768, opened 2021-12-23
Remove build matrix for Node versions

ahuth:
### Summary:

Remove build matrix for Node versions.

Testing this repo on different versions of Node isn't meaningful. The Node version isn't the thing that will make these tests fail or pass. More relevant would be something like operating system.

This PR simplifies things a bit, and has the side benefit of meaning things like #767 don't have to update the branch protection rules to get required checks working right.

If/when approved, I'll update the branch protection rules and the 14.x checks will no longer show up as required.

### Test Plan:

- CI

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,17 +5,12 @@ on: push
 jobs:
   test:
     runs-on: ubuntu-latest
-
-    strategy:
-      matrix:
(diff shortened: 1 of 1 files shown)


ahuth:
Merging so I don't have to change the branch protection rules around anymore. Let me know though if you see any issues with this.

PR #800, opened 2022-01-21
chore: extend props for Text and Heading

jinlee93:
[sc-66248]

### Summary:
- extends Heading props to have html heading attributes
- extends Text props to have global html attributes

### Test Plan:
- no typescript complaints when html attributes exist

Previously:
previousTextTyping
previousHeading

Now no errors:
textTyping
typedHeading

### DIFF
--- a/packages/components/src/Text/Text.tsx
+++ b/packages/components/src/Text/Text.tsx
@@ -16,7 +16,7 @@ type Props = {
   spacing?: TypographyProps<TextElement>["spacing"];
   tabIndex?: number;
   weight?: TypographyProps<TextElement>["weight"];
-};
+} & React.HTMLAttributes<HTMLElement>;
 
 const Text = forwardRef<HTMLElement, Props>(
   (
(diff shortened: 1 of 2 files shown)


packages/components/src/Text/Text.tsx
```diff
@@ -16,7 +16,7 @@ type Props = {
   spacing?: TypographyProps<TextElement>["spacing"];
   tabIndex?: number;
   weight?: TypographyProps<TextElement>["weight"];
-};
+} & React.HTMLAttributes<HTMLElement>;
```
ahuth, line 19: Is this mostly so it has access to things like `id`, `hidden`, `style`, etc?

ahuth (approved):

packages/components/src/Text/Text.tsx
jinlee93: Yea, was trying to use this as a target for `aria-labelledby`

PR #801, opened 2022-01-21
feat: migrate Tooltip component

jinlee93:
### Summary:

[sc-169594]

Copies over Tooltip from traject EDSCandidates.
Converts to Typescript and CSF storybook format.

### Test Plan:
- Stories match traject stories
- Publish alpha release [0.5.1-alpha.2](https://www.npmjs.com/package/@chanzuckerberg/eds-components/v/0.5.1-alpha.2) for Traject testing
  - left window local storybook with alpha release, right window main

https://user-images.githubusercontent.com/86632227/150617368-6d5d11a4-a60c-4225-bfee-210d1376188c.mov

https://user-images.githubusercontent.com/86632227/150617373-1eaa358b-d70e-4876-9879-1744bf8f515b.mov

### DIFF
--- a/packages/components/src/Tooltip/Tooltip.module.css
+++ b/packages/components/src/Tooltip/Tooltip.module.css
@@ -1,0 +1,88 @@
+/* stylelint-disable selector-pseudo-class-no-unknown */
+
+.tooltip {
+  @apply border border-solid shadow-2 rounded-lg;
+  @apply motion-reduce:transition-none;
+}
+
+/* Enables opacity fade animation */
+.tooltip[data-state='hidden'] {
+  @apply opacity-0;
+}
+
+.tooltip :global(.tippy-content) {
+  @apply px-4 py-2;
+}
+
+/* Color Variants */
+.variantLight {
+  @apply border-neutral-300 text-neutral-700 bg-neutral-100;
+  --arrow-color: var(--eds-color-neutral-300);
+}
+
+.variantDark {
+  @apply border-neutral-700 text-white bg-neutral-700;
+  --arrow-color: var(--eds-color-neutral-700);
+}
+
+/* Add Arrows */
+.tooltip :global(.tippy-arrow) {
+  @apply absolute w-4 h-4;
+}
+
+.tooltip :global(.tippy-arrow::before) {
+  @apply absolute border-solid border-transparent border-8;
+
+  content: '';
+}
+
+.tooltip[data-placement^='top'] :global(.tippy-arrow) {
+  @apply left-0;
+
+  transform: translate3d(73px, 0, 0);
+}
+
+.tooltip[data-placement^='bottom'] :global(.tippy-arrow) {
+  @apply top-0 left-0;
+
+  transform: translate3d(73px, 0, 0);
+}
+
+.tooltip[data-placement^='left'] :global(.tippy-arrow) {
+  @apply top-0 right-0;
+
+  transform: translate3d(0, 19px, 0);
+}
+
+.tooltip[data-placement^='right'] :global(.tippy-arrow) {
+  @apply top-0 left-0;
+
+  transform: translate3d(0, 25px, 0);
+}
+
+.tooltip[data-placement^='top'] :global(.tippy-arrow::before) {
+  @apply left-0;
+
+  border-top-color: var(--arrow-color);
+  border-bottom-width: 0;
+}
+
+.tooltip[data-placement^='bottom'] :global(.tippy-arrow::before) {
+  @apply left-0;
+
+  border-bottom-color: var(--arrow-color);
+  border-top-width: 0;
+  top: -7px;
+}
+
+.tooltip[data-placement^='left'] :global(.tippy-arrow::before) {
+  border-left-color: var(--arrow-color);
+  border-right-width: 0;
+  right: -7px;
+}
+
+.tooltip[data-placement^='right'] :global(.tippy-arrow::before) {
+  border-right-color: var(--arrow-color);
+  border-left-width: 0;
+  left: -7px;
+}
--- a/packages/components/src/Tooltip/Tooltip.stories.tsx
+++ b/packages/components/src/Tooltip/Tooltip.stories.tsx
@@ -1,5 +1,3 @@
-// @flow
-
 import type { StoryObj } from "@storybook/react";
 import * as React from "react";
 import Button from "../Button";
--- a/packages/components/src/Tooltip/Tooltip.stories.tsx
+++ b/packages/components/src/Tooltip/Tooltip.stories.tsx
@@ -1,0 +1,111 @@
+// @flow
+
+import type { StoryObj } from "@storybook/react";
+import * as React from "react";
+import Button from "../Button";
+import Tooltip from "./Tooltip";
+
+const defaultArgs = {
+  content: (
+    <span>
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit.{" "}
+      <b>Donec a erat eu augue consequat eleifend non vel sem.</b> Praesent
+      efficitur mauris ac leo semper accumsan.
+    </span>
+  ),
+  children: <Button>Tooltip trigger</Button>,
+  placement: "right",
+  visible: true,
+};
+
+export default {
+  title: "Tooltip",
+  component: Tooltip,
+  args: defaultArgs,
+  argTypes: {
+    variant: {
+      control: {
+        type: "select",
+        options: ["light", "dark"],
+      },
+    },
+    placement: {
+      control: {
+        type: "radio",
+        options: ["left", "right", "top", "bottom"],
+      },
+    },
+  },
+};
+
+type Args = React.ComponentProps<typeof Tooltip>;
+
+export const LightVariant: StoryObj<Args> = {};
+
+export const DarkVariant: StoryObj<Args> = {
+  ...LightVariant,
+  args: {
+    variant: "dark",
+  },
+};
+
+export const LeftPlacement: StoryObj<Args> = {
+  ...LightVariant,
+  args: {
+    placement: "left",
+    children: <Button style={{ marginLeft: "23rem" }}>Tooltip trigger</Button>,
+  },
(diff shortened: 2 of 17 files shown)


packages/components/src/Tooltip/Tooltip.module.css
```diff
@@ -0,0 +1,88 @@
+/* stylelint-disable selector-pseudo-class-no-unknown */
```
jinlee93, line 1: Better way via stylelintrc? Or would this be the only file with `:global` use?

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,86 @@
...
+   */
+  children?: React.ReactElement;
+  /**
+   * The trigger element the tooltip appears next to.
+   *
+   * Use this instead of `children` if the trigger element is being
+   * stored in state or a ref. Most cases will use `children` and not
+   * `reference`.
+   */
+  reference?: React.RefObject<Element>;
```
jinlee93, line 21: This was `React.Node` in traject, but `React.ReactNode` not compatible with TippyProp's `reference`: (`React.RefObject<Element> | Element | null`)

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,84 @@
...
+   */
+  children?: React.ReactElement;
+  /**
+   * The trigger element the tooltip appears next to.
+   *
+   * Use this instead of `children` if the trigger element is being
+   * stored in state or a ref. Most cases will use `children` and not
+   * `reference`.
+   */
+  reference?: React.RefObject<Element> | Element;
```
jinlee93, line 19: This was `React.Node` in traject, but `React.ReactNode` not compatible with `TippyProp`'s reference: `React.RefObject<Element> | Element | null`.

jinlee93:
All storybook styling changes were made to avoid issues Chromatic snaps cut off the tooltip, e.g.:
Screen Shot 2022-01-21 at 6 23 33 PM
Screen Shot 2022-01-21 at 6 23 13 PM
 
This happens because Chromatic snaps element `<div id="root">` and Tippy mounts the tooltip in `document.body`, a sibling element

packages/components/src/Tooltip/__snapshots__/Tooltip.spec.ts.snap
```diff
@@ -0,0 +1,378 @@
...
+  <div>
+    <button
+      class="mx-32 my-32 button sizeLarge variantFlat colorBrand"
+      type="button"
+    >
+      Tooltip trigger
+    </button>
+  </div>
+</body>
+`;
```
dierat, line 118: Oops, I just realized that the `Interactive` story isn't producing a useful snapshot because it's originally closed. I added it for manual interactive testing, so I feel like we should just tell this story to not make a snapshot. But I can also see the argument that we could use this story to verify that the tooltip does appear on hover or focus by adding a test that makes it appear on hover or focus and getting the snapshot. Thoughts?

packages/components/src/Tooltip/Tooltip.module.css
dierat: I think this is fine. 🤷 

> Or would this be the only file with :global use?

My guess is that we'll eventually end up doing it again as we start building on top of other components that are not from Headless UI and require us to dig in and target a specific element that's not exposed to us. But I don't have any specific examples; it's just a feeling.

jinlee93: Sounds good, I think we can add to stylelint when we have to get across that bridge

packages/components/src/Tooltip/__snapshots__/Tooltip.spec.ts.snap
jinlee93: Hmmm, I'll mess around with the `play` function to see if I can get a meaningful snapshot while keeping it interactive, but if too much of a struggle, I can "unsnap" this

packages/components/src/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+        options: ["light", "dark"],
+      },
+    },
+    placement: {
+      control: {
+        type: "radio",
+        options: ["left", "right", "top", "bottom"],
+      },
+    },
+  },
```
dierat, line 36: I think you don't need these in this repo. Try taking them out and see if you still see the controls on the first story on the `docs` tab.

packages/components/src/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+} as Meta<Args>;
+
+type Args = React.ComponentProps<typeof Tooltip>;
+
+export const LightVariant: StoryObj<Args> = {};
+
+export const DarkVariant: StoryObj<Args> = {
+  ...LightVariant,
```
dierat, line 44: What does using the `LightVariant` here do?

packages/components/src/Tooltip/Tooltip.stories.tsx
jinlee93: Ohh you are right, removed

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * The trigger element the tooltip appears next to.
+   */
+  children?: React.ReactElement;
+  /**
+   * The trigger element the tooltip appears next to.
+   *
+   * Use this instead of `children` if the trigger element is being
+   * stored in state or a ref. Most cases will use `children` and not
+   * `reference`.
```
ahuth, line 17: Nice prop doc!

Did the idea/reference/wording for storing an element in `state` come from us or from Tippy?

For my own knowledge, I'm curious as to the intended use case. Ref makes sense, but storing an element in state isn't usually necessary (and according to my own mental pattern matching would still work with `children` anyway).

packages/components/src/Tooltip/Tooltip.stories.tsx
jinlee93: Not much, oops, removed from all the stories

ahuth (approved):
LGTM

packages/components/src/Tooltip/Tooltip.tsx
jinlee93: Think @dierat deserves all the praise for the work, I'm just migrating to EDS haha. Any thought's on Andrew's question?

dierat: I don't remember exactly where "state" came from here; I'm fine with dropping it. This is we're currently using `reference` in the platform (and why we added this documentation): https://github.com/FB-PLP/traject/blob/7ab73ba57ce427d8d9b82ec587d89e6bd8121e8d/app/assets/javascripts/v2/teachers/schoolData/components/MetricCard/HorizontalBar.jsx#L171

packages/components/src/Tooltip/Tooltip.stories.tsx
```diff
@@ -100,13 +81,22 @@ export const LongButtonText: StoryObj<Args> = {
...
+    ),
   },
   parameters: {
     chromatic: {
       disableSnapshot: true,
     },
   },
+  play: async () => {
+    const trigger = await screen.findByRole("button");
+    await userEvent.hover(trigger);
```
dierat, line 100: Nice!

packages/components/src/Tooltip/Tooltip.tsx
ahuth: Makes sense. I say lets drop reference to `state`.

packages/components/src/Tooltip/Tooltip.stories.tsx
```diff
@@ -100,13 +81,22 @@ export const LongButtonText: StoryObj<Args> = {
 };
 
 export const Interactive = {
-  ...LightVariant,
   args: {
     visible: undefined,
+    children: (
+      <Button className="mx-32 my-32">
+        Tooltip trigger with hover done by storybook. <br />
+        Click within canvas to make interactive.
```
dierat, line 89: Is this specific to this repo? I don't have to click into the canvas in `traject` to get the hover to work. (You do need to have focus on that browser window though.)

dierat (approved):

packages/components/src/Tooltip/Tooltip.stories.tsx
jinlee93: The story loads and immediately hovers the button to spawn to the tooltip, so it doesn't look interactive. Have to click  to despawn the tooltip created by the play function.

How should I word this to make such message come across?

dierat: Ah, I see. So you have to click to dismiss the tooltip, and then you can hover over it?

jinlee93: Yup. Hovering and after hover also despawns it, but basically have to interact with it a little

dierat: I think this is fine as-is, but I'll throw out a suggestion for your consideration.

I'm wondering if non-devs will understand the "canvas" part here, so I was thinking you could move the text out of the button and change it to "Click somewhere in this area to dismiss the tooltip, then hover over the button to make it reappear." and then you could change the button text to something like "Hover here to see tooltip after clicking somewhere outside".

I'm probably overthinking this, so don't spend too much time dwelling on it. Thanks for figuring out how to make this story work! ❤️

ahuth: Great point 👍 

I think devs will probably be confused by this as well, since (I believe) "canvas" here refers to the Storybook "canvas", not an actual HTML canvas.

Most folks probably won't be that familiar with that terminology.

jinlee93: Updated story as advised, thanks!

PR #809, opened 2022-01-25
feat: add FormGroup component

dierat:
### Summary:
Clubhouse ticket: https://app.shortcut.com/czi-edu/story/169595/button-group-form-group-move-to-eds
Location in `traject`: https://github.com/FB-PLP/traject/tree/master/app/assets/javascripts/v2/core/EDSCandidates/FormGroup

### Screenshots

FormGroup in storybook

### Test Plan:
- [x] Manually tested the component in storybook
- [x] Verified the snapshot tests output looks right

### DIFF
--- a/packages/components/src/FormGroup/FormGroup.tsx
+++ b/packages/components/src/FormGroup/FormGroup.tsx
@@ -1,0 +1,107 @@
+import clsx from "clsx";
+import React, { ElementType, ReactNode } from "react";
+import Text from "../Text";
+import styles from "./FormGroup.module.css";
+
+type FormGroupProps = {
+  /**
+   * The contents of the form group. We suggest a FormGroup.Title followed by
+   * interactive elements.
+   *
+   * Should be wrapped in a fragment to allow our styling to control the spacing
+   * between elements.
+   */
+  children: ReactNode;
+  /**
+   * Type of element the immediate wrapper around the contents should be.
+   * @default 'fieldset'
+   */
+  as?: ElementType;
+  /**
+   * Additional classnames passed in for styling.
+   */
+  className?: string;
+};
+
+type TitleProps = {
+  /**
+   * The title of the form group.
+   */
+  children: ReactNode;
+  /**
+   * Type of element the wrapper should be.
+   * @default 'legend'
+   */
+  as?: ElementType;
+  /**
+   * Additional classnames passed in for styling.
+   */
+  className?: string;
+};
+
+type ItemsProps = {
+  /**
+   * The content of the helper form elements in the form group.
+   */
+  children: ReactNode;
+  /**
+   * Type of element the immediate wrapper around the contents should be.
+   * @default 'div'
+   */
+  as?: ElementType;
+  /**
+   * Additional classnames passed in for styling.
+   */
+  className?: string;
+};
+
+/**
+ * Helper sub-component for styling the title at the top of the component.
+ */
+const Title = ({
+  children,
+  as: Component = "legend",
+  className,
+}: TitleProps) => {
+  return (
+    <Component className={clsx(className, styles.title)}>
+      <Text size="sm" weight="bold">
+        {children}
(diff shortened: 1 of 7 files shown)


packages/components/src/FormGroup/index.ts
```diff
@@ -0,0 +1,3 @@
+// @flow
```
jeremiah-clothier, line 1: Should this be removed?

```suggestion

```

jeremiah-clothier:
Is it intentional that this is being merged into `diedra/button-group/add` instead of the master/main branch?
image

packages/components/src/FormGroup/FormGroup.tsx
```diff
@@ -0,0 +1,107 @@
...
+  children,
+  as: Component = "legend",
+  className,
+}: TitleProps) => {
+  return (
+    <Component className={clsx(className, styles.title)}>
+      <Text size="sm" weight="bold">
+        {children}
+      </Text>
+    </Component>
```
ahuth, line 71: Does this result in HTML like with an extra P tag:

```html
<legend>
  <p class="stuff">Some text</p>
</legend>
```

?

If so, I wonder if it's worth thinking about reducing unnecessary elements. Maybe with something like

```jsx
<Text as={Component} className={clsx(className, styles.title)} size="sm" weight="bold">
  {children}
</Text>
```

Although that may not really be worth it. And would probably require updating the Text component to accept more things for `as`.

packages/components/src/FormGroup/FormGroup.tsx
```diff
@@ -0,0 +1,107 @@
...
+  );
+};
+
+/**
+ * Helper sub-component for styling the form elements in the component.
+ */
+const Items = ({ children, as: Component = "div", className }: ItemsProps) => {
```
ahuth, line 78: I'm surprised to see the `Component` annotation here, since `ItemProps` defines `as` as `ElementType`.

Do we need that because TypeScript isn't happy with trying to do `<Component>` with intrinsic JSX elements or something?

jeremiah-clothier: In this instance I believe `as: Component` is just renaming the variable and not providing a type annotation. Then you can use `<Component .../>` instead of `<as ... />`

ahuth: Yep, you're right. It's not even possible to have a type annotation in the middle of object destructuring like that.

dierat:
> Is it intentional that this is being merged into `diedra/button-group/add` instead of the master/main branch?

Yeah; I stacked them so it was easier to make an alpha version with both of them on it. I'll merge them into `main` separately.

packages/components/src/FormGroup/FormGroup.tsx
dierat: https://app.shortcut.com/czi-edu/story/182120/allow-text-component-to-use-more-elements-for-as-prop

ahuth (approved):
LGTM

dierat:
Any idea what's causing this?
screenshot of github actions post above stating that this PR is increasing the package size by 1.37% and needs changed

ahuth:
No idea why a size increase of 1.37% would cause the size limit bot to block merging.

Couple options:
1. [Easiest] Dismiss the bot's review
2. Delete the size-limit-action. I don't think we really need it.
3. See if we can tune it's parameters

dierat:
> No idea why a size increase of 1.37% would cause the size limit action to block merging. Maybe it's enforcing a total size limit of the package or something.
> 
> Couple options:
> 
> 1. [Easiest] Dismiss the action's review, and see if subsequent PRs have this same issue
> 2. Delete the size-limit-action. I don't think we really need it.
> 3. See if we can tune it's parameters

Cool, I'll just dismiss it for now. Thanks for verifying! 👍

PR #813, opened 2022-01-26
RFC: Auto-generate story titles

ahuth:
As of Storybook 6.4, Storybook can auto-generate story titles for us.

Is this a good idea for us? It means less work for us, and letting the library take care of things it can handle on its own. Having said that, writing titles isn't exactly the most challenging thing.

For example, if we just leave off the `title` from a story file

```diff
export default {
-  title: 'Button',
  component: Button,
}
```

storybook will infer it from the filesystem.

See the results of this, before (manual titles) on the left and after (generated titles) on the right:
storybook titles

Note that
1. I changed the config so that the typography and tokens docs came first. This isn't related to generating the titles, but felt right.
2. We're now getting an extra level of nesting for each component. Instead of
    ```
    Button
    ┣━ foo
    ┗━ bar
    ```
    this will have
    ```
    📁 Button
    ┗━ Button
       ┣━ foo
       ┗━ bar
    ```
    This happens because the auto-generated titles are getting basically `Directory / ComponentName / all the stories`.

Number 2 could be weird, and let me know if it's too weird.

### DIFF
--- a/packages/components/.storybook/main.js
+++ b/packages/components/.storybook/main.js
@@ -1,5 +1,5 @@
 module.exports = {
-  stories: ["../src", "../docs"],
+  stories: ["../docs", "../src"],
   addons: [
     "@storybook/addon-essentials",
     "@storybook/addon-a11y",
--- a/packages/components/.storybook/manager.js
+++ b/packages/components/.storybook/manager.js
@@ -1,0 +1,11 @@
+import { addons } from "@storybook/addons"; // eslint-disable-line import/no-extraneous-dependencies
+
+addons.setConfig({
+  sidebar: {
+    // Show top-level directories (under components/src) as directories instead of separate
+    // sections. It's a little nicer seeing them as directories instead of their own sections.
+    // We only need this when auto-generating story titles. If we supply our own titles, we should
+    // delete this config.
+    showRoots: false,
+  },
+});
--- a/packages/components/src/Banner/Banner.stories.tsx
+++ b/packages/components/src/Banner/Banner.stories.tsx
@@ -5,7 +5,6 @@ import Heading from "../Heading";
 import Banner from "./Banner";
 
 export default {
-  title: "Banner",
   component: Banner,
   argTypes: {
     elevation: {
--- a/packages/components/src/Button/button.stories.tsx
+++ b/packages/components/src/Button/button.stories.tsx
@@ -13,7 +13,6 @@ import {
 import Button from "./button";
 
 export default {
-  title: "Button",
   component: Button,
   argTypes: {
     children: {
--- a/packages/components/src/ButtonGroup/ButtonGroup.stories.tsx
+++ b/packages/components/src/ButtonGroup/ButtonGroup.stories.tsx
@@ -4,7 +4,6 @@ import Button from "../Button";
 import ButtonGroup from "./ButtonGroup";
 
 export default {
-  title: "ButtonGroup",
   component: ButtonGroup,
   args: {
     spacing: "1x" as const,
--- a/packages/components/src/Checkbox/Checkbox.stories.tsx
+++ b/packages/components/src/Checkbox/Checkbox.stories.tsx
@@ -8,7 +8,6 @@ const defaultArgs = {
 };
 
 export default {
-  title: "Checkbox",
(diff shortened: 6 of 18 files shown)


packages/components/.storybook/main.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  stories: ["../src", "../docs"],
+  stories: ["../docs", "../src"],
```
ahuth, line 2: Just swapped docs and src here, so that the "docs" stories (typograhpy and tokens) show up first.

No reason to do that, but it felt weird having them buried at the bottom.

packages/components/.storybook/manager.js
```diff
@@ -0,0 +1,11 @@
+import { addons } from "@storybook/addons"; // eslint-disable-line import/no-extraneous-dependencies
+
+addons.setConfig({
+  sidebar: {
+    // Show top-level directories (under components/src) as directories instead of separate
+    // sections. It's a little nicer seeing them as directories instead of their own sections.
+    // We only need this when auto-generating story titles. If we supply our own titles, we should
+    // delete this config.
+    showRoots: false,
```
ahuth, line 9: Without this, when using auto-generated titles we get each top-level directory as a separate section in storybook, which looks weird.

Screen Shot 2022-01-26 at 12 44 04 PM

An alternative to this is the same thing mentioned in https://github.com/chanzuckerberg/edu-design-system/pull/813/files#r793032073.

jinlee93:
So within a story folder, can there be multiple titles? e.g. 
📁 Button
┗━ Button
   ┣━ foo
   ┗━ bar
┗━ Button2
   ┣━ foo
   ┗━ bar

ahuth:
> So within a story folder, can there be multiple titles?

Yes, if there were multiple story files in a directory. So like both `button/Primary.stories.ts` and `button/Secondary.stories.ts`.

jinlee93:
> > So within a story folder, can there be multiple titles?
> 
> Yes, if there were multiple story files in a directory. So like both `button/Primary.stories.ts` and `button/Secondary.stories.ts`.

Ah sounds good to me. Then do you think Storybook intends for devs to group their stories in the same folder? Such as `button/Button.stories.ts` and `button/ButtonGroup.stories.ts`?

ahuth:
> do you think Storybook intends for devs to group their stories in the same folder?

Not necessarily, although there's nothing wrong with it. Probably the most common scenario is a bunch of files under `src/` without any more subdirectories

```
src
  button.tsx
  button.stories.tsx
  checkbox.tsx
  checkbox.stories.tsx
```

packages/components/.storybook/main.js
jeremiah-clothier: I like having the tokens first, having the components nested under a section named `src` feels a bit awkward. That's just with `stories: ["../"],` right?

jeremiah-clothier:
> 2. We're now getting an extra level of nesting for each component. Instead of
> Button
> ┣━ foo
> ┗━ bar
> this will have
> 📁 Button
> ┗━ Button
>    ┣━ foo
>    ┗━ bar
> 
> This happens because the auto-generated titles are getting basically Directory / ComponentName / all the stories.
> Number 2 could be weird, and let me know if it's too weird.

I'm fine with Number 2. Only slight unfortunate thing I notice when looking at the storybook is the additional level of indentation for people all small screens. They might not be able to read the whole story title without making the left column tree view wider of the Storybook UI wider

packages/components/.storybook/main.js
ahuth: > That's just with stories: ["../"], right?

That's right

ahuth: Could alternatively make this change (putting docs/ first) without moving to auto-generated titles.

jinlee93: I think if the choice is between auto-generated titles and nested stories, I'm more down for auto-generated titles

jeremiah-clothier (approved):
Approved. I'm okay with auto generated titles

jinlee93 (approved):

ahuth:
Closing. Not sure auto-generating titles is super intuitive or even useful.

PR #814, opened 2022-01-27
docs(read-me): change yarn generateFlowtypes to yarn generateFlowTypes

dierat:
### Summary:
Fixing a typo?

### Test Plan:
- Ran `yarn generateFlowTypes DropdownButton`; verified that worked.
- Tried `yarn generateFlowtypes DropdownButton`; verified that did not work.
- Previewed the read me file; verified it said `yarn generateFlowTypes <Component>`.

### DIFF
--- a/packages/components/README.md
+++ b/packages/components/README.md
@@ -59,7 +59,7 @@ and then use them in your React components
 For any component, run
 
 ```bash
-yarn generateFlowtypes <Component>
+yarn generateFlowTypes <Component>
 ```
 
 to get automatically generated flowtypes.


dierat:
🤔 
screenshot of github-actions post above about size limit increase by 0 and requesting changes to the PR based on that

ahuth (approved):
LGTM 🤘 

I bet we're over some kind of size limit for the package. Although it's actually kind of challenging looking at https://github.com/chanzuckerberg/size-limit-action and trying to figure out what the limit would actually be...

jinlee93 (approved):
Thanks for catching this

ahuth:
Oh, actually the size limit is defined at https://github.com/chanzuckerberg/edu-design-system/blob/50565f9307bdb68b66ef58ee115bc7063c852022/package.json#L32-L45

Should be 70kb, which we're real close to?

dierat:
> EDIT: wait, now I'm confused. That config looks like it would apply to single files, not the entire package? But then again, 70kb would be reaallllllly big for 1 file 😬

It's gotta be checking the whole package size based on the comment on this PR 🤔

PR #817, opened 2022-01-28
chore(form-group): add display names to components

dierat:
### Summary:
The `FormGroup` component has 2 subcomponents: `Form.Title` and `Form.Items`. But in the `code` feature of the docs tab in storybook, these components are shown as just `Title` and `Items`, which makes it little confusing for devs trying to import and use the component. The solution is to add display names for all three (`FormGroup`, `FormGroup.Title`, and `FormGroup.Items`).

Similar PR for the `Banner` component: https://github.com/chanzuckerberg/edu-design-system/pull/734

#### Before
FormGroup component in storybook, before this change

#### After
FormGroup component in storybook, after this change

### Test Plan:
Manually tested in storybook.

### DIFF
--- a/packages/components/src/FormGroup/FormGroup.tsx
+++ b/packages/components/src/FormGroup/FormGroup.tsx
@@ -58,7 +58,7 @@ type ItemsProps = {
 /**
  * Helper sub-component for styling the title at the top of the component.
  */
-const Title = ({
+const Title: React.FC<TitleProps> = ({
   children,
   as: Component = "legend",
   className,
@@ -75,7 +75,11 @@ const Title = ({
 /**
  * Helper sub-component for styling the form elements in the component.
  */
-const Items = ({ children, as: Component = "div", className }: ItemsProps) => {
+const Items: React.FC<ItemsProps> = ({
+  children,
+  as: Component = "div",
+  className,
+}: ItemsProps) => {
   return (
     <Component className={clsx(className, styles.items)}>{children}</Component>
   );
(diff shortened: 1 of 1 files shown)


packages/components/src/FormGroup/FormGroup.tsx
```diff
@@ -58,7 +58,7 @@ type ItemsProps = {
 /**
  * Helper sub-component for styling the title at the top of the component.
  */
-const Title = ({
+const Title: React.FC<TitleProps> = ({
```
dierat, line 61: This `React.FC<TitleProps>` stuff is required to convince typescript that this is actually a component with a `displayName` property for us to override. Discussion on how to get around this issue in the `Banner` component: https://github.com/chanzuckerberg/edu-design-system/pull/734/files#r759746305

packages/components/src/FormGroup/FormGroup.tsx
```diff
@@ -104,4 +108,8 @@ function FormGroup({
 FormGroup.Title = Title;
 FormGroup.Items = Items;
 
+FormGroup.displayName = "FormGroup";
```
dierat, line 111: We have to set the `displayName` for the `FormGroup` itself in order for the docstring to continue to appear in storybook.

packages/components/src/FormGroup/FormGroup.tsx
ahuth: When we pass `TitleProps` as a type param, can we remove it from the annotation from the props?

```diff
  className,
-}: TitleProps) => {
+}) => {
```

dierat: Looks like we can! Update, thanks!

PR #816, opened 2022-01-27
Focus indicator prototypes

ahuth:
[sc-181369]

Prototype focus indicators.

focus prototypes

PR #818, opened 2022-01-28
Double size limit to 140kb

ahuth:
### Summary:

We're now over the 70kb size limit from the size-limit-action. This PR doubles it to 140kb, since it's expected that the package will keep getting bigger.

I'm not sure we should even use this action. But maybe it's nice to get the report on how much size increases per PR.

### Test Plan:

- CI

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,13 +32,13 @@
   "size-limit": [
     {
       "path": "packages/components/lib/**/*([a-zA-Z-_]).js",
-      "limit": "70KB",
+      "limit": "140KB",
       "name": "components",
       "webpack": false
     },
     {
       "path": "packages/components/lib/**/*([a-zA-Z-_]).css",
-      "limit": "70KB",
+      "limit": "140KB",
       "name": "styles",
       "webpack": false
     }


package.json
```diff
@@ -32,13 +32,13 @@
...
       "path": "packages/components/lib/**/*([a-zA-Z-_]).js",
-      "limit": "70KB",
+      "limit": "140KB",
       "name": "components",
       "webpack": false
     },
     {
       "path": "packages/components/lib/**/*([a-zA-Z-_]).css",
-      "limit": "70KB",
+      "limit": "140KB",
```
ahuth, line 41: I'm not sure this config is right.

It seems to indicate that JS and CSS should each be able to be 70kb in size. But the CI failures (for example, see #816) seem like it's the JS + CSS that's putting us over 70kb.

package.json
```diff
@@ -32,13 +32,13 @@
   "size-limit": [
     {
       "path": "packages/components/lib/**/*([a-zA-Z-_]).js",
-      "limit": "70KB",
+      "limit": "140KB",
```
ahuth, line 35: Unrelated to this change, but why are we using a forked version of size-limit-action? 🤔 

Something to think about for the future is seeing if we can get the non-forked version working for us.

ahuth: Ah, looks like Devin added some code to not post more than 1 comment from the action: https://github.com/chanzuckerberg/size-limit-action/commit/afba6038fb5322674ca9669c5482a47aac7a31c0

lesliecodes (approved):

ahuth:
Thanks!

PR #830, opened 2022-02-01
Updates

ahuth:
### Summary:

Updates
- Many dev deps
- Tailwind to the latest v2 (not v3)

### Test Plan:

- CI
- Local testing
- Chromatic


packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -146,6 +146,8 @@ export const CheckboxInput = React.forwardRef<HTMLInputElement, InputProps>(
   },
 );
 
+CheckboxInput.displayName = "CheckboxInput";
```
ahuth, line 149: There's now a React lint rule enforcing providing a `displayName` where it doesn't think one will be inferred (such as when using `forwardRef`).

packages/components/package.json
```diff
@@ -56,35 +56,35 @@
...
     "postcss-import": "^14.0.2",
     "postcss-nested": "^5.0.6",
     "react": "^16.14.0",
     "react-docgen-typescript-loader": "^3.7.2",
     "react-dom": "^16.14.0",
     "storybook-css-modules-preset": "^1.1.1",
-    "tailwindcss": "^2.2.7",
-    "ts-jest": "^27.0.7",
-    "typescript": "^4.5.4"
+    "tailwindcss": "^2.2.19",
```
anniehu4, line 86: did you see issues with moving to v3? or just safer to be in a separate PR? we have this ticket but I haven't looked into the changelog much: https://app.shortcut.com/czi-edu/story/176262/upgrade-to-tailwind-3-0

anniehu4 (approved):

packages/components/package.json
ahuth: I didn't look at moving to v3. Wasn't sure if there were reasons to stay on the same major version as LP or not.

But I can't think of a reason why it would be problematic (since we ship compiled styles, not Tailwind classnames).

ahuth:
Thanks!

PR #832, opened 2022-02-02
chore(dropdown-button): wrap children in span

dierat:
### Summary:
Part of https://app.shortcut.com/czi-edu/story/169970/dropdown-api-improvements

Wrapping the `DropdownButton` children in a `<span>` in case label contains more than just text

### Test Plan:
Verify the `<span>` is showing up in the snapshot file for the `DropdownButton` component.

PR #833, opened 2022-02-02
chore: remove --only-changed from chromatic github action

dierat:
### Summary:
After making a very small change in a component in https://github.com/chanzuckerberg/edu-design-system/pull/832, I hit a chromatic build issue involving our chromatic version and Turbosnap. To unblock development, we can downgrade chromatic or remove the `--only-changed` flag from the chromatic script. We are opting to do the latter.

### Test plan:
Make a change in some component (like adding a fragment or wrapping span or something),
then run `yarn workspace @chanzuckerberg/eds-components chromatic --project-token=${{ secrets.CHROMATIC_PROJECT_TOKEN }} --output-dir storybook-static --exit-zero-on-changes --skip 'dependabot/**'` (replacing `${{ secrets.CHROMATIC_PROJECT_TOKEN }}` with the actual token, which can be found in chromatic),
and verify you don't get an error.

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -30,7 +30,7 @@
     "build:styles": "postcss \"src/**/*.css\" --dir lib/ --base src/ --verbose",
     "build:watch": "yarn run build:js --watch",
     "build:storybook": "build-storybook -o storybook-static",
-    "chromatic": "chromatic --only-changed",
+    "chromatic": "chromatic",
     "deploy:docs": "storybook-to-ghpages --ci",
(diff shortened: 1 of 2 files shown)


jinlee93:
I'm down for removing `--only-changed` via this PR to mitigate this issue

jinlee93 (approved):
thanks for finding and addressing

anniehu4 (approved):

ahuth (approved):

PR #835, opened 2022-02-03
feat: add Dropdown component

dierat:
### Summary:
Part of https://app.shortcut.com/czi-edu/story/169970/dropdown-api-improvements

`Dropdown` component in `traject`: https://github.com/FB-PLP/traject/blob/master/app/assets/javascripts/v2/core/EDSCandidates/Dropdown/Dropdown.tsx

### Test Plan:
Test in storybook, verify the snapshots look good, tested in `traject` with an alpha build.


packages/components/src/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,199 @@
...
+              )}
+              <span className={styles.dropdownOptionContent}>{children}</span>
+            </li>
+          )}
+    </Listbox.Option>
+  );
+};
+
+Dropdown.Button = DropdownTrigger;
+Dropdown.Label = Listbox.Label;
```
dierat, line 195: I'm getting an error for this line when trying to rebuild the package:

>src/Dropdown/Dropdown.tsx:195:1 - error TS4032: Property 'Label' of exported interface has or is using name 'LabelRenderPropArg' from private module '"/Users/drater/edu-design-system/node_modules/@headlessui/react/dist/components/listbox/listbox"'.
>
>195 Dropdown.Label = Listbox.Label;

error output in terminal

jeremiah-clothier: That's a weird TS error I haven't seen before

Tagging @andremalan in case he's ran into something like this when converting files

ahuth: This may be basically the same as https://github.com/microsoft/TypeScript/issues/6307, since we are using the `--declaration` option (in this project's tsconfig.json).

Presumably LP doesn't have that problem because it doesn't output any declaration files.

Fix might be to re-export `LabelRenderPropArg` from Headless UI? Although that does seem unfortunate.

dierat: I don't think we can import `LabelRenderPropArg` from Headless UI to begin with because it's not being exported 🤔 https://github.com/tailwindlabs/headlessui/blob/fbaa1ae9da8788910ce8b240851f9779630babcb/packages/%40headlessui-react/src/components/listbox/listbox.tsx#L417

dierat: Looks like one hack around this is to add a `DropdownLabel` component that just wraps the `Listbox.Label` and sets the prop types to `any`:
```
const DropdownLabel = (props: any) => <Listbox.Label {...props} />;

Dropdown.Button = DropdownTrigger;
Dropdown.Label = DropdownLabel;
Dropdown.Option = DropdownOption;
Dropdown.Options = DropdownOptions;

export default Dropdown;
```

jeremiah-clothier: > Looks like one hack around this is to add a `DropdownLabel` component that just wraps the `Listbox.Label` and sets the prop types to `any`

That could work if there's a way to replace the `any` type with types that more closely align with `Listbox.Label`. Does this give us better typing for Props?

`type Props = React.ComponentProps<typeof Listbox.Label>;`

If that works, then you could check that the [render props of Listbox.Label](https://headlessui.dev/react/listbox#listbox-label) are inferred correctly as boolean values. Or you can try checking another prop just to see if TS throws an error when you try passing an invalid value

packages/components/src/Dropdown/index.ts
```diff
@@ -0,0 +1,3 @@
+// @flow
```
jeremiah-clothier, line 1:
```suggestion

```

packages/components/src/Dropdown/Dropdown.tsx
dierat: Using `React.ComponentProps<typeof Listbox.Label>;` unfortunately brings the issue back. 😔

Code block:
Screen Shot 2022-02-04 at 9 15 33 AM

Code block with type error popover visible:
Screen Shot 2022-02-04 at 9 15 43 AM

packages/components/src/Dropdown/index.ts
dierat: Removed, thanks!

packages/components/src/Dropdown/Dropdown.tsx
anniehu4: I don't really expect us to use the render props for this Label element 🤔  from the [Headless documentation](https://headlessui.dev/react/listbox#listbox-label), the only prop it expects is `as`

Could we type as
```
const DropdownLabel = (props: {
  as?: React.ElementType;
  className?: string;
}) => <Listbox.Label {...props} />;
```
?

dierat: > I don't really expect us to use the render props for this Label element 🤔 from the [Headless documentation](https://headlessui.dev/react/listbox#listbox-label), the only prop it expects is `as`
> 
> Could we type as
> 
> ```
> const DropdownLabel = (props: {
>   as?: React.ElementType;
>   className?: string;
> }) => <Listbox.Label {...props} />;
> ```

That change produces a type error in the `.stories` file usage of the `Dropdown.Label` for some reason:

>➤ YN0000: src/Dropdown/Dropdown.stories.tsx(37,8): error TS2559: Type '{ children: string; }' has no properties in common with type 'IntrinsicAttributes & { as?: ElementType<any> | undefined; className?: string | undefined; }'.
>➤ YN0000: src/Dropdown/Dropdown.stories.tsx(62,8): error TS2559: Type '{ children: string; }' has no properties in common with type 'IntrinsicAttributes & { as?: ElementType<any> | undefined; className?: string | undefined; }'.

type error in visual studio code editor

jeremiah-clothier: @dierat What if you add a type for `children`?
```
const DropdownLabel = (props: {
  as?: React.ElementType;
  className?: string;
  children?: React.ReactNode
}) => <Listbox.Label {...props} />;
```

anniehu4: +1 (and good catch), we could also make that prop required

dierat: Ah of course, thanks! I think this solves the issue ❤️ ❤️ ❤️

jeremiah-clothier (approved):

anniehu4 (approved):

PR #837, opened 2022-02-04
Exit Chromatic once the build is uploaded

ahuth:
Exit Chromatic's build step once the Storybook build is uploaded.

Previously this would wait until the build was uploaded AND Chromatic finished processing the build. But there's no reason for the job to wait for all the processing, since the UI Tests step won't be green until that processing is complete (and any diffs are approved).

Making this change should make the build a little more resilient. Long-running Chromatic processing times can still be an issue, but at least they won't cause the upload job to time out.


lesliecodes (approved):
makes sense to me

anniehu4 (approved):

ahuth:
Thanks!

PR #847, opened 2022-02-10
feat: remove unused legacy color tokens

anniehu4:
### Summary:
bit of cleanup! these tokens are all fully unused in traject now.

we should migrate the remaining ones being used at some point, but it's not too pressing since
- we have well defined mappings for them
- most are on the marketing pages, which I'm less comfortable changing

### Test Plan:
[smaller tokens section](https://61313967cde49b003ae2a860-fllxdlthwa.chromatic.com/?path=/story/tokens--page#legacy-claro-colors)


lesliecodes (approved):
yay cleanup

dierat (approved):
haha nice

text showing 9 lines were added to this PR and 233 lines were removed

PR #859, opened 2022-02-16
Escape button closes tooltip

jinlee93:
[sc-181028]

### Summary:
- Escape button closes tooltip
- Implemented as demo'd in docs: https://atomiks.github.io/tippyjs/v6/plugins/#hideonesc
### Test Plan:
- Unit test written
- Manual testing on storybook


packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -56,6 +56,9 @@ type TooltipProps = {
   delay?: number | [number | null, number | null];
 };
 
+type Plugins = NonNullable<React.ComponentProps<typeof Tippy>["plugins"]>;
+type Plugin = Plugins[0];
```
jinlee93, line 60: `@tippyjs/react` does not expose `tippy.js` types, have to extract via props and grab element type from array type

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -68,6 +71,25 @@ export default function Tooltip({
   className,
   ...rest
 }: TooltipProps) {
+  const hideOnEsc: Plugin = {
+    name: "hideOnEsc",
+    defaultValue: true,
+    fn: ({ hide }) => {
+      function onKeyDown(event: KeyboardEvent) {
+        if (event.code === "Escape") {
```
jinlee93, line 79: `event.keyCode` is deprecated, went with `event.code`

packages/components/src/Tooltip/Tooltip.spec.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+describe("<Tooltip /> snaps", () => {
+  generateSnapshots(TooltipStoryFile, {
+    getElement: (wrapper) => {
+      return wrapper.baseElement;
+    },
+  });
+});
+describe("<Tooltip />", () => {
+  it("should close tooltip via escape key", async () => {
```
anniehu4, line 18: You can group these in the same `describe` block:

```suggestion
describe("<Tooltip />", () => {
  generateSnapshots(TooltipStoryFile, {
    getElement: (wrapper) => {
      return wrapper.baseElement;
    },
  });
 
  it("should close tooltip via escape key", async () => {
```

see the `button` spec file for an example

packages/components/src/Tooltip/Tooltip.tsx
anniehu4: I thiiink `event.key` is [slightly more recommended](https://developers.google.com/web/updates/2016/04/keyboardevent-keys-codes). seeing docs around that "code" is the physical key being pressed while "key" is the meaning of the key (in case someone applied a separate keyboard layout or something)

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -68,6 +71,25 @@ export default function Tooltip({
   className,
   ...rest
 }: TooltipProps) {
+  const hideOnEsc: Plugin = {
```
anniehu4, line 74: nit: let's leave a comment & link to the tippy documentation:
```suggestion
  // Hides tooltip when escape key is pressed, following:
  // https://atomiks.github.io/tippyjs/v6/plugins/#hideonesc
  const hideOnEsc: Plugin = {
```

packages/components/src/Tooltip/Tooltip.tsx
anniehu4: ooh fancy -- would `Plugins[number]` work here? (theoretically should generate a union of the possible element types, and is a bit more general)

also, let's leave your comment as a code comment 🙂

anniehu4 (approved):

packages/components/src/Tooltip/Tooltip.tsx
jinlee93: Will do, and also changed to `number`

packages/components/src/Tooltip/Tooltip.tsx
jinlee93: Changed to `key`, thanks

packages/components/src/Tooltip/Tooltip.spec.tsx
jinlee93: Grouped into same `describe` block, thanks

packages/components/src/Tooltip/Tooltip.tsx
jinlee93: Added comment as recommended, thanks

packages/components/src/Tooltip/Tooltip.spec.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+describe("<Tooltip />", () => {
+  generateSnapshots(TooltipStoryFile, {
+    getElement: (wrapper) => {
+      return wrapper.baseElement;
+    },
+  });
+
+  it("should close tooltip via escape key", async () => {
+    // @ts-expect-error Animation prop exists on Tippy but not exposed on Tooltip,
```
dierat, line 18: Would another solution here be to add `animation` to the props list?

dierat (approved):
Nice!

packages/components/src/Tooltip/Tooltip.spec.tsx
jinlee93: Oh that would be a good option but: 
1. I didn't want to leave the scope of this ticket 
2. Are there animation guidelines that we can default animations for?

dierat: I don't think we have any guidelines yet, but [animation type from Tippy](https://github.com/atomiks/tippyjs/blob/8f7b2df156ae39c4df23c481b77ab59fe7f08c44/src/types.ts#L50) is `string | boolean` and I believe the default is `"fade"`. I think if we wanted to use a different one, [it would need to be imported](https://atomiks.github.io/tippyjs/v6/animations/). I don't expect people to really be playing with different animations, but I could see there being other situations where someone needed to remove it.

jinlee93: Ah reasonable that animation might want to be disabled. Added the `animation` prop

packages/components/src/Tooltip/Tooltip.spec.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { render, screen } from "@testing-library/react";
+import React from "react";
+import * as TooltipStoryFile from "./Tooltip.stories";
+
+const { Interactive } = composeStories(TooltipStoryFile);
+
+describe("<Tooltip />", () => {
+  generateSnapshots(TooltipStoryFile, {
+    getElement: (wrapper) => {
+      return wrapper.baseElement;
```
ahuth, line 13: May want to leave a comment that Tippy renders the tooltip as a child of `<body>`, so that people don't unnecessarily copy and paste this.

jinlee93: Added comment, thanks

PR #864, opened 2022-02-18
chore: add import docstrings to components

jinlee93:
[sc-184355]
### Summary:
- add import path on top of docstrings
- https://github.com/FB-PLP/traject/pull/10141 but for EDS

### Test Plan:
- vscode intellisense picks up import path
- storybook (mostly) has import path


packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -175,6 +175,13 @@ export const Label = ({
 };
 
 /**
+ * ```ts
+ * import {Checkbox} from "@chanzuckerberg/eds-components";
+ * ```
+ * ```ts
+ * import Checkbox, {CheckboxInput, Label} from '@chanzuckerberg/eds-components/lib/Checkbox';
+ * ```
+ *
```
jinlee93, line 184: Interestingly `Checkbox` and `Tooltip` docstrings don't show up in Storybook (even before this PR). Import paths still added for VSCode hover
single code blcok

packages/components/src/Checkbox/Checkbox.tsx
```diff
@@ -175,6 +175,13 @@ export const Label = ({
 };
 
 /**
+ * ```ts
+ * import {Checkbox} from "@chanzuckerberg/eds-components";
+ * ```
+ * ```ts
+ * import Checkbox, {CheckboxInput, Label} from '@chanzuckerberg/eds-components/lib/Checkbox';
+ * ```
```
jinlee93, line 183: Needs to be separated like this or weird styling issues occur:
single code block

anniehu4 (approved):
filed https://app.shortcut.com/czi-edu/story/185968/component-docstring-issues-in-storybook

packages/components/src/CloseButton/CloseButton.tsx
```diff
@@ -29,6 +29,10 @@ type CloseButtonProps = {
 };
 
 /**
+ * ```ts
+ * import {CloseButton} from "@chanzuckerberg/eds-components";
```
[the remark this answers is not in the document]
jinlee93: yup, added

packages/components/src/Checkbox/Checkbox.tsx
dierat: Oh, I think I've hit this issue before. There's a strange bug with regards to `displayName` – try moving the display name lines to the bottom of the file, above the export line, and make sure the main component is last.

Like
```
[checkbox component code]

Checkbox.displayName = "Checkbox";

CheckboxInput.displayName = "CheckboxInput";

export default Checkbox;
```

dierat: At least for `Tooltip`, just removing the display name seems to fix it. I don't remember why we added that in the first place. For the `Checkbox`, we need it because of the `forwardRef`, but I can only get the `CheckboxInput` docstring to show up on the docs page.

dierat: Adding `export` in front of `const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(` and moving the checkbox + display name above the checkbox input seems to fix it for me 🤷 
checkbox in storybook

packages/components/src/Banner/Banner.tsx
```diff
@@ -48,6 +48,10 @@ export type BannerProps = {
 };
 
 /**
+ * ```ts
+ * import {Banner} from "@chanzuckerberg/eds-components";
```
ahuth, line 52: Can/should this be part of the usage example?

```
* Example usage:
*
* ```tsx
* import {Banner} from "@chanzuckerberg/eds-components";
*
* <Banner
*   onDismiss={handleDismiss}
```

That feels more natural to me, since they go together.

jinlee93: I think the intention was to have it at the top of the page to be seen easily. I can also add to example block if you think the block could also use it

packages/components/src/Checkbox/Checkbox.tsx
jinlee93: Did similarly as suggested, and docstrings are back! thanks

PR #873, opened 2022-02-25
chore: disable dependabot version updates

anniehu4:
### Summary:
https://app.shortcut.com/czi-edu/story/183913/figure-out-what-to-do-with-dependabot

All dependabot version update PRs fail the build ([example](https://github.com/chanzuckerberg/edu-design-system/runs/5314913517?check_suite_focus=true)) because [dependabot does not support yarn v2](https://github.com/dependabot/dependabot-core/issues/2030)

this turns off the PRs (which we did, and then un-did awhile ago)

### Test Plan:
not sure how to test, I followed https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/enabling-and-disabling-dependabot-version-updates#disabling-dependabot-version-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,7 +4,8 @@ updates:
     directory: "/"
     schedule:
       interval: daily
-    open-pull-requests-limit: 10
+    # dependabot does not work with yarn v2, disabling for now
+    open-pull-requests-limit: 0
     commit-message:
       prefix: chore
       include: scope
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -4,23 +4,26 @@ updates:
...
     commit-message:
       prefix: chore
       include: scope
   - package-ecosystem: npm
     directory: "/packages/components"
     schedule:
       interval: daily
-    open-pull-requests-limit: 10
+    # dependabot does not work with yarn v2, disabling for now
+    open-pull-requests-limit: 0
```
ahuth, line 26: Should we delete this file and not even run the job?

ahuth (approved):
LGTM

.github/dependabot.yml
anniehu4: I had hope that one day dependabot would fix this issue and then we could edit the config to work better :')

PR #874, opened 2022-02-25
upgrade stylelint to v14

anniehu4:
### Summary:
v13 is no longer supported by the VSCode extension. a rare case where `traject` has more updated dependencies than EDS 😁 

I also noticed that `yarn lint:styles` was throwing errors (is `lint-staged` not working 🤔 ), so fixed those

### Test Plan:
- `yarn lint:styles` shows no errors
- tested that nesting errors are still flaggged

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -19,7 +19,8 @@
     "custom-property-empty-line-before": "never",
     "no-eol-whitespace": null,
     "selector-nested-pattern": "^(&(:[\\w-]+|\\.state(Hover|Focus|FocusWithin|Active)),?\\s*)+$",
-    "max-nesting-depth": 1
+    "max-nesting-depth": 1,
+    "selector-class-pattern": null
   },
   "ignoreFiles": [
     "packages/tokens/lib/**"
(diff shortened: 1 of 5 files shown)


.stylelintrc
```diff
@@ -19,7 +19,8 @@
     "custom-property-empty-line-before": "never",
     "no-eol-whitespace": null,
     "selector-nested-pattern": "^(&(:[\\w-]+|\\.state(Hover|Focus|FocusWithin|Active)),?\\s*)+$",
-    "max-nesting-depth": 1
+    "max-nesting-depth": 1,
+    "selector-class-pattern": null
```
anniehu4, line 23: [stylelint-config-standard enforces kebab case](https://github.com/stylelint/stylelint-config-standard/blob/main/index.js#L129), which we don't care about rn

package.json
```diff
@@ -82,8 +82,7 @@
   "lint-staged": {
     "*.{js,jsx,ts,tsx}": [
       "prettier --write",
-      "eslint --fix",
-      "stylelint --fix"
```
anniehu4, line 86: I think this is a relic from our Linaria days, and we don't need to run stylelint on js/ts 🤔

jeremiah-clothier (approved):

PR #881, opened 2022-03-02
chore: update deps and resolve

jinlee93:
### Summary:
- update storybook, plop

### Test Plan:
- storybook, plop, watch commands work


anniehu4 (approved):
if these are fixing vulnerabilities, could you link them in the PR summary? thanks!

jinlee93:
> > if these are fixing vulnerabilities, could you link them in the PR summary? thanks!
> 
> Yup, will do

After discussion with Andrew, will be dropping the resolutions. Since dev use only, should have minimal risk and is not worth resolving for.

A somewhat related consideration could be using [`chokidar-cli`](https://www.npmjs.com/package/chokidar-cli) over [`watch`](https://www.npmjs.com/package/watch) as it seems to be a little more supported. However, didn't include as part of this pr since there it brings its own [vulnerabilities](https://github.com/open-cli-tools/chokidar-cli/pull/106)

PR #885, opened 2022-03-07
chore(dropdown): use key instead of id in options

dierat:
### Summary:
While working on migrating the existing `Dropdown` usage in `traject` to use the new eds component, I realized that the types for the `options` prop is inconsistent with the actual usage. In the stories, we were using `id`, but they're using `key` in the app. Changing the component to be consistent with actual usage will make it easier to use this prop.

### Test Plan:
I published an alpha build (0.8.1-alpha.1), updated `traject` locally to use that build, updated an example dropdown (`PDSearchBar`), and verified the conversion from using `<Dropdown.Options>` to `options` was smooth.


dierat:
Thanks!

PR #887, opened 2022-03-08
Move tooltip to next 

jinlee93:
[sc-187512]
### Summary:
- moves Tooltip from v0.8.1 to next
- remove TW use
- use CSS custom properties
- convert styles to BEM
- add storybook testing libraries

### Test Plan:
- tests pass
- manual testing

https://user-images.githubusercontent.com/86632227/157141485-c1fa6e7d-c395-4ddc-95cb-ada53831ea5f.mov

### DIFF
--- a/docs/CODE_GUIDELINES.md
+++ b/docs/CODE_GUIDELINES.md
@@ -88,7 +88,7 @@ Here's another example:
 
 ### CSS Nesting
 
-EDS uses [PostCSS Nested](https://github.com/postcss/postcss-nested) to provide some developer ergononics. As a general principle, nesting should be used sparingly and is only used in the following situations:
+EDS uses [PostCSS Nested](https://github.com/postcss/postcss-nested) to provide some developer ergonomics. As a general principle, nesting should be used sparingly and is only used in the following situations:
 
 - Media queries
 - States and pseudo-selectors
@@ -127,7 +127,7 @@ EDS uses [PostCSS Nested](https://github.com/postcss/postcss-nested) to provide 
 
 #### Parent selectors
 
-Use [parent selectors](https://sass-lang.com/documentation/style-rules/parent-selector) to target a selector when it appears inside a specific parent element. Use parent selectors instead of child selectors in order to co-locate all styles around a specific selector, which improves maintability and findability.
+Use [parent selectors](https://sass-lang.com/documentation/style-rules/parent-selector) to target a selector when it appears inside a specific parent element. Use parent selectors instead of child selectors in order to co-locate all styles around a specific selector, which improves maintainability and findability.
 
 Use the following conventions:
 
@@ -188,7 +188,7 @@ For example:
 }
 ```
 
-Not all CSS declarations warrant a comment, but non-obvious declarations (like context-specific styles, magic numbers, or styles dependant on other selector styles) should be accompanied by a comment.
+Not all CSS declarations warrant a comment, but non-obvious declarations (like context-specific styles, magic numbers, or styles dependent on other selector styles) should be accompanied by a comment.
 
 ### Other CSS Rules
 
@@ -261,7 +261,7 @@ Please refer to the [design tokens documentation](./TOKENS.md) to learn how to u
 
 - **Presentational Components Only** - EDS provides a library of reusable [presentational UI components](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0) that are consumed by CZI applications. These presentational components are "dumb" and don't contain any application business logic and aren't hooked up to any data models.
 - **Predictable APIs** - EDS provides consistent, clear [component APIs](#component-naming) in order to provide a consistent and intuitive user developer experience.
-- **Composition over inheritence** EDS adheres to the [composition over inheritence](https://en.wikipedia.org/wiki/Composition_over_inheritance) principle in order to create clean, extensible components that aren't tied to specific contexts or content.
+- **Composition over inheritance** EDS adheres to the [composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance) principle in order to create clean, extensible components that aren't tied to specific contexts or content.
 
 ## JavaScript Tools <a name="js-tools"></a>
 
@@ -320,7 +320,7 @@ export interface Props {
 }
 ```
 
-All component props must be defined with appropriate [TypeScript type](https://www.typescriptlang.org/docs/handbook/basic-types.html) applied. Each prop must contain a comment above the prop declaration to document the prop's function. These comments and prop declarations are automatically converted into prop documentation in Storybook. As a general guideline, try to organize component prop definitiones alphabetically.
+All component props must be defined with appropriate [TypeScript type](https://www.typescriptlang.org/docs/handbook/basic-types.html) applied. Each prop must contain a comment above the prop declaration to document the prop's function. These comments and prop declarations are automatically converted into prop documentation in Storybook. As a general guideline, try to organize component prop definitions alphabetically.
 
 ### Export module
 
@@ -333,7 +333,7 @@ export const ComponentName: React.FC<Props> = ({
 }) => {
 ```
 
-This defines the component name and passses in all the `Props`.
+This defines the component name and passes in all the `Props`.
 
 ### Variables, Methods, and Hooks (if applicable)
 
@@ -371,7 +371,7 @@ Components in this library exist in a flat structure so each component directory
 
 Certain components (such as `<Tabs />` and `<Table />`) require splitting up into smaller subcomponents.
 
-By default, we err towards more centralized control over the component architecture in order to prevent undesired results (for instance, we don't want users to put a `<Card />` inside of `<Breadcrumbs />`). However, certain components will require more flexibility and will therefore be architected to be composible and flexible (such as `<Card>`).
+By default, we err towards more centralized control over the component architecture in order to prevent undesired results (for instance, we don't want users to put a `<Card />` inside of `<Breadcrumbs />`). However, certain components will require more flexibility and will therefore be architected to be composable and flexible (such as `<Card>`).
 
 #### Conventions for compound components
 
@@ -391,17 +391,17 @@ By default, we err towards more centralized control over the component architect
 EDS follows specific front-end API naming conventions. Authoring a consistent API language provides many benefits:
 
 - **More efficient development** - Because the API language is consistent across components, user developers can spend more time coding rather than reading API documentation. Also, library contributors don't have to think as much about component API naming when creating new components/variants.
-- **Shared vocuabulary between designers and developers** - When the code library and design library use the same language, designers and developers can spend more time collaborating rather than futzing over what things are named. This improves team velocity and product quality. It also positions the team to benefit from future tooling that can bring design and code closer together (something many startups and plugins are trying to solve right now!)
+- **Shared vocabulary between designers and developers** - When the code library and design library use the same language, designers and developers can spend more time collaborating rather than futzing over what things are named. This improves team velocity and product quality. It also positions the team to benefit from future tooling that can bring design and code closer together (something many startups and plugins are trying to solve right now!)
 - **Future changes** - Utilizing a consistent language means that future changes and improvements are as easy as find-and-replace.
 
-EDS adhreres to the following API naming conventions:
+EDS adheres to the following API naming conventions:
 
 ### Variants
 
 - `variant` should be used for primary _stylistic_ variations of a component, such as (e.g. `<Card variant="bordered">` or `<Button variant="secondary">`). `variant` should be used if there is primarily one variable used to manipulate the component style.
 - `inverted` should be used consistently for stylistic `variation`s that "invert" the color schemes (e.g. `inverted=true`) to work on a darker background.
 - `size` should be used for adjusting size attributes (e.g. `<Button variant="secondary" size="sm">` or `<Button size="lg"`>). Default to `sm` and `lg`, with "md" being the default.
-- `behavior` should be used for funcitonal variations of a pattern, such as `<Banner behavior="dismissable">`. Additional non-exclusive behaviors should be handled using boolean props prefixed with `is` (e.g. `isSticky` and `isDismissable`).
+- `behavior` should be used for functional variations of a pattern, such as `<Banner behavior="dismissable">`. Additional non-exclusive behaviors should be handled using boolean props prefixed with `is` (e.g. `isSticky` and `isDismissable`).
 - `orientation` should be used for controlling the layout or orientation of a component (e.g. `<ButtonGroup orientation="stacked">`)
 - `align` should be used for aligning content, and should include `left` (default), `center`, `right` if needed.
 - `verticalAlign` should be used for vertically aligning content, and should include `top`, `middle`, `bottom` if needed.
@@ -410,7 +410,7 @@ EDS adhreres to the following API naming conventions:
 
 - Default to `text` for strings of text, such as `<Button text="First Name">`.
 - For headings, default to `title`, such as `<PageHeader title="My Page Title">`.
-- For form-related comnp, use the semantic `label` or `legend` (e.g. `<TextField label="first name" />`).
+- For form-related component, use the semantic `label` or `legend` (e.g. `<TextField label="first name" />`).
 
 ### Tag name
 
--- a/src/components/Counter/Counter.stories.tsx
+++ b/src/components/Counter/Counter.stories.tsx
@@ -1,6 +1,7 @@
 import { Story, Meta } from '@storybook/react';
 import React from 'react';
 import { Counter, Props } from './Counter';
+import { Button } from '../Button/Button';
 import { Tooltip } from '../Tooltip/Tooltip';
 
 export default {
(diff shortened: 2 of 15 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,140 +1,121 @@
...
+        variant === 'light' && styles['tooltip--light'],
+        variant === 'dark' && styles['tooltip--dark'],
+      )}
+      duration={200}
+      placement={align}
+      plugins={[hideOnEsc]}
+    />
   );
 };
+export default Tooltip;
```
jinlee93, line 121: Noticed the original components were not exporting `default`, is that something we should also do?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,140 +1,121 @@
...
- * Primary UI component for user interaction
- */
-export const Tooltip: React.FC<Props> = ({
-  className,
-  children,
-  align,
-  buttonText,
-  ...other
-}) => {
+  variant?: 'light' | 'dark';
```
jinlee93, line 47: Should this be `inverted`?

src/components/Tooltip/Tooltip.module.css
```diff
@@ -1,72 +1,126 @@
...
  */
+
 .tooltip {
-  display: inline-block;
-  position: relative;
-  overflow: visible;
+  border-style: solid;
+  border-width: var(--eds-border-width-sm);
+  border-radius: var(--eds-border-radius-lg);
+  box-shadow: var(--eds-theme-box-shadow);
```
jinlee93, line 16: I think this shadow is thinner than the original shadow, but using what is available for consistency
original on the left, this pr on the right:
image

src/components/Tooltip/Tooltip.stories.module.css
```diff
@@ -0,0 +1,20 @@
+.trigger--spacing {
+  margin: 9rem;
```
jinlee93, line 2: Weird number but it's appropriate required for the tooltip to place where alignment is indicated, otherwise Tippy places somewhere else due to lack of space

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -1,41 +1,153 @@
...
+import styles from './Tooltip.stories.module.css';
+import { Button, Props } from '../Button/Button';
 
-import { Tooltip, Props } from './Tooltip';
-import { TextPassage } from '../TextPassage/TextPassage';
+/* TODO: replace ForwardRefButton with EDS button since it's already using forwardRef */
+const ForwardRefButton = React.forwardRef((props: Props, ref) => (
+  <Button {...props} buttonRef={ref} />
+));
+ForwardRefButton.displayName = 'ForwardRefButton';
```
jinlee93, line 13: This is [required](https://github.com/atomiks/tippyjs-react#component-children) for for Tippy. Button that will be imported from EDS already uses [forwardRef](https://github.com/chanzuckerberg/edu-design-system/blob/9aadd55d95623cd0d777c88edcc93bc8e3edbf13/packages/components/src/Button/button.tsx#L34)

docs/CODE_GUIDELINES.md
```diff
@@ -88,7 +88,7 @@ Here's another example:
 
 ### CSS Nesting
 
-EDS uses [PostCSS Nested](https://github.com/postcss/postcss-nested) to provide some developer ergononics. As a general principle, nesting should be used sparingly and is only used in the following situations:
+EDS uses [PostCSS Nested](https://github.com/postcss/postcss-nested) to provide some developer ergonomics. As a general principle, nesting should be used sparingly and is only used in the following situations:
```
dierat, line 91: Wow thanks for all the spelling updates! ❤️

dierat: I somehow feel like this was some sort of test and you passed.

src/components/Tooltip/Tooltip.module.css
dierat: I wonder if we need a ticket for updating the tokens 🤔

src/components/Tooltip/Tooltip.stories.module.css
dierat: Maybe leave that in a comment?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,140 +1,121 @@
...
+   * `reference`.
    */
-  buttonText?: string;
+  reference?: React.RefObject<Element> | Element;
   /**
-   * Child node(s) that can be nested inside component. This gets displayed inside a `TextPassage` inside the tooltip
+   * The content of the tooltip bubble.
    */
-  children?: ReactNode;
+  content?: React.ReactNode;
```
dierat, line 23: I wonder if this should be `text` based on the prop guidelines Big Medium uses.

src/components/Tooltip/Tooltip.tsx
dierat: I'm not sure 🤔 `inverted` is when the color scheme is inverted so it will have enough contrast on a dark background, but both of these variants work on light and dark backgrounds. Actually, this might be a good time to bring up whether we really need both of them, and, if we do, what are the guidelines around when to use which?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,140 +1,121 @@
...
+  children?: React.ReactElement;
   /**
-   * The visually-hidden button text for the tooltip trigger button
+   * The trigger element the tooltip appears next to.
+   *
+   * Use this instead of `children` if the trigger element is being
+   * stored in a ref. Most cases will use `children` and not
+   * `reference`.
    */
-  buttonText?: string;
```
dierat, line 24: I can't figure out what this prop was 🤔 Do you understand the comment?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,140 +1,121 @@
...
    */
-  function onClick(e: any) {
-    e.preventDefault();
-    setIsActive(!isActive);
-  }
+  animation?: string | boolean;
+};
 
-  useEffect(() => {
-    if (isActive) {
```
dierat, line 66: Interesting that their version of the component is focusable. I wonder why.

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -1,41 +1,153 @@
...
+
+const defaultArgs = {
+  content: (
+    <span data-testid="tooltip-content">
```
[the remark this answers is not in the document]
jinlee93: It's used in the unit test https://github.com/chanzuckerberg/edu-design-system/pull/887/files#diff-6bab1489fbf9d9e083fa98db5f1d8d512d1003477f42e50ab09064a57df4330dR22-R26

src/components/Tooltip/Tooltip.tsx
jinlee93: Hmmm in my understanding, tooltip should [not be focusable](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role#keyboard_interactions)

src/components/Tooltip/Tooltip.tsx
jinlee93: Uhh not sure what the comment means but the b/m tooltip also exported the trigger button, and `buttonText` is for the text inside the trigger button 
image
Screen Shot 2022-03-08 at 11 06 04 AM

src/components/Tooltip/Tooltip.stories.module.css
jinlee93: Sounds good, added comment

src/components/Tooltip/Tooltip.tsx
jinlee93: Changed to `text`, thanks

src/components/Tooltip/Tooltip.stories.tsx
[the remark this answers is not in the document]
dierat: Ah thanks!

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,140 +1,123 @@
...
-        {children}
-      </span>
-    </div>
+    <Tippy
+      {...rest}
+      className={clsx(
+        styles.tooltip,
+        className,
+        variant === 'light' && styles['tooltip--light'],
+        variant === 'dark' && styles['tooltip--dark'],
```
dierat, line 114: Do you know if we've cleared with BM if we're using ampersands or object literals for these subjective classes? The code guidelines still has it using the object literals, but I think we want to keep using the ampersands. https://github.com/chanzuckerberg/edu-design-system/blob/next/docs/CODE_GUIDELINES.md#define-componentclassname

dierat:
Very nice!

So the open questions I still have about our process based on seeing this change are:
- Do we need a ticket to update the tokens (ex: shadows)?
- Should `inverted` be used in a case like this where there are two light/dark colors schemes but it's not related to dark backgrounds or dark mode?
- Do we still need 2 colors schemes for tooltips?
- Should we export a `default`?
- Are we using objects literals or ampersands for subjective classes (ex: `size === 'lg' && styles['my-component--lg'` vs `{[styles['my-component--lg']]: size === 'lg'}`? (If using ampersands, we should update the code guidelines.)

Does that sound right?

src/components/Tooltip/Tooltip.tsx
jinlee93: Oooh good q, I do like ampersands b/c I feel like it's easier to read

jinlee93:
> Does that sound right?

Yup that sounds right, thanks for surfacing these questions

src/components/Counter/Counter.stories.tsx
```diff
@@ -20,6 +21,19 @@ Default.args = {
   plusButtonText: 'Add by 1',
 };
 
+/* TODO: replace <ForwardRefButton> with EDS button since it's already using forwardRef */
```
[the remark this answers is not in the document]
anniehu4: could we quickly add forwardRef to the existing button now? not sure how long EDS button conversion will take & it's nicer not having to hunt down TODOs

src/components/Tooltip/Tooltip.module.css
```diff
@@ -1,72 +1,126 @@
-@import '../../design-tokens/mixins.css';
+/* stylelint-disable selector-pseudo-class-no-unknown */
```
anniehu4, line 1: since we're here, probs worth configuring this rule to ignore `global`: https://stylelint.io/user-guide/rules/list/selector-pseudo-class-no-unknown/#ignorepseudoclasses-regex-string

src/components/Tooltip/Tooltip.module.css
anniehu4: I think all tokens are subject to change while design is doing discovery right now; so we can feel free to replace the values with our existing ones

src/components/Tooltip/Tooltip.stories.module.css
anniehu4: let's also do the footnote commenting thing brad had, i.e.:
```suggestion
  margin: 9rem; /* 1 */
```
to note that this specific line is connected to the 1) comment above

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,140 +1,123 @@
...
+type TooltipProps = {
   /**
-   * Alignment variations
-   * - **right** yields a tooltip that appears to the right of the trigger
-   * - **below** yields a tooltip that appears below the trigger
-   * - **left** yields a tooltip that appears to the left of the trigger
+   * The trigger element the tooltip appears next to.
    */
-  align?: 'right' | 'below' | 'left';
+  children?: React.ReactElement;
```
anniehu4, line 11: nit: can we keep the props alphabetized

src/components/Tooltip/Tooltip.tsx
anniehu4: yeah that's a great point @dierat

src/components/Tooltip/Tooltip.tsx
anniehu4: maybe they were trying to handle tooltips w/ interactive content 🤔  we don't want that tho

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,140 +1,123 @@
...
-        tabIndex={isActive ? 0 : -1}
-        onKeyDown={onKeyDown}
-      >
-        {children}
-      </span>
-    </div>
+    <Tippy
+      {...rest}
+      className={clsx(
+        styles.tooltip,
```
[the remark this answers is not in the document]
anniehu4: let's be consistent with the brackets + quotes

src/components/Tooltip/Tooltip.tsx
anniehu4: I don't think it hurts for us to have the extra `Button/index.ts` file, and like the extra flexibility it gives 🙂 if that's not reflected in the plop-templates could you update them as well?

src/components/Tooltip/Tooltip.tsx
anniehu4: I think they wanted to see an example in code first. So here it is! @bradfrost @ifrost1 @csnizik any concerns with this syntax?

src/components/Tooltip/Tooltip.module.css
jinlee93: Sorry, to clarify, are you saying to add a token to `shadows.json`, or not using a token for now and assigning a value to `box-shadow`?

src/components/Tooltip/Tooltip.stories.module.css
jinlee93: Will do

src/components/Tooltip/Tooltip.module.css
anniehu4: either add a token to `shadows.json` or update `eds-theme-box-shadow` to match ours 🙂 we should always be using a token, but the values are owned by us, and will probably evolve over the next few weeks. so not sure it's worth to update all of them up front, but if we notice discrepancies we can default to the values we were already using for the component

src/components/Counter/Counter.stories.tsx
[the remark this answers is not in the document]
jinlee93: Yep we can, and will do

src/components/Tooltip/Tooltip.module.css
jinlee93: Will add to `stylelintrc`

src/components/Tooltip/Tooltip.module.css
jinlee93: Ah thanks for the explanation, will update shadow token `md`,

src/components/Tooltip/Tooltip.tsx
jinlee93: Yes! Alphabetizing

src/components/Tooltip/Tooltip.tsx
bradfrost: @anniehu4 @dierat @jinlee93 This is all good stuff, and happy to adopt whatever ergonomic things like this help the cause. FWIW, I wrote a post that broke down the pros and cons of all these approaches: https://bradfrost.com/blog/post/this-or-that-component-names-index-js-or-component-js/

src/components/Tooltip/Tooltip.tsx
bradfrost: Flagged this conundrum with the team over here and we're hopefully helping make some calls here: https://bigmedium.slack.com/archives/C0310HJ9WEN/p1646949365214339

src/components/Tooltip/Tooltip.module.css
bradfrost: @anniehu4 @jinlee93 @dierat yep the token work is firming up, and I'm going to take the lead on an initial import of what they've put in place in Figma. Hoping to get that in there by the end of the day.

src/components/Counter/Counter.stories.tsx
```diff
@@ -30,8 +31,14 @@ DefaultWithTooltip.args = {
...
   labelAfter: (
-    <Tooltip buttonText="Select this button to trigger the tooltip">
-      Some text to help with a form field
+    <Tooltip text="Some text to help with a form field">
+      <Button
+        variant="bare"
+        iconPosition="after"
+        iconName="question-mark-circle"
+        aria-label="Hover this button to trigger the tooltip"
+        text="Hover this button to trigger the tooltip"
```
dierat, line 40: Nit - I don't think you need both the `text` and `aria-label` here since they're the same.

dierat (approved):
I saw go ahead without deprecating the dark variant; we can do that separately.

src/components/Counter/Counter.stories.tsx
jinlee93: fixed, thanks

jinlee93:
> I saw go ahead without deprecating the dark variant; we can do that separately.

Sounds good to me

src/components/Tooltip/Tooltip.tsx
jinlee93: Via offline eng sync, committing to this style of syntax

jinlee93:
Will go ahead and merge as comments have been addressed. Deprecating the dark variant can be done in a separate PR after clearer deprecation guidelines have been established.

PR #889, opened 2022-03-08
chore(button): clear out props we didn't already have in the Button

dierat:
### Summary:
This PR clears out all the `Button` props we don't have in our EDS `Button` and aren't using. I'm mostly putting this out as a request for input. I'll add comments for each of the props in the code in `Button.tsx` so we can discuss each one.

One prop we didn't have that I didn't remove is `type` – I figure people are probably using that one in the app even though we didn't have it listed.

### Test Plan:
Verify the stories I removed are no longer in storybook and the ones I left are unchanged.

### DIFF
--- a/src/components/Button/Button.module.css
+++ b/src/components/Button/Button.module.css
@@ -59,20 +59,6 @@
       transform: none;
     }
   }
-
-  /**
-   * Inverted button 
-   */
-  &.button--inverted {
-    border-color: var(--eds-theme-color-primary-border-inverted);
-    color: var(--eds-theme-color-primary-foreground-inverted);
-    /* stylelint-disable-next-line max-nesting-depth */
-    &:hover,
-    &:focus {
-      border-color: var(--eds-theme-color-primary-border-inverted-hover);
-      color: var(--eds-theme-color-primary-foreground-inverted-hover);
-    }
-  }
 }
 
 /**
@@ -110,20 +96,6 @@
       color: var(--eds-theme-color-disabled-foreground);
     }
   }
-
-  /**
-   * Inverted primary button 
-   */
-  &.button--inverted {
-    background-color: var(--eds-theme-color-primary-background);
-    color: var(--eds-theme-color-body-foreground-inverted);
-    /* stylelint-disable-next-line max-nesting-depth */
-    &:hover,
-    &:focus {
-      background-color: var(--eds-theme-color-primary-background-hover);
-      color: var(--eds-theme-color-body-foreground-inverted);
-    }
-  }
 }
 
 /**
@@ -143,13 +115,6 @@
   &:focus {
     background-color: transparent;
   }
-
-  /**
-   * Inverted link button 
-   */
-  &.button--inverted {
-    @mixin TextLinkInverted;
-  }
 }
 
 /**
@@ -166,19 +131,6 @@
   &:focus {
     color: var(--eds-theme-color-primary-foreground-hover);
     background-color: transparent;
-  }
-
-  /**
-   * Inverted bare button 
-   */
-  &.button--inverted {
-    color: var(--eds-theme-color-primary-foreground-inverted);
-    /* stylelint-disable-next-line max-nesting-depth */
-    &:hover,
-    &:focus {
-      color: var(--eds-theme-color-primary-foreground-inverted-hover);
-      background-color: transparent;
-    }
   }
 }
 
@@ -202,68 +154,4 @@
   padding-right: var(--eds-size-5);
   padding-bottom: var(--eds-size-2);
   padding-left: var(--eds-size-5);
-}
-
-/**
- * Full-width button
- * 1) A block button fills 100% of the width of its container
- */
-.button--full-width {
-  width: 100%;
-}
-
-/**
- * Button Icon
- * 1) Set SVG icon fill to the current text color
- * 2) Rotate loading icon
- */
-.button__icon {
-  fill: currentColor; /* 1 */
-
-  .button.eds-is-loading & {
-    animation: rotateIcon 2s linear infinite; /* 2 */
-  }
-}
-
-/**
- * Button text before icon spacing
- * 1) Controls the spacing between the button text
- *    and the icon that comes _before_ button text
- */
-.button__icon + .button__text:not(.u-is-vishidden) {
-  margin-left: var(--eds-size-1); /* 1 */
-
-  .button--sm & {
-    margin-left: var(--eds-size-half);
-  }
-}
-
-/**
- * Button after icon spacing
- * 1) Controls the spacing between the button text
- *    and the icon that comes _after_ button text
- */
-.button__text:not(.u-is-vishidden) + .button__icon {
-  margin-left: var(--eds-size-1); /* 1 */
-
-  /**
-   * Button icon after button text that's not vishidden within a link variant button
-   */
-  .button--link & {
-    margin-left: var(--eds-size-half);
-  }
-
-  .button--sm & {
-    margin-left: var(--eds-size-half);
-  }
-}
-
-/* The animation code */
-@keyframes rotateIcon {
-  0% {
-    transform: rotate(0deg);
-  }
-  100% {
-    transform: rotate(360deg);
-  }
 }
--- a/src/components/Button/Button.module.css
+++ b/src/components/Button/Button.module.css
@@ -59,6 +59,20 @@
       transform: none;
     }
   }
+
+  /**
+   * Inverted button 
+   */
+  &.button--inverted {
+    border-color: var(--eds-theme-color-primary-border-inverted);
+    color: var(--eds-theme-color-primary-foreground-inverted);
+    /* stylelint-disable-next-line max-nesting-depth */
+    &:hover,
+    &:focus {
+      border-color: var(--eds-theme-color-primary-border-inverted-hover);
+      color: var(--eds-theme-color-primary-foreground-inverted-hover);
+    }
+  }
 }
 
 /**
@@ -96,6 +110,20 @@
       color: var(--eds-theme-color-disabled-foreground);
     }
   }
+
+  /**
+   * Inverted primary button 
+   */
+  &.button--inverted {
+    background-color: var(--eds-theme-color-primary-background);
+    color: var(--eds-theme-color-body-foreground-inverted);
+    /* stylelint-disable-next-line max-nesting-depth */
+    &:hover,
+    &:focus {
+      background-color: var(--eds-theme-color-primary-background-hover);
+      color: var(--eds-theme-color-body-foreground-inverted);
+    }
+  }
 }
 
 /**
@@ -115,6 +143,13 @@
   &:focus {
     background-color: transparent;
   }
+
+  /**
+   * Inverted link button 
+   */
+  &.button--inverted {
+    @mixin TextLinkInverted;
+  }
 }
 
 /**
@@ -131,6 +166,19 @@
   &:focus {
     color: var(--eds-theme-color-primary-foreground-hover);
     background-color: transparent;
+  }
+
+  /**
+   * Inverted bare button 
+   */
+  &.button--inverted {
+    color: var(--eds-theme-color-primary-foreground-inverted);
+    /* stylelint-disable-next-line max-nesting-depth */
+    &:hover,
+    &:focus {
+      color: var(--eds-theme-color-primary-foreground-inverted-hover);
+      background-color: transparent;
+    }
   }
 }
 
@@ -154,4 +202,73 @@
   padding-right: var(--eds-size-5);
   padding-bottom: var(--eds-size-2);
   padding-left: var(--eds-size-5);
+}
+
+/**
+ * Full-width button
+ * 1) A block button fills 100% of the width of its container
+ */
+.button--full-width {
+  width: 100%;
+}
+
+/**
+ * Button Icon
+ * 1) Set SVG icon fill to the current text color
+ */
+.button__icon {
+  fill: currentColor; /* 1 */
+}
+
+/**
+ * Button text before icon spacing
+ * 1) Controls the spacing between the button text
+ *    and the icon that comes _before_ button text
+ */
+.button__icon + .button__text {
+  margin-left: var(--eds-size-1); /* 1 */
+
+  .button--sm & {
+    margin-left: var(--eds-size-half);
+  }
+}
+
+/**
+ * Button after icon spacing
+ * 1) Controls the spacing between the button text
+ *    and the icon that comes _after_ button text
+ */
+.button__text + .button__icon {
+  margin-left: var(--eds-size-1); /* 1 */
+
+  /**
+   * Button icon after button text within a link variant button
+   */
+  .button--link & {
+    margin-left: var(--eds-size-half);
+  }
+
+  .button--sm & {
+    margin-left: var(--eds-size-half);
+  }
+}
+
+/**
+ * Loading icon rotiation style
+ */
+.button.eds-is-loading > .button__icon {
+  animation: rotateIcon 2s linear infinite; /* 1 */
+  overflow: visible;
+}
+
+/**
+ * Loading icon rotation animation
+ */
+@keyframes rotateIcon {
+  0% {
+    transform: rotate(0deg);
+  }
+  100% {
+    transform: rotate(360deg);
+  }
 }
--- a/src/components/Button/Button.stories.tsx
+++ b/src/components/Button/Button.stories.tsx
@@ -10,11 +10,35 @@ export default {
 
 const Template: Story<Props> = (args) => <Button {...args} />;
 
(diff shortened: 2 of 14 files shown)


src/components/Button/Button.tsx
```diff
@@ -1,16 +1,12 @@
...
+
 export interface Props {
   /**
    * Button reference
    */
   buttonRef?: any;
-  /**
-   * Toggle between full width button (true) and default button width
-   */
-  block?: boolean;
```
dierat, line 13: For `block` and `fullWidth`, it's not clear to me why we would need both (I only see `full-width` in the css and the comment calls it a "block button"). More importantly, I haven't seen anything in the designs to suggest this is an option we'd want to support. However, I'm open to being convinced that we should keep it because it's a common paradigm in web design.

src/components/Button/Button.tsx
```diff
@@ -19,44 +15,14 @@ export interface Props {
    * Disables the field and prevents editing the contents
    */
   disabled?: boolean;
-  /**
-   * Toggles button that fills the full width of its container
-   */
-  fullWidth?: boolean;
```
dierat, line 25: See the comment on `block`.

src/components/Button/Button.tsx
```diff
@@ -19,44 +15,14 @@ export interface Props {
...
-  /**
-   * Visually hide button text (but text is still accessible to assistive technology)
-   */
-  hideText?: boolean;
```
dierat, line 29: I'm guessing this is for the case where you have just an icon with some screen reader text. The way we're doing this now is just passing in an icon and putting the screen reader text on the icon. I think that API is a little more obvious, but it hinges on people passing in their own icons. Another option is to just have people pass in screen reader-only text (in `traject` they use just put an `sr-only` class on the element) rather than passing in `text` and separately also passing in `hideText`.

src/components/Button/Button.tsx
```diff
@@ -19,44 +15,14 @@ export interface Props {
...
-  /**
-   * Name of SVG icon (i.e. caret-down, minus, warning)
-   */
-  iconName?: string;
-  /**
-   * Determines position of icon relative to button text.
-   * - **before** places icon before button text
-   * - **after** places icon after button text
-   */
-  iconPosition?: 'before' | 'after';
```
dierat, line 43: We considered using a similar API as well and decided against it. The reason why is because all of the icons are in square bounding boxes, but the icons themselves do not always have a square ratio, which means some skinny icons will have more space between the icon and the button text than others. So we decided to let devs pass in their own icons which would allow them to customize the space between the icon and the text. That being said, I'm open to the suggestion that I'm overthinking this and being nitpicky and we should just use this API because it's easier for devs. Here's the conversation where we discussed passing in icons via children vs using props: https://github.com/chanzuckerberg/edu-design-system/pull/643#discussion_r713295633

src/components/Button/Button.tsx
```diff
@@ -19,44 +15,14 @@ export interface Props {
...
-  /**
-   * Button rendered on a dark backgorund
-   */
-  inverted?: boolean;
```
dierat, line 47: We currently don't have any designs with light-colored buttons or links, and I don't think we even have any dark backgrounds in the app really (it's all white or light gray) *except* for the main sidebar navigation, which is a pretty unique case. We do have a similar prop in Along though because they do use dark backgrounds quite often.

src/components/Button/Button.tsx
```diff
@@ -19,44 +15,14 @@ export interface Props {
...
-  /**
-   * Loading state passed down from higher level used to trigger loader and text change
-   */
-  loading?: boolean;
```
dierat, line 51: I actually really like this paradigm of having a spinner in the button and would be happy to keep this one if design feels it's a solution they want to start using. Or maybe we're already doing it and we want to expose it in the design system. In any case, I'd like design to weigh in on this one.

src/components/Button/Button.tsx
```diff
@@ -19,44 +15,14 @@ export interface Props {
...
   /**
    * On click handler for component
    */
   onClick?: MouseEventHandler;
-  /**
-   * Visually hidden additional instruction text to help provide screen reader users additional context. For instance, "View details" might be the visible button text, but screenReaderText might add additional instructions such as "for confirmation number C1234567"
-   */
-  screenReaderText?: string;
```
dierat, line 59: We've been using `aria-label`, which I like because people already know it (and if they don't know it, want them to learn about it).

dierat:
The tests are broken because I haven't updated the other components because I want to collect feedback from others before putting too much work into this change.

src/components/Button/Button.tsx
```diff
@@ -1,16 +1,12 @@
 import clsx from 'clsx';
 import React, { MouseEventHandler, ReactNode } from 'react';
 import styles from './Button.module.css';
-import { Icon } from '../Icon/Icon';
+
 export interface Props {
   /**
    * Button reference
    */
   buttonRef?: any;
```
dierat, line 9: I'm curious why this is called `buttonRef` instead of `ref`. Is there a chance devs will think it's referring to the icon?

src/components/Button/Button.tsx
```diff
@@ -87,51 +53,22 @@ export const Button: React.FC<Props> = ({
...
+  const componentClassName = clsx(
+    styles['button'],
+    className,
+    variant === 'primary' && styles['button--primary'],
+    variant === 'bare' && styles['button--bare'],
+    variant === 'link' && styles['button--link'],
+    size === 'sm' && styles['button--sm'],
+    variant === 'table-header' && styles['button--table-header'],
   );
+  const TagName = href ? 'a' : 'button';
```
dierat, line 71: This is not related to clearing out the props, but this paradigm is very different than what we had. We chose to split the clickable components out into `Button` and `Link` so it was clearer what the expected functionality of the component was for developers. The downside is then you have the complexity around sharing styles, and it was also very confusing for designers since the distinction between these components came down to the tag and functionality, not the visual appearance. I'm assuming we want to keep using separate components, but if people want to collapse them, I'm personally okay with that.

src/components/Button/Button.tsx
ahuth: It's because you can't name a prop `ref`. React actually has special handling for `ref` and `key`, which look like props but are handled differently.

The only way to get something like `<Button ref={someRef} />` is to use `forwardRef`. But an alternative to that is to just name a prop something other than `ref`. Like `buttonRef`.

src/components/Button/Button.tsx
ahuth: Ah, the age old conflict.

We should stay with our current thinking - separate components. `Button` renders a button, and `Link` renders an `a`.

src/components/Button/Button.tsx
dierat: Ah right, thanks!

src/components/Button/Button.tsx
bradfrost: @dierat This is a good thing to flag with the broader group! In our experience over the years, we've never encountered a system that didn't utilize a `fullWidth` button. So my gut says that it's good to keep around, but we can also remove it until there's a tangible use case for it. 

As far as naming goes, we'd prefer `fullWidth` over `block` as it's more descriptive. We've worked with designers who interpret a "block" button as "filled" (read: primary), so `fullWidth` is more explicit of its purpose.

src/components/Button/Button.tsx
bradfrost: I buy it!

src/components/Button/Button.tsx
bradfrost: @dierat no this is an important architectural convention! And it's something that me and @anniehu4 touched on briefly. Icons is its own can of worms and ought to have some careful consideration.

We should talk about this as a broader group (including design), but it really comes down to control. Ceding control at a `Button` component level and giving devs the ability to pass in their own icons and control spacing means that each button needs design QA for every instance vs addressing it at the EDS level. There are no right answers here! I think we'll flag this with the broader group to get a sense of what design's expectations are, and then marry it with the technical considerations and requirements.

src/components/Button/Button.tsx
bradfrost: @dierat This is especially helpful for `link` and `bare` variants of a button, and can be useful for default and `primary` buttons that need to adjust their colors based on what backgrounds they sit against. I'd recommend keeping that in there, especially as the specifics of the design language are subject to change.

src/components/Button/Button.tsx
bradfrost: @dierat yep, this is related to the icon conversation above. It depends on how much you want to enforce a consistent loading button state at an EDS level vs leaving it to downstream teams to implement things in a way that suits them best.

src/components/Button/Button.tsx
bradfrost: We buy it! That specific language originated from another client who already had similar conventions in here; happy to change it to something that matches the HTML API in order to spread the good news of ARIA.

src/components/Button/Button.tsx
bradfrost: Age-old conflict indeed! Just going to share a few things as food for thought: (@ifrost1 pinging you here too for visibility).

Here's how we've historically handled things:
- We handle `button`s that need to look like hyperlinks using `<Button variant="link">`
- Our `Link` component specifically deals with rendering an `a` tag that looks like a hyperlink. We don't tend to handle `a` tags that need to stylistically look like buttons using this component as it requires a bigger lift to represent all the button styles into the component
- As you've seen, we've typically used the `<Button href="#">` approach to swap out the appropriate HTML tag to handle button-looking links. As devs, we know the difference between a `button` and `a` is strictly behavioral, so we've taken that work on at the `Button` level. The `<Button href="#">` convention fits with a non-dev vernacular: "We need a primary button labeled "Shop new arrivals" that takes users to the new arrivals landing page." Hopefully that provides some color to the convention.

Given that CZI's culture is more savvy, I think it's perfectly fine to expect user devs to have the know-how to reach for a `Link` if they need an `a` tag and a `Button` if they need a `button`.

If not going the `Button` tagName switch route, there are two ways of handling sharing the styles between these two components:
1. Preserve the `ClickableStyle` component to manage the shared styles
2. Define the shared styles between `Link` and `Button` as mixins and instantiate those mixins in their respective components.

Either is perfectly valid. However, if going the `ClickableStyle` route, we'd recommend storing `ClickableStyle` in a directory like `components/utils` to differentiate them from user-facing UI components. It can get a bit confusing and messy to have helper/utility/pass-through components intermingled with user-facing UI components, so giving those components their own space helps keep things tidy.

I could make a case for going the mixin route: sharing clusters of styles is specifically what mixins do, and we tend to try to minimize abstractions and keep things as direct as possible (e.g. editing `Button` styles should intuitively involve editing `Button.module.css`). But honestly either approach is perfectly fine.

Sorry for the big ol' text wall to effectively say "all these approaches are perfectly fine". But it's good to hold stuff like this up to the light and hit it from all angles!

dierat: This is great, thanks! I really appreciate the layout of the different options (and the different options they lead to, lol).

Based on how our variants work (with the multiplier of the `color` prop), I'm leaning really heavily towards keeping `ClickableStyle`, but we can discuss that in a PR with the actual styles in it so I don't have to try to describe the situation.

dierat:
I made a separate [ticket for resolving some of these open API questions that rely on design input](https://app.shortcut.com/czi-edu/story/188471/button-api-questions-for-design). I'll leave them in the component unless we get a definite no from design that we don't want to support them.

src/components/Button/Button.tsx
dierat: I'll revert this change and ask design whether we want to support this feature in this component: https://app.shortcut.com/czi-edu/story/188471/button-api-questions-for-design

src/components/Button/Button.tsx
dierat: I'll revert this change and ask design whether we want to support this feature in this component: https://app.shortcut.com/czi-edu/story/188471/button-api-questions-for-design

src/components/Button/Button.tsx
dierat: I'll revert this change and ask design whether we want to support this feature in this component: https://app.shortcut.com/czi-edu/story/188471/button-api-questions-for-design

src/components/Button/Button.tsx
dierat: I'll revert this change and ask design whether we want to support this feature in this component: https://app.shortcut.com/czi-edu/story/188471/button-api-questions-for-design

src/components/Button/Button.module.css
```diff
@@ -215,22 +215,17 @@
...
  * Button Icon
  * 1) Set SVG icon fill to the current text color
- * 2) Rotate loading icon
  */
 .button__icon {
   fill: currentColor; /* 1 */
-
-  .button.eds-is-loading & {
-    animation: rotateIcon 2s linear infinite; /* 2 */
-  }
```
dierat, line 225: I moved this `.eds-is-loading` stuff down because I was getting a lint error about descending specificity by having this before the `.button__icon + .buton__text` stuff:
the lint error in VS code

src/components/Button/Button.module.css
```diff
@@ -61,8 +61,8 @@
   }
 
   /**
-   * Inverted button 
-   */
+   * Inverted button 
+   */
```
dierat, line 65: Please ignore the whitespace changes on the inverted styles 🙈

src/components/Button/Button.module.css
```diff
@@ -258,7 +253,17 @@
   }
 }
 
-/* The animation code */
+/**
+ * Loading icon rotiation style
+ */
+.button.eds-is-loading > .button__icon {
+  animation: rotateIcon 2s linear infinite; /* 1 */
+  overflow: visible;
```
dierat, line 261: The loading icon was getting cut off and looked jank before adding `overflow: visible`.

src/components/Button/Button.stories.tsx
```diff
@@ -99,15 +97,13 @@ export const Loading = Template.bind({});
 Loading.args = {
   loading: true,
   disabled: true,
-  text: 'Loading...',
+  text: 'Loading Button',
```
dierat, line 100: I changed the `text` in the loading stories because I thought it kinda looked like the `Button` logic was replacing the `text` with "Loading..." because `loading` was `true`.

src/components/Button/Button.stories.tsx
```diff
@@ -99,15 +97,13 @@ export const Loading = Template.bind({});
...
 };
 
 export const PrimaryLoading = Template.bind({});
 PrimaryLoading.args = {
   variant: 'primary',
-  iconPosition: 'before',
-  iconName: 'spinner',
```
dierat, line 109: The `Button` component handles this internally.

src/components/Button/Button.tsx
```diff
@@ -4,13 +4,14 @@ import styles from './Button.module.css';
 import { Icon } from '../Icon/Icon';
 export interface Props {
   /**
-   * Button reference
+   * Visually hidden button text (but text is still accessible to assistive technology).
+   * This overrides `text`
    */
-  buttonRef?: any;
+  'aria-label'?: string;
```
dierat, line 10: I added `aria-label` to the props just in case since I'm removing the other hidden text props.

src/components/Button/Button.tsx
dierat: Remove this and added `aria-label` instead 👍

src/components/Button/Button.tsx
dierat: Remove this and added `aria-label` instead 👍

src/components/Button/Button.tsx
```diff
@@ -4,13 +4,14 @@ import styles from './Button.module.css';
...
   /**
-   * Toggle between full width button (true) and default button width
+   * Button reference
    */
-  block?: boolean;
```
dierat, line 13: Removed `block` but kept `fullWidth` (since it seems like they are intended to serve the same function).

dierat:
This one is ready for a proper review now 👍 

Also I accepted all the chromatic screenshots baselines; lemme know if we were planning to hold off on that for some reason.

src/components/Button/Button.tsx
```diff
@@ -4,13 +4,14 @@ import styles from './Button.module.css';
...
+  buttonRef?: any;
```
anniehu4, line 14: nit: are we getting rid of this in a future PR? think we'd prefer to have the component `forwardRef` to people can pass in `ref` directly

anniehu4 (approved):
> Also I accepted all the chromatic screenshots baselines; lemme know if we were planning to hold off on that for some reason.

No reason to hold off 🙂  I _think_ Chromatic compares against `main` so it'll come up again (could be wrong though)

PR #890, opened 2022-03-09
chore(button): update size prop and styles

dierat:
### Summary:
Just what it says – this PR updates the `size` prop and styles for the `Button` component, including `typography-presets.css` and `typography-usage.css`. It does not touch `tokens.json`.

The specific styles of our buttons are subject to change very soon, but I'm sticking to the existing design for now mostly because I want to see how this system works when we're trying to match specific rules. Also I'm guessing the new button styles won't be *that* different since this component is already being used in a bunch of different places.

### Test Plan:
Navigate to storybook,
verify the stories unrelated to size all use large buttons now and that size matches our designs,
verify the small story now matches our designs,
verify there is now a medium story,
and verify it matches our designs.


src/components/Button/Button.module.css
```diff
@@ -139,19 +138,27 @@
  */
 .button--sm {
   @mixin eds-theme-typography-button-sm;
-  padding-top: var(--eds-size-half);
-  padding-right: var(--eds-size-1);
-  padding-bottom: var(--eds-size-half);
-  padding-left: var(--eds-size-1);
+  height: var(--eds-size-3);
```
dierat, line 141: We're specifying height instead of top and bottom padding because of our icons, which are enlarged when placed in buttons so they match the text even though they would cause the button height to expand if not contained in this way.

src/components/Button/Button.module.css
```diff
@@ -10,7 +10,6 @@
  * 3) Margin 0 added becuase Safari has margin on the button
  */
 .button {
-  @mixin eds-theme-typography-button;
```
dierat, line 13: Removed this line because there is now a default size and the sizes all specify the typography styles.

src/design-tokens/tier-1-definitions/typography-presets.css
```diff
@@ -106,10 +106,9 @@
 @define-mixin eds-typography-preset-8 $important {
   font-family: var(--eds-font-family-tertiary) $important;
   font-weight: var(--eds-font-weight-regular) $important;
-  font-size: var(--eds-font-size-md) $important;
-  line-height: var(--eds-line-height-sm) $important;
+  font-size: var(--eds-font-size-xs) $important;
+  line-height: var(--eds-size-1-and-half) $important;
```
dierat, line 110: I didn't add tokens for the new line heights because, honestly, I'm lazy and I don't want to rabbit hole into type sizes right now.

src/design-tokens/tier-2-usage/typography-usage.css
```diff
@@ -90,11 +90,15 @@
 }
 
 @define-mixin eds-theme-typography-button-sm $important {
-   @mixin eds-typography-preset-8 $important;
+   @mixin eds-typography-preset-8-bold $important;
+}
+
+@define-mixin eds-theme-typography-button-md $important {
+   @mixin eds-typography-preset-8-bold $important;
```
dierat, line 97: Medium and small buttons use the same typography; the difference between them is in the padding and height.

src/components/Button/Button.module.css
```diff
@@ -10,15 +10,10 @@
...
   display: inline-flex;
   align-items: center;
   justify-content: center;
-  padding-top: var(--eds-size-1-and-half);
-  padding-right: var(--eds-size-5);
-  padding-bottom: var(--eds-size-1-and-half);
-  padding-left: var(--eds-size-5);
```
dierat, line 20: Same as above – the sizes now handle their own padding rather than having a default set and overriding it in the sizes.

src/components/Button/Button.tsx
```diff
@@ -78,7 +78,7 @@ export const Button: React.FC<Props> = ({
   buttonRef,
   className,
   variant,
-  size,
+  size = 'lg',
```
anniehu4, line 81: _technically_ in the docs now it suggests `size` should always default to `'md'`. we've talked about this at length before but maybe can have design take another look at the variants and maybe remove one of the smaller sizes 🥲

src/design-tokens/tier-1-definitions/typography-presets.css
```diff
@@ -120,10 +119,9 @@
 @define-mixin eds-typography-preset-9 $important {
   font-family: var(--eds-font-family-tertiary) $important;
   font-weight: var(--eds-font-weight-regular) $important;
-  font-size: var(--eds-font-size-md-2) $important;
-  line-height: var(--eds-line-height-sm) $important;
-  letter-spacing: 1.5px $important;
-  text-transform: uppercase $important;
+  font-size: var(--eds-font-size-sm) $important;
```
anniehu4, line 122: kinda weird that `preset-9` is larger than `preset-8` ... I think we want them to be in descending size

src/design-tokens/tier-1-definitions/typography-presets.css
anniehu4: that's fine lol, could you leave comments though that we should update with line-height tokens when they're ready

src/components/Button/Button.tsx
dierat: >technically in the docs now it suggests size should always default to 'md'

I dunno, that feels weirdly restrictive to me and I'd personally rather change the docs than force the button sizes to comply. But if we really wanted to make it work, another solution would be to bump the buttons sizes down a level ('large' -> 'md', 'medium' -> 'sm', 'small' -> 'xs').

dierat: But I don't want to add a new large size if we have no actual intention of using it.

src/design-tokens/tier-1-definitions/typography-presets.css
dierat: Oops I didn't realize that, thanks!

src/components/Button/Button.tsx
anniehu4: yeah, fair. Changing the docs sounds fine

src/components/Button/Button.tsx
```diff
@@ -68,7 +68,7 @@ export interface Props {
   /**
    * Available _stylistic_ variations available for the Button component
    */
-  variant?: 'primary' | 'bare' | 'link';
+  variant?: 'primary' | 'bare' | 'link' | 'table-header';
```
dierat, line 71: Dunno why the 'table-header' variant is missing. I remove it in the next PR, but at this point in the stack, it's still being used here and in another file.

docs/CODE_GUIDELINES.md
```diff
@@ -470,12 +470,13 @@ EDS adheres to the following API naming conventions:
...
+- `size` should be used for adjusting size attributes (e.g. `<Button variant="secondary" size="sm">` or `<Button size="md"`>). Use abbreviations for sizes (ex: `xs`, `sm`, `md`, `lg`).
 - `behavior` should be used for functional variations of a pattern, such as `<Banner behavior="dismissable">`. Additional non-exclusive behaviors should be handled using boolean props prefixed with `is` (e.g. `isSticky` and `isDismissable`).
 - `orientation` should be used for controlling the layout or orientation of a component (e.g. `<ButtonGroup orientation="stacked">`)
 - `disabled` boolean should be used to control the interactivity of a component (e.g. `<Button disabled={true} />`)
 - `align` should be used for aligning content, and should include `left` (default), `center`, `right` if needed.
 - `verticalAlign` should be used for vertically aligning content, and should include `top`, `middle`, `bottom` if needed.
+- The default option should be the one most commonly used in order to reduce friction for developers using the components.
```
dierat, line 479: This is how we've been doing it so far. One issue is, again, this is very closely tied to the LP product. But the primary purpose of this DS is to accelerate development for the LP team, so I feel like this pattern makes sense.

src/design-tokens/tier-1-definitions/typography-presets.css
```diff
@@ -106,9 +106,9 @@
 @define-mixin eds-typography-preset-8 $important {
   font-family: var(--eds-font-family-tertiary) $important;
   font-weight: var(--eds-font-weight-regular) $important;
-  font-size: var(--eds-font-size-md) $important;
+  font-size: var(--eds-font-size-md-2) $important;
   line-height: var(--eds-line-height-sm) $important;
-  letter-spacing: normal $important;
+  letter-spacing: 1.5px $important;
```
dierat, line 111: `preset-8` is equal to the old `preset-9`. The kicker typography is now larger than the button typography, so it gets 8 and 9 is now the large button styles.

src/design-tokens/tier-1-definitions/typography-presets.css
```diff
@@ -120,13 +120,27 @@
...
+  font-size: var(--eds-font-size-xs) $important;
+  /* TODO: add this to line height variables */
+  line-height: var(--eds-size-1-and-half) $important;
+  letter-spacing: normal $important;
+}
+
+@define-mixin eds-typography-preset-10-bold $important {
+  @mixin eds-typography-preset-10 $important;
+  font-weight: var(--eds-font-weight-bold) $important;
+}
```
dierat, line 146: `preset-10` is the new medium + small button styles (the only difference between them is padding).

dierat, line 146: `preset-10` is the new medium + small button styles (the only difference between them is padding).

src/components/AvatarImage/__snapshots__/AvatarImage.spec.tsx.snap
```diff
@@ -0,0 +1,47 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
dierat, line 1: This is being automatically generated. If I remove it from the PR, the tests fail. 🤷

package.json
```diff
@@ -45,6 +45,7 @@
     "react": ">= 16.8.0"
   },
   "dependencies": {
+    "@chanzuckerberg/story-utils": "^3.0.4",
```
dierat, line 48: Moved this because of a lint error that insists this is a main dependency, which doesn't seem right to me since the file it's pointing to is a test file 🤔 https://github.com/chanzuckerberg/edu-design-system/runs/5572780062

> /home/runner/work/edu-design-system/edu-design-system/src/components/AvatarImage/AvatarImage.spec.tsx
> Error:   1:1  error  '@chanzuckerberg/story-utils' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
> 
> ✖ 1 problem (1 error, 0 warnings)

src/components/Tooltip/__snapshots__/Tooltip.test.tsx.snap
```diff
@@ -5,7 +5,7 @@ exports[`<Tooltip /> BottomPlacement story renders snapshot 1`] = `
   <div>
     <button
       aria-describedby="tippy-5"
-      class="button trigger--spacing-bottom trigger--spacing-left"
+      class="button trigger--spacing-bottom trigger--spacing-left button--lg"
```
dierat, line 8: These snapshot changes are because of the default size prop value.

dierat:
This one should be ready for a proper review now 👍

package.json
jinlee93: I just started running into this error too (not sure why it wasn't flagging before) but it's because it's a `spec` file not a `test` file 
image

src/components/AvatarImage/__snapshots__/AvatarImage.test.tsx.snap
```diff
@@ -0,0 +1,47 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
dierat, line 1: These `AvatarImage` test files were renamed to handle a lint error.

anniehu4 (approved):

dierat:
I'm gonna go ahead and merge this to make my downstream work easier, but feel free to leave comments – I'm happy to revisit anything in this PR as I continue this work.

PR #891, opened 2022-03-09
Feature/page shell

bradfrost:
### Summary: 
This PR puts some broad strokes in place for the overall page shell of the application. There's still a good amount of detailed work to do in order to bring it to parity with the live application, but I wanted to get this in here soon in order to enable the team to begin an integration trial in `traject`.


src/design-tokens/tier-1-definitions/layout.json
```diff
@@ -4,6 +4,9 @@
       "max-width": {
         "value": "71.25rem"
       },
+      "sidebar-width": {
+        "value": "13.4375rem"
```
anniehu4, line 8: Could we round up to `13.5rem` instead? (it's a difference of 1 pixel, from 215 -> 216px)

anniehu4 (approved):
Could you also remove the `package-lock.json` file before landing?

PR #895, opened 2022-03-15
feat: introduce badges addon for storybook

jinlee93:
### Summary:
- This explores the badges addon for storybook 
image
- Can be configured to have simple or more complex tooltips
image
image
- Badges can be configured (color, border, text, etc)
- https://storybook.js.org/addons/@geometricpanda/storybook-addon-badges

### Test Plan:
- Example [story](https://61313967cde49b003ae2a860-fcoxvnsxcz.chromatic.com/?path=/story/badgesaddon--default) to show usage, to be removed before merge?

### DIFF
--- a/packages/components/.storybook/preview.tsx
+++ b/packages/components/.storybook/preview.tsx
@@ -3,6 +3,10 @@ import "../src/styles/fonts.css";
 import "../src/styles/global.css";
 import "../src/styles/tailwindUtilities.css";
 
+import {
+  BADGE,
+  defaultBadgesConfig,
+} from "@geometricpanda/storybook-addon-badges";
 import React from "react";
 
 export const decorators = [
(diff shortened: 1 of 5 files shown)


packages/components/.storybook/preview.tsx
```diff
@@ -26,4 +30,26 @@ export const parameters = {
...
     ],
   },
+  badgesConfig: {
+    deprecated: {
+      ...defaultBadgesConfig[BADGE.DEPRECATED],
+      tooltip: "This component is deprecated, please avoid using it.",
+    },
+    beta: {
+      ...defaultBadgesConfig[BADGE.BETA],
+      tooltip: {
```
anniehu4, line 40: nit: could we remove the tooltips for now, until we find a use for them? 🙂

packages/components/docs/badges-addon.stories.tsx
```diff
@@ -0,0 +1,15 @@
+import { BADGE } from "@geometricpanda/storybook-addon-badges";
+import { StoryObj } from "@storybook/react";
+import React from "react";
+
+const Example = () => <p>Example to show badges addon</p>;
+
+export default {
```
anniehu4, line 7: Is this mostly for internal documentation? Would suggest naming it `title: "Documentation/BadgesAddon"` (or once we merge with `next`, adding to one of the comprehensive docs files there)

anniehu4 (approved):

packages/components/.storybook/preview.tsx
jinlee93: I'm down for removing the tooltips, mostly to show configuration possibility

bradfrost:
@jinlee93 This is awesome, thanks! When you feel good about this (looks like there's a little cleanup), would you mind also getting this setup into `next` as well? That way we can have it all set up and be able to manage things the other way (alpha! beta! stuff like that!). This is super cool.

PR #896, opened 2022-03-16
chore: add storybook badges add-on

jinlee93:
### Summary:
- Add storybook badges addon to next branch
- Same as #895 but for next branch

### Test Plan:
- See it working manually
Screen Shot 2022-03-16 at 9 12 14 AM


jinlee93:
Type and lint errors are being addressed in #890

PR #897, opened 2022-03-16
[1] Move `packages/tokens` to `src/tokens`

anniehu4:
### Summary:
[ch187206]

Working on [consolidating to publish a single EDS package](https://docs.google.com/document/d/1JKgyVWybRlt-KnjQ1Y9c6m2eE9D-mv3q57Y070IK9mg/edit#)

1. (this PR): consolidate `tokens`
2. consolidate `components`
3. replace yarn workspace commands
4. use `standard-version` instead of lerna in publish scripts
5. publish a new `@chanzuckerberg/eds` package!

### Test Plan:
- Confirmed same tokens are now being output to `lib/tokens` as part of the build (committed in GitHub so we can keep an eye on unintended changes)
- Build failures are expected because I renamed the entire `packages/` folder to `src/` but haven't fixed the configs affecting `components` yet

PR #898, opened 2022-03-16
 [2] Move `packages/components` to `src/components`  

anniehu4:
### Summary:
[ch187209]

Working on [consolidating to publish a single EDS package](https://docs.google.com/document/d/1JKgyVWybRlt-KnjQ1Y9c6m2eE9D-mv3q57Y070IK9mg/edit#)

1. [consolidate `tokens`](https://github.com/chanzuckerberg/edu-design-system/pull/897)
2. (this PR) consolidate `components`
3. replace yarn workspace commands
4. use `standard-version` instead of lerna in publish scripts
5. publish a new `@chanzuckerberg/eds` package!

### Test Plan:

- Everything is failing in this PR -- will fix with the next one 🙃


.storybook/preview.tsx
```diff
@@ -1,4 +1,4 @@
-import "@chanzuckerberg/eds-tokens/lib/css/variables.css";
+import "../lib/css/variables.css";
```
anniehu4, line 1: we now read tokens from internally. if it's bad practice to read from `lib/`, can save these elsewhere in the repo

anniehu4 (commented):
again, mostly just moving files around 🤓

.storybook/preview.tsx
```diff
@@ -1,4 +1,4 @@
-import "@chanzuckerberg/eds-tokens/lib/css/variables.css";
+import "../lib/tokens/css/variables.css";
```
anniehu4, line 1: doesn't seem great importing from `lib/`, but this may change when we merge in the `next` branch anyways. alt approach is to also output tokens into `.storybook` and `src/` as needed for internal use

ahuth: Seems reasonable to import from lib. Or at least may be better than jumping through hoops.

src/components/SvgIcon/SvgIcon.stories.tsx
```diff
@@ -1,6 +1,6 @@
-import * as ColorTokens from "@chanzuckerberg/eds-tokens/lib/ts/colors";
 import type { StoryObj } from "@storybook/react";
 import React from "react";
+import * as ColorTokens from "../../../lib/tokens/ts/colors";
```
ahuth, line 3: Not totally sure if it's worth it, but we could create a path alias to turn this relative import into something like

```ts
import * as ColorTokens from "lib/tokens/ts/colors";
```

ahuth (approved):
LGTM 🤘

src/components/SvgIcon/SvgIcon.stories.tsx
anniehu4: mm yeah. maybe if these stick around 😁

PR #899, opened 2022-03-18
chore: move text to next

jinlee93:
[sc-187515]
### Summary:
- Moves `Text` component from v0 to next
- Removed the ref prop since typography is no longer a subcomponent (and hence can’t pass ref along)
- Stories basically directly ported over, only added a `Colors` story to represent the color options available
- Removed `yellow` color since v0 typography never styled it
- Moved spacing out of typography and use in text styling, seemed more appropriate to isolate typography tokens

breaking changes from v0:
- `overline` is bold since using `eds-theme-typography-kicker`
- Base, brand, success, neutral are not the same color
- Heading sizes removed

### Test Plan:
- Wrote story
- Manual testing
- Tests pass


src/components/Text/Text.tsx
```diff
@@ -1,37 +1,80 @@
...
-   * The child node(s)
-   */
-  children?: ReactNode;
+export type Size =
+  | 'h1'
+  | 'h2'
+  | 'h3'
+  | 'h4'
+  | 'h5'
+  | 'body'
```
jinlee93, line 11: No `md` size, but defaults to `body`, should `md` be added to follow recommendations?

src/components/Text/Text.tsx
```diff
@@ -1,37 +1,80 @@
...
   /**
-   * CSS class names that can be appended to the component.
+   * Controls whether to render text inline (defaults to "p");
    */
+  as?: 'p' | 'span';
+  children: React.ReactNode;
   className?: string;
-}
+  color?: Color;
+  size?: Size;
```
jinlee93, line 36: No `variant` prop as there are a lot of modifying props, should either `size` or `color` (or other prop) be the `variant` prop?

src/components/Text/Text.tsx
```diff
@@ -1,37 +1,80 @@
...
+  | 'warning'
+  | 'white';
+
+export type Props = {
   /**
-   * CSS class names that can be appended to the component.
+   * Controls whether to render text inline (defaults to "p");
    */
+  as?: 'p' | 'span';
+  children: React.ReactNode;
```
jinlee93, line 33: Stuck with using `children` to pass the content. Can be changed to `text` as per guidelines, but seems more natural to write
```
<Text>Lorem Ipsum</Text>
```
than
```
<Text text="Lorem Ipsum />
```
to mimic <p> or <span>, however can change.

Also easier to handle cases like [`TextColorInherit` story](https://github.com/chanzuckerberg/edu-design-system/pull/899/files#diff-5972c0a39278e3e54030442e0e738d4084181e92d647a0dcbd3ee36145e209faR130) using `children`

src/design-tokens/tier-2-usage/typography-usage.css
```diff
@@ -100,3 +100,12 @@
...
    @mixin eds-typography-preset-6 $important;
 }
+
+@define-mixin eds-theme-typography-caption $important {
+   @mixin eds-typography-preset-7 $important;
+}
+
+@define-mixin eds-theme-typography-overline $important {
+   @mixin eds-typography-preset-7 $important;
+   text-transform: uppercase $important;
```
jinlee93, line 110: Put this here for typography usage, please advise if should relocate

src/components/Text/Text.module.css
```diff
@@ -1,12 +1,101 @@
...
+  margin-bottom: var(--eds-size-1);
+}
+.text--mb-2x {
+  margin-bottom: var(--eds-size-2);
+}
+
+/**
+ * Weights
+ */
+.text--weight-normal {
```
jinlee93, line 96: Not sure if appropriate bem syntax for spacing and weights. Added the "mb-" and "weight-" to distinguish nonspecific descriptors such as `-normal`

src/components/Text/Text.module.css
```diff
@@ -1,12 +1,101 @@
 @import '../../design-tokens/mixins.css';
 
 /*------------------------------------*\
-    #TEXT
+    #Text
```
dierat, line 4: Was this intentional?

src/components/Text/Text.tsx
bradfrost: @jinlee93 we should talk more about this, but I think it's important to dedicate `Text` for non-heading level text (such as `p` and `span`). A `Heading` component should manage the heading-level semantics and styles. 

It would still be possible to achieve a `Text` component with a heading-type style using utility classes, but would recommend against building that into the component to not encourage that usage.

src/components/Text/Text.tsx
bradfrost: @jinlee93 Yeah, for `Text` (and `Heading`, `TextPassage` and other text-specific components), `children` is the way to go. Supporting inner HTML tags (`<strong>`, `<sup>`, and so on) is a lot more elegant than using `dangerouslySetInnerHTML` on the component internals. 

This component has a specific usage (namely, very short text strings) so having `children` in place is manageable.

bradfrost: Just flagging that it would be a good idea to have a session to go through some of the text-level components like `Text`, `Heading`, `TextPassage`, and some text-but-also-structural components like `PageHeader` and `Section`.

src/components/Text/Text.tsx
bradfrost: @jinlee93 I reckon `color` ought to be `variant`, as `size` is being properly used to control the text size and `color` is the primary stylistic differentiator.

src/components/Text/Text.module.css
dierat: Makes sense to me. Using the example `button--full-width` from the docs, I could also see the prop name and value reversed (ie `text--normal-weight`), but I dunno if we wanna be that persnickety.

src/components/Text/Text.stories.tsx
```diff
@@ -1,18 +1,141 @@
...
+    );
+  },
+  parameters: {
+    /* 2 */
+    axe: {
+      skip: true,
+    },
+    snapshot: {
+      skip: true,
+    },
```
dierat, line 118: Is this one also covered by #2?

src/components/Text/Text.tsx
dierat: Mmmmmmm I can't decide 😅 'body' is just such a common term when discussing text, but the text used for body is generally medium-sized, no? I think I like 'body' more because the usage is just much clearer. Kinda like 'overline' and 'caption', and even the headings – the size of the text isn't as important as how it's used.

src/components/Text/Text.tsx
dierat: Honestly it generally feels more natural to pass the content via `children` for most components :/

src/components/Text/Text.tsx
dierat: I could see `size` becoming `variant` because it actually controls a couple of different style properties (ie text transform and weight) in addition to size.

src/components/Text/Text.stories.tsx
jinlee93: Yea, should I also put `/* 2 */ ` there?

src/components/Text/Text.stories.tsx
```diff
@@ -1,18 +1,141 @@
...
+import clsx from 'clsx';
 import React from 'react';
 
-import { Text, Props } from './Text';
+import Text from './Text';
+import type { Color } from './Text';
+import styles from './Text.stories.module.css';
 
 export default {
-  title: 'Atoms/Text/Text',
```
anniehu4, line 7: nit: do we wanna keep this organization?

src/components/Text/Text.stories.tsx
anniehu4: do they do this style of commenting in TSX files also? I kinda like not having to look back & forth to figure out what a comment means

src/components/Text/Text.tsx
anniehu4: @bradfrost The `Text` component as-is can only render a `p` or a `span` semantically, but can be styled to look like anything (which is what this `Size` type controls).

I took a quick scan through the app, and people are styling Text to look like a heading pretty rarely, and if they do it's `"h3"` or `"h4"`. Are you suggesting instead of
```
<Text size="h3">
```
to encourage
```
<Text className="eds-u-typography-h3" >
```
instead?

anniehu4: and I'd agree with keeping `body` 👍

src/components/Text/Text.tsx
anniehu4: > Just flagging that it would be a good idea to have a session to go through some of the text-level components like Text, Heading, TextPassage, and some text-but-also-structural components like PageHeader and Section.

Added a ticket: https://app.shortcut.com/czi-edu/story/189697/text-level-components

src/components/Text/Text.tsx
```diff
@@ -1,37 +1,80 @@
...
   className?: string;
-}
+  color?: Color;
+  size?: Size;
+  spacing?: 'none' | 'half' | '1x' | '2x';
```
anniehu4, line 37: could we remove this prop? not so useful with TW utility classes now

src/components/Text/Text.tsx
```diff
@@ -1,37 +1,80 @@
...
+   * Components that wrap typography sometimes requires props such as event handlers
+   * to be passed down into the element. One example is the tooltip component.  It
+   * attaches a onHover and onFocus event to the element to determine when to
+   * trigger the overlay.
+   */ ...other
+}: Props) => {
   const TagName = as;
-
   return (
-    <TagName className={componentClassName} {...other}>
```
anniehu4, line 33: same comment abt `componentClassName`

anniehu4 (commented):
removing the `typography` dependence makes sense to me!

> Removed the ref prop since typography is no longer a subcomponent (and hence can’t pass ref along)�

I think we should move the forwardRef and set the `ref` on the Text component then?

src/components/Text/Text.stories.tsx
jinlee93: Hmm, it's not part of any story file but it's pretty common amongst component files. Could just have a block comment instead

src/components/Text/Text.stories.tsx
jinlee93: Aye aye good catch, will change

src/components/Text/Text.tsx
jinlee93: Yup, sounds good

src/components/Text/Text.tsx
jinlee93: Um can you please elaborate?

src/components/Text/Text.module.css
jinlee93: Will removing `spacing` as it can be applied with utility classes and as for weight, will reverse as per snickets

src/components/Text/Text.tsx
dierat: One nice thing about using props though is we're specifying which levels are appropriate for the component. We can't stop you from using 12px but we encourage you to pick from this specific set. When we were migrating over the headings in traject, I found it useful to have a short list to pick from.

src/components/Text/Text.tsx
dierat: @anniehu4 what were you referring to here?

anniehu4: omg sorry I totally missed this -- can we pull out the className logic into a `componentClassName` variable like they do

jinlee93: Will do

src/components/Text/Text.module.css
```diff
@@ -5,8 +5,78 @@
...
+  @mixin eds-theme-typography-body-text-sm;
+}
+.text--xs {
+  @mixin eds-theme-typography-body-text-xs;
+}
+.text--caption {
+  @mixin eds-theme-typography-caption-text-md;
+}
+.text--overline {
+  @mixin eds-theme-typography-kicker;
```
jinlee93, line 35: For some reason, this isn't being capitalized https://61313967cde49b003ae2a860-hxziyqzwqa.chromatic.com/?path=/story/atoms-text-text--overline, please lmk what I can fix here

src/components/Text/Text.module.css
```diff
@@ -5,8 +5,78 @@
...
+.text,
+.text--body, 
+.text--md {
+  @mixin reset;
+  
+  @mixin eds-theme-typography-body-text;
+}
+
+/**
+ * Sizes
```
jinlee93, line 20: removed heading sizes

src/components/Text/Text.tsx
jinlee93: Kept body, but also added `lg` (available) and `md` (same as body)

src/components/Text/Text.tsx
jinlee93: Keeping children

src/design-tokens/tokens.css
```diff
@@ -0,0 +1,205 @@
+:root {
```
jinlee93, line 1: autogen

src/components/Text/Text.module.css
```diff
@@ -5,8 +5,78 @@
...
+/**
+ * Weights
+ */
+.text--normal-weight {
+  /* TODO: replace with tier 2 token */
+  font-weight: var(--eds-font-weight-medium);
+}
+.text--bold-weight {
+  /* TODO: replace with tier 2 token */
+  font-weight: var(--eds-font-weight-bold);
```
jinlee93, line 81: Please inform me if these TODO's are correct, haven't found the correct ones for typography

src/components/Text/Text.module.css
```diff
@@ -5,8 +5,78 @@
...
+}
+.text--caption {
+  @mixin eds-theme-typography-caption-text-md;
+}
+.text--overline {
+  @mixin eds-theme-typography-kicker;
+}
+
+/**
+ * Color
```
jinlee93, line 39: `base`, `brand`, `success`, `neutral` colors differ from v0, but I'm assuming that's intended via design

src/components/Text/Text.tsx
```diff
@@ -1,32 +1,69 @@
...
+      size = 'body',
+      weight,
+      /**
+       * Components that wrap typography sometimes requires props such as event handlers
+       * to be passed down into the element. One example is the tooltip component.  It
+       * attaches a onHover and onFocus event to the element to determine when to
+       * trigger the overlay.
+       */ ...other
+    }: Props,
+    ref?: ForwardedRef<HTMLParagraphElement>, // Setting as HTMLParagraphElement to satisfy TS, but unit test covers both span and p cases for sanity
```
jinlee93, line 52: Ref tested with both `span` and `p`

src/components/Text/Text.tsx
anniehu4: I think other components more commonly control color through `variant`, so renaming `color` makes sense to me. Text is a special case though so I'd be fine keeping it as-is for now and making a note to check once all components are in whether this is the only place we're still using `color` (and if so, should definitely rename it)

src/design-tokens/tokens.css
anniehu4: could you rebase on the latest `next`? this file should be in `src/tokens-dist/` now

src/components/Text/Text.tsx
anniehu4: nit: I don't think you have to type the `ref` as optional

src/components/Text/Text.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+    const HelperComponent = ({ as }: { as: 'p' | 'span' }) => {
+      const refContainerSpan = React.useRef(null);
+      return (
+        <Text as={as} ref={refContainerSpan}>
+          Ref container parent test {as}
+        </Text>
+      );
+    };
+    render(<HelperComponent as="p" />);
+    expect(screen.getByText('Ref container parent test p')).toBeTruthy();
```
anniehu4, line 29: is it enough that the component rendered without error? should we also test setting focus on the `refContainer` and making sure it works?

src/components/Text/Text.module.css
```diff
@@ -5,8 +5,78 @@
...
- * 1) 
+ * 1) h1, h2, h3, h4, h5, h6
+ * 2) Styles are either controlled by as or size declaration
  */
-/* stylelint-disable-next-line block-no-empty */
-.text {
+.text,
+.text--body, 
+.text--md {
+  @mixin reset;
```
anniehu4, line 14: will this make it hard for people to set their own margins through tailwind? we could just set `margin-bottom` to match the existing component

src/components/Text/Text.module.css
anniehu4: The kicker uses preset `-008`, which isn't capitalized. you can see the full (WIP) typescale here: https://www.figma.com/file/IfMWAtWqNrQprImlco4pi0/EDS---UI-Foundations?node-id=62%3A3269

Would suggest updating `eds-theme-typography-kicker` to point to either preset `-011` or `-012` instead

src/components/Text/Text.module.css
```diff
@@ -5,8 +5,78 @@
...
+}
+.text--success {
+  color: var(--eds-theme-color-text-utility-success);
+}
+.text--warning {
+  color: var(--eds-theme-color-text-utility-warning);
+}
+.text--white {
+  /* TODO: replace with tier 2 token */
+  color: var(--eds-color-neutral-white);
```
anniehu4, line 69:
```suggestion
  color: var(--eds-theme-color-text-neutral-knockout);
```

src/components/Text/Text.module.css
```diff
@@ -5,8 +5,78 @@
...
+.text--base {
+  color: var(--eds-theme-color-text-neutral-default);
+}
+.text--brand {
+  /* TODO: replace with tier 2 token */
+  color: var(--eds-color-brand-grape-700);
+}
+.text--info {
+  /* TODO: doesn't exist, replace with tier 2 token */
+  color: var(--eds-color-other-blue-700);
```
anniehu4, line 53: actually I think the blue tokens fully no longer exist, see: https://www.figma.com/file/IfMWAtWqNrQprImlco4pi0/EDS---UI-Foundations?node-id=58%3A4827

we should ask designers what to do about deprecating the `info` variants

src/components/Text/Text.module.css
```diff
@@ -5,8 +5,78 @@
 \*------------------------------------*/
 
 /**
- * 1) 
+ * 1) h1, h2, h3, h4, h5, h6
```
anniehu4, line 8: might be missing something -- is this comment used?

anniehu4 (approved):

src/components/Text/Text.module.css
jinlee93: It is not, will remove

src/components/Text/Text.tsx
jinlee93: I'm down to rename it as the color selections themselves sound more "variant-y" anyways

src/components/Text/Text.stories.module.css
```diff
@@ -0,0 +1,40 @@
...
+\*------------------------------------*/
+
+/**
+ * 1) For the "color" story to display the different text color options.
+ * 2) Uses grid to neatly align the color variations.
+ */
+.color__table { /* 1 */
+  display: grid; /* 2 */
+  grid-template-columns: 1fr 1fr 1fr; 
+  gap: 4px;
```
dierat, line 12: Maybe use `0.25rem` here instead of px? (I don't know if we ever decided on a pattern in EDS but in Along we only use px for things like borders and drop shadows; spacing and sizing is also rem. I don't think it has a user-facing effect though so it's more for code consistency imo.)

src/components/Text/__snapshots__/Text.test.tsx.snap
```diff
@@ -0,0 +1,79 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
dierat, line 1: I thought snapshots were being skipped 🤔

src/components/Text/Text.module.css
jinlee93: Was pointing to `-011` but there was a duplicate with `-008`, removed

src/components/Text/Text.module.css
jinlee93: Will change to `--eds-theme-color-text-neutral-default-knockout`

src/components/Text/Text.stories.module.css
jinlee93: Will change to `eds-size-half` (0.25rem), thanks

src/components/Text/__snapshots__/Text.test.tsx.snap
jinlee93: Oh were they for `Text` or in general?

src/components/Text/Text.module.css
jinlee93: Will do

src/components/Text/Text.test.tsx
jinlee93: Since it's a `p`/`span` element, can't focus, but instead tested ref is that element

src/design-tokens/tier-2-usage/typography-usage.css
```diff
@@ -164,10 +163,6 @@
   @mixin eds-typography-preset-006-bold $important;
 }
 
-@define-mixin eds-theme-typography-kicker $important {
-  @mixin eds-typography-preset-008 $important;
-}
-
```
jinlee93, line 170: Removing legacy preset since it's being used in line 75

PR #900, opened 2022-03-18
feat: add CloseButton

dierat:
### Summary:
Ticket: https://app.shortcut.com/czi-edu/story/187521/move-closebutton-to-new-architecture
Close button in `main`: https://github.com/chanzuckerberg/edu-design-system/tree/main/packages/components/src/CloseButton

Adding the `CloseButton`. This will be used by the `Banner` and `Toast` in EDS and the `Modal` in `traject` (which will eventually be moved into EDS as well).

Some outstanding issues:
- there's a whole bunch of TODOs in here because I'm handling styles with CSS that we were using `Icon` props for before
- also the tests are broken because jest doesn't like [some syntax in the `Icon` component code](https://github.com/chanzuckerberg/edu-design-system/blob/next/src/components/Icon/Icon.tsx#L67)
- I'm still not sure how we're supposed to handle using color tokens that aren't being used in a `theme` yet. The `Banner`, `Toast`, `CloseButton`, and `NotificationIcon` are designed to go together, so I'm thinking I'll make a new theme tokens file and list them all out in there.

### Test Plan:
Manually tested in storybook.

https://user-images.githubusercontent.com/7761701/159060263-0c212f95-cab9-49dd-8ed7-f61eb436a40e.mp4

### DIFF
--- a/src/components/CloseButton/CloseButton.module.css
+++ b/src/components/CloseButton/CloseButton.module.css
@@ -1,0 +1,48 @@
+.close-button {
+  width: var(--eds-size-6);
+  height: var(--eds-size-6);
+  border-radius: 50%;
+
+  svg {
+    /* TODO(Icon): use size prop on Icon after that component is done */
+    height: var(--eds-size-4);
+    width: var(--eds-size-4);
+    /* TODO(Icon): use color prop on Icon after that component is done */
+    fill: var(--close-button-icon-color);
+  }
+
+  &:hover {
+    background-color: var(--close-button-icon-hover-color);
+  
+    /* TODO(Icon): fix after Icon component is done */
+    /* stylelint-disable-next-line max-nesting-depth */
+    svg {
+      fill: var(--eds-theme-color-neutral-bold-foreground-inverted);
+    }
+  }
+}
+
+.close-button--variant-brand {
+  --close-button-icon-color: var(--eds-color-brand-blue-700);
+  --close-button-icon-hover-color: var(--eds-theme-color-primary-foreground);
+}
+
+.close-button--variant-neutral {
+  --close-button-icon-color: var(--eds-theme-color-neutral-md-background-inverted);
+  --close-button-icon-hover-color: var(--eds-theme-color-neutral-md-foreground);
+}
+
+.close-button--variant-success {
+  --close-button-icon-color: var(--eds-theme-color-utility-success-foreground);
+  --close-button-icon-hover-color: var(--eds-color-other-green-600);
+}
+
+.close-button--variant-warning {
+  --close-button-icon-color: var(--eds-theme-color-utility-warning-foreground);
+  --close-button-icon-hover-color: var(--eds-color-other-orange-600);
+}
+
+.close-button--variant-alert {
+  --close-button-icon-color: var(--eds-theme-color-utility-error-foreground);
+  --close-button-icon-hover-color: var(--eds-color-other-red-600);
+}
--- a/src/components/CloseButton/CloseButton.tsx
+++ b/src/components/CloseButton/CloseButton.tsx
@@ -1,0 +1,76 @@
+import clsx from 'clsx';
+import React, { forwardRef } from 'react';
+import styles from './CloseButton.module.css';
+import { Button } from '../Button/Button';
+import { Icon } from '../Icon/Icon';
+
+type CloseButtonProps = {
+  /**
+   * Stylistic variations for the close button type.
+   * - **brand** - results in a blue background color on hover
+   * - **neutral** - results in a gray background color on hover
+   * - **success** - results in a green background color on hover
+   * - **warning** - results in an orange background color on hover
+   * - **alert** - results in a red background color on hover
+   */
+  variant?: 'brand' | 'neutral' | 'success' | 'warning' | 'alert';
+  // TODO(Icon): handle this when the Icon component is done
+  /**
+   * Size of the icon. Does not affect actual button size. The button is larger than the
+   * icon to ensure the hit box is large enough (for accessibility).
+   *
+   * The string should be some number + px, rem, em, vh, etc. Ex: 2rem.
+   * Recommendation: use "EdsSizeLineHeight" tokens from
+   * @chanzuckerberg/eds-tokens/json/variables.json
+   */
+  // iconSize?: string;
+  /**
+   * Custom aria-label. If undefined, "close" will be used.
+   */
+  'aria-label'?: string;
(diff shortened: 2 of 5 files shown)


bradfrost:
@dierat Woot! A question: do we have an idea where the success/alert/warning variants are used. Is it `Banner`? Or are there other instances?

dierat:
> @dierat Woot! A question: do we have an idea where the success/alert/warning variants are used. Is it `Banner`? Or are there other instances?

Great question! Currently we have:
- `Button`
- `Banner`
- `Toast`
- `CloseButton` (this component)
- `NotificationIcon` (helper specifically for `Toast` and `Banner`)

And there are 2 other notification components in figma using these: `PageLevelBanner` and `InlineNotification`. `Tag` is not in the list because it also includes yellow (with black text).

Originally `info` was floating around too, specifically in the `Banner` I think, but since that was dropped these are now all using the same 5 themes.

src/components/CloseButton/CloseButton.module.css
```diff
@@ -0,0 +1,48 @@
+.close-button {
+  width: var(--eds-size-6);
+  height: var(--eds-size-6);
+  border-radius: 50%;
```
anniehu4, line 4: nit:
```suggestion
  border-radius: var(--eds-border-radius-round);
```

src/components/CloseButton/CloseButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+      onClose,
+      // TODO(Icon): handle this when the Icon component is done
+      // iconSize = '2rem',
+      'aria-label': ariaLabel = 'close',
+      ...rest
+    },
+    ref,
+  ) => (
+    <Button
+      className={clsx(
```
anniehu4, line 56: nit: I'm just noticing this now, but let's pull out className logic into `componentClassName` like they have in the docs

src/components/CloseButton/CloseButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+        variant === 'brand' && styles['close-button--variant-brand'],
+        variant === 'neutral' && styles['close-button--variant-neutral'],
+        variant === 'success' && styles['close-button--variant-success'],
+        variant === 'warning' && styles['close-button--variant-warning'],
+        variant === 'alert' && styles['close-button--variant-alert'],
+      )}
+      onClick={onClose}
+      variant="bare"
+      ref={ref}
+      text={<Icon name="x" title={ariaLabel} />}
```
anniehu4, line 69: o .... ya if we're passing non-text stuff through, this `text` prop really doesn't make sense :eye-twitch:

src/components/CloseButton/CloseButton.module.css
```diff
@@ -0,0 +1,48 @@
...
+  
+    /* TODO(Icon): fix after Icon component is done */
+    /* stylelint-disable-next-line max-nesting-depth */
+    svg {
+      fill: var(--eds-theme-color-neutral-bold-foreground-inverted);
+    }
+  }
+}
+
+.close-button--variant-brand {
```
anniehu4, line 25: super super nit: could this be `close-button--brand`? or is `variant-brand` more consistent?

anniehu4 (commented):
> I'm still not sure how we're supposed to handle using color tokens that aren't being used in a theme yet.

I also wonder if this'll become easier with the token changes @bradfrost and @tpitre-czi are doing? Maybe (hopefully) it's already on their radar

Edit: here are the WIP token changes: https://github.com/chanzuckerberg/edu-design-system/pull/903

src/components/CloseButton/CloseButton.module.css
```diff
@@ -0,0 +1,48 @@
...
+  width: var(--eds-size-6);
+  height: var(--eds-size-6);
+  border-radius: 50%;
+
+  svg {
+    /* TODO(Icon): use size prop on Icon after that component is done */
+    height: var(--eds-size-4);
+    width: var(--eds-size-4);
+    /* TODO(Icon): use color prop on Icon after that component is done */
+    fill: var(--close-button-icon-color);
```
anniehu4, line 11: I also wonder BM's thoughts on duplication vs. clarity / themability. Looking at their banner component, it seems they would do something like:
```
svg {
  // parent selector
  .close-button--variant-brand & {
     fill: ...
  }
  // and so on
```

instead of pulling out a `--close-button-icon-color` variable

anniehu4 (approved):

src/components/CloseButton/CloseButton.tsx
dierat: Another way to do it is use one of the icon props and remove the margin that's automatically applied. I passed it in as `text` because it feels weird to me to remove the margin if we don't need to, but I think that might be the preferred way to do it based on existing patterns?

src/components/CloseButton/CloseButton.module.css
dierat: In this case, I expect this to be a short-term solution because our `Icon` component has internal color support, but it's good to have the conversation around whether we prefer duplication even if css variables would make it shorter and cleaner.

src/components/CloseButton/CloseButton.tsx
dierat: Ah, will do, thanks!

src/components/CloseButton/CloseButton.module.css
csnizik: BM avoids using HTML elements as selectors. Use a class instead of `svg` here.

csnizik: Actually now that I looked at this in context, we're modifying the svg that is rendered by the IconComponent, which in turn is being rendered as the text prop of the ButtonComponent. This feels like it's going against the Single Responsibility Principle to me.

dierat: @csnizik as the TODO comments indicate, this is temporary styling that will be replaced when the `Icon` component can support the `color` and `size` props and therefore apply this styling itself. On `main` we're changing the text color on the button, which the svg inherits. https://github.com/chanzuckerberg/edu-design-system/blob/main/packages/components/src/CloseButton/CloseButton.module.css

That being said, I'll talk to Sean about whether we can use the standard icon button styles for this. I personally like this button's styling but I also want it to be consistent; we don't use this visual pattern anywhere else, which is why we have to do it manually instead of relying on the button component.

dierat:
I talked to Sean about this component and he's okay with just using a plain button with an icon for these situations. Closing!

PR #902, opened 2022-03-18
feat: add NotificationIcon

dierat:
### Summary:
Ticket: https://app.shortcut.com/czi-edu/story/187520/move-notificationicon-to-new-architecture
NotificationIcon on `main`: https://github.com/chanzuckerberg/edu-design-system/tree/main/packages/components/src/common/Notifications/NotificationIcon

This adds the `NotificationIcon` that will be used in the `Banner` and `Toast` components. This PR is very similar to https://github.com/chanzuckerberg/edu-design-system/pull/900

Outstanding issues:
- Lots of TODO comments again for the `Icon` component because I'm adding the styles with CSS whereas before we were passing down props and the `Icon` was handling it
- No test file because jest doesn't like [some syntax in the `Icon` component code](https://github.com/chanzuckerberg/edu-design-system/blob/next/src/components/Icon/Icon.tsx#L67)
- I'm still not sure how we're supposed to handle using color tokens that aren't being used in a `theme` yet. The `Banner`, `Toast`, `CloseButton`, and `NotificationIcon` are designed to go together, so I'm thinking I'll make a new theme tokens file and list them all out in there.
- Also, we're probably going to reuse that variant type across some of these components. We used to export it from this component and use it in other places, but I feel like it should be in its own file instead of in a component because the `CloseButton` uses the same type as well and it doesn't rely on this component.

### Test Plan:
Manually tested in storybook. I currently have this component in `Molecules/Messaging`, but since it's a util component and not intended to be consumed externally, I'm planning to remove the `.stories` file when it's been integrated into the `Banner` and `Toast`.

notification icon in storybook

### DIFF
--- a/src/components/Utilities/NotificationIcon/NotificationIcon.module.css
+++ b/src/components/Utilities/NotificationIcon/NotificationIcon.module.css
@@ -1,0 +1,48 @@
+/*------------------------------------*\
+    # NOTIFICATION ICON
+\*------------------------------------*/
+
+/* TODO(Icon): update styles to use Icon component props after Icon component is done */
+
+/**
+ * Notification icon
+ */
+.notification-icon {
+  width: var(--eds-size-2-and-half);
+  height: var(--eds-size-2-and-half);
+}
+
+/**
+ * Brand notification icon
+ */
+.notification-icon--brand {
+  fill: var(--eds-color-brand-blue-500);
+}
+
+/**
+ * Neutral notification icon
+ */
+.notification-icon--neutral {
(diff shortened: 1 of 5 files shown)


bradfrost:
@dierat Looking at this now, and think we'll want to leave this sit here until we have our icon discussion on Thursday. I'm confident we'll land on some good conventions we can use for all icons, including these ones!

dierat:
Oops, meant to close this. We can come back to this question later; the code duplication here isn't a big deal.

PR #905, opened 2022-03-20
feat(dropdown): add compact button as option

silvestry:
### Summary:

Addresses: https://app.shortcut.com/czi-edu/story/189746/notebook-c3-compact-mode-variant-for-eds-drop-down

This PR adds compact variant to EDS drop-down:

- Dropdown button is only as large as content requires
- Dropdown can have differing button width and options menu width
- Dropdown menu has wider width to accomodate
- Dropdown menu can be right aligned to the button

Screen Shot 2022-03-19 at 6 43 49 PM

Alternative approaches considered:
1. Using CSS var to defined and pass width
```
.compact > .options {
  width: var(--eds-dropdown-options-width);
}
```
2. Passing ref back to caller to style options menu width

### Test Plan:

See storybook for Dropdown component:
1. Compact
2. Compact With Options Right Aligned
3. Separate Button And Menu Width
4. Compact Using Children Prop And No Visible Label

### DIFF
--- a/packages/components/src/Dropdown/Dropdown.module.css
+++ b/packages/components/src/Dropdown/Dropdown.module.css
@@ -2,6 +2,14 @@
   @apply relative bg-white;
 }
 
+.compact {
+  @apply w-min;
+}
+
+.compact > button {
+  @apply p-1;
+}
+
 .label {
   @apply text-sm text-neutral-500 font-bold mb-1;
 }
@@ -11,7 +19,15 @@
 }
 
 .options {
-  @apply w-full absolute bg-white mt-2 py-2 rounded shadow-2 z-10;
+  @apply absolute bg-white mt-2 py-2 rounded shadow-2 z-10;
+}
+
+.optionsFullWidth {
+  @apply w-full;
+}
+
+.optionsAlignRight {
+  @apply right-0;
 }
 
 .option {
--- a/packages/components/src/Dropdown/Dropdown.module.css
+++ b/packages/components/src/Dropdown/Dropdown.module.css
@@ -2,11 +2,11 @@
   @apply relative bg-white;
 }
 
-.compact {
+.compactDropdown {
   @apply w-min;
 }
 
-.compact > button {
+.compactDropdownButton {
   @apply p-1;
 }
 
--- a/packages/components/src/Dropdown/Dropdown.tsx
+++ b/packages/components/src/Dropdown/Dropdown.tsx
@@ -68,6 +68,9 @@ type DropdownProps = ListboxProps & {
   optionsAlign?: OptionsAlignType;
   /**
    * Optional className for additional options menu styling.
+   *
+   * When not using the compact variant, if optionsClassName is provided please
+   * include the width property to define the options menu width.
    */
   optionsClassName?: string;
 };
--- a/packages/components/src/Dropdown/Dropdown.tsx
+++ b/packages/components/src/Dropdown/Dropdown.tsx
@@ -284,7 +284,7 @@ function Dropdown(props: DropdownProps) {
     className: clsx(
       styles.dropdown,
       className,
-      variant === "compact" && styles.compact,
+      variant === "compact" && styles.compactDropdown,
     ),
     // Provide a wrapping <div> element for the dropdown. This is needed so that any props
     // passed directly to this component have a corresponding DOM element to receive them.
@@ -374,6 +374,8 @@ const DropdownTrigger = function (
   props: PropsWithRenderProp<{ disabled: boolean; open: boolean }>,
 ) {
   const { children, className, ...rest } = props;
+  const { compact } = useContext(DropdownContext);
+
   return (
     <Listbox.Button
       // Render as a fragment instead of the default element. We're rendering our own element in
@@ -385,7 +387,11 @@ const DropdownTrigger = function (
       {typeof children === "function" ? (
         children
       ) : (
-        <DropdownButton className={className}>{children}</DropdownButton>
+        <DropdownButton
+          className={clsx(className, compact && styles.compactDropdownButton)}
+        >
+          {children}
+        </DropdownButton>
       )}
     </Listbox.Button>
   );
--- a/packages/components/src/Dropdown/Dropdown.tsx
+++ b/packages/components/src/Dropdown/Dropdown.tsx
@@ -11,6 +11,7 @@ import CheckRoundedIcon from "../Icons/CheckRounded";
 import styles from "./Dropdown.module.css";
 
 export type OptionsAlignType = "left" | "right";
+export type VariantType = "compact" | "full";
 
 type ListboxProps = ComponentProps<typeof Listbox>;
 type DropdownProps = ListboxProps & {
@@ -56,15 +57,13 @@ type DropdownProps = ListboxProps & {
    */
   className?: string;
   /**
-   * Render dropdown button that is only as wide as the content.
+   * The style of the dropdown.
    *
-   * When defining compact dropdown need to provide optionsWidth to
-   * define width of options menu, and optionally provide optionsAlign
-   * if desired to right align to dropdow button.
+   * Compact renders dropdown button that is only as wide as the content.
    */
-  compact?: boolean;
+  variant?: VariantType;
   /**
-   * Align dropdown menu to the left (default) or right of dropdown button
+   * Align dropdown menu to the left (default) or right of dropdown button.
    */
   optionsAlign?: OptionsAlignType;
   /**
@@ -190,7 +189,7 @@ function childrenHaveLabelComponent(children?: ReactNode): boolean {
  * );
  * ```
  *
- * For compact dropdown button, add compact and optionsWidth props to
+ * For compact variant, add variant="compact" and optionsWidth props to
  * <Dropdown>.
  *
  * Examples:
@@ -200,9 +199,9 @@ function childrenHaveLabelComponent(children?: ReactNode): boolean {
  * return (
  *   <Dropdown
  *     aria-label="Options"
- *     compact
  *     optionsAlign="right"
  *     optionsWidth="w-96"
+ *     variant="compact"
  *   >
  *     <Dropdown.Options>
  *       <Dropdown.Option>Option 1</Dropdown.Option>
@@ -221,10 +220,10 @@ function childrenHaveLabelComponent(children?: ReactNode): boolean {
  * return (
  *   <Dropdown
  *     aria-label="Options"
- *     compact
  *     options={options}
  *     optionsAlign="right"
  *     optionsWidth="w-96"
+ *     variant="compact"
  *   />
  * );
  * ```
@@ -244,10 +243,10 @@ function childrenHaveLabelComponent(children?: ReactNode): boolean {
  *   <Dropdown
  *     aria-label="Options"
  *     className="w-60"
- *     compact
  *     options={options}
  *     optionsAlign="right"
  *     optionsWidth="w-96"
+ *     variant="compact"
  *   />
  * );
  * ```
@@ -255,12 +254,12 @@ function childrenHaveLabelComponent(children?: ReactNode): boolean {
 function Dropdown(props: DropdownProps) {
   const {
     className,
+    variant,
     labelText,
     buttonText,
     options,
     children,
     "aria-label": ariaLabel,
-    compact,
     optionsAlign,
     optionsWidth,
     ...rest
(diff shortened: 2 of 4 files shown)


packages/components/src/Dropdown/Dropdown.tsx
```diff
@@ -48,6 +55,23 @@ type DropdownProps = ListboxProps & {
...
+  compact?: boolean;
+  /**
+   * Align dropdown menu to the left (default) or right of dropdown button
+   */
+  optionsAlign?: OptionsAlignType;
+  /**
+   * Render dropdown menu with width (specify using tailwind utility string)
+   * independent of dropdown button width
+   */
+  optionsWidth?: string;
```
anniehu4, line 74: Do these need to be separate props independent of `compact`? i.e.:
- Will a `compact` dropdown always have right aligned options and a specific width?
- Or, is this accounting for possible flexibility with the Dropdown options, independent of what the button looks like?

If the first, I'd recommend removing the external props and enforcing the styling for `compact`. If the latter, that's basically what the existing Dropdown is already -- our advice (to avoid complicating the existing component), is to apply your own Tailwind classes on `Dropdown.Options` if you need to customize the styles further.

Does that make sense? Or we could surface `optionsClassname` in the overall Dropdown if you anticipate _every_ `compact` Dropdown is going to need some customization

packages/components/src/Dropdown/Dropdown.module.css
```diff
@@ -2,6 +2,14 @@
   @apply relative bg-white;
 }
 
+.compact {
+  @apply w-min;
+}
+
+.compact > button {
+  @apply p-1;
+}
```
anniehu4, line 11: Maybe we could thread the `compact` prop through `DropdownButton` as well? It's a shared component between other things that use this "dropdown style" (like Popovers)

anniehu4 (changes requested):
Thanks David!! Using context makes sense to me 👍  And the `compact` change looks good -- requesting changes because I'm not sure the extra complication of the `optionsAlign / optionsWidth` props are worth it

packages/components/src/Dropdown/Dropdown.tsx
silvestry: Thanks for the review @anniehu4 .

So, users of compact variant will always need to style the options menu.

The first reason is that compact sizes to the length of the selected option text and then the options menu sizes according to that button width. So if options are: { 'abc', 'abcde', 'abcdefg' } where 'abc' is selected, the options menu would open up too narrow to display text. I suppose we might calculate the widest option and pass that back to size the button, but that seems overly complicated.

As for left and right aligned, it is the same problem as described above. If 'abc' is selected then the options menu will need to be wider and will stick out. Depending on context (ex: if the dropdown button is on the left of the comment card or to the right), we'll want the options menu to stick out to the left or to the right.

Given that, I like to have specific props that tells the developer they need to specify at least width if they selected compact. Also on a side note I was having trouble with `w-full` being applied to options and that seemed to take precedence over classname provided to Dropdown.Options.

That said, imo we can be specific with this variant, or maybe not add this variant as developer can already (I think) do this with composition and providing classname in Dropdown.Options. I'm on the fence on this and would appreciate your thoughts on this. Thanks!

anniehu4: hi @silvestry! sorry, catching up from HQ week still. thanks for explaining the behavior of this variant more. Some of this styling does seem a bit specific (especially for the right aligned options), so I think it's worth handling it here and saving some dev effort. Based on what you said, I'd consider:
- keeping the `optionsAlign` prop 👍 
- renaming `optionsWidth` to `optionsClassname`
    - the reasoning here is EDS should work independently of a consumer app using tailwind or not, so it's a little strange to specify that a prop must pass in a tailwind utility class. `optionsClassname` still would handle a utility class, but can also handle a regular className -- which may be helpful extra flexibility anyways. We can document the custom width use case in storybook + zeroheight, and I'm assuming devs are testing the changes they're making and would notice if custom width is needed. If you're really worried, we could throw a console warning if this prop isn't provided when `"compact"` is
    - a different option could be keeping it named `optionsWidth` but either:
        - taking a raw size value (like `"6rem"`), and passing that through to CSS
        - setting a union of possible values (may be annoying to maintain)

packages/components/src/Dropdown/Dropdown.tsx
```diff
@@ -48,6 +55,23 @@ type DropdownProps = ListboxProps & {
...
    */
   className?: string;
+  /**
+   * Render dropdown button that is only as wide as the content.
+   *
+   * When defining compact dropdown need to provide optionsWidth to
+   * define width of options menu, and optionally provide optionsAlign
+   * if desired to right align to dropdow button.
+   */
+  compact?: boolean;
```
anniehu4, line 65: _Should we make this a variant="compact" prop instead?_, I wondered to myself. Right now I think since this is only toggling sizing, it's fine as-is -- but if the "compact" style also had e.g. a background color change, then the `variant` prop could make sense

anniehu4: `size?: "compact" | "full"` could make sense though 🤔

packages/components/src/Dropdown/Dropdown.tsx
```diff
@@ -48,6 +55,23 @@ type DropdownProps = ListboxProps & {
    * Optional className for additional styling.
    */
   className?: string;
+  /**
+   * Render dropdown button that is only as wide as the content.
+   *
+   * When defining compact dropdown need to provide optionsWidth to
+   * define width of options menu, and optionally provide optionsAlign
+   * if desired to right align to dropdow button.
```
dierat, line 63: Got a little typo in there.
```suggestion
   * if desired to right align to dropdown button.
```

packages/components/src/Dropdown/Dropdown.tsx
```diff
@@ -228,7 +319,10 @@ function Dropdown(props: DropdownProps) {
   const trigger = buttonText && <DropdownTrigger>{buttonText}</DropdownTrigger>;
 
   const optionsList = options && (
-    <DropdownOptions>
+    <DropdownOptions
+    // optionsAlign={optionsAlign}
+    // optionsWidth={optionsWidth}
```
dierat, line 324: Are these lines supposed to be commented out here?

packages/components/src/Dropdown/Dropdown.tsx
```diff
@@ -165,6 +189,68 @@ function childrenHaveLabelComponent(children?: ReactNode): boolean {
...
+ *     compact
+ *     options={options}
+ *     optionsAlign="right"
+ *     optionsWidth="w-96"
+ *   />
+ * );
+ * ```
+ *
+ * For dropdown that differs in button and options menu width, style <Dropdown>
+ * with className for the button with and provide optionsWidth for the options
```
dierat, line 233: Should this say "button width"?
```suggestion
 * with className for the button width and provide optionsWidth for the options
```

dierat:
Sorry, late to the party. After reading through the conversation, I'm still not totally sure I understand the need for the `optionsWidth` prop. In playing with it locally, it kinda looked to me like the options menu would be fine for the `compact` version without specifying the width as long as we removed the `optionsFullWidth` styling, which we could do by tracking `compact` in the `DropdownContext` instead of the `optionsWidth`. Or maybe there's something else about the situation that I'm missing?

Test PR: https://github.com/chanzuckerberg/edu-design-system/pull/915

packages/components/src/Dropdown/Dropdown.tsx
silvestry: @anniehu4 : Cool. Sounds good. I've kept `optionsAlign` prop and renamed `optionsWidth` to `optionsClassName` (notice the casing... not sure what is the standard in EDS). I've also added check to ensure `optionsClassName` under the `compact` variant.

packages/components/src/Dropdown/Dropdown.tsx
silvestry: I've landed with `variant` because it is unclear by size if that meant the button or the whole component.

packages/components/src/Dropdown/Dropdown.tsx
silvestry: Thanks for catching. I am removing that particular documentation actually.

packages/components/src/Dropdown/Dropdown.tsx
silvestry: Nope! Good catch!

silvestry:
> Sorry, late to the party. After reading through the conversation, I'm still not totally sure I understand the need for the `optionsWidth` prop. In playing with it locally, it kinda looked to me like the options menu would be fine for the `compact` version without specifying the width as long as we removed the `optionsFullWidth` styling, which we could do by tracking `compact` in the `DropdownContext` instead of the `optionsWidth`. Or maybe there's something else about the situation that I'm missing?
> 
> Test PR: #915

@dierat : Yup yup! This is a good comment. I've taken Annie's advice and adopted`optionsClassName`. That said, you are right, it is not necessary to provide anything if we just remove `optionsFullWidth`. I'll update the documentation. Thanks!

packages/components/src/Dropdown/Dropdown.tsx
silvestry: Actually, given @dierat 's comment below, I'll remove the check as `optionsClassName` is not necessary with `compact`.

packages/components/src/Dropdown/Dropdown.module.css
silvestry: Instead of adding variant for DropdownButton, I just styled it on `DropdownTrigger`. Let me know if you think otherwise.

packages/components/src/Dropdown/Dropdown.tsx
```diff
@@ -287,8 +401,19 @@ const DropdownOptions = function (
...
+    useContext(DropdownContext);
+
   return (
-    <Listbox.Options className={clsx(styles.options, className)} {...rest} />
+    <Listbox.Options
+      className={clsx(
+        styles.options,
+        className,
+        optionsAlign === "right" && styles.optionsAlignRight,
+        optionsClassName || (!compact && styles.optionsFullWidth),
```
dierat, line 413: Mm, I think it's a bit odd that if you pass in an `optionsClassName` and you're not using the `compact` variant, the width will suddenly be off because the full width styling was removed. Maybe the concatenation method would work better here where `optionsClassName`, if it exists, is added to `styles.optionsFullWidth` if it's not in compact mode.

silvestry: Thanks for the comment @dierat . This is why I started with `optionsWidth` to distinguish between either user defined width or full width. If I concatenate, then the full width will take precedence when dev supplied optionsClassName with inline Tailwind (ex: `optionsClassName="w-96"`). Even if dev defined the style in a module, it is unclear which should take precedence, unless dev adds `!important`. I think this ambiguity is not that great. That said, I totally agree with you that it is somewhat opaque that providing `optionsClassName` will remove full width. But I suppose 1. I can add more documentation and 2. it is more visually evident no width is provided and that dev needs to add width to `optionsClassName`. What do you think @dierat @anniehu4 ?

dierat: Ah, I see. Maybe it would be enough to add a comment on the `optionsClassName` prop on line 72 explaining that if they're not using the `compact` variant, they will need to explicitly set the width in those styles?

silvestry: Sounds good. I've added documentation on line 72.

packages/components/src/Dropdown/Dropdown.module.css
dierat: I'm also wondering about whether it should be a prop, but I'm not super opinionated about it. I'm fine with waiting for a `Popover` to need a `compact` button before refactoring this.

dierat (approved):
I'm happy with this implementation 👍 Thanks for working through our concerns! ❤️

silvestry:
@anniehu4 : If I can get your re-review and approval. Thank you!

packages/components/src/Dropdown/Dropdown.tsx
```diff
@@ -197,7 +284,11 @@ function Dropdown(props: DropdownProps) {
   }
 
   const sharedProps = {
-    className: clsx(styles.dropdown, className),
+    className: clsx(
+      styles.dropdown,
+      className,
+      variant === "compact" && styles.compactDropdown,
```
anniehu4, line 290: nit: could be
```suggestion
      compact && styles.compactDropdown,
```

packages/components/src/Dropdown/Dropdown.tsx
```diff
@@ -249,9 +344,25 @@ function Dropdown(props: DropdownProps) {
     </>
   );
 
-  return <Listbox {...sharedProps}>{childrenToUse}</Listbox>;
+  const contextValue = Object.assign(
```
anniehu4, line 347: nit: should we re-use this for the case at line 301?

anniehu4 (approved):
thanks so much for the iteration here ❤️

packages/components/src/Dropdown/Dropdown.tsx
silvestry: Good catch. Thanks Annie!

PR #909, opened 2022-03-25
chore(list detail): adding variants

csnizik:
### Summary:
Shortcut: https://app.shortcut.com/czi-edu/story/189947/list-detail-component-sequence-variant

This PR adds a 'sequence' variant to the List Detail Component; sequence variant adds icons for 'success', 'warning' and 'error' and a vertical line to the list detail component's unordered list;

NOTES:
- The colors need to be replaced
- The line between list elements has vert space above and below it, but there is irregularity when the text in adjacent list elements are varying heights (see screenshot below)
- I fought the linter and the linter won. I'm open to suggestions on how to refactor those CSS selectors (src/components/ListDetail/ListDetail.module.css:107, 161)

Screen Shot 2022-03-24 at 10 59 06 PM

### DIFF
--- a/.storybook/pages/ProjectOverview/ProjectOverview.stories.tsx
+++ b/.storybook/pages/ProjectOverview/ProjectOverview.stories.tsx
@@ -10,5 +10,5 @@ export default {
 
 const Template: Story<Props> = (args) => <ProjectOverview {...args} />;
 
-export const Default = Template.bind({});
-Default.args = {};
+export const StudentView = Template.bind({});
+StudentView.args = {};
--- a/.storybook/pages/ProjectOverview/ProjectOverview.stories.tsx
+++ b/.storybook/pages/ProjectOverview/ProjectOverview.stories.tsx
@@ -12,8 +12,3 @@ const Template: Story<Props> = (args) => <ProjectOverview {...args} />;
 
 export const Default = Template.bind({});
 Default.args = {};
-
-export const Sequence = Template.bind({});
-Sequence.args = {
-  listVariant: 'sequence'
-}
--- a/.storybook/pages/ProjectOverview/ProjectOverview.stories.tsx
+++ b/.storybook/pages/ProjectOverview/ProjectOverview.stories.tsx
@@ -10,5 +10,10 @@ export default {
 
 const Template: Story<Props> = (args) => <ProjectOverview {...args} />;
 
-export const StudentView = Template.bind({});
-StudentView.args = {};
+export const Default = Template.bind({});
+Default.args = {};
+
+export const Sequence = Template.bind({});
+Sequence.args = {
+  listVariant: 'sequence'
+}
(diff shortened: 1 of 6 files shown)


src/components/ListDetail/ListDetail.tsx
```diff
@@ -193,33 +201,44 @@ export const ListDetail: React.FC<Props> = ({
...
-        <ul className={styles['list-detail__list']} role="tablist">
+        <ul
+          className={clsx(styles['list-detail__list'], {
+            [styles['list-detail__list--sequence']]: variant === 'sequence',
+          })}
+          role="tablist"
+        >
           {listDetailItems().map((tab: any, i: any) => {
             const isActive = activeIndexState === i;
+            const itemVariant = variant && tab.props.variant;
```
frostyweather, line 232: @csnizik What was your reasoning around changing this to a `const`? Just curious if it was more straightforward of a name or something. Thanks!

src/components/ListDetail/ListDetail.tsx
```diff
@@ -193,33 +201,44 @@ export const ListDetail: React.FC<Props> = ({
...
-
-  const componentClassName = clsx(styles['list-detail'], className, {
-  });
+  const componentClassName = clsx(styles['list-detail'], className, {});
 
   return (
     <div className={componentClassName} {...other}>
       <div className={styles['list-detail__header']}>
-        <ul className={styles['list-detail__list']} role="tablist">
+        <ul
```
frostyweather, line 224: Let's allow for the `ordered` variant to render an `ol` instead of a `ul`

.storybook/pages/ProjectOverview/ProjectOverview.stories.tsx
```diff
@@ -10,5 +10,10 @@ export default {
...
 const Template: Story<Props> = (args) => <ProjectOverview {...args} />;
 
-export const StudentView = Template.bind({});
-StudentView.args = {};
+export const Default = Template.bind({});
+Default.args = {};
+
+export const Sequence = Template.bind({});
+Sequence.args = {
+  listVariant: 'sequence'
```
frostyweather, line 18: Change this to `Ordered` instead of `Sequence` when you make that prop change. Thanks!

frostyweather (changes requested):
This all looks good! Just a few things:
- Adding numbers
- Changing `sequence` to `ordered` and rendering an `ol` instead of a `ul` in that instance.

Maybe another comment or two. Thanks for this!

src/components/ListDetail/ListDetail.tsx
csnizik: Did that to avoid having to add another condition to the logic for classNames; I originally had double conditionals on the li classNames to check if variant was sequence and if tab.props.variant had a value, which looked awkward. I used this way so that `itemVariant` will only hold a value if both conditions are true (list === sequence && this tab's variant is set)

PR #912, opened 2022-03-28
Ifrost/global header

ifrost1:
### Summary:

- Get the general structure and responsiveness of the header in place
- Get the `ListDetail` and `ListDetailPanel` structure in place (sidebar tabs) for project overview
- Hide icon on last `BreadcrumbsItem`
- Put page header in place on the project overview page

### Test Plan:

`yarn test` passes

`yarn lint` passes

`yarn start` runs

`yarn build` runs properly


anniehu4 (approved):

src/components/Header/Header.tsx
```diff
@@ -21,15 +39,119 @@ export interface HeaderProps {
 }
 
 export const Header: React.FC<HeaderProps> = ({
+  behavior,
   className,
   children,
+  pinnedStop = 152,
```
bradfrost, line 45: @ifrost1 Not sure why 152 specifically, but if it's not needing that level of accuracy may I suggest a rounder number?
```suggestion
  pinnedStop = 150,
```

src/components/AvatarBlock/AvatarBlock.module.css
```diff
@@ -0,0 +1,25 @@
...
+  border: none;
+  appearance: none;
+  color: inherit;
+  cursor: pointer;
+}
+
+.avatar-block__body {
+  flex: 1;
+  white-space: nowrap;
+}
```
bradfrost, line 25: For visibility, the avatar work is largely happening elsewhere and will be finalized in another PR. Since this is just getting the header structure in place we'll leave this be for now.

.storybook/recipes/GlobalHeader/GlobalHeader.tsx
```diff
@@ -26,42 +26,56 @@ export const GlobalHeader: React.FC<Props> = ({ className, ...other }) => {
...
-          inverted={true}
-          onClick={toggleMenu}
-        />
-      </div>
+      <Button
+        className={styles['global-header__menu-button']}
+        variant="bare"
+        text={isActive ? 'Close' : 'Menu'}
+        iconPosition="before"
+        hideText={true}
```
anniehu4, line 48: @ifrost1  instead of adding the `hideText` prop (which we've already decided to remove in https://github.com/chanzuckerberg/edu-design-system/pull/889/files#r822192445) -- could this be
```suggestion
        aria-label={isActive ? 'Close' : 'Menu'}
        iconPosition="before"
```
? I don't think this is dependent on Diedra's upcoming button work

dierat (changes requested):
Requesting changes for the `hideText` prop addition to `Button`.

bradfrost (approved):

.storybook/recipes/GlobalHeader/GlobalHeader.tsx
```diff
@@ -25,42 +26,55 @@ export const GlobalHeader: React.FC<Props> = ({ className, ...other }) => {
...
-          iconPosition="before"
-          iconName={isActive ? 'x' : 'menu'}
-          inverted={true}
-          onClick={toggleMenu}
-        />
-      </div>
+      <Button
+        className={styles['global-header__menu-button']}
+        variant="bare"
+        aria-label={isActive ? 'Close' : 'Menu'}
```
bradfrost, line 46: @anniehu4 Ah, I _knew_ there was a reason why we didn't use `aria-label` , but found the article: https://heydonworks.com/article/aria-label-is-a-xenophobe/

I'm not sure how true that post is (it's from 2018, so things could have improved), but at least now you know why we had things set up the way we had them.

dierat: Oh interesting! I hadn't heard about that. Thanks for the link!

dierat (approved):

PR #910, opened 2022-03-28
Bvoran/sv next

bgreater:
### Summary:

Adds commitizen and standard version to project.

### Test Plan:

TK

PR #914, opened 2022-03-28
Test publish with standard-version

anniehu4:
### Summary:

Builds on https://github.com/chanzuckerberg/edu-design-system/pull/910, and:
- removes commit hook for now (we can turn it back on once the repo is a bit more stable)
- uses `"files"` to allowlist instead of `.npmignore`'s denylist
- updates publishing docs

### Test Plan:
- `yarn commit` uses commitizen:
Screen Shot 2022-03-28 at 6 00 12 PM
- confirm output of `yarn release:alpha`:
Screen Shot 2022-03-28 at 6 01 30 PM
- test published to https://www.npmjs.com/package/@chanzuckerberg/eds/v/0.1.0-alpha.2


package.json
```diff
@@ -9,6 +9,9 @@
   "publishConfig": {
     "access": "public"
   },
+  "files": [
+    "/dist"
```
anniehu4, line 13: this is more explicit than having to denylist all the random config files

docs/PUBLISHING.md
```diff
@@ -12,11 +12,32 @@ Look to [this helpful document](http://designsystem.morningstar.com/about/versio
...
+We use [standard-version](https://github.com/conventional-changelog/standard-version) to increment the version number in `package.json`, create a git tag for the new release, and update `CHANGELOG.md` based on the commit log. The package is not published, yet.
+
+4. Run the last command output by `standard-version`. It will look something like: `git push --follow-tags origin <branch> && npm publish`.
+5. Communicate the changes via all appropriate channels. <TODO: determine communication strategy for new release>
+
+**Before the first time you publish**, make sure to:
+
+- set up Two Factor Authentication for your npm account
+- run `npm login` in your terminal to generate an access token for publishing
+
```
dierat, line 35: Could we move this above the main instructions? I actually ran into this at the bee workshop where they had the hot tips after the main instructions, so I didn't read them until I was done with that section and had found out the hard way.

anniehu4 (commented):
test installed in `traject`: https://github.com/FB-PLP/traject/tree/ahu/eds-next-alpha-1

PR #915, opened 2022-03-28
chore(dropdown): remove optionsWidth

dierat:
### Summary:
Just a test PR to see if we can get away with not using a specified width for the options menu when using a compact dropdown button. The changes to the options list in `.stories` is just to make it easier to verify that the options menu width is working as expected.

Screen Shot 2022-03-28 at 3 33 34 PM


dierat:
This was just an example that I didn't intend to ship; closing.

PR #916, opened 2022-03-29
chore(components): add index.ts files

csnizik:
### Summary:
Associated Shortcut ticket: https://app.shortcut.com/czi-edu/story/188968/add-index-ts-file-to-each-component

This PR:

1.  Adds an `index.ts` file to each component directory. Each `index.ts` file simply imports the exported component and exports it as the default export
1.  Changes import statements to use the default export where needed
1.  Adds an `index.ts` to `plop-templates/Component`

### Test Plan:

- All exported modules should remain unbroken with their bindings still intact. This can be checked within Storybook.
- When creating a new Component via plop, the new directory structure should be created.


src/components/Badge/Badge.tsx
```diff
@@ -20,7 +20,7 @@ export interface Props {
    * - **teal** results in a badge with teal state
    * - **dark-gray** results in a sparkline text bubble with normal results
    */
-  variant: 'warning' | 'error' | 'success';
+  variant?: 'warning' | 'error' | 'success';
```
csnizik, line 23: I changed this because of a TS warning I saw when I opened this file; it doesn't relate to the import work in this PR and can be moved to a separate PR if requested.

src/components/Table/Table.tsx
```diff
@@ -8,7 +8,7 @@ export interface Props {
    * - **overflow-sm** Removes the min-width from the table for tables with few columns
    * - **overflow-lg** Used for tables with lots of table rows to overflow at a larger breakpoint
    */
-  behavior?: 'overflow';
+  behavior?: 'overflow' | 'overflow-lg' | 'overflow-sm';
```
csnizik, line 11: I changed this because of a TS warning I saw when I opened this file; it doesn't relate to the import work in this PR and can be moved to a separate PR if requested.

dierat:
In `main`, we're using `export default` in the component file and `export { default } from "./ComponentFile";` whereas this PR is not exporting default in the file and using `export { Component as default } from "./ComponentFile";`. I'm personally unopinionated about which method we use, but we should probably pick one to be consistent about it. @chanzuckerberg/edu-design-system what do you think?

src/components/Card/Card.stories.tsx
```diff
@@ -1,13 +1,13 @@
...
 import React from 'react';
-import { Card } from './Card';
-import { Button } from '../Button/Button';
-import { ButtonGroup } from '../ButtonGroup/ButtonGroup';
-import { CardBody } from '../CardBody/CardBody';
-import { CardFooter } from '../CardFooter/CardFooter';
-import { CardHeader } from '../CardHeader/CardHeader';
-import { Heading } from '../Heading/Heading';
-import { TextPassage } from '../TextPassage/TextPassage';
+import { Card, Props } from './Card';
```
dierat, line 3: Sorry, where is `Props` being used?

src/components/Table/Table.tsx
dierat: Nice, thanks!

anniehu4 (commented):
sweet, thanks! looks good to approve besides the `Tooltip` stories change

> In main, we're using export default in the component file and export { default } from "./ComponentFile"; whereas this PR is not exporting default in the file and using export { Component as default } from "./ComponentFile";. I'm personally unopinionated about which method we use, but we should probably pick one to be consistent about it. https://github.com/orgs/chanzuckerberg/teams/edu-design-system what do you think?

@dierat I feel fine going with this approach; kind of nice to not type an extra `export default` line. I think the main difference is we'll still have to use the named import in stories & tests, which the `plop` templates take care of

@csnizik could we also update `src/index.ts`? this could be a follow-up. thinking
```
// old
export { Accordion } from './components/Accordion/Accordion';

// becomes
export { default as Accordion } from './components/Accordion';
```

dierat:
> @dierat I feel fine going with this approach; kind of nice to not type an extra export default line. I think the main difference is we'll still have to use the named import in stories & tests, which the plop templates take care of

Cool, then I think there's just one `export default` we've added that can be taken out: https://github.com/chanzuckerberg/edu-design-system/blob/next/src/components/Tooltip/Tooltip.tsx#L123

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -3,7 +3,7 @@ import { within, userEvent } from '@storybook/testing-library';
 import clsx from 'clsx';
 import React from 'react';
 import Tooltip from './Tooltip';
-import styles from './Tooltip.stories.module.css';
```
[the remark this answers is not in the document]
anniehu4: @csnizik can you revert this change? It ensures certain Tooltip stories show up correctly

src/components/Card/Card.stories.tsx
csnizik: Good catch

src/components/Tooltip/Tooltip.stories.tsx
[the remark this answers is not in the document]
csnizik: @dierat @annihu4 I made the change to quiet a TypeScript warning: "Cannot find module './Tooltip.stories.module.css' or its corresponding type declarations." Is that a misfiring warning? If so, what change needs to be made to quiet it?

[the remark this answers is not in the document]
anniehu4: and here's a PR to fix the typechecking 🙂 https://github.com/chanzuckerberg/edu-design-system/pull/925

anniehu4 (approved):

dierat:
Weird test failure about `Legend` not having a default export even though you updated that one 🤔 
https://github.com/chanzuckerberg/edu-design-system/runs/5777045113
>Run yarn run types
Error: src/index.ts([4](https://github.com/chanzuckerberg/edu-design-system/runs/5777045113#step:5:4)8,10): error TS230[5](https://github.com/chanzuckerberg/edu-design-system/runs/5777045113#step:5:5): Module '"./components/Legend/Legend"' has no exported member 'default'.
Error: Process completed with exit code 2.

dierat:
Mmm student overview's on the struggle bus
https://www.chromatic.com/build?appId=61313967cde49b003ae2a860&number=709
https://61313967cde49b003ae2a860-yistlvbijb.chromatic.com/?path=/story/pages-projects-overview--student-view
Screen Shot 2022-03-31 at 12 54 01 PM

PR #918, opened 2022-03-29
Ifrost/project overview content stubbing

ifrost1:
### Summary:

- Add existing `TextPassage`, `Card`, `Heading`, and `Section` components to the project overview page
- Add ability to make a `Card` an `a` tag and add hover styles.
- Add `titleBefore` slot for image/avatar to go in before the `Section` title

### Test Plan:

`yarn test` passes

`yarn lint` runs clean

`yarn start` spins up properly.

`yarn build` works


src/components/Section/Section.tsx
```diff
@@ -24,6 +24,10 @@ export interface Props {
    * Section heading text string
    */
   title?: string;
+  /**
+   * Section tile before container to place items like images or avatars
```
anniehu4, line 28: nit
```suggestion
   * Section title before container to place items like images or avatars
```

.storybook/pages/ProjectOverview/ProjectOverview.tsx
```diff
@@ -18,33 +19,53 @@ export const ProjectOverview: React.FC = () => (
...
+						<li>How are structures in the body organized? More text blah blah blah.</li>
+						<li>How are structures in the body organized?</li>
+					</ol>
+					</TextPassage>
+        <Hr className={utilityStyles['u-margin-bottom-none']}  />
+        <Section titleBefore={<img style={{ display: 'block', borderRadius: '50%'}} src="https://via.placeholder.com/32x32" alt="placeholder" />} title="Power Focus Areas">
+          <Card href="#">
+						<CardBody>
+							<Heading className={utilityStyles['u-margin-bottom-sm']} as="h3" size={4}>What Was Medieval Japan Like?</Heading>
+							<TextPassage>In this Focus Area, you explore the main question: How does the structures of organisms enable life's functions?</TextPassage>
```
anniehu4: looks like lint is unhappy about these single quotes https://github.com/chanzuckerberg/edu-design-system/runs/5741773933?check_suite_focus=true

you could disable the rule for this whole file

anniehu4 (approved):

src/components/Button/Button.tsx
```diff
@@ -24,6 +25,10 @@ export interface Props {
    * Toggles button that fills the full width of its container
    */
   fullWidth?: boolean;
+  /**
+   * Visually hides button text for icon-only buttons for accessibility
+   */
+  hideText?: boolean;
```
dierat, line 31: We explicitly removed `hideText` in https://github.com/chanzuckerberg/edu-design-system/pull/889 because you can use `aria-label` instead of `text` to get the same effect.

src/components/Card/Card.tsx
```diff
@@ -24,14 +28,18 @@ export const Card: React.FC<Props> = ({
...
   children,
   inverted,
+  href,
   ...other
 }) => {
   const componentClassName = clsx(styles['card'], className, {
     [styles['card--inverted']]: inverted,
   });
+
+  const TagName = href ? 'a' : 'div';
```
dierat, line 38: Should we be concerned at all about `Card`s having interactive elements inside `children`? Also a link probably shouldn't have that much text for screen readers.

The way we're handling this in `traject` is we have an `onClick` on the card that correlates to some link in the card's contents (usually the header text). https://github.com/FB-PLP/traject/blob/80fa0f387e3c510dd4d070b1210269a1dd74297f/app/assets/javascripts/v2/core/EDSCandidates/Card/Card.tsx#L74 This is from the "redundant click solution" in the [inclusive components Card post](https://inclusive-components.design/cards/).

src/components/Header/Header.module.css
```diff
@@ -5,36 +5,74 @@
...
+  }
+}
+
+/**
+* Sticky header with scrolled class applied
+* 1) Class is applied when the sticky header gets pinned/unpinned and animation is needed
+*/
+.header--sticky.eds-is-scrolled {
+  @media all and (max-width: $eds-bp-lg) {
+    transition: transform var(--eds-anim-fade-quick) var(--eds-anim-ease);
```
dierat, line 65: Can we get a [prefers-reduced-motion](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) check here?
```suggestion
    transition: transform var(--eds-anim-fade-quick) var(--eds-anim-ease);
    
    @media (prefers-reduced-motion) {
      transition: none;
    }
```

src/components/Header/Header.module.css
```diff
@@ -5,36 +5,74 @@
...
   }
 }
 
-.header__inner {
-  display: flex;
-  align-items: center;
-  gap: var(--eds-size-2);
+/**
+ * Container around sticky header
+ * 1) This is added to
```
dierat, line 30: Are we missing the end of the sentence here by any chance?

src/components/Header/Header.module.css
```diff
@@ -5,36 +5,74 @@
...
+ */
+.header-sticky-container {
+  height: 100vh;
+  transition: all var(--eds-anim-move-quick) var(--eds-anim-ease);
```
dierat, line 34: Should we be transitioning `all`? Could we target the specific properties instead?

.storybook/recipes/GlobalHeader/GlobalHeader.module.css
```diff
@@ -6,58 +6,48 @@
  * 1) 
  */
 .global-header {
+  position: relative;
   margin-bottom: var(--eds-size-4);
 
+  &:hover,
+  &:focus-within {
+    @media all and (min-width: $eds-bp-lg) {
```
dierat, line 14: Just curious – why are we using `all`? Isn't that the default? Do we need to specify it?

.storybook/recipes/GlobalHeader/GlobalHeader.module.css
```diff
@@ -6,58 +6,48 @@
...
-    margin-bottom: 0;
+    position: fixed;
     display: flex;
     flex-direction: column;
-    max-width: var(--eds-l-sidebar-width);
+    align-items: flex-start;
+    overflow: hidden;
+    margin-bottom: 0;
+    width: 3.75rem;
+    transition: width var(--eds-anim-move-quick) var(--eds-anim-ease);
```
dierat, line 28: Prefers reduced motion again. Also, should we be concerned about the performance of transitioning the `width` property?

src/components/AvatarBlock/AvatarBlock.tsx
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * URL or a URL fragment string that the hyperlink points to
+   */
+  href?: string;
+}
+
+/**
+ * Primary UI component for user interaction
+ */
+export const AvatarBlock: React.FC<Props> = ({
```
dierat, line 25: I believe we're moving away from using `React.FC` https://github.com/chanzuckerberg/edu-design-system/pull/919

src/components/Layout/Layout.module.css
```diff
@@ -10,5 +10,67 @@
...
+    width: 3.625rem;
+  }
+}
+
+/**
+ * Sidebar layout section in expandable layout
+ */
+.layout--expandable .layout__section--sidebar {
+  @media all and (min-width: $eds-bp-lg) {
+    transition: width var(--eds-anim-move-quick) var(--eds-anim-ease);
```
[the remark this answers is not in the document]
bradfrost: @dierat One of the many reasons that "move everything on hover" is maybe not the best treatment. This has been removed elsewhere so resolving it here.

PR #919, opened 2022-03-29
chore(components): replace React.FC

csnizik:
### Summary: 
Shortcut ticket: https://app.shortcut.com/czi-edu/story/189120/sweep-through-the-existing-components-and-implement-the-changes-to-code-guidelines

This PR:
1. Switches existing components away from using React.FunctionalComponent (React.FC)
2. Changes the plop template for a Component to match
3. Adjusts the example for typing useRef() in CODE_GUIDELINES to match the places in the codebase where it is already being used
4. Adds `import { MutableRefObject } from 'react';` in files where it was missing

### Test Plan:
Confirm that TypeScript does not produce warnings or errors in the project.


docs/CODE_GUIDELINES.md
```diff
@@ -407,9 +406,8 @@ import React, { useRef, MutableRefObject, RefObject } from 'react';
 
 ...
 
-const mutableRef: MutableRefObject<HTMLDivElement | null> = useRef<HTMLDivElement | null>(null);
+const ref = useRef() as MutableRefObject<HTMLInputElement>;
```
anniehu4, line 409: thanks for updating the docs to match existing examples 👍 Can we remove these sentences from the paragraph at line 399 now?

> For mutable refs, use a union with `null` when initializing the ref; otherwise, TypeScript will complain when you change the value of `.current`. Similarly, *do not* create a union with `null` for non-mutable refs. 

also, are there use cases of the non-mutable `RefObject` currently?

anniehu4 (approved):

dierat (approved):

docs/CODE_GUIDELINES.md
csnizik: re: are there any non-mutable instances now? I don't know if there are or not

PR #920, opened 2022-03-29
docs(code guidelines): revert change re: state

csnizik:
### Summary:
This PR proposes removing the section in CODE_GUIDELINES that was recently added re: the typing of state. Rather than refactor components that aren't broken, I amended the `useState()` section to suggest the pattern of initializing state with a prop when explicit typing is necessary (and allowing TypeScript's type inference to handle the rest of the cases.)

### Test Plan:
Confirm that the suggested patterns conform with best practices.


anniehu4 (approved):

dierat (approved):
Thanks!

PR #921, opened 2022-03-30
chore(components): refactor conditionals

csnizik:
### Summary:
Shortcut: https://app.shortcut.com/czi-edu/story/188969/update-conditional-to-use-ampersands-instead-of-object-literals

This PR contains work that refactors conditionals used as arguments passed to the `clsx` utility to determine classnames. Specifically, this pattern:
`{[styles['component--style']]: prop === value}`

is replaced with this pattern:
`prop === value && styles['component--style']`

### Test Plan:
All refactored arguments should match the conditions in the original arguments; no regressions should be seen, i.e. missing or improperly assigned styles on components.


anniehu4 (approved):
thanks! are any changes needed to the [Component plop file](https://github.com/chanzuckerberg/edu-design-system/blob/next/plop-templates/Component/Component.tsx.hbs)?

src/components/Hr/Hr.tsx
```diff
@@ -24,10 +24,12 @@ export interface Props {
...
 export const Hr: React.FC<Props> = ({ className, size, variant, ...other }) => {
-  const componentClassName = clsx(styles['hr'], className, {
-    [styles['hr--lg']]: size === 'lg',
-    [styles['hr--brand']]: variant === 'brand',
-  });
+  const componentClassName = clsx(
+    styles['hr'],
+    className,
+    size === 'lg' && styles['hr--lg'],
+    size === 'brand' && styles['hr--brand'],
```
anniehu4, line 31: I think this should be
```suggestion
    variant === 'brand' && styles['hr--brand'],
```
and the prop type doesn't need to change because there's already a `variant` prop

src/components/Banner/Banner.tsx
```diff
@@ -46,7 +46,7 @@ export interface Props {
    * - **error** - results in a red banner
    * - **info** - results in a blue banner
    */
-  variant?: 'success' | 'warning' | 'error' | 'info' | 'brand';
+  variant?: 'success' | 'warning' | 'error' | 'info' | 'brand' | 'vertical';
```
dierat, line 49: I don't think we want `vertical` to be a type of `variant` because a banner can be vertical and also have one of the other variants (success, warning, etc).

src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -18,7 +18,7 @@ export interface Props {
    * - **responsive** results in a ButtonGroup that stacks on small screens, but displays side by side on larger screens
    * - **stacked** results in a ButtonGroup that stacks on all screen sizes
    */
-  behavior?: 'responsive' | 'stacked';
+  behavior?: 'responsive' | 'stacked' | 'right';
```
dierat, line 21: Can a button group be stacked and also aligned right?

src/components/TextField/TextField.tsx
```diff
@@ -189,11 +189,13 @@ export const TextField = ({
...
-    [styles['text-field--inverted']]: inverted === true,
-    [styles['eds-is-error']]: isError,
-    [styles['eds-is-disabled']]: disabled,
-  });
+  const componentClassName = clsx(
+    styles['text-field'],
+    className,
+    inverted && styles['text-field--inverted'],
+    isError && styles['eds-is-error'],
+    disabled && styles['eds-is-disabled'],
```
dierat, line 197: I'm just now realizing it's kind of weird that we use `disabled` and not `isDisabled` 🤔

src/components/ButtonGroup/ButtonGroup.tsx
bradfrost: @csnizik @dierat yeah that's right, the `align` prop should be used to align something to the right.

src/components/Banner/Banner.tsx
bradfrost: @dierat Yep I agree. Hey, this is a good exercise! May I recommend capturing these as tickets or something so we can address that cleanup independently of this PR? This PR is just flipping things around, but I think it makes sense to use this as an audit exercise and tackle any issues outside of this PR.

src/components/ButtonGroup/ButtonGroup.tsx
bradfrost: @dierat Same note as above here! https://github.com/chanzuckerberg/edu-design-system/pull/921/files#r838967593

Let's capture these things as tickets and address those things we catch in other work.

src/components/TextField/TextField.tsx
bradfrost: @dierat Yeah I've thought about that conundrum before too! It is weird that it's not more explicit. Ultimately, I default to `disabled` simply because that's what the native HTML attribute is. I don't know if that's good or bad but that's what I think of anyways!

bradfrost (approved):
@csnizik Great work!

bradfrost:
Merging this in since it gets things in the proper format. @dierat @anniehu4 you both raised some great points on some of the component API! Let's capture those as items to tackle and work through as we touch these component builds.

src/components/Banner/Banner.tsx
anniehu4: I'd generally agree with making tickets for pre-existing issues. I think in this case though there was a typo below:
```
[styles['banner--vertical']]: orientation === 'vertical',
```
turned into
```
variant === 'vertical' && styles['banner--vertical'],
```
, where `variant` should have been `orientation` if things were properly flipped. I bet the props change was to resolve the resulting TS issue (and it's probably what happened in the `align` case as well)

anniehu4:
Ticket here: https://app.shortcut.com/czi-edu/story/190821/styling-prop-cleanup

I assigned it to @csnizik for now but we can re-assign if needed

src/components/Banner/Banner.tsx
csnizik: @anniehu4 Yes, this was the result of a typo. I inadvertently changed:

`[styles['banner--vertical']]: orientation === 'vertical'` 

into

`variant === 'vertical' && styles['banner--vertical']`

which then generated a TypeScript error:

`This condition will always return 'false' since the types '"success" | "warning" | "error" | "info" | "brand"' and '"vertical"' have no overlap.`

Rather than realize that I had a typo in the refactored logic, I mistakenly corrected the TS error by piping "vertical" into the type union assigned to `variant`. My bad.

PR #923, opened 2022-03-30
chore(button): rename 'text' prop to 'children'

dierat:
### Summary:
Following up on our EDS eng sync conversation on 3/29, this PR renames the `text` prop in the `Button` component to `children` to be consistent with what the users of this design system expect.

### Test Plan:
Spot check storybook, especially the docs changes since automated tests won't catch typos in there.

### DIFF
--- a/docs/CODE_GUIDELINES.md
+++ b/docs/CODE_GUIDELINES.md
@@ -291,7 +291,6 @@ The design system's component directory contains all of the design system's comp
 
 ### Imports
 
-
 The framework follows a specific ordering/clustering for importing modules into a component. This is enforced through the [`import/order` lint rule](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md).
 
 Here's an example:
@@ -339,6 +338,7 @@ export const ComponentName = ({
 This defines the component name and passes in all the `Props`.
 
 ### Children
+
 When a component uses `children` as a prop, use the type `React.ReactNode` unless context dictates otherwise, including `ReactNode` as a named import.
 
 ```tsx
@@ -356,16 +356,18 @@ export const ComponentName = ({ children }: { children: ReactNode }) => {
 Interactive components likely require defining necessary [state and lifecycle](https://reactjs.org/docs/state-and-lifecycle.html) functions in addition to defining any other necessary variables and functions.
 
 #### useState()
+
 When using the `useState()` hook, TypeScript will infer the correct type from the initial value. If explicit typing of a variable in state is necessary and a hard-coded initial value is not set, you can use a prop as the initial value.
 
 #### useEffect()
+
 `useEffect()` hooks do not require any typing. TypeScript expects them to either return nothing or a Destructor-typed function (a function that cleans up any side effects and returns void.)
 
 #### useRef()
-`useRef()` hooks access underlying DOM elements to perform imperative actions. The resulting `ref` object can either be *mutable* or *not mutable*. (If the value store in its' `.current` property may be changed, the ref needs to be `mutable`.) Explictly convey the intended mutability status of each `ref` by using either `React.MutableObject` and `React.RefObject`, in a generic type definition. 
+
+`useRef()` hooks access underlying DOM elements to perform imperative actions. The resulting `ref` object can either be _mutable_ or _not mutable_. (If the value store in its' `.current` property may be changed, the ref needs to be `mutable`.) Explictly convey the intended mutability status of each `ref` by using either `React.MutableObject` and `React.RefObject`, in a generic type definition.
 
 For consistency, include `MutableRefObject` and `RefObject` with your import statements.
-
 
 ```tsx
 import React, { useRef, MutableRefObject, RefObject } from 'react';
@@ -375,6 +377,7 @@ import React, { useRef, MutableRefObject, RefObject } from 'react';
 const ref = useRef() as MutableRefObject<HTMLInputElement>;
 
 ```
+
 ### Define `componentClassName`
 
 The last thing that appears above the `return` statement is the `componentClassName`, which defines the CSS block for the component in addition to any modifier CSS class names using the [`clsx` library](https://www.npmjs.com/package/clsx). CSS Modules' bracket syntax (e.g. `styles['my-component']`) is used to enable BEM conventions (which uses dashes).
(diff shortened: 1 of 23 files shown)


docs/CODE_GUIDELINES.md
```diff
@@ -480,7 +480,7 @@ EDS adheres to the following API naming conventions:
 
 ### Text, Labels, Titles
 
-- Default to `text` for short strings of text, such as `<Button text="Click here">` or `<Badge text="Overdue" />`.
+- Default to `text` for short strings of text, such as `<BreadCrumbsItem text="My Courses">` or `<Badge text="Overdue" />`.
```
dierat, line 483: I just picked another component that still uses `text`. I dunno if we want to add something in the docs about the `children` paradigm, which is now being used by `Button` and `Text`.

src/components/Button/Button.tsx
```diff
@@ -5,7 +5,7 @@ import { Icon } from '../Icon/Icon';
 export interface Props {
   /**
    * Visually hidden button text (but text is still accessible to assistive technology).
-   * This overrides `text`
+   * This overrides `children` for screen readers
```
dierat, line 8: Added "for screen readers" because I thought this might be confusing for people who are new to accessibility.

docs/CODE_GUIDELINES.md
csnizik: @dierat if I understand correctly, your concern stems from the way in which the Button component handles its contents; i.e. it is being switched from this pattern:

`<Button>Press me</Button>`

to this pattern:

`<Button text="Press me" />`

and not just the prop name itself being `text`. Basically now you're taking a component that held content between its opening and closing tags and making devs relearn it as a self-closing tag that only accepts named props.

JSX (and by extension, TypeScript) reserve the word "children" to be a catch-all term for... the kitchen sink. Every React component has access to `props.children` containing the contents of anything between the opening and closing tags. Unless you specify otherwise, TypeScript makes assumptions that children can be anything, including another React component. Is that the intention? Or do the contents of the button component need to be constrained to a text string only? I'm assuming the latter, but I shouldn't assume.

csnizik: I should have added: one of the changes that I put in the TypeScript section of CODE_GUIDELINES was to move components away from using `React.FC` as their source of typing, and the reasoning behind that change was actually to reign in the free-for-all use of `children`. Components that are typed with `React.FC` receive implicit permission to receive `children`... they are typed for you as `ReactNode` even if you don't explicitly include them in your explicit typing of Props. Taking that off of the components forces us to now explicitly type `children`, which is what you would need to do as part of this ticket, and going forward, devs would have to be reminded that `children` on buttons is treated differently and needs to be typed as a string.

anniehu4: I think it'd be good to include something about `children` and when we'd prefer that

@csnizik great points 🙂 Our main motivation is that we _want_ to allow more than just strings in buttons. We've seen cases in the Learning Platform where the extra flexibility is needed, and we're concerned that making the API tighter (to only allow strings) would hinder adoption. So we're okay with the more relaxed typing of `ReactNode` here

csnizik: @anniehu4 so for instance if someone wants a button with an icon instead of text, or an icon positioned before/after a text string, you want them to have the flexibility to do that? I'm not sure that the flexibility of allowing children of type `ReactNode` would give them the results you want style-wise. i.e., how do you set up button styles to properly display any component or html tag that gets passed in?

anniehu4: yep, exactly! That's how our current edu-design-system button does it. We set up some padding / icon sizing styles to ensure things look good by default ([see code file here](https://github.com/chanzuckerberg/edu-design-system/blob/main/packages/components/src/ClickableStyle/ClickableStyle.module.css#L16)) -- @dierat probably knows better the exact relevant styles, if you're curious.

We have a good amount of usage of the Buttons in the platform now and haven't run into styling complaints yet, so we feel pretty good about this approach. Will note that all the usage has been passing a string or an icon with a string -- so we don't anticipate people doing things that crazy with it, but do want to allow for the existing usages

anniehu4:
looks good, let's make sure to add some docs about when we're using `children`

src/components/Button/Button.tsx
```diff
@@ -33,9 +33,9 @@ export interface Props {
    */
   iconName?: string;
   /**
-   * Determines position of icon relative to button text.
-   * - **before** places icon before button text
-   * - **after** places icon after button text
+   * Determines position of icon relative to button children.
+   * - **before** places icon before button children
+   * - **after** places icon after button children
```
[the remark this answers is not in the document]
dierat: I originally removed them in https://github.com/chanzuckerberg/edu-design-system/pull/889#pullrequestreview-903732508 but reverted that change because of push back.

My plan was to ask Sean how he feels about having the unreliable spacing between the icon and the text. If he's fine with it, we can leave these props and the set margin between the icon and the text. I believe that issue with the spacing is the only reason we wanted to avoid this API and let devs pass in the icons themselves (because then they can also add custom styling).

The only other downside I can think of is if there are custom icons in `traject` that are not in EDS, they'll need to pass those in via `children` anyway. But I think right now the only custom icon not in EDS is the LP logo, which I don't think is being used in `Button`...

anniehu4 (approved):

docs/CODE_GUIDELINES.md
dierat: Something I just realized is that the `text` prop is already accepting a `ReactNode` in the type, so the argument that we should keep `text` in order to restrict the type doesn't make much sense to me. If the type is the same, shouldn't we just pick the name and structure that makes sense for our team?

https://github.com/chanzuckerberg/edu-design-system/blob/next/src/components/Button/Button.tsx#L60

dierat:
Type failures are expected as they are also appearing on `next`; fixed here: https://github.com/chanzuckerberg/edu-design-system/pull/934

dierat:
@anniehu4 I wasn't totally sure what to write in the docs, so I put down my current view of when to use `children` based on the current components. I'm going to merge this in to reduce conflicts, but let me know if you want any changes there and I'll come back to it.

PR #926, opened 2022-04-01
fix(list detail): replace missing types

csnizik:
### Summary: 
This is a fix for some missing types; it was causing TypeScript errors.

### Test Plan:
Ensure that TS errors as seen here are no longer present by running `yarn run types`
Screen Shot 2022-04-01 at 11 49 20 AM

### DIFF
--- a/src/components/ListDetail/ListDetail.tsx
+++ b/src/components/ListDetail/ListDetail.tsx
@@ -86,6 +86,7 @@ export const ListDetail = ({
   overflow,
   onChange,
   required,
+  title,
   ...other
 }: Props) => {
   /**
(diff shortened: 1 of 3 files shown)


dierat (approved):

src/components/ListDetailPanel/ListDetailPanel.tsx
```diff
@@ -29,6 +37,8 @@ export const ListDetailPanel = ({
   className,
   id,
   ariaLabelledBy,
+  variant,
+  title,
```
dierat, line 41: Are these being used anywhere though?

src/components/ListDetail/ListDetail.tsx
```diff
@@ -86,6 +86,7 @@ export const ListDetail = ({
   overflow,
   onChange,
   required,
+  title,
```
dierat, line 89: I don't see this one being used either.

csnizik: ... as a matter of fact... now that you mention it... <scratchesHead/> ... no... taking another look to see why they quieted TS

PR #927, opened 2022-04-01
[1/2] Add existing EDS icons

anniehu4:
### Summary:
[ch191086]

- Ran [convert-icons.js](https://gist.github.com/anniehu4/1b68cd46a0884b95952b5fd84e690050) to auto-convert existing component files into SVG files (manually converted few more complicated svgs)
- Removed icons that were unused in either `next` or `main` for now

There are still **20** icons that I haven't moved over because the current name conflicts with a name that's used in this branch. Will put that up in a separate PR for visibility

### Test Plan:
- Verified [IconGrid in Storybook](https://61313967cde49b003ae2a860-wiulwbeeto.chromatic.com/?path=/story/atoms-icons-icongrid--default)


dierat (approved):
I see the `ListDetail` icon colors are b0rked but I assume we'll come back to that.

https://www.chromatic.com/test?appId=61313967cde49b003ae2a860&id=624766413c273a003a24f2fd
list detail component in chromatic

anniehu4:
mm good catch. Looks like that change happened here: https://github.com/chanzuckerberg/edu-design-system/pull/926 but wasn't reviewed / accepted in Chromatic 😬 lemme see if it's a quick fix

PR #929, opened 2022-04-01
chore(banner): use plain button instead of CloseButton

dierat:
### Summary:
The `CloseButton` is kind of a snowflake because it has a hover effect for the icon button that we don't use anywhere else. I talked to Sean and we're good to remove it to make the close buttons consistent with our other icon buttons. But before we can delete the component, we need to update the locations using it.

This PR updates the `Banner` component to use a plain `Button` instead of the customized `CloseButton`. The placement of the icon is a little different than before, but we paired on it and decided this is fine for now.

### Test Plan:
Tested in storybook.

https://user-images.githubusercontent.com/7761701/161342193-d8d57a08-abde-41ff-9aca-2ff9186cba39.mp4

### DIFF
--- a/packages/components/src/Banner/__snapshots__/Banner.spec.tsx.snap
+++ b/packages/components/src/Banner/__snapshots__/Banner.spec.tsx.snap
@@ -53,7 +53,7 @@ exports[`<Banner /> AlertDismissable story renders snapshot 1`] = `
   class="bannerDialog horizontal dismissable colorAlert"
 >
   <button
-    class="button dismiss colorAlert button variantLink colorAlert"
+    class="dismiss button sizeLarge variantPlain colorAlert"
     type="button"
   >
     <svg
@@ -67,7 +67,7 @@ exports[`<Banner /> AlertDismissable story renders snapshot 1`] = `
       xmlns="http://www.w3.org/2000/svg"
     >
       <title>
-        close
+        dismiss banner
       </title>
       <path
         d="M18.3 5.71a.9959.9959 0 00-1.41 0L12 10.59 7.11 5.7a.9959.9959 0 00-1.41 0c-.39.39-.39 1.02 0 1.41L10.59 12 5.7 16.89c-.39.39-.39 1.02 0 1.41.39.39 1.02.39 1.41 0L12 13.41l4.89 4.89c.39.39 1.02.39 1.41 0 .39-.39.39-1.02 0-1.41L13.41 12l4.89-4.89c.38-.38.38-1.02 0-1.4z"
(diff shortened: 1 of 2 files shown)


packages/components/src/Banner/__snapshots__/Banner.spec.tsx.snap
```diff
@@ -53,7 +53,7 @@ exports[`<Banner /> AlertDismissable story renders snapshot 1`] = `
   class="bannerDialog horizontal dismissable colorAlert"
 >
   <button
-    class="button dismiss colorAlert button variantLink colorAlert"
```
dierat, line 56: The `CloseButton` is using the "link" variant because we hadn't even added the "plain" variant yet, hence the disparity.

packages/components/src/Banner/__snapshots__/Banner.spec.tsx.snap
```diff
@@ -67,7 +67,7 @@ exports[`<Banner /> AlertDismissable story renders snapshot 1`] = `
       xmlns="http://www.w3.org/2000/svg"
     >
       <title>
-        close
```
dierat, line 70: "close" feels pretty weird in the context of a banner. Close what?

anniehu4 (approved):
pour one out for the old hover effect, you were a good one 🥲  consistency is also good tho

dierat:
> pour one out for the old hover effect, you were a good one 🥲 consistency is also good tho
> 
> actually ... hate to say this lol but I wonder if we should introduce an `"icon" / "bare"` size or variant that doesn't use so much padding? feel free to cancel if that's the wrong direction of complexity

No, that's a good question. We could just make the padding equal on all sides for the "bare" variant, but then the question is whether that would be weird when there's text + an icon. Or we could do some automagical work where if there's only an `Icon` in the `children`, we adjust the padding. Or have another prop that's like `squarePadding` or something that controls the padding, but then we just might as well just make another variant.

The way we're doing it on Along (in case you're curious) is the "icon" variant is square with a round hover effect (the close button hover state was inspired by that). The key difference there is we never use it with text. Text buttons only come in the "link" variant, even if it's hovering in space or whatever. (No one's ever complained about it, so it's still confusing to me why having an underlined text link not in the context of a paragraph is considered so horrid on LP.) It would make our lives so much simpler if we could do that on LP, but I have a feeling that's never going to happen.

In any case, I'll make a ticket with these notes.

dierat:
Icon button padding ticket: https://app.shortcut.com/czi-edu/story/191523/figure-out-padding-on-icon-only-bare-buttons

PR #930, opened 2022-04-01
chore(toast): use plain button instead of CloseButton

dierat:
### Summary:
The `CloseButton` is kind of a snowflake because it has a hover effect for the icon button that we don't use anywhere else. I talked to Sean and we're good to remove it to make the close buttons consistent with our other icon buttons. But before we can delete the component, we need to update the locations using it.

This PR updates the `Toast` component to use a plain `Button` instead of the customized `CloseButton`. The placement of the icon is a little different than before, but we paired on it and decided this is fine for now.

### Test Plan:
Tested in storybook.

https://user-images.githubusercontent.com/7761701/161343986-affc8360-0251-4fb7-8b24-f15baea6e6b1.mp4

### DIFF
--- a/packages/components/src/Toast/Toast.tsx
+++ b/packages/components/src/Toast/Toast.tsx
@@ -1,6 +1,7 @@
 import clsx from "clsx";
 import React from "react";
-import CloseButton from "../CloseButton";
+import Button from "../Button";
+import CloseRoundedIcon from "../Icons/CloseRounded";
 import Text from "../Text";
 import colorStyles from "../common/Notifications/Notification.module.css";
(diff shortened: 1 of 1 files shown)


jinlee93 (approved):
Looks good to me, but yea the spacing of the button not being symmetrical(? idk the word) is kind of interesting

dierat:
> Looks good to me, but yea the spacing of the button not being symmetrical(? idk the word) is kind of interesting

You mean it not being evenly spaced on all sides? Yeah that has always bugged me about our icon buttons because there is no real "icon" version; they're all designed for text.

PR #933, opened 2022-04-02
chore(list detail): update tokens

csnizik:
### Summary:
This PR updates the tokens used in ListDetail. 

### Test Plan:
Confirm that styles are applied correctly to ListDetail in Storybook.

### DIFF
--- a/src/components/ListDetail/ListDetail.module.css
+++ b/src/components/ListDetail/ListDetail.module.css
@@ -1,17 +1,17 @@
 @import '../../design-tokens/mixins.css';
 
 /*------------------------------------*\
-    #List Detail
+ #List Detail
 \*------------------------------------*/
 
 /**
  * 1) List of of links where each link toggles open associated information
  */
 .list-detail {
-  background: var(--eds-color-neutral-white); /* TODO */
+  background: var(--eds-theme-color-background-neutral-default);
   border-width: var(--eds-theme-border-width);
   border-style: solid;
-  border-color: var(--eds-theme-color-neutral-subtle-border);
+  border-color: var(--eds-theme-color-border-neutral-subtle);
   border-radius: var(--eds-theme-border-radius);
 
   @media all and (min-width: $eds-bp-lg) {
@@ -46,7 +46,7 @@
   @media all and (min-width: $eds-bp-lg) {
     flex: 1; /* 1 */
     padding: var(--eds-size-5) var(--eds-size-8);
-    box-shadow: -6px 0px 10px -2px var(--eds-theme-color-neutral-subtle-border); /* 1 */
+    box-shadow: -6px 0px 10px -2px var(--eds-theme-color-border-neutral-subtle); /* 1 */
   }
 }
 
@@ -79,17 +79,20 @@
 /**
   * List detail item after
   * 1) When list variant is 'ordered', add a line between items that shouldn't link out unliike the bullet
+  * 2) This sets the size of the gap above the vertical line by determining the top of the line relative to the top of the current list item; a value of 0 would position the line at the top of the current list item; 100% would position it at the bottom of the current list item. We want a small visual gap between the bullet/icon and the top of the line 
+  * 3) This sets the height of the vertical line relative to the height of the current list item. We want a small visual gap between the bottom of the line and the next bullet/icon
+  * 4) Using a border color here because this is a vertical line, similar to a border
   */
 .list-detail__list--ordered .list-detail__item {
   &:after {
     content: '';
     position: absolute;
-    top: 1.5rem;
     left: 0.25rem;
+    top: 50%; /* 2 */
+    height: 45%; /* 3 */ 
     width: 0.0625rem;
-    height: 70%;
-    background: var(--eds-theme-color-neutral-subtle-background);
-    border-radius: 50%;
+    background:  var(--eds-theme-color-border-neutral-subtle); /* 4 */
+    border-radius: var(--eds-theme-border-radius);
   }
 
   /**
@@ -100,7 +103,7 @@
 
     /**
       * List list detail item after
-      * 1) Don't add a line between items
+      * 1) Don't add a line after the last list item
       */
     /* stylelint-disable-next-line max-nesting-depth */
     &:after {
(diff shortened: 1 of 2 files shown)


src/components/ListDetail/ListDetail.module.css
```diff
@@ -79,17 +79,20 @@
...
+  * 3) This sets the height of the vertical line relative to the height of the current list item. We want a small visual gap between the bottom of the line and the next bullet/icon
+  * 4) Using a border color here because this is a vertical line, similar to a border
   */
 .list-detail__list--ordered .list-detail__item {
   &:after {
     content: '';
     position: absolute;
-    top: 1.5rem;
     left: 0.25rem;
+    top: 50%; /* 2 */
```
csnizik: The gaps above and below the line are subject to change because: a) the icons (success, warning, error) and number are larger than the grey bullet, so when a bullet is present the gap will be slightly larger than when an icon or number is used; b) if the text in a list item takes more than a single line, the gap at the bottom will be larger than the gap at the top.

src/components/ListDetail/ListDetail.module.css
```diff
@@ -79,17 +79,20 @@
...
+  * 3) This sets the height of the vertical line relative to the height of the current list item. We want a small visual gap between the bottom of the line and the next bullet/icon
+  * 4) Using a border color here because this is a vertical line, similar to a border
   */
 .list-detail__list--ordered .list-detail__item {
   &:after {
     content: '';
     position: absolute;
-    top: 1.5rem;
     left: 0.25rem;
+    top: 1.75rem; /* 2 */
```
csnizik, line 91: The gaps above and below the line are subject to change because: a) the icons (success, warning, error) and number are larger than the grey bullet, so when a bullet is present the gap will be slightly larger than when an icon or number is used; b) if the text in a list item takes more than a single line, the gap at the bottom will be larger than the gap at the top.

src/components/ListDetail/ListDetail.module.css
```diff
@@ -167,49 +174,58 @@
...
+  background: var(---eds-theme-color-background-neutral-default); /* 1*/
   z-index: var(--eds-z-index-100);
 
   /**
   * List detail icon within success list detail item
   */
   .list-detail__item--success & {
-    fill: var(--eds-theme-color-utility-success-foreground);
-  } /**
+    fill: var(--eds-theme-color-background-grade-complete-default);
```
csnizik, line 215: Using background colors here because they looked like the closest to what is shown in Figma

src/components/ListDetail/ListDetail.module.css
```diff
@@ -167,49 +174,58 @@
...
+  } 
+  
+  /**
   * List detail icon within warning list detail item
   */
   .list-detail__item--warning & {
-    fill: var(--eds-theme-color-utility-warning-foreground);
-  } /**
+    fill: var(--eds-theme-color-background-grade-revise-default);
```
csnizik, line 222: The exclamation mark used in this icon should be black. Adding that flexibility to the Icon component seems like the best way to handle it? Adding a workaround here like a black background color doesn't work because the bounding box is rectangular; getting around that by setting border-radius still results in the background being larger than the svg and giving the appearance of a black border. I'm open to suggestions here, or do we just mark it as tech debt?

anniehu4: is "text color" in this case white? I wonder if https://github.com/chanzuckerberg/edu-design-system/pull/940 will make this easier to handle

src/components/ListDetail/ListDetail.module.css
```diff
@@ -167,49 +174,58 @@
...
+  justify-content: center;
+  align-items: center;
   left: -4px;
-  width: 12px;
-  height: 12px;
-  border-radius: 50%;
-  border: 1px solid var(--eds-theme-color-neutral-subtle-border);
+  width: 1.5rem; /* 3 */ 
+  height: 1.5rem; /* 3 */
+  border-radius: 50%; /* 4 */
```
anniehu4, line 190: couldn't this still be `eds-border-radius-round`? or should we avoid it b/c that's a tier-1 token definition?

src/components/ListDetail/ListDetail.module.css
```diff
@@ -46,7 +46,7 @@
   @media all and (min-width: $eds-bp-lg) {
     flex: 1; /* 1 */
     padding: var(--eds-size-5) var(--eds-size-8);
-    box-shadow: -6px 0px 10px -2px var(--eds-theme-color-neutral-subtle-border); /* 1 */
+    box-shadow: -6px 0px 10px -2px var(--eds-theme-color-border-neutral-subtle); /* 1 */
```
anniehu4, line 49: should we define this as a theme box shadow?

src/components/ListDetail/ListDetail.module.css
```diff
@@ -111,12 +114,14 @@
...
  * List detail link
+ * 1) Base style for each link in the list
+ * 2) Using disabled text color here because all links are "inactive" until selected and this color is overridden
  */
 .list-detail__link {
   position: relative;
   display: flex;
   text-decoration: none;
-  color: var(--eds-theme-color-neutral-subtle-foreground);
+  color: var(--eds-theme-color-text-disabled); /* 2 */
```
anniehu4, line 124: Flagging that this is reallyy hard to read color contrast wise. We may want to pick a darker disabled text color

anniehu4 (approved):

src/components/ListDetail/ListDetail.module.css
csnizik: I'm about to review #940 , that would be great if it did

src/components/ListDetail/ListDetail.module.css
csnizik: I personally like the idea of decoupling this from the border token simply because this value needs to remain unchanged here.

src/components/ListDetail/ListDetail.module.css
csnizik: Maybe a tech debt item should be to sweep through components to find outliers like this to discuss with design, i.e. should this really be a unique box shadow anyway?

src/components/ListDetail/ListDetail.module.css
csnizik: Good point.

PR #935, opened 2022-04-03
feat(button): remove icon props

dierat:
### Summary:
On the CZI side, we've been discussing retaining the existing `Button` API where devs pass in an icon via `children` instead of the `iconName` and `iconPosition` props. This PR explores what that would look like.

Our reasoning is:
- we can get the spacing between icons and text with `gap`; we don't need to manually add margin to icons before and after the text
- this API feels much more intuitive and natural to us, and we feel confident the devs using the components would agree
- the other API felt especially weird to us when the button only contained an icon; you still had to pass in an `iconPosition` and it would add the margin on to one side, making it lop-sided; `gap` only adds the spacing when there's more than one child

The only downsides imo are:
- if the `children` text is broken up into multiple spans or something for some reason, that would need to be wrapped in another `span` to avoid weird spacing in the text (this seems pretty rare though)
- it was kinda nice that if there was an icon before the text and the `loading` prop was set to `true`, the loading spinner would automatically replace the icon to avoid a double icon (though I don't know if we really care about that)

For more context on the `loading` prop, we don't have that in the current EDS `Button` and I've only seen one or two places in the product that are using a loading spinner in a button like this, but I talked to Sean and we agreed this would be a nice paradigm to start using because it's so simple and consistent. We're planning to write up a little guidance on when to use it. So although it's not currently a thing in LP, we would like for it to become one.

### Test Plan:
Check out the buttons in storybook that have icons and verify they're mostly unchanged (the icon placement did shift a little).

### DIFF
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -95,10 +95,14 @@ export const Banner = ({
           className={styles['banner__close-btn']}
           variant="bare"
           aria-label={closeButtonText}
-          iconName="close"
-          iconPosition="after"
           onClick={(e: any) => onDismiss(e)}
-        />
+        >
+          <Icon
+            name="x"
+            title={closeButtonText}
+            className={styles['banner__icon']}
+          />
+        </Button>
       )}
     </div>
   );
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -97,7 +97,7 @@ export const Banner = ({
           onClick={(e: any) => onDismiss(e)}
         >
           <Icon
-            name="x"
+            name="close"
             title={closeButtonText}
             className={styles['banner__icon']}
             purpose="informative"
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -94,7 +94,6 @@ export const Banner = ({
         <Button
           className={styles['banner__close-btn']}
           variant="bare"
-          aria-label={closeButtonText}
           onClick={(e: any) => onDismiss(e)}
         >
           <Icon
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -101,6 +101,7 @@ export const Banner = ({
             name="x"
             title={closeButtonText}
             className={styles['banner__icon']}
+            purpose="informative"
           />
         </Button>
       )}
--- a/src/components/Button/Button.module.css
+++ b/src/components/Button/Button.module.css
@@ -8,12 +8,14 @@
  * 1) Button or link that has functionality to it
  * 2) Defaults to primary button
  * 3) Margin 0 added becuase Safari has margin on the button
+ * 4) Adds a gap between children to automatically handle spacing between icons and text
  */
 .button {
   display: inline-flex;
   align-items: center;
   justify-content: center;
   margin: 0; /* 3 */
+  gap: 0.75rem; /* 4 */
   border-width: var(--eds-theme-border-width);
   border-style: solid;
   border-color: var(--eds-theme-color-primary-border);
@@ -22,6 +24,11 @@
   background-color: transparent;
   cursor: pointer;
   transition: all var(--eds-anim-fade-quick) var(--eds-anim-ease);
+
+  /* TODO: make default color prop in Icon inherit or something? */
+  svg {
+    fill: currentColor;
+  }
 
   &:hover {
     color: var(--eds-theme-color-primary-foreground-hover);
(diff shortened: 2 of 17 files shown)


src/components/Banner/Banner.tsx
```diff
@@ -93,11 +93,14 @@ export const Banner = ({
...
           variant="bare"
-          aria-label={closeButtonText}
-          iconName="x"
-          iconPosition="after"
           onClick={(e: any) => onDismiss(e)}
-        />
+        >
+          <Icon
+            name="x"
+            title={closeButtonText}
```
dierat, line 100: Something I noticed while going through this that's a little odd is sometimes the `Icon`s by themselves use `title` but in the `Button` component it's using `aria-label`. (Originally the Big Medium `Button` was using visually hidden text within the `children` and we've switched to `aria-label`, but either way it's a different solution than using `title`.) I'm thinking we should switch to using `title` on the `Icon` component (and maybe inside the `Icon` it uses `aria-label` or visually hidden text or whatever) for consistency.

src/components/Button/Button.module.css
```diff
@@ -23,6 +25,11 @@
   cursor: pointer;
   transition: all var(--eds-anim-fade-quick) var(--eds-anim-ease);
 
+  /* TODO: make default color prop in Icon inherit or something? */
+  svg {
+    fill: currentColor;
+  }
```
dierat, line 31: Not sure what we want to do here. I figure using the default `Icon` color prop would be best because then it's easy for the developer to override it if needed (of course this is dependent on the `color` prop being implemented in the `Icon` component).

src/components/Button/Button.module.css
```diff
@@ -8,12 +8,14 @@
...
  * 2) Defaults to primary button
  * 3) Margin 0 added becuase Safari has margin on the button
+ * 4) Adds a gap between children to automatically handle spacing between icons and text
  */
 .button {
   display: inline-flex;
   align-items: center;
   justify-content: center;
   margin: 0; /* 3 */
+  gap: 0.75rem; /* 4 */
```
dierat, line 18: If you look at the deleted code below, this number varied depending on the button variant and size, but it's always the same in our mocks. (Also the font size no longer changes at different button sizes, which may have been why the gap shrank with smaller buttons.)

src/components/Button/Button.module.css
```diff
@@ -229,45 +236,16 @@
...
  * Loading icon rotiation style
  */
 .button.eds-is-loading > .button__icon {
-  animation: rotateIcon 2s linear infinite; /* 1 */
+  animation: rotateIcon 2s linear infinite;
   overflow: visible;
+
+  @media screen and (prefers-reduced-motion) {
+    animation: none;
+  }
```
dierat, line 248: We should add this even if this PR is dropped.

src/components/Button/Button.module.css
```diff
@@ -229,45 +236,16 @@
...
-
-  .button--sm & {
-    margin-left: var(--eds-size-half);
-  }
-}
-
 /**
  * Loading icon rotiation style
  */
 .button.eds-is-loading > .button__icon {
```
dierat, line 242: TODO: I should remove the `.button__icon` class name since these two are only being used together now.

src/components/Button/Button.stories.tsx
```diff
@@ -22,16 +23,22 @@ Default.args = { children: 'Button' };
 
 export const DefaultWithIconBefore = Template.bind({});
 DefaultWithIconBefore.args = {
-  children: 'Button',
-  iconPosition: 'before',
-  iconName: 'chevron-left',
+  children: (
+    <>
+      <Icon aria-hidden="true" focusable={false} name="chevron-left" />
```
dierat, line 28: Side note: when would we want the icon to be focusable anyway? Why not put it in a `Button`? We recently found a bug where tooltips aren't read out in NVDA unless the trigger is in a `Button`, so I don't think tooltips is the intended use case here.

src/components/TableHeaderCell/TableHeaderCell.tsx
```diff
@@ -82,13 +82,9 @@ export const TableHeaderCell = ({
...
     >
-      <Button
-        variant="table-header"
-        iconPosition="after"
-        iconName="arrow-narrow-down"
-        onClick={onClick}
-      >
-        {children}
+      <Button variant="table-header" onClick={onClick}>
+        {text}
```
dierat, line 86: The change here from `children` to `text` is cleaning up a mistake from when I renamed the `text` prop in `Button` to `children`.

src/components/Banner/Banner.tsx
anniehu4: makes sense to me

src/components/Button/Button.module.css
anniehu4: ask and ye shall receive 🙂 https://github.com/chanzuckerberg/edu-design-system/pull/940/files#diff-08acd9c461dd2b3858a54d0dd361b9164bdc650ee0050ace49c108d9dc6e2059

src/components/Button/Button.stories.tsx
anniehu4: I had the same Q while merging in SvgIcon props & decided to just remove `focusable` 😬  I think it should default to `false` tho so you shouldn't need to pass it here. also secretly the component passes `aria-hidden={!title}` so you shouldn't need to explicitly set aria-hidden to true either

anniehu4 (approved):

src/components/Button/Button.stories.tsx
dierat: Great, thanks! I'll wait until your PR merges so I can also add the `purpose` prop to these new icons. I'll also clean up the `focusable`, `aria-hidden`, and `aria-label` vs `title` stuff.

anniehu4: just merged, fyi!

dierat: Thanks!

PR #936, opened 2022-04-04
feat(button): update variants to match new tokens

dierat:
### Summary:
What this PR does:
- update the `Button` variants to match the new button token structure

What this PR does *not* do:
- update the styles of the variants, including newly added ones, to use the new tokens
- add stories for the new variants

I wanted to put this change out first to make sure we're aligned on this API change before digging into the actual styles.

### Test Plan:
- `yarn types` returns no errors
- no changes to existing button usage in storybook

### DIFF
--- a/src/components/Button/Button.module.css
+++ b/src/components/Button/Button.module.css
@@ -46,20 +46,14 @@
   /**
    * Disabled buttons
    */
-  &:disabled {
+   &:disabled,
+   &:disabled:hover,
+   &:disabled:focus {
     background: transparent;
     border-color: var(--eds-theme-color-disabled-border);
     color: var(--eds-theme-color-disabled-foreground);
     cursor: not-allowed;
-
-    /* stylelint-disable-next-line max-nesting-depth */
-    &:hover,
-    &:focus {
-      background: transparent;
-      border-color: var(--eds-theme-color-disabled-border);
-      color: var(--eds-theme-color-disabled-foreground);
-      transform: none;
-    }
+    transform: none;
   }
 
   /**
@@ -87,11 +81,7 @@
   background-color: var(--eds-theme-color-primary-background);
   color: var(--eds-theme-color-body-foreground-inverted);
 
-  &:hover {
-    background-color: var(--eds-theme-color-primary-background-hover);
-    color: var(--eds-theme-color-body-foreground-inverted);
-  }
-
+  &:hover,
   &:focus {
     background-color: var(--eds-theme-color-primary-background-hover);
     color: var(--eds-theme-color-body-foreground-inverted);
@@ -101,16 +91,11 @@
     @mixin focus;
   }
 
-  &:disabled {
+  &:disabled,
+  &:disabled:hover {
     background-color: var(--eds-theme-color-disabled-background);
     border-color: var(--eds-theme-color-disabled-border);
     color: var(--eds-theme-color-disabled-foreground);
-    /* stylelint-disable-next-line max-nesting-depth */
-    &:hover {
-      background-color: var(--eds-theme-color-disabled-background);
-      border-color: var(--eds-theme-color-disabled-border);
-      color: var(--eds-theme-color-disabled-foreground);
-    }
   }
 
   /**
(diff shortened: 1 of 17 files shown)


src/components/Button/Button.module.css
```diff
@@ -121,6 +121,106 @@
   }
 }
 
+/**
```
dierat, line 124: The secondary and tertiary button styles are just copies of the primary one for now as placeholders. Will address in a follow-up PR.

src/components/Button/Button.module.css
```diff
@@ -177,6 +277,57 @@
...
```
dierat, line 280: Also a copy of the primary variant styles.

src/components/Button/Button.stories.tsx
```diff
@@ -57,18 +57,18 @@ PrimaryInverted.args = {
   text: 'Primary Button',
 };
 
-export const BareIcon = Template.bind({});
-BareIcon.args = {
-  variant: 'bare',
+export const IconButton = Template.bind({});
```
dierat, line 60: I only updated the story names and types; will add stories for the new variants in a follow-up PR.

src/components/Button/Button.module.css
```diff
@@ -121,6 +121,106 @@
...
+  &:focus-visible {
+    @mixin focus;
+  }
+
+  &:disabled {
+    background-color: var(--eds-theme-color-disabled-background);
+    border-color: var(--eds-theme-color-disabled-border);
+    color: var(--eds-theme-color-disabled-foreground);
+    /* stylelint-disable-next-line max-nesting-depth */
+    &:hover {
```
anniehu4, line 152: to DRY some (since these are the exact same styles), could this be
```
&:disabled,
&:disabled:hover {
```

src/components/Button/Button.module.css
```diff
@@ -121,6 +121,106 @@
...
+    color: var(--eds-theme-color-body-foreground-inverted);
+  }
+
+  &:focus {
+    background-color: var(--eds-theme-color-primary-background-hover);
+    color: var(--eds-theme-color-body-foreground-inverted);
+  }
+
+  &:focus-visible {
+    @mixin focus;
```
anniehu4, line 144: what's the difference between this & the `focus` styles above? (should there be a difference)?

src/components/Button/Button.module.css
```diff
@@ -121,6 +121,106 @@
...
+  border-color: var(--eds-theme-color-primary-border);
+  background-color: var(--eds-theme-color-primary-background);
+  color: var(--eds-theme-color-body-foreground-inverted);
+
+  &:hover {
+    background-color: var(--eds-theme-color-primary-background-hover);
+    color: var(--eds-theme-color-body-foreground-inverted);
+  }
+
+  &:focus {
```
anniehu4, line 138: could this be
```
&:hover,
&:focus {
```

anniehu4 (approved):
API makes sense to me (with the perennial "what will we do about the `color` multiplier" q 😩 )

left some thoughts on DRY-ing styles a bit

src/components/Button/Button.module.css
dierat: I believe they're only showing the actual focus outline when you tab to a button (as opposed to clicking on it with a mouse) because people typically find those outlines ugly and don't like to see them unless it's necessary for the user. I guess they're okay with the text and background color changes, but they don't like the outline.

This is [Along's implementation](https://github.com/chanzuckerberg/along/blob/c2f1572541dd1c2e35b4e35fa68c5f6ec44fda96/client/src/ui/theme/index.tsx#L47):
```
  /**
   * Hide focus outlines from mouse interaction, but still use them for keyboard interaction.
   * This relies on the https://github.com/WICG/focus-visible polyfill.
   */
  .js-focus-visible :focus:not(.focus-visible) {
    outline: none;
  }
```
They have that styling at the top level, and then in the actual component styles they just target `:focus`.

I kinda wanna just use `:focus` for now and make a ticket to come back and consider using this because there could be serious accessibility issues if we throw it in and mess up the implementation. `traject` doesn't currently use this, so there's no regression if we skip it.

dierat: Mmm but it's used in 21 places in 14 files. 🤔

src/components/Button/Button.module.css
dierat: Done, thanks!

src/components/Button/Button.module.css
dierat: Also done, thanks again!

PR #938, opened 2022-04-04
chore(popover): add missing index.ts files

csnizik:
### Summary:
This PR adds `index.ts` with default exports to each of the Popover components and updates import statements to use the default export.

### Test Plan:
Confirm that no errors related to missing modules exist in affected files.


anniehu4 (approved):
thanks!

PR #940, opened 2022-04-04
Add `SvgIcon` props to `Icon`

anniehu4:
### Summary:
[ch187509]

- Adds props from `SvgIcon` to `Icon` component
- Updates existing callsites
- Adds extra stories from EDS

**To handle later** (aka I tried but it took too much time & wanted to timebox it):
- Making `ALL_ICONS` a union type for the Icon `name` prop to read from
- Adding snapshot tests (there's an error rn because the stories read from `ALL_ICONS`, and jest doesn't like `require.context`)
https://app.shortcut.com/czi-edu/story/191277/make-icon-name-prop-a-union-of-available-icon-types

### Test Plan:
- No type errors
- See visual changes in Chromatic
- Make sure Icon Docs page controls work
Screen Shot 2022-04-04 at 6 59 41 PM

### DIFF
--- a/src/components/Icon/Icon.module.css
+++ b/src/components/Icon/Icon.module.css
@@ -8,9 +8,17 @@
  * 1) Small graphic that represents functionality
  */
 .icon {
-  display: block;
-  height: var(--eds-size-2);
-  width: var(--eds-size-2);
+  display: inline-block;
+  /**
+   * Size Priority:
+   * 1: --icon-size, passed through props
+   * 2: --icon-size-default, determined by context (text, button, pill)
+   * default: 1em, same size as surrounding text
+   *
+   * Inspired by https://www.youtube.com/watch?v=EDyiaDJJu-4
+   */
+   width: var(--icon-size, var(--icon-size-default, 1em));
+   height: var(--icon-size, var(--icon-size-default, 1em));
 
   /**
    * Icon within text passage
@@ -19,21 +27,19 @@
   .text-passage & {
     display: inline; /* 1 */
   }
-
-  /**
-   * Icon within small text passage
-   * 1) Sets icon to smaller height and width to match font size
-   */
-  .text-passage--sm & {
-    height: 12px; /* 1 */
-    width: 12px; /* 1 */
-  }
 }
 
 /**
-   * Inverted icon
-   * 1) Sets icon to smaller height and width to match font size
-   */
+ * Inverted icon
+ */
 .icon--inverted {
   fill: var(--eds-theme-color-body-foreground-inverted);
+}
+
+/**
+ * Full-width icon
+ * 1) A block icon fills 100% of the width of its container
+ */
+.icon--full-width {
+  display: block
 }
--- a/src/components/Icon/Icon.module.css
+++ b/src/components/Icon/Icon.module.css
@@ -30,13 +30,6 @@
 }
 
 /**
- * Inverted icon
- */
-.icon--inverted {
-  fill: var(--eds-theme-color-body-foreground-inverted);
-}
-
-/**
  * Full-width icon
  * 1) A block icon fills 100% of the width of its container
  */
--- a/src/components/Icon/Icon.tsx
+++ b/src/components/Icon/Icon.tsx
@@ -5,68 +5,149 @@ import svg4everybody from 'svg4everybody';
 import styles from './Icon.module.css';
 import icons from '../../icons/spritemap/spritemap.svg';
 
-export interface Props {
+interface IconPropsBase {
   /**
    * CSS class names that can be appended to the component.
    */
   className?: string;
   /**
-   * Declares whether the icon is focusable or not for accessibility
+   * Either a fragment of svg elements (g, path, circle, etc.), or a single
+   * svg element. Useful for creating specific icon components.
+   *
+   * @example
+   * function CircleIcon(props: IconProps) {
+   *   return (
+   *     <Icon {...props}>
+   *       <circle cx="50" cy="50" r="50" />
+   *     </Icon>
+   *   )
+   * }
    */
-  focusable?: boolean;
+  children?: React.ReactNode;
+  /**
+   * The SVG Color, expects a valid css color (hex, rgb, etc.).
+   * Recommendation: if `currentColor` isn't sufficient, use color tokens from
+   * @chanzuckerberg/eds/lib/tokens/colors.ts
+   */
+  color?: string;
+  /**
+   * If true, the element will be displayed as a block, otherwise
+   * default is inline like images
+   */
+  fullWidth?: boolean;
   /**
    * ID used so the svg can read the title of the SVG icon to the user when accessibility is needed
    */
-  id?: any;
-  /**
-   * Name of icon to reference in icon sprite
-   */
-  name?: string;
+  id?: string;
   /**
    * Inverted variant for dark backgrounds
+   *
+   * TODO: investigate if needed
    */
   inverted?: boolean;
   /**
-   * SVG title which serves as alt text for the SVG.
+   * Name of icon to reference in icon sprite
+   *
+   * TODO: add typing of possible icon names
    */
-  title?: string;
+  name?: string;
+  /**
+   * Width/Height string (px, rem, em, vh, etc.)
+   * Recommendation: use "EdsFontSize" tokens from
+   * @chanzuckerberg/eds/lib/tokens/variables.json
+   */
+  size?: string;
+}
+
+interface InformativeIconProps extends IconPropsBase {
+  /**
+   * The role of the icon.
+   *
+   * Use "informative" when the icon **_does_** provide additional meaning to other text on the
+   * page. You'll be required to pass in a title to label the icon.
+   */
+  purpose: 'informative';
+  title: string;
+}
+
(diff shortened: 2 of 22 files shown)


src/components/Icon/Icon.module.css
```diff
@@ -19,21 +27,19 @@
   .text-passage & {
     display: inline; /* 1 */
   }
-
-  /**
-   * Icon within small text passage
-   * 1) Sets icon to smaller height and width to match font size
-   */
-  .text-passage--sm & {
```
anniehu4, line 27: Icon default size is now `1em`, which makes this size

src/components/Icon/Icon.tsx
```diff
@@ -5,68 +5,149 @@ import svg4everybody from 'svg4everybody';
...
   /**
    * CSS class names that can be appended to the component.
    */
   className?: string;
   /**
-   * Declares whether the icon is focusable or not for accessibility
+   * Either a fragment of svg elements (g, path, circle, etc.), or a single
+   * svg element. Useful for creating specific icon components.
+   *
+   * @example
```
anniehu4, line 17: we support this use case in the existing component, so figured it doesn't hurt to keep it

src/components/Icon/Icon.tsx
```diff
@@ -5,68 +5,149 @@ import svg4everybody from 'svg4everybody';
...
+   * function CircleIcon(props: IconProps) {
+   *   return (
+   *     <Icon {...props}>
+   *       <circle cx="50" cy="50" r="50" />
+   *     </Icon>
+   *   )
+   * }
    */
-  focusable?: boolean;
```
anniehu4, line 16: the only use cases were setting this to `false` but the default was already `false`, so prop didn't appear to be needed 👀  lmk if there are use cases I didn't think of though

src/design-tokens/css/base/media.css
```diff
@@ -10,3 +10,10 @@ img {
   max-width: 100%;
   height: auto;
 }
+
+/**
+ * Align SVG elements to be vertically centered by default
+ */
+svg {
+  vertical-align: middle;
```
anniehu4, line 18: traject already has this through [Tailwind Preflight](https://tailwindcss.com/docs/preflight#images-are-block-level), and I figured this would be lower-specificity to override if needed. Could set the styling in `Icon.module.css` though

src/util/allIcons.ts
```diff
@@ -0,0 +1,4 @@
+export const ALL_ICONS: Array<string> = require
```
anniehu4, line 1: ideally we could do some typescript magic here and generate a string union :') but it can't infer the result of this function operation. will investigate in the future

src/components/Icon/Icon.tsx
```diff
@@ -5,68 +5,149 @@ import svg4everybody from 'svg4everybody';
...
+  children?: React.ReactNode;
+  /**
+   * The SVG Color, expects a valid css color (hex, rgb, etc.).
+   * Recommendation: if `currentColor` isn't sufficient, use color tokens from
+   * @chanzuckerberg/eds/lib/tokens/colors.ts
```
anniehu4, line 30: hmmmm I guess these should surface the tier-2 color tokens instead. will add that as a ticket

src/components/Icon/Icon.tsx
```diff
@@ -5,68 +5,149 @@ import svg4everybody from 'svg4everybody';
...
    * ID used so the svg can read the title of the SVG icon to the user when accessibility is needed
    */
-  id?: any;
+  id?: string;
+  /**
+   * Inverted variant for dark backgrounds
+   *
+   * TODO: investigate if needed
+   */
+  inverted?: boolean;
```
dierat, line 47: I really don't think so, especially since you have the `color` prop now and devs can just pass in white. Also I talked to Sean about inverted buttons and we agreed there is no use case for them in LP at the moment or the foreseeable future; I think that applies to inverted icons and text too.

csnizik:
Reviewing this now

src/components/LinkListItem/LinkListItem.tsx
```diff
@@ -73,7 +73,7 @@ export const LinkListItem = ({
       >
         {iconPosition === 'before' && (
           <Icon
-            aria-hidden="true"
+            purpose="decorative"
             name={iconName}
             className={styles['link-list__icon']}
```
dierat, line 78: I bet some of these have colors set in the styles.. Maybe make a ticket to go back and clean those up?

src/components/Icon/Icon.tsx
```diff
@@ -5,68 +5,149 @@ import svg4everybody from 'svg4everybody';
...
+    inverted,
+    purpose,
+    size,
+  } = props;
+  const [idVar, setId] = useState('');
 
   useEffect(() => {
     setId(id || nanoid());
-    svg4everybody(); //Required to get IE to render icon sprites
+    svg4everybody(); // Required to get IE to render icon sprites
```
dierat, line 111: Nice :)

src/components/Icon/Icon.tsx
dierat: Maybe ask BM in case there's something we're missing? I'm sure they had a specific use case in mind when they added this prop.

src/components/Icon/Icon.tsx
```diff
@@ -5,68 +5,149 @@ import svg4everybody from 'svg4everybody';
...
+   * Use "decorative" when the icon **_does not_** provide any additional context or meaning to
+   * associated text. Basically the icon is for show and people don't need it to understand what's
+   * on the page.
+   */
+  purpose: 'decorative';
+}
+
+export type IconProps = DecorativeIconProps | InformativeIconProps;
+
+interface SvgStyle extends React.CSSProperties {
```
csnizik, line 86: nit: add to named imports

src/components/Icon/Icon.tsx
```diff
@@ -5,68 +5,149 @@ import svg4everybody from 'svg4everybody';
...
+   * }
    */
-  focusable?: boolean;
+  children?: React.ReactNode;
+  /**
+   * The SVG Color, expects a valid css color (hex, rgb, etc.).
+   * Recommendation: if `currentColor` isn't sufficient, use color tokens from
+   * @chanzuckerberg/eds/lib/tokens/colors.ts
+   */
+  color?: string;
```
csnizik, line 32: I looked at places where `Icon` is being used and each time its fill color is being set by adding a `__icon` subclass to the parent component, and that fill color is overriding any value pass to the `color` prop of `Icon` (and in keeping with that pattern, I wasn't able to find any uses of `Icon` that actually did pass a value to `color`. But using this pattern of modifying one component from within another component via a subclass is icky, not SOLID, etc.

If we take this on and refactor the way props are sent to Icons, it would help to include something in CODE_GUIDELINES to say "do not use classes on one component to alter the behavior of a different component", which would mean no more adding classnames with other component names in them like "__icon"

src/components/Icon/Icon.tsx
anniehu4: per [slack discussion](https://czi-edu.slack.com/archives/C0310HJ9WEN/p1649173028243059), moving forward with removing it 👍

src/components/Icon/Icon.tsx
```diff
@@ -5,68 +5,149 @@ import svg4everybody from 'svg4everybody';
...
+    /**
+     * height/width html properties are overriden by the defaults applied in CSS module
+     */
+    style,
+    width: size,
+    xmlns: 'http://www.w3.org/2000/svg',
+  };
+  // allow passing custom SVGs to render, otherwise
+  // load from the spritemap of EDS icons
+  const svgToUse = children || <use xlinkHref={`${icons}#${name}`} />;
```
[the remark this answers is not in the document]
anniehu4: ooh good call, I like it

src/components/Icon/Icon.tsx
anniehu4: yep, this is a prop from the `main` branch. and agree that I like this more than the `__icon` class pattern -- will file a ticket to do this refactor and update the guidelines

src/components/Icon/Icon.tsx
```diff
@@ -5,68 +5,149 @@ import svg4everybody from 'svg4everybody';
...
+   * The role of the icon.
+   *
+   * Use "decorative" when the icon **_does not_** provide any additional context or meaning to
+   * associated text. Basically the icon is for show and people don't need it to understand what's
+   * on the page.
+   */
+  purpose: 'decorative';
+}
+
+export type IconProps = DecorativeIconProps | InformativeIconProps;
```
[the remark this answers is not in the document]
anniehu4: I thiiink unions have to be declared as `type`, otherwise this would need to look something like `interface IconProps extends Decorative..., Informative... {}`

but I don't really know the differences between type and interface in the first place 😅

anniehu4:
followup work:
- https://app.shortcut.com/czi-edu/story/191478/surface-tier-2-color-tokens-for-use-with-color-prop-in-icon-component
- https://app.shortcut.com/czi-edu/story/191630/add-icon-story-snapshots
- https://app.shortcut.com/czi-edu/story/191277/make-icon-name-prop-a-union-of-available-icon-types

PR #941, opened 2022-04-05
chore: move tag to next

jinlee93:
[sc-187510]
### Summary:
- move tag from v0 to next
- rename `badge` to `tag` 
- add tier 1 token other-yellow-300
- add tier 2 token text color brand primary
- Color variants renamed:
  - Neutral -> default
  - Alert -> stop
  - Yellow -> yield
  - Success -> go
- `flat` | `outline` Variant changed to boolean `hasOutline`
- `color` changed to `variant`
- Text should be passed as `text` prop and not `children`

### Test Plan:
- snap testing


src/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,83 @@
...
+  className?: string;
+  /**
+   * The tag icon (shouldn't provide color or size since those are determined
+   * by the color prop).
+   */
+  icon?: React.ReactNode;
+  /**
+   * Style of the tag.
+   */
+  variant?: 'flat' | 'outline';
```
jinlee93, line 45: Kept variant as `flat` | `outline` but since `text` variant is color, should we change `color` to `variant` here and this to something else?

src/design-tokens/tier-2-usage/colors-text.json
```diff
@@ -49,6 +49,11 @@
           },
           "disabled": {
             "value": "{eds.color.neutral.300}"
+          },
+          "brand": {
+            "primary": {
+              "value": "{eds.color.brand.grape.700}"
```
jinlee93, line 55: text color brand added for this case, thought it would be broader use so added in tier2, lmk if should be tier3

src/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,83 @@
...
+  | 'brand';
+
+export const stylesByColor = {
+  neutral: styles[`tag--color-neutral`],
+  warning: styles[`tag--color-warning`],
+  brand: styles[`tag--color-brand`],
+  alert: styles[`tag--color-alert`],
+  revise: styles[`tag--color-revise`],
+  success: styles[`tag--color-success`],
+};
```
jinlee93, line 21: Brought over from v0. I assume the use case for this was to export the tag styles around but it's only being used in the story file as a list of colors, could we remove and replace in the story file as a simple enum/array?

src/components/Tag/Tag.module.css
```diff
@@ -0,0 +1,99 @@
...
+  display: inline-flex;
+  align-items: center;
+
+  border: var(--tag-secondary-color) solid var(--eds-theme-border-width); /* 1 */
+  border-radius: 9999px; /* 2 */
+
+  /** 
+   * TODO: (dcwither) when adding SvgIcon to EDS, include this in tailwind font 
+   * size macro.
+   */
```
dierat, line 29: I think we can remove this comment since we're not using tailwind anymore.

src/components/Tag/Tag.stories.tsx
```diff
@@ -0,0 +1,86 @@
+import type { StoryObj } from '@storybook/react';
+import React from 'react';
+import Tag, { stylesByColor } from './Tag';
+import type { Color } from './Tag';
+import styles from './Tag.stories.module.css';
+import Icon from '../Icon';
+
+// todo (Andrew): look into getting rid of the `as` cast. The `stylesByColor` object's keys are
+// members of Color. Can TypeScript understand that?
```
dierat, line 9: Could you change the beginning of this comment to `TODO(Andrew):` to be consistent with other todo comments? Just so it's easier to find.

src/components/Tag/Tag.module.css
jinlee93: What about the line under it `--svg-icon-size-default: 1.25rem;` ? Are we using that variable in icons?

src/components/Tag/Tag.stories.tsx
jinlee93: Will do

src/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,83 @@
...
+import React from 'react';
+import styles from './Tag.module.css';
+import Text from '../Text';
+
+export type Color =
+  | 'neutral'
+  | 'alert'
+  | 'success'
+  | 'warning'
+  | 'revise'
```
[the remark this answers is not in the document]
dierat: We should make a note to make the code names and the design names for these things the same – in figma they're called "default", "stop", "go", "warning", "yield", and "brand" (only "warning" and "brand" are the same).

https://www.figma.com/file/P9VPyI821gLq832tuU2WeY/UI-Kit-Master-(Core)?node-id=2525%3A662
the tag component in figma

src/components/Tag/Tag.tsx
dierat: I personally like the deduplication mostly because that means if we change the props here at some point (like we add a new one), it's automatically updated in storybook as well.

jinlee93: `stylesByColor` isn't being used to type props tho, and I think this problem is part of Andrew's todo haha

src/components/Tag/Tag.tsx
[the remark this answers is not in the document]
jinlee93: Should I change these to "default", "stop", "go", "warning", "yield", and "brand"?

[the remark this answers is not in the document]
jinlee93: Will change to "default", "stop", "go", "warning", "yield", and "brand"

src/components/Tag/Tag.module.css
```diff
@@ -0,0 +1,99 @@
...
+  --svg-icon-size-default: 1.25rem;
```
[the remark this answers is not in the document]
jinlee93: Discussing in https://github.com/chanzuckerberg/edu-design-system/pull/941#discussion_r843047983, resolving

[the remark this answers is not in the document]
dierat: I think it is relevant but now it's called `--icon-size` and maybe we should be passing it in via the `size` prop instead of setting the css variable directly?

https://github.com/chanzuckerberg/edu-design-system/pull/940/files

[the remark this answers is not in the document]
jinlee93: I will change it to `--icon-size` to match, but don't think allowing icon variations in the tag is necessary since there are no size variations for the text and the `Icon` is being passed as a prop and can be styled indepedently

src/components/Tag/Tag.module.css
jinlee93: As per https://github.com/chanzuckerberg/edu-design-system/pull/941#discussion_r842993194, will remove comment and change the css variable name

src/components/Tag/Tag.module.css
```diff
@@ -0,0 +1,94 @@
...
+  padding-top: var(--eds-size-half);
+  padding-bottom: var(--eds-size-half);
+  padding-left: var(--eds-size-1-and-half);
+  padding-right: var(--eds-size-1-and-half);
+
+  display: inline-flex;
+  align-items: center;
+
+  border: var(--tag-secondary-color) solid var(--eds-theme-border-width); /* 1 */
+  border-radius: 9999px; /* 2 */
```
dierat, line 24: Does this have the same effect as `border-radius: 50%`? Could we use `eds-border-radius-round` or something like that?

jinlee93: Does not, `border-radius: 50%` creates ovals: 
image

dierat: Awesome, thanks for confirming 👍

src/components/Tag/Tag.module.css
```diff
@@ -0,0 +1,99 @@
...
+  max-width: 20rem;
+  padding-top: var(--eds-size-half);
+  padding-bottom: var(--eds-size-half);
+  padding-left: var(--eds-size-1-and-half);
+  padding-right: var(--eds-size-1-and-half);
+
+  display: inline-flex;
+  align-items: center;
+
+  border: var(--tag-secondary-color) solid var(--eds-theme-border-width); /* 1 */
```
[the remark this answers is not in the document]
dierat: I'm personally in favor of using internal css variables to reduce repetition. Interestingly, the `--tag-primary-color` is only used in one place (`color`), so we're actually not getting any benefit from that one (might as well just set the property instead of setting the variable value). I wonder if we're setting all 3 for consistency.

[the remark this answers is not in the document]
jinlee93: Agree for both using internal css variables and that setting `--tag-primary-color` for consistency. Can change that since its used once, but I like the consistency

src/components/Tag/Tag.tsx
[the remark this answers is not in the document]
dierat: We're still going to need to normalize these with design because they're using the other pattern for `Banner` 😭 I'll make a ticket.
Banner component in figma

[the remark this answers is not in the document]
jinlee93: Sweet, thanks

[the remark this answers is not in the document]
dierat: Ticket: https://app.shortcut.com/czi-edu/story/191591/make-banner-and-tag-variant-names-consistent

src/components/Tag/Tag.tsx
dierat: Ideas:
 - Change the prop name to something else so `color` can have `variant`. We could switch to something like `hasOutline` and make it a bool, for example.
 - Talk to design about possibly deprecating the flat one. We agreed to deprecate the flat `Banner` because the "elevated" one looks good everywhere and now we'll have consistency, and I feel like `Tag` should go in the same direction. I think it's actually more important because someone might think and outlined tag means or does something differently than a flat one (like one is button and the other isn't).

jinlee93: Quick update: messaged Ashley about deprecating `flat` variant, who will connect with the other designers regarding the deprecation. Waiting on updates from design

dierat: Awesome, thanks! I'm wondering if, even if they decide to deprecate it, we'll still need to keep that variant around and do the whole deprecation procedure that the `Tooltip`'s dark variant is going through.

src/components/Tag/Tag.module.css
```diff
@@ -0,0 +1,94 @@
...
+  border-radius: 9999px; /* 2 */
+
+  background-color: var(--tag-secondary-color); /* 1 */
+  color: var(--tag-primary-color);
+  --icon-size: 1.25rem;
```
anniehu4, line 28: I think we should keep this `--icon-size-default`. Icon size follows this logic:
```
var(--svg-icon-size, var(--svg-icon-size-default, 1em));
```
(in order of priority from left to right).

So this would mean:
- By default, icons in the tag will be 1.25rem
- But if someone chooses to pass in an Icon with a specific size, that will then update the value of `--icon-size`, which is highest priority and will thus be used

I should add that to CODE_GUIDELINES

src/components/Tag/Tag.tsx
anniehu4: I feel ambivalent haha. If we declared the color array in `Tag.tsx` I'm pretty sure you could then generate the types, i.e.:
```
export const COLORS = [...] as const;

export type Color = typeof COLORS[number];
```
per https://melvingeorge.me/blog/convert-array-into-string-literal-union-type-typescript

src/design-tokens/tier-2-usage/colors-border.json
```diff
@@ -70,7 +70,12 @@
               "value": "{eds.color.other.mint.300}"
             },
             "revise": {
-              "value": "{eds.color.other.orange.100}"
+              "default": {
```
anniehu4, line 73: should these be flipped so `subtle` is lighter than the `default`?

src/components/Tags/Tags.tsx
```diff
@@ -1,26 +0,0 @@
...
-  /**
-   * CSS class names that can be appended to the component.
-   */
-  className?: string;
-}
-
-/**
- * Primary UI component for user interaction
- */
-export const Tags = ({ children, className, ...other }: Props) => {
```
anniehu4, line 19: I think this is handling a separate use case (multiple Tags rather than a single Tag), so would keep it around for now

anniehu4 (commented):
nice, looks great!!

src/components/Tag/Tag.module.css
jinlee93: Right, saw that coming in from the Icon pr. Will change to `--icon-size-default`

src/components/Tag/Tag.module.css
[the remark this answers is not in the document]
jinlee93: Changing to `--icon-size-default` as per https://github.com/chanzuckerberg/edu-design-system/pull/941#discussion_r843303384

src/design-tokens/tier-2-usage/colors-border.json
jinlee93: Yea, that makes sense, will flip

src/components/Tag/Tag.tsx
jinlee93: Ah I like that approach, will use that, update story file, and remove Andrew todo comment

src/components/Tags/Tags.tsx
jinlee93: Got it, will bring back `Tags`, might be a little confusing with naming tho

anniehu4: ya tru, the naming is unfortunate 🤔  I think our Tag _does_ replace their `Badge` component though, so maybe we can get rid of that?

jinlee93: Always down to remove duplicates, will remove `Badge`

src/components/Tag/Tag.tsx
jinlee93: Mmmm true. In that case, shouldn't be a blocker for this PR and can deprecate later

src/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,76 @@
...
+        styles['tag'],
+        styles[`tag--${variant}`],
+        hasOutline && styles['tag--outline'],
+      )}
+      size="sm"
+      weight="bold"
+    >
+      {icon}
+      {/* No width space to ensure height of contents */}
+      {'\u200B'}
```
dierat, line 70: Can you double-check that we still need this? I think we also had it in the `Button` but I removed it (because it was causing issues with tests finding the component for some reason but also it didn't seem to be having an effect).

dierat (approved):

src/components/Tag/Tag.tsx
jinlee93: Yah will do

src/components/Tag/Tag.module.css
```diff
@@ -0,0 +1,95 @@
...
+ * 1) Grabs colors from whichever variant is selected.
+ * 2) Rounds the corners of the tag.
+ */
+.tag {
+  width: max-content;
+  max-width: 20rem;
+  padding-top: var(--eds-size-half);
+  padding-bottom: var(--eds-size-half);
+  padding-left: var(--eds-size-1-and-half);
+  padding-right: var(--eds-size-1-and-half);
```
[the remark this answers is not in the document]
jinlee93: Oh yea definitely can shorthand the padding, thanks for the suggestion

src/components/Tag/Tag.tsx
jinlee93: Made no difference for me as well, removing

PR #942, opened 2022-04-05
chore(templates): update plop templates

csnizik:
### Summary:
This PR contains changes to the plop templates that were missed in the initial pass. Specifically:

1. Add `index.ts` to the Page and Recipe templates to match established directory structure
2. Remove React.FC typing in Page and Recipe templates to match Component template

### Test Plan:
Run `yarn run plop`, `yarn run plop:page` and `yarn run plop:recipe` and confirm that all files are generated correctly with no errors. 

Screen Shot 2022-04-05 at 8 36 47 AM

### DIFF
--- a/plop-templates/Page/Page.tsx.hbs
+++ b/plop-templates/Page/Page.tsx.hbs
@@ -6,7 +6,7 @@ import {
 
 import { PageShell } from '../../recipes/PageShell/PageShell';
 
-export const {{pascalCase name}}: React.FC = () => (
+export const {{pascalCase name}} = () => (
   <PageShell>
     <PageHeader title="Page title" />
   </PageShell>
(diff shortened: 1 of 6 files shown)


plop-templates/Page/Page.tsx.hbs
```diff
@@ -6,7 +6,7 @@ import {
 
 import { PageShell } from '../../recipes/PageShell/PageShell';
 
-export const {{pascalCase name}}: React.FC = () => (
+export const {{pascalCase name}} = () => (
```
anniehu4, line 9: could you also update `Page.stories.tsx.hbs` to no longer use `Props`, since this file doesn't export any? it's caused a few TS errors -- I think the ProjectOverview file now has an example with `Props` removed

anniehu4 (approved):

dierat (approved):

PR #943, opened 2022-04-05
chore: deprecate dark tooltip

jinlee93:
[sc-189391]
### Summary:
- deprecates dark tooltip via:
  - tsdoc `@deprecated` flag (doesn't do anything with VSCode now, but hopefully will be better supported in future TS releases https://github.com/microsoft/tsdoc/issues/131)
  - `console.warn` in non prod
  - badge in storybook
### Test Plan:
- manual testing
  - storybook has badge
  - console warns

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -52,7 +52,10 @@ type TooltipProps = {
   /**
    * Whether the tooltip has a light or dark background.
    */
-  variant?: 'light' | 'dark';
+  variant?:
+    | 'light'
+    /** @deprecated */
+    | 'dark';
   /**
    * Whether the tooltip is always visible or always invisible.
    *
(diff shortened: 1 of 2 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -83,6 +86,9 @@ export const Tooltip = ({
   text,
   ...rest
 }: TooltipProps) => {
+  if (variant === 'dark' && process.env.NODE_ENV !== 'production') {
+    console.warn('Dark variant is deprecated.');
+  }
```
dierat, line 91: Maybe we could add a little more explanation, like "The dark variant is deprecated and will be removed in an upcoming release. Please use the light variant instead."? I'm afraid people might not know what "deprecated" means in this context because this is the first time we're doing it.

dierat (approved):

src/components/Tooltip/Tooltip.tsx
jinlee93: Sounds good, thanks for the suggestion, will change

PR #944, opened 2022-04-05
chore: replace flaky image url with a cuter image url

dierat:
### Summary:
The `MediaBlock` and `StackedBlock` stories have been flaky in chromatic because they're linking to an image that doesn't exist (`https://placehold.it/600x400` takes forever and eventually just loads an error page), so sometimes we get no image and sometimes we get a broken image icon. We shouldn't link to external images anyway because it can cause flakiness, but in this case it's really obvious why this particular image is flaky, so I'm just replacing it with an actual image url for now.

### Test Plan:
Check out the `MediaBlock` and `StackedBlock` stories in storybook and verify they are now 1000% cuter.

default media block story in storybook which has a photo of two kittens being super cute
reserved media block story in storybook which has a photo of two kittens being super cute
stacked block story in storybook which has a photo of two kittens being super cute

### DIFF
--- a/src/components/MediaBlock/MediaBlock.stories.tsx
+++ b/src/components/MediaBlock/MediaBlock.stories.tsx
@@ -26,13 +26,15 @@ const Template: Story<Props> = (args) => (
 
 export const Default = Template.bind({});
 Default.args = {
-  imgSrc: 'https://placehold.it/600x400',
+  // eslint-disable-next-line @chanzuckerberg/stories/no-ext-resources-in-stories
+  imgSrc: 'https://st1.latestly.com/wp-content/uploads/2021/08/31-6.jpg',
   imgAlt: 'placeholder image',
 };
 
 export const Reversed = Template.bind({});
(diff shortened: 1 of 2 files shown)


anniehu4 (approved):
blessed 😂 

> We shouldn't link to external images anyway because it can cause flakiness, but in this case it's really obvious why this particular image is flaky, so I'm just replacing it with an actual image url for now.

argh, yeah I filed https://app.shortcut.com/czi-edu/story/191080/turn-off-image-loading-in-stories-that-are-passing-imgsrc for this. How do we avoid linking to external images? Link to internal files instead?

dierat:
> argh, yeah I filed https://app.shortcut.com/czi-edu/story/191080/turn-off-image-loading-in-stories-that-are-passing-imgsrc for this. How do we avoid linking to external images? Link to internal files instead?

Yeah, I think we should just add a placeholder image in the codebase, import it, and pass it into `src`. I'll update the ticket.

PR #947, opened 2022-04-06
feat: generate Icon name union type

anniehu4:
### Summary:
[ch191277]

- Generate list of icon names whenever `yarn build:icons` is run
- TS can then automagically build a union type, since it's a `const` array

### Test Plan:
- `yarn types` passes (and existing errors proved the typing is working)


src/components/Banner/Banner.tsx
```diff
@@ -1,6 +1,7 @@
 import clsx from 'clsx';
 import React, { ReactNode, useState } from 'react';
 import styles from './Banner.module.css';
+import type { IconName } from '../../util/allIcons';
```
anniehu4, line 4: we could re-export this from the `Icon` component, but this doesn't seem bad to me

anniehu4: actually no I change my mind

src/components/FieldNote/FieldNote.tsx
```diff
@@ -60,7 +60,7 @@ export const FieldNote = ({
           className={styles['field-note__icon']}
           title="error"
           purpose="informative"
-          name={isError && 'cancel'}
+          name={isError ? 'cancel' : undefined}
```
anniehu4, line 63: I guess previously this would pass an empty string, which typescript didn't like

src/components/Icon/Icon.tsx
```diff
@@ -4,6 +4,9 @@ import React, { useEffect, useState, ReactNode, CSSProperties } from 'react';
 import svg4everybody from 'svg4everybody';
 import styles from './Icon.module.css';
 import icons from '../../icons/spritemap/spritemap.svg';
+import { ALL_ICONS } from '../../util/allIcons';
+
+export type IconName = typeof ALL_ICONS[number];
```
anniehu4, line 9:
```suggestion
// generates a union type of all possible icon names
export type IconName = typeof ALL_ICONS[number];
```

scripts/generateIconTypes.js
```diff
@@ -0,0 +1,34 @@
...
+
+const iconFiles = fs
+  .readdirSync(ICON_FOLDER, { withFileTypes: true })
+  .filter((dirent) => dirent.isFile());
+
+// convert 'close.svg' => 'close'
+const iconNames = iconFiles.map((dirent) => {
+  const name = dirent.name.match(/([\w\s-]*)\.svg$/)[1];
+  return `'${name}'`;
+});
```
dierat, line 16: Very cool!

dierat (approved):
This is awesome, thanks!

PR #948, opened 2022-04-06
feat(button): removed inverted prop

dierat:
### Summary:
I talked to Sean about the `inverted` `Button` prop, and we decided to remove it. The reasoning is that:
- LP has no dark backgrounds (except the sidebar/header, which is unique and will need unique styling anyway)
- we have no plans to implement dark mode (maybe one day but it's not in the foreseeable future)

The inverted button tokens were also removed, so there's nothing for us to use in the styles anyway.

### Test Plan:
Verify there are no type errors and there is no change to the icons in the`GlobalHeader` recipe.

### DIFF
--- a/.storybook/recipes/GlobalHeader/GlobalHeader.tsx
+++ b/.storybook/recipes/GlobalHeader/GlobalHeader.tsx
@@ -44,7 +44,6 @@ export const GlobalHeader = ({ className, ...other }: Props) => {
       <Button
         className={styles['global-header__menu-button']}
         variant="icon"
-        inverted={true}
         onClick={toggleMenu}
       >
         <Icon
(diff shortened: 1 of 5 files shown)


.storybook/recipes/GlobalHeader/GlobalHeader.tsx
```diff
@@ -44,7 +44,6 @@ export const GlobalHeader = ({ className, ...other }: Props) => {
       <Button
         className={styles['global-header__menu-button']}
         variant="icon"
-        inverted={true}
```
dierat, line 47: Removing this prop here has no effect on this button because the icon has `fill: currentColor` (inside the Icon component) and the `Header` has its color set to white.

anniehu4 (approved):

ahuth (approved):

PR #949, opened 2022-04-06
Enable icon snapshot tests

anniehu4:
### Summary:
[ch191630]

Turns out the error was mocking `svg` files & some specific modules correctly. This adds back `Icon`, `Toast`, and `Tag` snapshot tests 🎉 

### Test Plan:
- `yarn test` passes

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -6,11 +6,10 @@ module.exports = {
   coverageDirectory: 'coverage',
   moduleNameMapper: {
     '\\.css$': 'identity-obj-proxy',
-    svg4everybody: 'identity-obj-proxy',
     'react-portal': 'identity-obj-proxy',
     shortid: 'identity-obj-proxy',
     '.*\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
-      'identity-obj-proxy',
+      '<rootDir>/jest/mocks/fileMock.js',
   },
   preset: 'ts-jest',
   roots: ['<rootDir>/src'],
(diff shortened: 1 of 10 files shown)


jest.config.js
```diff
@@ -6,11 +6,10 @@ module.exports = {
   coverageDirectory: 'coverage',
   moduleNameMapper: {
     '\\.css$': 'identity-obj-proxy',
-    svg4everybody: 'identity-obj-proxy',
     'react-portal': 'identity-obj-proxy',
     shortid: 'identity-obj-proxy',
     '.*\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
-      'identity-obj-proxy',
+      '<rootDir>/jest/mocks/fileMock.js',
```
anniehu4, line 12: lesson learned -- test errors that look something like
```
Error: Uncaught [TypeError: symbol is not a function]
```
as was this case, are usually due to files or functions that are loaded through webpack and thus `jest` can't natively handle. Will almost always need some extra mocking

jinlee93:
I think `Tag` also had some icon snaps being skipped, but doesn't seem to be present in your target branch

anniehu4:
> I think Tag also had some icon snaps being skipped, but doesn't seem to be present in your target branch

ah you right, updating

jest.config.js
jinlee93: Is there a place to record error cases like these and various approaches to solve?

jinlee93 (approved):
LGTM, thanks for fixing and re-enabling these snaps

ahuth (approved):

dierat (approved):

jest.config.js
anniehu4: good Q! do you think one of our existing FE wiki pages would make sense? if not could make a new one like "Jest debugging"

PR #952, opened 2022-04-06
feat(button): use new button tokens, update storybook

dierat:
### Summary:
I previously updated the variants in the `Button` component to match the new color tokens. This PR updates the `Button` styles to use those tokens. I also updated the stories to match the UI Kit, so it's now very close to the figma representation of the component. There's a little more work to do, mostly around sizing (icon size in the button, link text size, space between link text and icon), so I left a couple of todo comments.

### Test Plan:
Verify the button stories and styles now mostly match the UI Kit in figma, and make sure the hover/focus states look right (especially be sure to verify the disabled buttons don't change color or anything on hover/click).

### DIFF
--- a/.storybook/data/tokens.json
+++ b/.storybook/data/tokens.json
@@ -207,7 +207,7 @@
   "eds-theme-color-button-tertiary-border": "#707070",
   "eds-theme-color-button-tertiary-border-hover": "#707070",
   "eds-theme-color-button-tertiary-border-active": "#21272D",
-  "eds-theme-color-button-tertiary-text": "#6B65E2",
+  "eds-theme-color-button-tertiary-text": "#707070",
   "eds-theme-color-button-tertiary-text-hover": "#ffffff",
   "eds-theme-color-button-tertiary-text-active": "#ffffff",
   "eds-theme-color-button-icon-background": "rgba(0, 0, 0, 0)",
--- a/src/components/Button/Button.module.css
+++ b/src/components/Button/Button.module.css
@@ -11,6 +11,7 @@
  * 4) Adds a gap between children to automatically handle spacing between icons and text
  */
 .button {
+  @mixin eds-theme-typography-button-label;
   display: inline-flex;
   align-items: center;
   justify-content: center;
@@ -18,28 +19,11 @@
   gap: 0.75rem; /* 4 */
   border-width: var(--eds-theme-border-width);
   border-style: solid;
-  border-color: var(--eds-theme-color-primary-border);
   border-radius: var(--eds-theme-border-radius);
-  color: var(--eds-theme-color-primary-foreground);
-  background-color: transparent;
   cursor: pointer;
   transition: all var(--eds-anim-fade-quick) var(--eds-anim-ease);
 
-  /* TODO: make default color prop in Icon inherit or something? */
-  svg {
-    fill: currentColor;
-  }
-
-  &:hover {
-    color: var(--eds-theme-color-primary-foreground-hover);
-    border-color: var(--eds-theme-color-primary-border-hover);
-  }
-
   &:focus {
-    background-color: transparent;
-  }
-
-  &:focus-visible {
     @mixin focus;
   }
 
@@ -47,13 +31,8 @@
    * Disabled buttons
    */
    &:disabled,
-   &:disabled:hover,
-   &:disabled:focus {
-    background: transparent;
-    border-color: var(--eds-theme-color-disabled-border);
-    color: var(--eds-theme-color-disabled-foreground);
+   &:disabled:hover {
     cursor: not-allowed;
-    transform: none;
   }
 }
 
@@ -63,25 +42,28 @@
  *    the most important action of a page.
  */
 .button--primary {
-  border-color: var(--eds-theme-color-primary-border);
-  background-color: var(--eds-theme-color-primary-background);
-  color: var(--eds-theme-color-body-foreground-inverted);
+  border-color: var(--eds-theme-color-button-primary-border);
+  background-color: var(--eds-theme-color-button-primary-background);
+  color: var(--eds-theme-color-button-primary-text);
 
   &:hover,
   &:focus {
-    background-color: var(--eds-theme-color-primary-background-hover);
-    color: var(--eds-theme-color-body-foreground-inverted);
+    border-color: var(--eds-theme-color-button-primary-border-hover);
+    background-color: var(--eds-theme-color-button-primary-background-hover);
+    color: var(--eds-theme-color-button-primary-text-hover);
   }
 
-  &:focus-visible {
-    @mixin focus;
+  &:active {
+    border-color: var(--eds-theme-color-button-primary-border-active);
+    background-color: var(--eds-theme-color-button-primary-background-active);
+    color: var(--eds-theme-color-button-primary-text-active);
   }
 
   &:disabled,
   &:disabled:hover {
-    background-color: var(--eds-theme-color-disabled-background);
-    border-color: var(--eds-theme-color-disabled-border);
-    color: var(--eds-theme-color-disabled-foreground);
+    border-color: var(--eds-theme-color-border-disabled);
+    background-color: var(--eds-theme-color-background-disabled);
+    color: var(--eds-theme-color-button-primary-text);
   }
 }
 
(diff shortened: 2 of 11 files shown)


.storybook/data/tokens.json
```diff
@@ -207,7 +207,7 @@
   "eds-theme-color-button-tertiary-border": "#707070",
   "eds-theme-color-button-tertiary-border-hover": "#707070",
   "eds-theme-color-button-tertiary-border-active": "#21272D",
-  "eds-theme-color-button-tertiary-text": "#6B65E2",
+  "eds-theme-color-button-tertiary-text": "#707070",
```
dierat, line 210: There's one token update in here because the tertiary text was set to purple instead of gray. I assume that was a copy/pasta typo.

src/components/Button/Button.module.css
```diff
@@ -11,49 +11,28 @@
  * 4) Adds a gap between children to automatically handle spacing between icons and text
  */
 .button {
+  @mixin eds-theme-typography-button-label;
```
dierat, line 14: All the button sizes use the same typography styles now, so I moved it into the main styles.

src/components/Button/Button.module.css
```diff
@@ -11,49 +11,28 @@
...
   align-items: center;
   justify-content: center;
   margin: 0; /* 3 */
   gap: 0.75rem; /* 4 */
   border-width: var(--eds-theme-border-width);
   border-style: solid;
-  border-color: var(--eds-theme-color-primary-border);
   border-radius: var(--eds-theme-border-radius);
-  color: var(--eds-theme-color-primary-foreground);
-  background-color: transparent;
```
dierat, line 24: I removed the default color styles; the variant styles handle this now.

src/components/Button/Button.module.css
```diff
@@ -11,49 +11,28 @@
...
   cursor: pointer;
   transition: all var(--eds-anim-fade-quick) var(--eds-anim-ease);
 
-  /* TODO: make default color prop in Icon inherit or something? */
-  svg {
-    fill: currentColor;
-  }
```
dierat, line 31: The `Icon` now has a default color of `currentColor`, which devs can override by passing in some other color into `Icon`.

src/components/Button/Button.module.css
```diff
@@ -11,49 +11,28 @@
...
-  &:hover {
-    color: var(--eds-theme-color-primary-foreground-hover);
-    border-color: var(--eds-theme-color-primary-border-hover);
-  }
-
   &:focus {
-    background-color: transparent;
-  }
-
-  &:focus-visible {
```
dierat, line 42: `focus-visible` is used so the focus outline only appears when a user tabs to a button and not when a user clicks on a button. This is theoretically fine, but I don't want this in here unless we're sure we don't need to do anything else to make sure keyboard users do in fact get the outline. I believe `Along` is using a polyfill for this, and I don't know what the ramifications are of not using that (it might just be that Internet Explorer doesn't get the same functionality).

src/components/Button/Button.module.css
```diff
@@ -63,25 +42,28 @@
...
   &:focus {
-    background-color: var(--eds-theme-color-primary-background-hover);
-    color: var(--eds-theme-color-body-foreground-inverted);
+    border-color: var(--eds-theme-color-button-primary-border-hover);
+    background-color: var(--eds-theme-color-button-primary-background-hover);
+    color: var(--eds-theme-color-button-primary-text-hover);
   }
 
-  &:focus-visible {
-    @mixin focus;
```
dierat, line 77: We don't need this in the individual variants because it's up in the main styles.

src/components/Button/Button.module.css
```diff
@@ -63,25 +42,28 @@
...
+    border-color: var(--eds-theme-color-button-primary-border-active);
+    background-color: var(--eds-theme-color-button-primary-background-active);
+    color: var(--eds-theme-color-button-primary-text-active);
   }
 
   &:disabled,
   &:disabled:hover {
-    background-color: var(--eds-theme-color-disabled-background);
-    border-color: var(--eds-theme-color-disabled-border);
-    color: var(--eds-theme-color-disabled-foreground);
```
dierat, line 84: I also reordered the styles so it always goes border color, background color, text color.

src/components/Button/Button.module.css
```diff
@@ -63,25 +42,28 @@
...
+    border-color: var(--eds-theme-color-border-disabled);
+    background-color: var(--eds-theme-color-background-disabled);
+    color: var(--eds-theme-color-button-primary-text);
```
dierat, line 66: Just FYI there are no disabled tokens per variant; we just have the general disabled tokens (which are all set to neutral 300 I think).

src/components/Button/Button.module.css
```diff
@@ -145,16 +133,15 @@
  */
 .button--link {
   @mixin textLink;
-  align-items: flex-start;
```
dierat, line 148: This was causing the icon to be higher than the text.

src/components/Button/Button.module.css
```diff
@@ -145,16 +133,15 @@
...
+
   padding: 0;
   border: 0;
-  border-radius: 0;
```
dierat, line 151: I like the rounded focus outline for links.

src/components/Button/Button.module.css
```diff
@@ -209,7 +213,6 @@
  *    cause the button height to expand
  */
 .button--sm {
-  @mixin eds-theme-typography-button-label;
```
dierat, line 212: These are now in the top level button styles.

src/components/UtilityNav/UtilityNav.module.css
```diff
@@ -49,7 +49,8 @@
  * Utility nav link
  */
 .utility-nav__link {
-  @mixin textLink;
+  @mixin TextLinkInverted;
+  text-decoration: none;
```
dierat, line 53: The link token was broken before so this was inheriting the text color; now the links are the same color as the nav background. I used the same styling solution as the primary nav: https://github.com/chanzuckerberg/edu-design-system/blob/next/src/components/PrimaryNav/PrimaryNav.module.css#L36

src/design-tokens/mixins.css
```diff
@@ -7,20 +7,24 @@
 }
 
 @define-mixin textLink {
+  /* TODO: inherit size */
   @mixin eds-theme-typography-body-text-md;
-  color: var(--eds-theme-color-link-foreground);
+  /* TODO: underline change on hover */
   text-decoration: underline;
+  color: var(--eds-theme-color-text-link-strong);
```
dierat, line 14: The "strong" link tokens are purple whereas the default ones are gray. I need to verify that we're not using gray links in the platform; that might be the only variant we're not supporting that is being used.

src/components/Button/Button.module.css
anniehu4: mm yeah good call. we should check browser support for `focus-visible` now (at one point I thought it was only Chrome)

src/components/Button/Button.stories.tsx
```diff
@@ -11,74 +11,233 @@ export default {
 
 const Template: Story<Props> = (args) => <Button {...args} />;
 
-export const Default = Template.bind({});
-Default.args = { children: 'Button' };
+export const PrimaryNoIcon = Template.bind({});
+PrimaryNoIcon.args = { children: 'Button' };
```
anniehu4, line 15: nit: could we keep this `Primary`, `Secondary`, etc. (since they're the most common cases)

src/design-tokens/mixins.css
anniehu4: does that mean the future design direction is to have gray links by default? 🤔

anniehu4 (approved):

src/design-tokens/mixins.css
dierat: I'm not sure 🤔 I can't find any updated button/link designs, only the tokens page and some placeholder pages for buttons and links that I think are waiting on Sean's work. I wonder if we'll end up with two link variants, like `link-strong` and `link-default`?

dierat: I did notice gray links in the presentation today.

PR #953, opened 2022-04-06
Update existing icons to use `size`, `color` props

anniehu4:
### Summary:
[ch191478]

just some cleanup work 🙂 

### Test Plan:
- verify w/ Chromatic tests

### DIFF
--- a/docs/CODE_GUIDELINES.md
+++ b/docs/CODE_GUIDELINES.md
@@ -232,6 +232,20 @@ Use:
 }
 ```
 
+- Avoid using classes on one component to alter the styling of a different component. For example, instead of:
+
+```css
+.banner__icon {
+  fill: var(--eds-theme-color-utility-success-foreground);
+}
+```
+
+Use:
+
+```tsx
+<Icon color={EdsThemeColorUtilitySuccessForeground} />
+```
+
 ## Utility classes <a name="utility-classes"></a>
 
 EDS provides a number of utility classes (e.g. `u-margin-bottom-lg` and `u-padding-none`) that can be appended to components (e.g. `<TableCell className="u-padding-none">`) in order to achieve certain style results. Utility classes provide an additional layer to help "massage" components into place without having to overwhelm component styles with too many style variants.
--- a/src/components/AccordionPanel/AccordionPanel.module.css
+++ b/src/components/AccordionPanel/AccordionPanel.module.css
@@ -171,18 +171,7 @@
  */
 .accordion-panel__icon {
   margin-left: auto;
-  height: var(--eds-size-4);
-  width: var(--eds-size-4);
-  fill: var(--eds-theme-color-primary-foreground);
   transition: transform var(--eds-anim-fade-quick) var(--eds-anim-ease);
-}
-
-/**
- * Accordion panel icon within inverted accordion
- * 1) Change the icon fill color lighter
- */
- .accordion--inverted .accordion-panel__icon {
-  fill: var(--eds-theme-color-primary-foreground-inverted); /* 1 */
 }
 
 /**
--- a/src/components/AccordionPanel/AccordionPanel.tsx
+++ b/src/components/AccordionPanel/AccordionPanel.tsx
@@ -9,6 +9,7 @@ import React, {
   ReactNode,
 } from 'react';
 import styles from './AccordionPanel.module.css';
+import { EdsThemeColorIconNeutralStrong } from '../../tokens-dist/colors';
 import Icon from '../Icon';
 
 export interface Props {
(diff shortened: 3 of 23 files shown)


src/components/AccordionPanel/AccordionPanel.module.css
```diff
@@ -171,20 +171,9 @@
...
-  width: var(--eds-size-4);
-  fill: var(--eds-theme-color-primary-foreground);
   transition: transform var(--eds-anim-fade-quick) var(--eds-anim-ease);
 }
 
-/**
- * Accordion panel icon within inverted accordion
- * 1) Change the icon fill color lighter
- */
- .accordion--inverted .accordion-panel__icon {
```
anniehu4, line 184: unused

src/components/AccordionPanel/AccordionPanel.tsx
```diff
@@ -9,6 +9,7 @@ import React, {
   ReactNode,
 } from 'react';
 import styles from './AccordionPanel.module.css';
+import { EdsThemeColorIconNeutralStrong } from '../../tokens-dist/colors';
```
anniehu4, line 12: I double-checked that this gets included when we build to publish. It is a little weird to import from `tokens-dist` internally though (external imports will use `dist/tokens`) 🤔  Alternatively, we could rename tokens-dist -> tokens but then we have to live with a "design-tokens" folder and a "tokens" folder...

src/components/AccordionPanel/AccordionPanel.module.css
```diff
@@ -171,20 +171,9 @@
  */
 .accordion-panel__icon {
   margin-left: auto;
-  height: var(--eds-size-4);
-  width: var(--eds-size-4);
-  fill: var(--eds-theme-color-primary-foreground);
```
anniehu4, line 176: this token no longer exists so I replaced it with `icon-neutral-strong` throughout 🤷‍♀️  sure we'll polish it later

src/components/Button/Button.module.css
```diff
@@ -342,7 +337,6 @@
  * 1) Set SVG icon fill to the current text color
  */
 .button__icon-loading {
-  fill: currentColor; /* 1 */
```
anniehu4, line 345: this is now the default

src/components/FileUploadField/FileUploadField.stories.tsx
```diff
@@ -61,23 +61,23 @@ DefaultWithMultipleFiles.args = {
 export const Uploading = Template.bind({});
 Uploading.args = {
   files: files,
-  status: 'uploading',
+  statusLabel: 'uploading',
```
anniehu4, line 64: gets the stories to load correctly

src/components/ListDetail/ListDetail.module.css
```diff
@@ -200,32 +200,11 @@
 * 1) Used so the transparent icon doesn't show the line behind. 
 */
 .list-detail__icon {
+  --icon-size: var(--eds-size-3);
```
anniehu4, line 203: kept here b/c adding the `size` prop 3 different times in `ListDetail` felt unfortunate. but could add it there for consistency

src/components/LoadingIndicator/LoadingIndicator.tsx
```diff
@@ -1,26 +1,33 @@
...
   return (
     <div className={componentClassName} {...other}>
       <Icon
-        purpose="decorative"
-        name="spinner"
         className={styles['loading-indicator__icon']}
+        color={EdsThemeColorTextNeutralDefault}
+        name="spinner"
+        purpose="decorative"
+        size={size || '3rem'}
```
ahuth, line 30: Would "3rem" be better as a default value when destructuring? It's possible Storybook will do something smart and pick it up as a default value (haven't verified this myself, though).

ahuth (approved):

docs/CODE_GUIDELINES.md
```diff
@@ -232,6 +232,22 @@ Use:
...
+  .banner--success & {
+    fill: var(--eds-theme-color-utility-success-foreground);
+  }
+}
+```
+
+Use:
+
+```tsx
+<Icon color={EdsThemeColorUtilitySuccessForeground} />
```
dierat, line 248: May be worth having some sort of a check in here since the example CSS is targeting a specific variant, or maybe just drop the `banner--success` part in the example styles.
```suggestion
<Icon color={variant === 'success' && EdsThemeColorUtilitySuccessForeground} />
```

dierat (approved):

src/components/LoadingIndicator/LoadingIndicator.tsx
anniehu4: yup, can do

PR #954, opened 2022-04-06
chore: move heading to next

jinlee93:
[sc-187514]

### Summary:
- BM had ‘as’ prop required and ‘size’ optional, but changed to ‘size’ required ‘as’ optional to match v0
- BM had size as number, changed to same as tag ('h1', 'h2', 'h3', 'h4', 'h5', 'h6') to match v0
- `color` prop renamed to `variant`
- `yellow` removed since never had styling
- `info` deprecated
- Removed `spacing` prop
- Removed `weight` prop since we don’t recommend in v0 typography stylesheet to begin with
  - around 15 cases of it used in LP but 13 set to bold (which is what it is in next), 2 set to normal in <Cell> (maybe advise to change)
  - can custom classname style if needed (but shouldn’t)
 
### Test Plan:
- visual regression
image
- snap test
- unit test


src/components/Heading/Heading.tsx
```diff
@@ -1,56 +1,82 @@
...
+       */ ...other
+    }: Props,
+    ref: React.ForwardedRef<HTMLHeadingElement>,
+  ) => {
+    if (variant === 'info' && process.env.NODE_ENV !== 'production') {
+      console.warn(
+        'Info variant is deprecated and will be removed in an upcoming release. Please use the consider another variant instead.',
+      );
+    }
+    const TagName = as || size;
```
jinlee93, line 65: BM had `as` as required and `size` optional, and `size` would fall back to `as` if `size` not passed.
EDS v0 had `size` required and set `as` optional, and `as` would fall back to `size` if `as` not passed.
Kept the EDS way and fixed `Heading` usage throughout the repo, lmk if we should consider the BM way

src/components/Heading/Heading.tsx
```diff
@@ -1,56 +1,82 @@
...
+  children: React.ReactNode;
   className?: string;
-  /**
-   * Id of the heading
-   */
-  id?: string;
-  /**
-   * Size variants of the Heading component, ranging from 1-7. These numbers map to typographic sizing/style presets. This prop is used in instances where a Heading needs to use a specific tag (such as `h3`) for semantic reasons, but needs to be styled like something other than its default styles (such as an `h2` or `h1`).
-   */
-  size?: 1 | 2 | 3 | 4 | 5 | 6;
```
jinlee93, line 21: I used same typing for `size` as `as`. As used in EDS v0 and I feel would be less confusing

src/components/Heading/Heading.module.css
```diff
@@ -5,81 +5,64 @@
...
 }
 
-/**
- * H2 Heading
- * 1) Use this for the default heading if as = h2
- */
-h2.heading {
+.heading--size-h2 {
   @mixin eds-theme-typography-heading-2;
 }
```
dierat, line 24: This is interesting. They're ascribing styles based on the element, but then they override those with styles based on the size prop? I definitely prefer the EDS method where you define the style with one prop and then you can override the element with another prop.

src/components/Heading/Heading.module.css
```diff
@@ -5,81 +5,64 @@
 \*------------------------------------*/
 
 /**
- * 1) h1, h2, h3, h4, h5, h6
- * 2) Styles are either controlled by as or size declaration
+ * Remove top and bottom margins applied by browser for headings.
  */
 .heading {
-  @mixin reset;
```
dierat, line 12: I wonder if we were getting anything else from the reset mixin besides removing margin.

src/components/Heading/Heading.tsx
```diff
@@ -1,56 +1,82 @@
...
+  'white',
   /**
-   * Child node(s) that can be nested inside component
-   */
-  children: ReactNode;
+   * @deprecated Info variant is deprecated.
+   */ 'info',
+] as const;
+export type Variant = typeof VARIANTS[number];
+export type HeadingElement = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
```
dierat, line 19: I vaguely remember them specifically bringing up the difference between the EDS sizes vs the BM sizes, but I don't remember if there was a specific reason why they preferred one over the other – maybe because it's easier to insert or remove a size in the middle of the list? I personally like this one a lot better because it's somehow easier for me to remember than single digit numbers.

src/components/Heading/Heading.tsx
```diff
@@ -1,56 +1,82 @@
...
+      as,
+      children,
+      className,
+      variant,
+      size,
+      /**
+       * Components that wrap typography sometimes require props such as
+       * event handlers, tabIndex, etc. to be passed down into the element.
+       * TODO: add better typing or documentation for optional props,
+       * e.g. React.HTMLAttributes<HTMLHeadingElement>
```
dierat, line 55: I wonder if this todo comment is still relevant when we're letting all components pass down `other` now.

src/components/Heading/Heading.tsx
```diff
@@ -1,56 +1,82 @@
...
+  'neutral',
+  'success',
+  'warning',
+  'white',
   /**
-   * Child node(s) that can be nested inside component
-   */
-  children: ReactNode;
+   * @deprecated Info variant is deprecated.
+   */ 'info',
```
dierat, line 16: I didn't even realize blue was an option for text lol

dierat (approved):

src/components/Heading/Heading.module.css
jinlee93: Same here

src/components/Heading/Heading.module.css
jinlee93: It also removes padding, which never existed on heading

src/components/Heading/Heading.tsx
jinlee93: Hmm you're right, I'll change the comment to solely why we're passing other and rename `rest` to `other`, thanks

src/components/Heading/Heading.tsx
jinlee93: Hmm I had made a ticket to "deprecate the dark tooltip variant" which I used to deprecate in code, but not exactly fully remove. Maybe we can add onto the breaking changes doc?

src/components/Heading/Heading.tsx
jinlee93: Hmmm I think it may be that more sizes can be made ( >6 ) but heading tags are just `h1` - `h6`.  Which would also explain why the `as` prop is required and not the `size` tag. 
Thinking this, do you think we should go towards that kind of syntax?

src/components/Heading/Heading.tsx
dierat: I thought for the dark tooltip variant, we want to announce that we're removing it and then wait a little while so designers have time to adjust. I'm afraid we'll forget in that interim period.

For this blue text, I see it's being used in 3 places, all in the nux popover (`nuxPopover.tsx`, `nuxOnboardingTourBody.tsx`, `nuxOnboardingPopoverBody.tsx`). Those nuxes have blue backgrounds, so we're going to need to decide whether to override the the text to be blue for those (and then phase them out over time) or pick a different color.

src/components/Heading/Heading.tsx
dierat: Some random thoughts:
- Right now the way we handle `size` and `as` works better if they're using the same levels (so you only specify `as` if it's different than `size`). If we changed size to ints, I imagine you would specify both? ie `<Heading as="h4" size={4}>`
- On the other hand, making people specify `as` might make them think harder about what element a heading should actually be.
- I feel like the BM size levels are basically h levels but they just dropped the h because size 4 does correlate to h4, etc. Because they're basically the same, dropping the h seems pretty pointless to me.
- Limiting headings to only have 6 levels doesn't sound like a bad idea.

src/components/Heading/Heading.tsx
jinlee93: Ah that makes sense. Then should I add `info` styling (not sure how to pick a color/token here) or since styling can be done via `className`, have the product engineer pass in a styled class?

dierat: >Then should I add info styling (not sure how to pick a color/token here) or since styling can be done via className, or have the product engineer pass in a styled class?

Sorry, I'm not sure what you mean. Add info styling where/to what?

jinlee93: Ah to clarify:
- Currently, there's no styling for `.heading--info` like there are for the other color variants.
- If the `info` variant is being used, but deprecated, does that mean the styling should exist (via `.heading--info`) until we fully announce removal of `info`?
  - If so, which color should I assign to `.heading--info`?

dierat: Mmm, good question. Since it's only used in 3 places, I would just remove the variant here and override the styles in `traject` since the css variable will still exist there. I don't think we need to keep the variant around because no one's using it anyway. (I'm pretty sure we actually added those 3 headings when we did a heading migration project ourselves.)

anniehu4: > For this blue text, I see it's being used in 3 places, all in the nux popover (nuxPopover.tsx, nuxOnboardingTourBody.tsx, nuxOnboardingPopoverBody.tsx). Those nuxes have blue backgrounds, so we're going to need to decide whether to override the the text to be blue for those (and then phase them out over time) or pick a different color.

And some extra context -- on the design side, they want to move the NUXes to be purple. Not sure why they haven't yet though

dierat: That's really helpful, thanks @anniehu4 !

PR #956, opened 2022-04-07
feat(list detail): add responsive

csnizik:
### Summary:

Shortcut: https://app.shortcut.com/czi-edu/story/190151/list-detail-build-responsive-behavior

### DIFF
--- a/src/components/ListDetail/ListDetail.module.css
+++ b/src/components/ListDetail/ListDetail.module.css
@@ -14,7 +14,7 @@
   border-color: var(--eds-theme-color-border-neutral-subtle);
   border-radius: var(--eds-theme-border-radius);
 
-  @media all and (min-width: $eds-bp-lg) {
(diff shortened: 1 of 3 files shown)


ifrost1:
@csnizik 
- Let's remove that duplicate breadcrumb/page header stuff inside of the white panel. Let's have a discussion about that. All of this functionality looks right. Things to change:
- I think what we'll want is a `position: relative` to the parent `ListDetail` component. Then apply `position: absolute; top: 0` to the `list-detail__panel` That will take care of the wonkiness around z-index stuff being tucked under the header
- You likely have a `max-width` media query at 768px that is causing the back button to not show up. This is why we use `min-width` in most places for media queries.

I think the functionality looks good overall. Just tightening up styles so that the list detail panel stays within the box and doesn't overlay the entire window is the remaining stuff from a high level.

csnizik:
Closing this PR to open a new one for this ticket

PR #957, opened 2022-04-07
feat(button): polish button styles

dierat:
### Summary:
I think there are going to be changes when Sean gets back because he's been working on buttons, but this PR achieves parity with our current component on `main`.

Figma: https://www.figma.com/file/P9VPyI821gLq832tuU2WeY/UI-Kit-Master-(Core)?node-id=3072%3A11792

Little bit of polish details on the `Button` component:
- make icons bigger by default
- make icons smaller for link variant
- make gap between icon and text smaller for link variant
- make link text inherit surrounding text styles
- make link underline taller on hover/focus
- remove transition for motion reduce a11y OS setting
- port over tests from v0

### Screenshots
primary
secondary
tertiary
icon
link
destructive
other

### Test Plan:
Chromatic tests and test the hover/focus states in storybook.


src/components/Button/Button.module.css
```diff
@@ -34,6 +39,10 @@
    &:disabled:hover {
     cursor: not-allowed;
   }
+
+  @media screen and (prefers-reduced-motion) {
+    transition: none;
+  }
```
dierat, line 45: This isn't really important because the button is just changing color and it's really fast, but our current guideline is to add it whenever you have a transition.

src/components/Button/Button.test.tsx
```diff
@@ -1,24 +1,55 @@
+// import { generateSnapshots } from '@chanzuckerberg/story-utils';
```
dierat, line 1: This file's easier to review if you use the "Hide whitespace" github setting.

src/design-tokens/mixins.css
```diff
@@ -7,15 +7,14 @@
 }
 
 @define-mixin textLink {
-  /* TODO: inherit size */
```
dierat, line 10: This was added in the `Button` CSS.

dierat:
https://app.shortcut.com/czi-edu/story/187502/move-button-to-new-architecture

src/design-tokens/mixins.css
anniehu4: should we also add it here? for whatever other places might use this mixin?

anniehu4 (approved):

src/design-tokens/mixins.css
dierat: Oo, good call 👍 I moved it to the mixin, thanks!

PR #959, opened 2022-04-07
chore: move checkbox to next

jinlee93:
[sc-187506]
### Summary:
- move `Checkbox` from v0 to next
- deleted the <InlineCheckbox> components since this checkbox fulfills that role
- updates `Checkbox` use in `CheckboxFieldItem`

Screen Shot 2022-04-07 at 1 59 27 PM
Screen Shot 2022-04-07 at 1 59 29 PM
Screen Shot 2022-04-07 at 1 59 24 PM

### Test Plan:
- imported unit/snap tests
- CheckboxField stories render


src/components/Checkbox/Checkbox.module.css
```diff
@@ -7,158 +7,109 @@
...
+
+/**
+ * Text labelling the checkbox.
+ */
+.label {
+  color: var(--eds-theme-color-form-label-text);
+
+  /* Shift label so the checkbox always appears vertically centered
+  with the *first* line of text, in case label wraps. */
+  /* margin-top: calc((var(--checkbox-svg-size) - var(--label-line-height)) / 2); */
```
jinlee93, line 31: Label text used to be centered via a margin-top like so, but I changed centering by setting line-height (see `.label--sm` and `.label--md` styling). Lmk if should revert to this calculated margin-top styling, otherwise will remove these comments.

src/components/Checkbox/Checkbox.module.css
```diff
@@ -7,158 +7,109 @@
...
+}
+
+/**
+ * Size variants for the label.
+ * 1) line-height is set to be same as checkbox size to center label text relative to the checkbox.
+ */
+.label--sm {
+  @mixin eds-theme-typography-label-sm;
```
jinlee93, line 39: these label tokens are smaller than the v0 label sizes for both small and medium sizes. Should the checkbox itself be smaller? (is [20px](https://github.com/chanzuckerberg/edu-design-system/pull/959/files#diff-204ed52cdcdeb2aade99c93d518d9783f564e65ff7df2d65976757fa27f03a31R81) right now), or is there a mock I couldn't find?

src/components/Checkbox/Checkbox.module.css
```diff
@@ -7,158 +7,109 @@
...
+  @mixin eds-theme-typography-label-md;
+  line-height: var(--checkbox-svg-size); /* 1 */
+}
+
+/**
+ * Wrapper for the checkbox, including the visually hidden input element.
+ * 1) line-height is set to be same as checkbox size to equal checkbox size and prevent browser styling.
+ */
+.input__wrapper {
+  /* position: relative; */
```
jinlee93, line 53: Not sure what this was used for?

src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,117 +1,227 @@
...
-   * On Keydown
-   */
-  function onKeyDown(e: any) {
-    /**
-     * If enter key is pressed, trigger on change event
-     */
-    if (onChange && e.code === ENTER_KEYCODE) {
-      onChange(e);
-    }
-  }
```
jinlee93, line 86: There used to be an 'enter' key listener but I never got it to work, so just removed it


https://user-images.githubusercontent.com/86632227/162302708-b5b82f9d-c501-4d7f-ba5e-f14ed9c89097.mov

src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,117 +1,227 @@
...
-   * HTML id for the component
+   * Text to render in label.
    */
-  id?: string;
+  text: React.ReactNode;
   /**
-   * Indeterminate prop that turns the checkbox indeterminate state on
+   * Additional classnames passed in for styling.
    */
-  indeterminate?: boolean;
```
jinlee93, line 35: BM used a prop for indeterminate state but v0 uses checked prop = `'indeterminate'`, stuck with v0 style since we don't seem to have any extra styling for it besides an svg change for the checkmark

src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,117 +1,227 @@
...
+  },
+);
 
+Checkbox.displayName = 'Checkbox';
+
+/**
+ * Checkbox input element, exported for greater flexibility.
+ * You must provide an `id` prop and connect it to a visible label.
+ */
+const CheckboxInput = React.forwardRef<HTMLInputElement, InputProps>(
```
anniehu4, line 166: so currently, their [guidance](https://github.com/chanzuckerberg/edu-design-system/blob/next/docs/CODE_GUIDELINES.md#exactly-one-component-per-directory-and-per-folder) says that sub-components should go in their own folder, e.g. `CheckboxInput/CheckboxInput.tsx`.

What do you think / is that do-able here? If it takes too much work to re-finagle we could file a ticket to come back to it later

src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,117 +1,227 @@
...
+      htmlFor={htmlFor}
+      style={svgStyle}
+    >
+      {text}
+    </label>
   );
 };
+
+export default Checkbox;
+export { Checkbox, CheckboxInput, Label };
```
anniehu4, line 227: I believe we decided to stop doing `export` default so we could replace the `const` declarations above with `export const`

src/components/Checkbox/index.ts
```diff
@@ -1 +1,2 @@
-export { Checkbox as default } from './Checkbox';
+export { default, Checkbox, CheckboxInput, Label } from './Checkbox';
```
anniehu4, line 1:
```suggestion
export { Checkbox as default, CheckboxInput, Label } from './Checkbox';
```

src/components/CheckboxFieldItem/CheckboxFieldItem.tsx
```diff
@@ -112,25 +113,21 @@ export const CheckboxFieldItem = ({
     className,
     inverted && styles['checkbox-field__item--inverted'],
   );
+  const label = (
```
anniehu4, line 116: could we just pass `text` and remove the styling? Styling should happen through the `Checkbox` component now right

src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -1,23 +1,131 @@
...
+    label: undefined,
+  },
+};
+
+export const LongLabels = {
+  render: () => {
+    const label = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit';
+
+    return (
+      <div className="w-80 grid grid-cols-2 gap-4">
```
anniehu4, line 116: gotta remove the tailwind utilities 🥲

src/components/Checkbox/Checkbox.module.css
anniehu4: oh nice! idk why I didn't think of that 🙂  maybe we can set the line height in this `.label` class to DRY? and let's make sure to call out in the comment that the label is vertically centered with the first line of text (not the entire block of text)

anniehu4 (commented):
looks great! some food for thought but nothing blocking

src/components/Checkbox/Checkbox.module.css
jinlee93: I tried setting line-height in `.label` for aridness but if the `.label` block is placed higher than `.label--sm` or `.label--md` blocks, the line-heights being set in those variants override the line-height in `.label`. Didn't think having styling dependent on order in stylesheet was good, but lmk if we should proceed with that since I can place the `.label` lower.

anniehu4: ohhh right of course, that makes sense! thanks for explaining 👍

src/components/Checkbox/Checkbox.module.css
jinlee93: Sounds good, will message Ashley

src/components/Checkbox/Checkbox.module.css
jinlee93: Ah that makes sense, will reintroduce

src/components/Checkbox/Checkbox.module.css
```diff
@@ -7,158 +7,109 @@
...
-      opacity: 1;
-      transform: rotate(0deg); /* 2 */
-      border: 0; /* 2 */
-      width: 10px; /* 3 */
-      height: 2px; /* 3 */
-      background: var(--eds-theme-color-primary-foreground);
-      top: 9px; /* 4 */
-    }
-  }
+.checkbox__input+svg {
```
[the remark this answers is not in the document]
jinlee93: Will add whitespace for the combo selectors

src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -1,23 +1,131 @@
...
+};
 
 export default {
   title: 'Atoms/Forms/Checkbox',
   component: Checkbox,
-} as Meta;
+  args: defaultArgs,
+  argTypes: {
+    // For some reason Storybook does not infer all props correctly;
+    // we manually include the most relevant controls here.
```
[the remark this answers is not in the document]
jinlee93: Will check

src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,117 +1,227 @@
 import clsx from 'clsx';
-import React, {
-  useRef,
-  useEffect,
-  ChangeEventHandler,
-  MutableRefObject,
-} from 'react';
+import React from 'react';
+import { useUID } from 'react-uid';
```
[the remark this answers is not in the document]
jinlee93: Thanks!

src/components/Checkbox/Checkbox.tsx
jinlee93: Hmmmmm, I'll take a quick stab at it

src/components/Checkbox/Checkbox.tsx
jinlee93: Sounds good, will remove and update exports

src/components/Checkbox/Checkbox.stories.tsx
jinlee93: Ah shoot good catch

src/components/CheckboxFieldItem/CheckboxFieldItem.tsx
jinlee93: Definitely could, and will change to do so, was just being respectful of the custom styling that was here

src/components/Checkbox/index.ts
jinlee93: Will commit as advised, thanks!

src/components/Checkbox/Checkbox.stories.tsx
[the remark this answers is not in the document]
jinlee93: Looks like the size is being inferred properly, but `checked` isn't:
image
Should be: 
image

Will change to only have `argType` for `checked`

src/components/Checkbox/Checkbox.tsx
jinlee93: Separated `Checkbox`, `CheckboxInput`, and `CheckboxLabel`. Lmk how you feel about it and the separated `CheckboxLabel`. Would be easy to revert since I committed separately

src/components/CheckboxInput/CheckboxInput.module.css
```diff
@@ -0,0 +1,76 @@
+@import '../../design-tokens/mixins.css';
+
+
+/*------------------------------------*\
+    #CHECKBOXINPUT
```
dierat, line 5: I don't know how BM has been doing it, but can we put a space in here between "CHECKBOX" and "INPUT"? Caps are hard enough to read as it is 😭

jinlee93: Definitely could, will add space here and in `CheckboxLabel`

src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,117 +1,79 @@
...
 import styles from './Checkbox.module.css';
-import { ENTER_KEYCODE } from '../../util/keycodes';
+import CheckboxInput from '../CheckboxInput';
+import type { CheckboxInputProps } from '../CheckboxInput';
+import CheckboxLabel from '../CheckboxLabel';
+import type { CheckboxLabelProps } from '../CheckboxLabel';
```
anniehu4, line 8: nit, could be
```suggestion
import CheckboxInput, { CheckboxInputProps } from '../CheckboxInput';
import CheckboxLabel, { CheckboxLabelProps } from '../CheckboxLabel';
```

though if you want to be more explicit that's fine

src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,117 +1,79 @@
...
+  size?: CheckboxLabelProps['size'];
+};
 
 /**
- * Primary UI component for user interaction
+ * ```ts
+ * import {Checkbox} from "@chanzuckerberg/eds-components";
+ * ```
+ * ```ts
+ * import Checkbox, {CheckboxInput, Label} from '@chanzuckerberg/eds-components/lib/Checkbox';
```
anniehu4, line 33:
```suggestion
 * import {CheckboxInput, Label} from '@chanzuckerberg/eds-components';
```

also added to Breaking changes doc

src/components/CheckboxInput/CheckboxInput.module.css
```diff
@@ -0,0 +1,75 @@
...
+   */
+  .checkbox__input + svg {
+    vertical-align: top;
+  
+    display: inline-block;
+  
+    color: var(--eds-theme-color-icon-neutral-default);
+    --checked-color: var(--eds-theme-color-icon-brand-primary);
+  }
+  .checkbox__input:checked + svg>.checkmark__outline {
```
anniehu4, line 55:
```suggestion
  .checkbox__input:checked + svg > .checkmark__outline {
```

src/components/CheckboxInput/CheckboxInput.module.css
```diff
@@ -0,0 +1,75 @@
...
+    fill: var(--checked-color);
+    stroke: var(--checked-color);
+  }
+  .checkbox__input:checked + svg>.checkmark__content {
```
anniehu4, line 59:
```suggestion
  .checkbox__input:checked + svg > .checkmark__content {
```

src/components/CheckboxInput/CheckboxInput.stories.tsx
```diff
@@ -0,0 +1,75 @@
+import { StoryObj, Meta } from '@storybook/react';
+import React from 'react';
+import { CheckboxInput } from './CheckboxInput';
+
+export default {
```
anniehu4, line 5: Thanks for splitting these out! 💪 I think we can keep all the stories in the top-level file (`Checkbox.stories.tsx`) for easier discoverability. That matches their code guidelines as well

src/components/CheckboxLabel/CheckboxLabel.stories.tsx
```diff
@@ -0,0 +1,68 @@
+import { StoryObj } from '@storybook/react';
+import React from 'react';
+import { CheckboxLabel } from './CheckboxLabel';
+
+export default {
```
anniehu4, line 5: would also keep these stories in `Checkbox.stories.tsx`

anniehu4 (approved):
let's keep all the stories in `Checkbox.stories.tsx`, otherwise looks good 👍

src/components/CheckboxInput/CheckboxInput.module.css
jinlee93: Thanks for catching these :oof:

src/components/Checkbox/Checkbox.tsx
jinlee93: Thanks for adding to the breaking changes doc!

src/components/Checkbox/Checkbox.tsx
jinlee93: Will condense imports, thanks

PR #960, opened 2022-04-07
chore: update sb and remove unused dep

jinlee93:
### Summary:
- removes `react-docgen-typescript-loader` which was probably used for storybook typescript compatibility before it supported ts
- updates storybook from 6.4.19 to 6.4.20

### Test Plan:
- storybook dev server runs
  - docs is visible
- storybook dependent ci tasks pass
  - this also means storybook build works
  - docs is visible and [working](https://61313967cde49b003ae2a860-tiibezmyzj.chromatic.com/?path=/docs/checkbox--default)

### Additional thoughts:
- Consider using [webpack 5 for storybook](https://storybook.js.org/blog/storybook-for-webpack-5/), with the oncoming webpack5 upgrade for LP, as I think this would improve upon nested dependencies with security issues (e.g. https://github.com/storybookjs/storybook/blob/v6.4.20/lib/builder-webpack5/package.json#L88 and current webpack4 relying on a security advised `ansi-regex`:
image)


ahuth (approved):
Nice, LGTM

Also, dang, we should have github show yarn.lock as a generated file

anniehu4:
> Consider using [webpack 5 for storybook](https://storybook.js.org/blog/storybook-for-webpack-5/), with the oncoming webpack5 upgrade for LP, as I think this would improve upon nested dependencies with security issues (e.g. https://github.com/storybookjs/storybook/blob/v6.4.20/lib/builder-webpack5/package.json#L88 and current webpack4 relying on a security advised ansi-regex:

ooh that'd be cool. could you make a ticket in the "Design System Backlog" epic?

jinlee93:
> ooh that'd be cool. could you make a ticket in the "Design System Backlog" epic?

Ya will do

PR #962, opened 2022-04-07
feat(clickable): add ClickableStyle, use it to style Button and Link

dierat:
### Summary:
This PR takes the `Button` code, moves it into a new component called `Clickable`, then uses that component to style the existing `Button` and a new `Link` component. This uses the same structure as the existing EDS components but with the updated props and CSS from the Big Medium `Button`.

Some notes:
- `ClickableStyle` has no tests because they're covered by `Button` and `Link`
- `ClickableStyle` is almost a direct copy of `Button`; the things changed are:
  - naming (of course)
  - types (to accept different tags)
  - docstring
  - I alphabetized the props being passed into `Icon` and `Component` just because 🤷 

The components on `main` (for comparison):
- [ClickableStyle](https://github.com/chanzuckerberg/edu-design-system/tree/main/packages/components/src/ClickableStyle)
- [Button](https://github.com/chanzuckerberg/edu-design-system/tree/main/packages/components/src/Button)
- [Link](https://github.com/chanzuckerberg/edu-design-system/tree/main/packages/components/src/Link)

Sorry this is just a big PR – I just feel like the `ClickableStyle` doesn't make any sense unless you see the 3 components together. And most of it is copy/pasted, so hopefully it won't be too tiresome.

### Test Plan:
Verify chromatic didn't pick up any `Button` changes and all 3 look ok in storybook.


src/components/Button/Button.test.tsx
```diff
@@ -22,12 +22,6 @@ describe('<Button />', () => {
     expect(onClick).toHaveBeenCalled();
   });
 
-  it('renders as a link when provided href', () => {
-    render(<Button href="/test-url">test link</Button>);
-
-    expect(screen.getByRole('link')).toBeDefined();
-  });
-
```
dierat, line 30: The component doesn't support this anymore.

src/components/Button/Button.tsx
```diff
@@ -1,134 +1,37 @@
...
+ *
+ * This component is called `Button` because it should be used to make `<button>` elements;
+ * however, it can be styled to look like a link.
+ *
+ * If you need to style an `<a>` element to look like a button, please use the `Link` component.
+ * If you need to style a different element or component (e.g. `Link` from `react-router`) to
+ * look like a button or link, you can use the `ClickableStyle` component.
+ *
+ * In terms of the look and feel of the component in the UI, the `Button`, and `Link`, and `ClickableStyle`
+ * components are exactly the same.
```
dierat, line 16: This is the docstring in v0.

src/components/Button/Button.tsx
```diff
@@ -1,134 +1,37 @@
...
-      href,
-      loading,
-      onClick,
-      size = 'lg',
-      children,
-      type,
-      variant = 'primary',
-      ...other
-    }: Props,
+    { forwardRef, size = 'lg', type = 'button', variant = 'primary', ...other },
```
dierat, line 23: The `Button` uses `primary` as the default variant while `Link` uses `link`.

src/components/ClickableStyle/ClickableStyle.module.css
```diff
@@ -22,6 +22,7 @@
   border-style: solid;
   border-radius: var(--eds-theme-border-radius);
   cursor: pointer;
+  text-decoration: none;
```
dierat, line 25: I added this line because the link stories were showing underlines in the button variants. (This didn't come up before because none of the stories had an `href` prop, which is what activates the link styles.)

src/components/ClickableStyle/ClickableStyle.module.css
```diff
@@ -228,51 +229,52 @@
...
+.clickable-style--full-width {
   width: 100%;
 }
 
 /**
  * Loading icon rotiation style
  * 1) Set SVG icon fill to the current text color
  */
-.button__icon-loading {
+.eds-is-loading svg {
```
dierat, line 276: This was the original name for this class in the Big Medium component. It got changed around when I removed the `icon` style, so here I'm just putting it back the how it should have stayed.

src/tokens-dist/colors.ts
```diff
@@ -150,7 +150,7 @@ export const EdsThemeColorButtonTertiaryBackgroundActive = '#21272D';
 export const EdsThemeColorButtonTertiaryBorder = '#707070';
 export const EdsThemeColorButtonTertiaryBorderHover = '#707070';
 export const EdsThemeColorButtonTertiaryBorderActive = '#21272D';
-export const EdsThemeColorButtonTertiaryText = '#6B65E2';
+export const EdsThemeColorButtonTertiaryText = '#707070';
```
dierat, line 153: I don't know why this is showing up in this PR.. All the other tertiary text tokens are correct ('#707070'), but for some reason this one is still purple.

dierat: Also, I didn't manually make this change, or the other change in this file, but the updates are correct.

src/components/ClickableStyle/ClickableStyle.tsx
```diff
@@ -0,0 +1,134 @@
...
+        tabIndex={disabled ? -1 : undefined}
+        {...other}
+      >
+        {loading && (
+          <Icon
+            className={styles['clickable__icon-loading']}
+            name="spinner"
+            purpose="informative"
+            title="loading"
+            size="1.5em"
```
dierat, line 125: This icon is really big compared to our other ones because it fills the whole bounding box. When we clean up the whitespace on the other icons, we'll probably want them all to be 1.5em instead of 2.

src/components/Drawer/DrawerExample.tsx
```diff
@@ -24,7 +24,7 @@ export interface Props {
  */
 export const DrawerExample = ({ className, ...other }: Props) => {
   const [drawerOpen, setDrawerOpen] = useState(false);
-  const drawerButton = useRef() as MutableRefObject<HTMLSpanElement>;
```
dierat, line 27: There are 3 changes like this. I don't know why a button ref claimed to be a span or why it's only complaining now, but changing it to a button element makes sense to me 🤷

dierat:
https://app.shortcut.com/czi-edu/story/187518/decide-what-to-do-with-clickablestyle

dierat:
Ugh forgot to put reviewers on these 🤦 I do that all the time.

src/components/ClickableStyle/ClickableStyle.stories.tsx
```diff
@@ -0,0 +1,268 @@
...
+export default {
+  title: 'Molecules/Buttons/ClickableStyle',
+  component: ClickableStyle,
+} as Meta;
+
+const Template: Story<ClickableStyleProps<React.ElementType>> = (args) => (
+  <ClickableStyle {...args} as="button" />
+);
+
+export const Primary = Template.bind({});
```
anniehu4, line 16: we got rid of the `ClickableStyle` stories in EDS `main`, do we want them back now?

src/components/Button/Button.stories.tsx
```diff
@@ -1,15 +1,18 @@
...
+import type { ClickableStyleProps } from '../ClickableStyle';
 import Icon from '../Icon';
 
 export default {
   title: 'Molecules/Buttons/Button',
   component: Button,
 } as Meta;
 
-const Template: Story<Props> = (args) => <Button {...args} />;
+const Template: Story<ClickableStyleProps<React.ElementType>> = (args) => (
```
anniehu4, line 13: nit: may be easier to do
```
type Args = React.ComponentProps<typeof Button>;

const Template: Story<Args> = (args) ...
```

src/components/ClickableStyle/ClickableStyle.tsx
```diff
@@ -0,0 +1,134 @@
...
+   */
+  className?: string;
+  /**
+   * Disables the field and prevents editing the contents
+   */
+  disabled?: boolean;
+  /**
+   * ClickableStyle reference
+   */
+  forwardRef?: any;
```
anniehu4, line 23: why do we need the `forwardRef` prop? wouldn't people always pass `ref`

src/components/ClickableStyle/ClickableStyle.tsx
```diff
@@ -0,0 +1,134 @@
...
+   * The visible clickable children
+   */
+  children: string | ReactNode;
+  /**
+   * Determines type of clickable
+   * - **button** The clickable is a clickable button.
+   * - **submit** The clickable is a submit clickable (submits form data).
+   * - **reset** The clickable is a reset clickable (resets the form-data to its initial values)
+   */
+  type?: 'button' | 'reset' | 'submit';
```
anniehu4, line 50: The native HTML attributes are already included in the prop intersection `& React.ComponentProps<IComponent>`, so we don't _have_ to specify them here. It'd be more for documentation purposes then typing

Given that, wonder if some of these should only be declared in the `Button` component? specifically:
- disabled
- loading
- onClick
- type

don't seem relevant to `<a>` tags.

anniehu4 (approved):

src/components/ClickableStyle/ClickableStyle.stories.tsx
dierat: I was mostly using them for testing and didn't think too much about whether we should keep them. I actually do wonder if we should have them because it is possible for devs to use it in `traject` for `NavLink`s and such, but then we should probably add it to zeroheight too. What do you think?

src/components/ClickableStyle/ClickableStyle.module.css
dierat: You got it 👍

src/components/ClickableStyle/ClickableStyle.tsx
dierat: This was copied over from `Button`. I don't think we need it; removing.

src/components/ClickableStyle/ClickableStyle.tsx
dierat: I do wonder if that's making too many assumptions about what components might be passed into `ClickableStyle`, but I can't think of anything aside from `NavLink` and `Link` from react-router. I'll leave `onClick` though since there could be other logic in there for links (like analytics).

src/components/ClickableStyle/ClickableStyle.stories.tsx
dierat: I'm going to go ahead and merge this for now, but let's talk about this some more later outside of this PR.

PR #963, opened 2022-04-08
chore: add grid styling to checkbox long labels story

jinlee93:
### Summary:
- This was accidentally removed from original merged move checkbox pr
- Reintroduces grid to show that label wraps while `CheckboxInput` centers with first line of label

### Test Plan:
- `CheckboxInput` centers with first line of label in Storybook/Chromatic


jinlee93:
Thought this would work but it doesn't, will just create new branch off `next`

PR #965, opened 2022-04-08
Ifrost/189630 map new tokens components

ifrost1:
### Summary:
- Updated token mapping to component CSS to reflect the new architecture in Figma
- Updated logic around inverse tokens displaying (add dark background to inverse token names)
- Updated the way form fields are handled. Reduce CSS used by passing `isError` and `disabled` props down to inputs themselves.
- Add larger `yellow` ramp per Courtney's request

### Test Plan:

`yarn test` runs with deprecation message only
`yarn lint` runs without errors
`yarn types` runs without errors
`yarn start` runs Storybook


src/components/Avatar/Avatar.module.css
```diff
@@ -6,5 +6,3 @@
  * 1) 
  */
 /* stylelint-disable-next-line block-no-empty */
```
dierat, line 8: Should we just remove the whole file if there are no styles in it?

src/components/Accordion/Accordion.module.css
```diff
@@ -13,10 +13,149 @@
...
 .accordion--inverted {
-  color: var(--eds-theme-color-inverted-text);
-  border-bottom-color: var(--eds-theme-color-neutral-subtle-border-inverted);
+  color: var(--eds-theme-color-text-neutral-default-inverse);
+  border-bottom-color: var(--eds-theme-color-border-neutral-strong);
+}
+
+/**
+ * Individual accordion panel
+ */
```
anniehu4, line 30: currently, in the CODE_GUIDELINES, we have:

> Compound component children have an associated `.module.css` file, and child component styles will contain styles relevant to the subcomponent element (e.g. `AccordionPanel.module.css` would begin with `.accordion__panel { ... }`).

To be consistent, could we keep the accordion panel styles in `AccordionPanel.module.css`?

src/components/Panel/Panel.module.css
```diff
@@ -8,10 +8,10 @@
  * 1) A container for content
  */
 .panel {
-  background: var(--eds-color-neutral-white); /* TODO */
+  background: var(--eds-theme-color-background-neutral-default); /* TODO */
```
anniehu4, line 11: nit: is this still a TODO?

src/components/Breadcrumbs/Breadcrumbs.module.css
```diff
@@ -57,6 +57,7 @@
 .breadcrumbs__icon {
   margin-left: var(--eds-size-1);
   margin-right: var(--eds-size-1);
+  fill: var(--eds-theme-color-icon-neutral-strong);
```
anniehu4, line 60: as of https://github.com/chanzuckerberg/edu-design-system/pull/953, Icon `color` and `size` should be set as props from the component. In `BreadcrumbsItem`, it's already invoking
```
      <Icon
        className={styles['breadcrumbs__icon']}
        color={EdsThemeColorIconNeutralStrong}
        name="chevron-right"
        purpose="decorative"
        size="0.75rem"
      />
```
so the extra `fill` styling isn't needed

src/components/FieldNote/FieldNote.module.css
```diff
@@ -11,31 +11,37 @@
...
+  position: relative;
+  bottom: 1px;
+
   .eds-is-error & {
     display: inline-block;
-    margin-right: var(--eds-size-1);
     position: relative;
     top: 3px;
+    margin-right: var(--eds-size-1);
+    fill: currentColor; /* 1 */
```
anniehu4, line 38: This is the default behavior for icons now, so extra `fill` styling not needed. See prop destructuring in `Icon.tsx`

src/components/LoadingIndicator/LoadingIndicator.module.css
```diff
@@ -12,6 +12,9 @@
 }
 
 .loading-indicator__icon {
+  height: 3rem;
+  width: 3rem;
+  fill: var(--eds-theme-color-icon-brand-primary);
```
anniehu4, line 17: `color` and `size` props should be updated in `LoadingIndicator.tsx` instead, when calling `<Icon />`

src/components/PrimaryNav/PrimaryNav.module.css
```diff
@@ -61,5 +63,8 @@
  * Primary nav icon
  */
 .primary-nav__icon {
+  height: var(--eds-size-2-and-half);
+  width: var(--eds-size-2-and-half);
+  fill: var(--eds-theme-color-icon-neutral-default-inverse);
```
anniehu4, line 68: `color` and `size` props should be updated in `PrimaryNavItem.tsx` instead, when calling `<Icon />`

src/components/SelectField/SelectField.module.css
```diff
@@ -72,7 +54,10 @@
 .select-field__icon {
   position: absolute;
   top: 50%;
+  height: var(--eds-size-3);
+  width: var(--eds-size-3);
   transform: translateY(-50%);
   right: var(--eds-size-1);
   pointer-events: none; /* 1 */
+  fill: var(--eds-theme-color-icon-brand-primary);
```
anniehu4, line 62: `color` and `size` props should be updated in `SelectField.tsx` instead, when calling `<Icon />`

anniehu4 (commented):
couple of comments about CSS structure and recent changes to the `Icon` component API (happy to discuss more if you want! those PRs just landed on Friday so you might not have seen them yet). otherwise looks good 🙂

src/components/Accordion/Accordion.module.css
ifrost1: That's fine. I saw `PrimaryNav` and `UtilityNav` not have that and I think I rolled with that. Happy to break that back out

ifrost1: I can do a pass at all of the ones that are using the parent to make sure they're all consistent

src/components/Breadcrumbs/Breadcrumbs.module.css
ifrost1: @anniehu4 More of a higher-level question: What's the reasoning behind using a `color` prop here instead of a custom property in the CSS? Is this more of a "support the existing system" prop?

anniehu4: yep, that's the main motivation! It also cleans up the CSS in certain situations (e.g. different colored icons in different banner variants don't need extra CSS). Don't remember if there were other considerations when we first added the prop

anniehu4 (approved):
thanks! think this is good to go after you finish updates / build passes

src/components/Accordion/Accordion.module.css
ifrost1: Took a pass at the ones I worked on. Should be good now I think

PR #966, opened 2022-04-09
chore: move dropdown to next

jinlee93:
[sc-187516]
[sc-187517]
### Summary:
- moves dropdown and dropdown button to next
- DropdownTrigger `children` since more complex button triggers can be used, such as in ‘UsingFunctionChildrenProp’ story

### Test Plan:
- Unit/snap test
- Chromatic


src/components/Dropdown/Dropdown.module.css
```diff
@@ -0,0 +1,139 @@
...
+
+  margin-top: var(--eds-size-1);
+  padding: var(--eds-size-1) 0 var(--eds-size-1);
+
+  background-color: var(--eds-theme-color-background-neutral-default);
+
+  border-radius: var(--eds-theme-border-radius);
+  box-shadow: var(--eds-theme-box-shadow);
+
+  z-index: var(--eds-z-index-100); /* 1 */
```
jinlee93, line 57: This used to be 10, but 100 seems to be lowest in tokens

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,464 @@
...
+   *
+   * When not using the compact variant, if optionsClassName is provided please
+   * include the width property to define the options menu width.
+   */
+  optionsClassName?: string;
+};
+
+type RenderProp<Arg> = (arg: Arg) => ReactNode;
+type PropsWithRenderProp<RenderPropArg> = {
+  children?: ReactNode | RenderProp<RenderPropArg>;
```
jinlee93, line 80: This is for `<DropdownTrigger>`. Kept the `children` prop since more complex button triggers can be used, such as in ‘UsingFunctionChildrenProp’ story.

jinlee93:
Pointer cursors, background colors and font body's were added for explicitness, used to rely on body/browser styling.

src/components/DropdownButton/DropdownButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+    return (
+      <button
+        className={clsx(styles['dropdown-button'], className)}
+        ref={ref}
+        {...rest}
+      >
+        {/* Wrapping span ensures that `children` and icon will be correctly pushed to
+            either side of the button even if `children` contains more than one element. */}
+        <span>{text}</span>
+        <Icon name="expand-more" purpose="decorative" size="1.25rem" />
```
dierat, line 29: This isn't a regression, but I'm just wondering – should we be rotating thing icon on open/close? It's kinda weird that it's always pointed down or whatever, right? Example in the new `Accordion`: https://github.com/FB-PLP/traject/blob/c344cc8526a5d1d1a957e17bcb0779e56da58508/app/assets/javascripts/v2/core/EDSCandidates/Accordion/Accordion.tsx#L82

src/components/DropdownButton/DropdownButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * ```ts
+ * import {DropdownButton} from "@chanzuckerberg/eds-components";
+ * ```
+ *
+ * A styled button with an expand icon to be used in triggering Popovers, Dropdowns, etc.
+ */
+export const DropdownButton = forwardRef<HTMLButtonElement, Props>(
+  ({ className, text, ...rest }, ref) => {
```
dierat, line 19:
```suggestion
  ({ className, text, ...other }, ref) => {
```

src/components/DropdownButton/DropdownButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+    return (
+      <button
+        className={clsx(styles['dropdown-button'], className)}
+        ref={ref}
+        {...rest}
```
dierat, line 24:
```suggestion
        {...other}
```

src/components/DropdownButton/DropdownButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+ * import {DropdownButton} from "@chanzuckerberg/eds-components";
+ * ```
+ *
+ * A styled button with an expand icon to be used in triggering Popovers, Dropdowns, etc.
+ */
+export const DropdownButton = forwardRef<HTMLButtonElement, Props>(
+  ({ className, text, ...rest }, ref) => {
+    return (
+      <button
+        className={clsx(styles['dropdown-button'], className)}
```
dierat, line 22: Do we want to pull out the `clsx` logic into a `componentClassName` variable every time, for consistency?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,464 @@
...
+export type VariantType = 'compact' | 'full';
+
+type ListboxProps = ComponentProps<typeof Listbox>;
+type DropdownProps = ListboxProps & {
+  /**
+   * Text for the dropdown label.
+   *
+   * This is an alternative to passing <Dropdown.Label> in via `children`.
+   * If you pass in `labelText`, we expect `buttonText`, and `options` props
+   * as well and no `children`.
```
dierat, line 23:
```suggestion
   * This is an alternative to passing <Dropdown.Label> in via `children`.
   * If you pass in `labelText`, we expect `buttonText` and `options` props
   * as well and no `children`.
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,464 @@
...
+   * and `options` props as well and no `children`.
+   */
+  buttonText?: ReactNode | string;
+  /**
+   * All options to be displayed in the dropdown list, passed in as
+   * an array of objects.
+   *
+   * This is an alternative to passing in the options via children.
+   * If you pass in `options`, we expect `labelText` (or `aria-label`),
+   * and `buttonText` props as well and no `children`.
```
dierat, line 49:
```suggestion
   * This is an alternative to passing in the options via children.
   * If you pass in `options`, we expect `labelText` (or `aria-label`)
   * and `buttonText` props as well and no `children`.
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,464 @@
...
+ *
+ * ```
+ * const options = [
+ *   ...
+ * ];
+ *
+ * return (
+ *   <Dropdown
+ *     aria-label="Options"
+ *     className="w-60"
```
dierat, line 243: I wonder if we should remove tailwind from the examples since we're not using it anymore even through `traject` still is 🤔

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,464 @@
...
+    className: clsx(
+      styles['dropdown'],
+      className,
+      compact && styles['dropdown--compact'],
+    ),
+    // Provide a wrapping <div> element for the dropdown. This is needed so that any props
+    // passed directly to this component have a corresponding DOM element to receive them.
+    // Otherwise we get an error.
+    as: 'div' as const,
+    ...rest,
```
dierat, line 296:
```suggestion
    ...other,
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,464 @@
...
+    <Listbox.Label className={clsx(styles['label'], className)}>
+      {children}
+    </Listbox.Label>
+  );
+};
+
+const DropdownTrigger = function (
+  props: PropsWithRenderProp<{ disabled: boolean; open: boolean }>,
+) {
+  const { children, className, ...rest } = props;
```
dierat, line 377:
```suggestion
  const { children, className, ...other } = props;
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,464 @@
...
+  const { compact } = useContext(DropdownContext);
+
+  return (
+    <Listbox.Button
+      // Render as a fragment instead of the default element. We're rendering our own element in
+      // the render prop to control styling and positiong, and we don't want to end up with
+      // duplicate buttons.
+      as={React.Fragment}
+      {...rest}
```
dierat, line 386:
```suggestion
      {...other}
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,464 @@
...
+  const { compact, optionsAlign, optionsClassName } =
+    useContext(DropdownContext);
+
+  const computedClassName = clsx(
+    styles['dropdown__options'],
+    className,
+    optionsAlign === 'right' && styles['dropdown__options--align-right'],
+    optionsClassName || (!compact && styles['dropdown__options--full-width']),
+  );
+  return <Listbox.Options className={computedClassName} {...rest} />;
```
dierat, line 416: Just curious – why did you use `computedClassName` instead of `componentClassName`?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,464 @@
...
+};
+
+const DropdownOption = function (props: DropdownOptionProps) {
+  const { children, className, ...rest } = props;
```
dierat, line 420: Damn so much prop deconstruction in this component huh lol
```suggestion
  const { children, className, ...other } = props;
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,464 @@
...
+    selected: boolean;
+  };
+
+  return (
+    <Listbox.Option
+      // Render as a fragment instead of the default <li>. We're rendering our own <li> in the
+      // render prop to control active/selected styling, and we don't want to end up with duplicate
+      // <li>'s.
+      as={React.Fragment}
+      {...rest}
```
dierat, line 434:
```suggestion
      {...other}
```

src/components/Dropdown/Dropdown.module.css
```diff
@@ -0,0 +1,139 @@
...
+  margin-bottom: var(--eds-size-half);
+}
+
+/**
+ * Visually hides the label element if a aria-label is passed,
+ * but element is kept for screen reading accessibility.
+ */
+.sr-only {
+  @mixin visuallyHide;
+}
```
dierat, line 40: Do we need this? Could we just put the `u-is-vishidden` class name on the element?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,464 @@
...
+   * visible label, you must provide an `aria-label` for screen readers.
+   * If you pass in an `aria-label`, you don't need `labelText` or <Label>.
+   */
+  'aria-label'?: string;
+  /**
+   * Text for the dropdown label.
+   *
+   * This is an alternative to passing <Dropdown.Button> in via children.
+   * If you pass in `buttonText`, we expect `labelText` (or `aria-label`),
+   * and `options` props as well and no `children`.
```
dierat, line 40:
```suggestion
   * This is an alternative to passing <Dropdown.Button> in via children.
   * If you pass in `buttonText`, we expect `labelText` (or `aria-label`)
   * and `options` props as well and no `children`.
```

dierat:
There's another dropdown component in the `next` branch called `Select`. It's a barebones HTML dropdown with no substantial styling.

select component in storybook with the dropdown open

@anniehu4 and I talked about this and we think the main reason to use a library like `HeadlessUI` for a dropdown is how easy they make it to style the component. `select` is notoriously difficult to style because it looks very different in different browsers. We had talked about investigating how hard it would be to style the Big Medium `Select`, because it would be nice to have such a simple component, but looking around at conversations online about the difficulties of styling `select` makes me think that we should just leave this component as-is. In which case, we probably want to delete `Select` to avoid confusion over which component to use.

src/components/DropdownButton/DropdownButton.tsx
jinlee93: Ya will do!

src/components/Dropdown/Dropdown.tsx
jinlee93: Hmmmmm, good question. I'll change the className to something "BEM"-ey to show styling capability via classname passing

src/components/Dropdown/Dropdown.tsx
jinlee93: Dear god I missed so many of these, thanks for catching

src/components/Dropdown/Dropdown.module.css
jinlee93: I think this is needed for component functionality.
As for hiding via `u-is-vishidden` I think I would need to import that, which is the same as `@mixin visuallyHide` anyways iirc?

jinlee93:
> @anniehu4 and I talked about this and we think the main reason to use a library like `HeadlessUI` for a dropdown is how easy they make it to style the component. `select` is notoriously difficult to style because it looks very different in different browsers. We had talked about investigating how hard it would be to style the Big Medium `Select`, because it would be nice to have such a simple component, but looking around at conversations online about the difficulties of styling `select` makes me think that we should just leave this component as-is. In which case, we probably want to delete `Select` to avoid confusion over which component to use.

Sounds good, can remove `Select` and re-map imports to use this `Dropdown`.

src/components/DropdownButton/DropdownButton.tsx
jinlee93: Mmmm good q, I'll rotate the icon when opening/closing

jinlee93: Mmmm good q, I'll rotate the icon when opening/closing

src/components/Dropdown/Dropdown.tsx
jinlee93: I doh'd and thought `computed` was `component`, will switch to `componentClassName`

src/components/Dropdown/Dropdown.tsx
jinlee93: Yea...

src/components/DropdownButton/DropdownButton.tsx
jinlee93: Rotated in https://github.com/chanzuckerberg/edu-design-system/pull/966/files#diff-dd4dd3af268081f171000b5868e6a442be7aa697fe47f8f999e6d456020fc1ccR30-R45, lmk if you think the rotation is too fast (available tokens were 150ms or 400ms)

src/components/DropdownButton/DropdownButton.module.css
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * The caret icon to decorate the dropdown button, animated to rotate.
+ * 1) Turns off animation if user prefers reduced motion
+ */
+.dropdown-button__icon {
+  transform: rotate(0);
+
+  transition: transform var(--eds-anim-move-quick) ease-out;
+  @media (prefers-reduced-motion) {
+    transition-property: none; /* 1 */
```
dierat, line 39: Just curious – why did you decide to use `transition-property: none` (as opposed to `transition: none` or `transition-duration: 0`)? I don't think any specific way is superior, but I do kinda want us to pick one and stick with it. (I did hear an argument to use `transition-duration: 0.0001` or something which is that if there's some code that is checking for when the transition ends, reducing the duration will still trigger that code whereas removing the transition altogether won't. But we don't usually have logic like that, so I dunno if I really want to optimize for it.)

src/components/Dropdown/Dropdown.module.css
dierat: How about we keep this but rename it to `u-is-vishidden` for consistency?

src/components/DropdownButton/DropdownButton.tsx
dierat: >lmk if you think the rotation is too fast (available tokens were 150ms or 400ms)

I definitely prefer 150 to 400, but 150 does feel a bit fast. What if we added a third token between them (like `--eds-anim-move-medium`) and set it to 250 or 300?

jinlee93: I'm down to do that, will add a 300 token and use that

src/components/Dropdown/Dropdown.module.css
jinlee93: Sounds good, will rename to `u-is-vishidden`

src/components/DropdownButton/DropdownButton.module.css
jinlee93: Good question, the `transition-none` TW utility class is `transition-property: none`, so I just used that, but if `transition: none` also disables transition animations, I will go with that since it feels more self explanatory imo

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -118,6 +118,10 @@ export const LongButtonText: StoryObj<Args> = {
       </Button>
     ),
   },
+  parameters: {
+    // Is flaky on chromatic, disabling for now
```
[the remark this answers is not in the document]
jinlee93: Think we can! My initial thought was to disable the flakiness first and then play with the delay timing to find a good time, but now I feel that since people haven't been complaining about this flakiness (afaik) I'll just set a delay time instead.

dierat (approved):
Looks great, thanks!

PR #968, opened 2022-04-11
feat(clickable-style): split link variant into default and strong

dierat:
### Summary:
One last thing for the `Button` component for now – there are two link token types (default and strong), but we only have one link variant. This PR splits that variant into two different ones: `linkDefault` and `linkStrong`. The default link is gray and the strong one is purple.

I decided to call the default link variant `linkDefault` instead of `link` because I felt that this name better communicates that there are other link variants. I'm afraid that because there has only been one link variant and people used to pass in a color, it will be confusing that `link` only produces a gray link, and they'll instinctively override the color instead of looking to see if there's another variant to get the purple color.

I'm open to opinions on whether this API direction makes sense and whether we should consider other options. But I really don't want to add the `color` prop back if possible.

### Test Plan:
Storybook and chromatic.


dierat:
https://app.shortcut.com/czi-edu/story/187503/move-link-to-new-architecture

src/components/ClickableStyle/ClickableStyle.module.css
```diff
@@ -155,15 +156,27 @@
...
+
   &:disabled,
   &:disabled:hover {
     color: var(--eds-theme-color-text-disabled);
   }
 }
 
-/* Style icons that may be in the clickable style children */
-.clickable-style--link svg {
-  --icon-size-default: 1.25em;
```
dierat, line 166: I moved this into the main style block because that felt more consistent with the rest of the codebase.

src/design-tokens/mixins.css
```diff
@@ -9,11 +9,11 @@
 @define-mixin textLink {
   @mixin eds-theme-typography-body-text-md;
   text-decoration: underline;
-  color: var(--eds-theme-color-text-link-strong);
+  color: var(--eds-theme-color-text-link);
```
dierat, line 12: I changed this one back to the default tokens. I feel like this mixin is kind of pointless now because our components are set up in such a way that you can just use one of those, but I'll leave this for now in case I'm missing something.

.storybook/pages/ProjectOverview/ProjectOverview.tsx
```diff
@@ -52,7 +52,7 @@ export const ProjectOverview = () => {
           </TextPassage>
           <Button
             className={utilityStyles['u-margin-bottom-lg']}
-            variant="link"
+            variant="linkDefault"
```
dierat, line 55: This is the one place that was using a `link` variant.

src/design-tokens/mixins.css
jinlee93: I think I agree with `--eds-theme-color-text-link` as far as nomenclature goes and if the underlying tier 1 token itself changes, this should stay the same

jinlee93:
>  I'm afraid that because there has only been one link variant and people used to pass in a color, it will be confusing that link only produces a gray link, and they'll instinctively override the color instead of looking to see if there's another variant to get the purple color.
I agree with your sentiment here but wonder what the consuming engineers think

src/components/ClickableStyle/ClickableStyle.tsx
```diff
@@ -56,7 +56,8 @@ export type ClickableStyleProps<IComponent extends React.ElementType> = {
     | 'secondary'
     | 'tertiary'
     | 'icon'
-    | 'link'
+    | 'linkDefault'
+    | 'linkStrong'
```
anniehu4, line 60: the (rare) times I'm seeing multi-word variants in this code base, they've been dash-separated (e.g. `link-default`, `link-strong`). We should use that for consistency unless you feel strongly

src/components/ClickableStyle/ClickableStyle.stories.tsx
```diff
@@ -209,12 +209,12 @@ IconClickableStyleIconOnlySmall.args = {
 };
 
 export const Link = Template.bind({});
-Link.args = { children: 'Clickable Style', variant: 'link' };
+Link.args = { children: 'Clickable Style', variant: 'linkDefault' };
```
anniehu4, line 212: can totally table this discussion -- we got rid of the `ClickableStyle` stories in EDS `main`, do we want them back now?

src/components/ClickableStyle/ClickableStyle.tsx
dierat: Ah, I didn't see that; happy to change it 👍

dierat:
This PR is now irrelevant due to https://github.com/chanzuckerberg/edu-design-system/pull/977. I'll pull the change in the mixin into a different PR.

PR #970, opened 2022-04-12
chore: move buttongroup to next

jinlee93:
[sc-187505]
### Summary:
- Move buttongroup to next

### Test Plan:
- Unit/snap tests
- Chromatic

### DIFF
--- a/src/components/Button/Button.stories.tsx
+++ b/src/components/Button/Button.stories.tsx
@@ -81,18 +81,18 @@ SecondaryMedium.args = { children: 'Button', variant: 'secondary', size: 'md' };
 export const SecondarySmall = Template.bind({});
 SecondarySmall.args = { children: 'Button', variant: 'secondary', size: 'sm' };
 
-export const Teritary = Template.bind({});
-Teritary.args = { children: 'Button', variant: 'tertiary' };
+export const Tertiary = Template.bind({});
+Tertiary.args = { children: 'Button', variant: 'tertiary' };
 
-export const TeritaryDisabled = Template.bind({});
-TeritaryDisabled.args = {
+export const TertiaryDisabled = Template.bind({});
+TertiaryDisabled.args = {
   children: 'Button',
   variant: 'tertiary',
   disabled: true,
 };
 
-export const TeritaryLeftIcon = Template.bind({});
-TeritaryLeftIcon.args = {
+export const TertiaryLeftIcon = Template.bind({});
+TertiaryLeftIcon.args = {
   children: (
     <>
       <Icon purpose="decorative" name="arrow-back" />
@@ -102,8 +102,8 @@ TeritaryLeftIcon.args = {
   variant: 'tertiary',
 };
 
-export const TeritaryRightIcon = Template.bind({});
-TeritaryRightIcon.args = {
+export const TertiaryRightIcon = Template.bind({});
+TertiaryRightIcon.args = {
   children: (
     <>
       Button
@@ -113,11 +113,11 @@ TeritaryRightIcon.args = {
   variant: 'tertiary',
 };
 
-export const TeritaryMedium = Template.bind({});
-TeritaryMedium.args = { children: 'Button', variant: 'tertiary', size: 'md' };
+export const TertiaryMedium = Template.bind({});
+TertiaryMedium.args = { children: 'Button', variant: 'tertiary', size: 'md' };
 
-export const TeritarySmall = Template.bind({});
-TeritarySmall.args = { children: 'Button', variant: 'tertiary', size: 'sm' };
+export const TertiarySmall = Template.bind({});
+TertiarySmall.args = { children: 'Button', variant: 'tertiary', size: 'sm' };
 
 export const IconButtonLeftIcon = Template.bind({});
 IconButtonLeftIcon.args = {
--- a/src/components/ButtonGroup/ButtonGroup.module.css
+++ b/src/components/ButtonGroup/ButtonGroup.module.css
@@ -3,31 +3,37 @@
 \*------------------------------------*/
 
 /**
- * 1) A group of buttons displayed in a horizontal fashion
+ * A group of buttons displayed in an organized fashion.
  */
 .button-group {
-  display: flex;
-  align-items: center;
-  gap: var(--eds-size-2);
+  display: inline-grid;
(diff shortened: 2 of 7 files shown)


src/components/ButtonGroup/ButtonGroup.module.css
```diff
@@ -3,31 +3,37 @@
...
+ * Hopefully no one actually ever uses this... 
  */
-.button-group--stacked {
-  display: flex;
-  flex-direction: column;
-  align-items: flex-start;
+.button-group--spacing-max.button-group--vertical {
+  width: 100%;
+  height: 100%;
+  align-content: center;
```
jinlee93, line 38: Is this to prevent breaking styling? Should we even include if we don't want people using it this way?

src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -2,53 +2,55 @@ import clsx from 'clsx';
...
+  spacing = '1x',
+  orientation = 'horizontal',
+}: Props) {
+  if (
+    spacing === 'max' &&
+    orientation === 'vertical' &&
+    process.env.NODE_ENV !== 'production'
+  ) {
+    console.warn(
+      "*** Weird prop combo warning ***: Are you sure you want max spacing *and* vertical orientation? It's a valid combination, but it's extremely unlikely to ever be used intentionally.",
```
jinlee93, line 45: see https://github.com/chanzuckerberg/edu-design-system/pull/970/files#r847854570

src/components/Modal/ModalExample.tsx
```diff
@@ -84,7 +84,7 @@ export const ModalExample = ({
           </TextPassage>
         </ModalBody>
         <ModalFooter>
-          <ButtonGroup align="right">
+          <ButtonGroup>
```
jinlee93, line 87: Didn't seem to make any change (left is `next`, right is this feature branch local):
image

src/components/Button/Button.stories.tsx
```diff
@@ -81,18 +81,18 @@ SecondaryMedium.args = { children: 'Button', variant: 'secondary', size: 'md' };
 export const SecondarySmall = Template.bind({});
 SecondarySmall.args = { children: 'Button', variant: 'secondary', size: 'sm' };
 
-export const Teritary = Template.bind({});
-Teritary.args = { children: 'Button', variant: 'tertiary' };
+export const Tertiary = Template.bind({});
```
dierat, line 84: lol ty

src/components/ButtonGroup/ButtonGroup.module.css
```diff
@@ -3,31 +3,37 @@
...
- * Responsive button group
- * 1) A button group that's stacked by default then moves side-by-side
- *    when space becomes available 
+ * Spacing variants.
  */
-.button-group--responsive {
-  @media all and (max-width: $eds-bp-md) {
-    align-items: flex-start;
-    flex-direction: column;
-  }
```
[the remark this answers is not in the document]
dierat: Mmm.. I think I'd want to see it in context. Like what kind of UI were these buttons being used in and how many were there? I personally feel like I would rather have them wrap.

src/components/ButtonGroup/ButtonGroup.module.css
dierat: Yeah, I was just trying to cover our bases. I can't see a use case where this is needed, but I didn't like the idea that there was a combo you could make with the props that we just didn't do even though it's not difficult to implement. But I say it's up to you – if you feel strongly that it should go, remove it and update the warning to say it's not supported.

src/components/Modal/ModalExample.tsx
dierat: I think theirs was broken. I would leave the `align="right"` and celebrate that the chromatic test will show it's now working.

dierat (approved):

src/components/Modal/ModalExample.tsx
jinlee93: I am down to keep not having `align` prop. It's a matter of adding `width: 100%` and `justify-content: end`, but since the `ButtonGroup` itself is a wrapping `div`, the parent can style it as such.

src/components/ButtonGroup/ButtonGroup.module.css
jinlee93: I can kind of see a case where on a vertical bar, there might be buttons spaced max and placed vertically.... (not a designer lol). Will keep

src/components/ButtonGroup/ButtonGroup.module.css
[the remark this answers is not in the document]
jinlee93: I feel the same about wrapping. I converted to `display: inline-flex` and added a `wrap` prop to control wrapping behavior. Let me know how you feel about it since it was originally implemented as grid.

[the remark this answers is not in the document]
jinlee93: Ack pushed a wrong branch (jlee/ButtonGroup not jlee/Buttongroup), pushed this branch now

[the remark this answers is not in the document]
anniehu4: When would we want the `'nowrap'` or `'wrap-reverse'` cases (or was that for illustration purposes)? I think it'd be enough to always `wrap` and if people need, they can override the style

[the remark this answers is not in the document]
jinlee93: I'm down to always `wrap`. Was trying to allow more options, but definitely do agree most cases would just use `wrap`. One case I can think of for `nowrap` is when the intention is for the users to scroll the buttons to prevent them from taking up too much space on the screen, but as you said, devs can override if necessary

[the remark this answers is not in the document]
anniehu4: hmm yeah I see -- not sure we want to encourage that design tho. and should be rare in any case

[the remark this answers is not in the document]
dierat: I like the updated solution of allowing it to wrap whenever it gets too small, regardless of the screen it's on. It feels very flexible and maintainable 👍

dierat (approved):

PR #971, opened 2022-04-12
chore: change FormGroup to Fieldset and move to next

jinlee93:
### Summary:
- Change `FormGroup` to `Fieldset` since that seems to be the previous default use case
- Change `Fieldset.title` from accepting `as = 'legend'` and just set it as `<legend>`

### Test Plan:
- Unit/snap tests
- Chromatic


jinlee93:
RFC for now pending some q's:
1. Original BM `fieldset` / `legend` has `required`, `requiredLabel`, and `optionalLabel`
  - Should we also support these?
2. Currently exporting `legend` as `Fieldset.Legend`, similarly as `FormGroup.Title` in v0.
  - The `<legend>` element should be only used in a `Fieldset`, but it seems like the BM component `Legend` was being used outside of their BM component `Fieldset`. I would like to replace the BM `Legend` with the `Fieldset.Legend` that is present here, but would need compatibility with existing `required` `optionalLabel` `requiredLabel` props or remove such usage in their respective stories. Then I would like to put the `Legend` stories within `Fieldset` stories since it should be a subcomponent and follow the code guidelines

src/components/Fieldset/Fieldset.tsx
```diff
@@ -1,70 +1,89 @@
...
-export const Fieldset = ({
-  className,
-  legend,
-  legendAfter,
-  hideLegend,
-  required,
+const Legend: React.FC<LegendProps> = ({ children, className }) => {
+  const componentClassName = clsx(className, styles['legend']);
+  return <legend className={componentClassName}>{children}</legend>;
+};
```
jinlee93, line 53: Code guidelines suggest moving this subcomponent out of this `Fieldset` file into its own file. However, there is an existing `Legend` which seems to be being used outside of the `Fieldset` component and in `<fieldset>` html tags. Should cases be supported? If so should `Fieldset` export its own `Legend`?

Also the existing `Legend` also accepts `required`, `requiredLabel`, and `optionalLabel`. Should we support those?

anniehu4: oh interesting ... it looks like only a few use cases, so I'd say easier to consolidate and only use the existing `Legend` component?

I'd be curious to see it used in context, but keeping `required` & `optionalLabel` makes sense to me (to match how we'd handle Inputs) -- less clear when we'd use `requiredLabel`

src/components/Fieldset/Fieldset.tsx
```diff
@@ -1,70 +1,89 @@
...
+
+/**
+ * Helper sub-component for styling the control elements in the component.
+ */
+const Items: React.FC<ItemsProps> = ({
```
anniehu4, line 58: let's remove FC syntax:
```
const Items = ({...}: ItemsProps) => {
```

and move this out to a `FieldsetItems` file

src/components/Fieldset/Fieldset.tsx
```diff
@@ -1,70 +1,89 @@
...
+ * ```ts
+ * import {Fieldset} from "@chanzuckerberg/eds-components";
+ * ```
+ *
+ * A container for a fieldset including a legend and one or more form inputs.
+ *
+ * Uses `fieldset` and `legend` elements by default but can be customized.
+ *
+ * If you're not using a `fieldset` element, please consider whether you should be
+ * using a `ul` instead.
```
anniehu4, line 77:
```suggestion
```

src/components/Fieldset/Fieldset.tsx
```diff
@@ -1,70 +1,89 @@
...
+  const componentClassName = clsx(className, styles['fieldset']);
+  return <fieldset className={componentClassName}>{children}</fieldset>;
+}
+
+Fieldset.Legend = Legend;
+Fieldset.Items = Items;
+
+Fieldset.displayName = 'Fieldset';
+Fieldset.Legend.displayName = 'Fieldset.Legend';
+Fieldset.Items.displayName = 'Fieldset.Items';
```
anniehu4, line 89: nit: do we need these since we aren't forwarding refs?

src/components/Fieldset/Fieldset.tsx
jinlee93: I'm down with a `FieldsetLegend` component but naming wise should we also consider moving/renaming `Fieldset.Items`?
Otherwise we'd end up with something like:
```
import {Fieldset, FieldsetLegend} from '@cz/eds';
<Fieldset>
  <FieldsetLegend>Legend</FieldsetLegend> //or text or something
  <Fieldset.Items>
    {inputs}
  </Fieldset.Items>
</Fieldset>
```

src/components/Fieldset/Fieldset.module.css
```diff
@@ -5,18 +5,28 @@
 \*------------------------------------*/
 
 /**
- * 1) Grouping of form fields with a legend
+ * The fieldset groups a legend and several controls.
+ * 1) Overrides default browser styling.
  */
 .fieldset {
-  @mixin fieldsetStyles;
```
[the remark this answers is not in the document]
jinlee93:
```
@define-mixin fieldsetStyles {
  display: block;
  margin: 0;
  padding: 0;
  border: none;
}

@define-mixin legendReset {
  display: inline-block;
  padding: 0;
  border: none;
}
```
Think I could be using these instead but default display for `fieldset` is block anyways.
As for legend, default display is also `block` but for this case don't think it would matter?

src/components/Fieldset/Fieldset.tsx
jinlee93: Right, sounds good

src/components/Fieldset/Fieldset.tsx
jinlee93: Probably not, will remove. These will also change with the components being extracted into own files

src/components/Fieldset/Fieldset.tsx
```diff
@@ -1,70 +1,35 @@
...
-  /**
-   * Required is passed down to the legend to display "optional" or not
-   */
-  required?: boolean;
-  /**
-   * String for the required label to add additional information if needed.
-   */
-  requiredLabel?: string;
-}
+} & React.HTMLAttributes<HTMLFieldSetElement>;
```
jinlee93, line 18: Appends the native html attributes of `fieldset` element

src/components/CheckboxField/CheckboxField.tsx
```diff
@@ -27,10 +29,6 @@ export interface Props {
    * Used as helper text or error message
    */
   fieldNote?: string | ReactNode;
-  /**
-   * Toggles the visibility of the form control legend
-   */
-  hideLegend?: boolean;
```
jinlee93, line 33: This has been removed since the dev can just not include the subcomponent `<FieldsetLegend>`

src/components/Fieldset/Fieldset.stories.tsx
```diff
@@ -1,18 +1,54 @@
...
-  </Fieldset>
-);
+export const CheckboxField = CheckboxFieldStory;
+CheckboxField.args.fieldNote =
+  'Checkbox fieldset usage. See CheckboxField Stories for more examples.';
+CheckboxField.parameters = { chromatic: { disableSnapshot: true } }; // Disabled since already being snapped in original story.
 
-export const Default = Template.bind({});
-Default.args = { legend: 'Fieldset Legend' };
+export const RadioField = RadioFieldStory;
```
jinlee93, line 51: RadioField and CheckboxField default stories were imported to display here to show usage of `Fieldset` and for regression sanity. Snap tests are left on for dev checking, chromatic tests are turned off since they're already being tested in their respective story file. 
Let me know if the stories should just be built in this file.

src/components/CheckboxField/CheckboxField.tsx
dierat: What was it even doing?

src/components/CheckboxField/CheckboxField.tsx
```diff
@@ -47,10 +45,6 @@ export interface Props {
    * HTML label text
    */
   label?: string;
-  /**
-   * Slot for node to appear directly after field legend. Typically used to include a Toolip
-   */
-  legendAfter?: ReactNode;
```
dierat, line 53: I wonder what this use case actually looked like. Wish they had an example somewhere.

src/components/Fieldset/Fieldset.stories.tsx
dierat: I personally appreciate the deduplication – less code and easier to maintain if there are changes to the `RadioField` component API.

src/components/FieldsetItems/FieldsetItems.module.css
```diff
@@ -0,0 +1,10 @@
+/*------------------------------------*\
+    #FIELDSET ITEMS
+\*------------------------------------*/
+
+/**
+ * The contents of the fieldset. Spaces them apart.
+ */
+ .items > :not(:last-child) {
```
dierat, line 8: Should this class be `.fieldset-items`?

src/components/FieldsetLegend/FieldsetLegend.module.css
```diff
@@ -0,0 +1,26 @@
...
+\*------------------------------------*/
+
+/**
+ * A label that's rendered as a <legend> for fieldsets.
+ * It contains the same characteristics as a label (ability to add flag for optional field, etc),
+ * but semantically/stylistically a bit different
+ * 1) Removes some default browser styles.
+ * 2) Adjust margin between legend and option list
+ */
+.legend {
```
dierat, line 14: Should this class be `.fieldset-legend`?

src/components/FieldsetLegend/FieldsetLegend.tsx
```diff
@@ -0,0 +1,43 @@
...
+  required = true,
+  ...other
+}: FieldsetLegendProps) => {
+  const componentClassName = clsx(styles['legend'], className);
+  return (
+    <legend className={componentClassName} {...other}>
+      {text}{' '}
+      {!required && (
+        <span className={styles['legend__flag']}>{optionalLabel}</span>
+      )}
```
dierat, line 40: This is interesting. It's assumed that all form groups are required unless specified otherwise? I think in the product they do it the other way – assume all controls are optional unless marked as required. But that's for individual controls; I'm not used to seeing form *groups* labeled as required or optional.

src/components/CheckboxField/CheckboxField.tsx
jinlee93: It was applying that `u-is-vishidden` class that hides the legend on the screen but is available for screen readers. I think simply applying `aria-label` or any similar aria naming attributes on the `fieldset` is a better alternative since it is more native

src/components/CheckboxField/CheckboxField.tsx
jinlee93: It was [appending this after](https://github.com/chanzuckerberg/edu-design-system/pull/971/files#diff-070bda2885e117424aa887bf9acfd459bd90ae98e74213a05058371eeeba02a8L68-L70) the `optionalLabel` and was probably used for elements such as a Tooltip as suggested above. For our tooltip at least we can pass the ref, add a trigger child element within the `Fieldset` wrapper, or any other approach that our Tooltip allows

src/components/FieldsetItems/FieldsetItems.module.css
jinlee93: Yup, will change

src/components/FieldsetLegend/FieldsetLegend.module.css
jinlee93: Yup, will change, thanks

src/components/FieldsetLegend/FieldsetLegend.tsx
jinlee93: Hmm thanks for bringing up this consideration.
Since the `required` prop is used for presentation of this `optionalLabel` only, will refactor to remove the `required` prop and make the `optionalLabel` (or rename more appropriately) an optional enum between `(required)` or `(optional)`.
Let me know if you think there's value in keeping the `required` label in case of using a red asterisk to indicate required status, or alternatively the `optionalLabel` enum should be `(required)`, `(optional)`, and `*`

jinlee93: Ok, how do you feel about https://github.com/chanzuckerberg/edu-design-system/pull/971/files#diff-2fbe26d11df22cb350e98e747e952b42398d0b34fe51b333efcbc6fa2adf3d16R13:

image
image

src/components/CheckboxField/CheckboxField.tsx
```diff
@@ -47,22 +45,10 @@ export interface Props {
...
-  optionalLabel?: string;
-  /**
-   * Indicates that field is required for form to be successfully submitted
-   */
-  required?: boolean;
-  /**
-   * String for the required label to add additional information if needed.
-   */
-  requiredLabel?: string;
+  optionalLabel?: '(required)' | '*' | '(optional)';
```
anniehu4, line 51: ambivalent on the prop naming, but I think we should get rid of the `'*'` option. in the a11y training they were saying it's extra cognitive load vs. just writing out the word 🙂

anniehu4 (approved):

src/components/CheckboxField/CheckboxField.tsx
jinlee93: Sounds good, will remove `'*'`

jinlee93:
Linking to storybook
[sc-183038]
[sc-187508]

PR #973, opened 2022-04-13
feat(banner): update banner main component styles

dierat:
### Summary:
First part of pulling the `Banner` code from `main` into the component on `next`.

What this PR does:
- updates the props
- updates the tokens
- updates icons to be determined by the variant
- supports horizontal and vertical orientations (including making the orientation vertical on narrow screens)
- makes the close button match the text color when not being interacted with
- adds the Title and Description subcomponents as separate components

What this PR does *not* do (these will be handled separately):
- handle close button hover/focus/active styles
- change color of links in the text
- change the color of buttons in the `action` prop
- pull out styles so they can be reused by the `Toast` component

#### All stories
https://user-images.githubusercontent.com/7761701/163277242-2f5b8fb5-7437-42d4-beb5-1495d0f88406.mp4

#### Responsive
https://user-images.githubusercontent.com/7761701/163277321-84c490f0-9a22-4210-8d34-85382bc2e2df.mp4

### Test Plan:
Chromatic + storybook.


src/components/Banner/Banner.module.css
```diff
@@ -6,163 +6,186 @@
...
+  border-top-color: var(--border-dark-color); /* 3 */
 }
 
 /**
- * Banner success
- * 1) The thick left border operates as a foreground band of color,
- *    so foreground rather than border is used
+ * Banner informational icon
+ * 1) Selector doubled up to bump specificity to not interact with .icon.
+ *    See code guidelines for more info.
```
anniehu4, line 40:
```suggestion
```

src/components/Banner/Banner.module.css
```diff
@@ -6,163 +6,186 @@
...
-   */
-  .banner--success & {
-    fill: var(--eds-theme-color-utility-success-foreground);
-  }
+/*------------------------------------*\
+    # HORIZONTAL ORIENTATION
+\*------------------------------------*/
 
+/* TODO is there a token for this? */
+@media screen and (min-width: 768px) {
```
anniehu4, line 85: ya you'll need to use the `$` variable syntax because native CSS variables won't work -- but check `breakpoints.js` for reference. I believe this would be `$eds-bp-md`

src/components/BannerDescription/BannerDescription.stories.tsx
```diff
@@ -0,0 +1,16 @@
+import { Story, Meta } from '@storybook/react';
+import React from 'react';
+
+import { BannerDescription, Props } from './BannerDescription';
+
+export default {
```
anniehu4, line 6: Based on [CODE_GUIDELINES](https://github.com/chanzuckerberg/edu-design-system/blob/next/docs/CODE_GUIDELINES.md#conventions-for-compound-components), these subcomponents shouldn't have their own story or test files -- rather, it'll be covered by the Parent component

src/components/Banner/Banner.tsx
```diff
@@ -2,108 +2,162 @@ import clsx from 'clsx';
...
+   *
+   * Vertical banners are used in narrow areas, like sidebars
    */
   orientation?: 'vertical';
   /**
-   * Optional heading text for the banner that appears above banner body message
+   * The child node(s) contains the banner message
    */
-  title?: string;
+  text?: ReactNode;
```
anniehu4, line 34: I worry that using `text` here contradicts our own guidance to "default to `text` for short strings of text". I know we haven't revisited this in a while 😅 , but could we consider passing
- `title` prop to replace `BannerTitle`
- `titleAs` or `headingAs` prop to toggle heading element if needed
-  `description` to replace `BannerDescription`

? right now the Banner API is kind of a snowflake, so trying to balance that with making sure people use the `action` prop

src/components/Banner/Banner.module.css
dierat: Sorry, I'm having trouble seeing the difference between the comment and your suggested change. Could you please explain it to me?

src/components/Banner/Banner.module.css
dierat: Awesome, thanks!

src/components/BannerDescription/BannerDescription.stories.tsx
dierat: Oops, I missed that, thanks!

src/components/Banner/Banner.tsx
dierat: Ah, thanks! I also felt like `text` didn't quite fit but also thought it fit better than `textContent`. I'm just wondering if the types for these props should be `string` or whether we want to allow them to pass in react nodes in the rare case that they need to override the text styles for some reason 🤔

dierat: Actually the `description` definitely needs to be a `React.Node` so they can embed links. I could see the `title` needing test ids or other attributes, so I think `React.Node` is the way to go in this case.

src/components/Banner/Banner.module.css
```diff
@@ -6,163 +6,203 @@
...
-  margin-right: var(--eds-size-2);
-  fill: var(--eds-theme-color-neutral-medium-foreground);
-  width: var(--eds-size-3);
-  height: var(--eds-size-3);
+.banner__close-btn.banner__close-btn {
+  position: absolute;
+  top: var(--eds-size-half);
+  right: var(--eds-size-half);
+  /* TODO: handle hover styles */
+  color: inherit; /* 2 */
```
dierat, line 77: The `Button` is being changed in https://github.com/chanzuckerberg/edu-design-system/pull/977 to accommodate this need. I could stack this PR on top of that one in order to use those styles, but I've been getting tired of rebasing from complex PRs, so I'm instead going to wait until that one merges.

src/components/Banner/Banner.tsx
anniehu4: sounds fine to me 👍

src/components/Banner/Banner.tsx
```diff
@@ -2,108 +2,205 @@ import clsx from 'clsx';
...
 export const Banner = ({
-  iconTitle,
+  action,
   className,
-  title,
-  children,
-  iconName,
-  dismissible,
+  description,
+  descriptionAs = 'p',
```
anniehu4, line 123: nit: seems unused right now

src/components/Banner/Banner.stories.tsx
```diff
@@ -1,156 +1,219 @@
...
+          href="/"
+          onClick={(event) => event.preventDefault()}
+          variant="link"
+        >
+          click into the course
+        </Button>
+        .
+      </>
+    ),
+    heading:
```
anniehu4, line 25: suuper nit bc I got confused 😅  is there a reason you didn't name these `description` & `title`?

anniehu4 (approved):

src/components/Banner/Banner.tsx
dierat: Whoops! Thanks for pointing that out! ❤️

src/components/Banner/Banner.stories.tsx
dierat: Nah, this was just leftover from before. Updating, thanks!

PR #974, opened 2022-04-13
[RFC] feat(banner): style links and buttons in Banner

dierat:
### Summary:
When we revamped the `Button` props, we slimmed down the potential variants to only purple and gray buttons. All the orange, red, and green buttons were removed because they're only used in the `Banner` and `Toast` components.

Since they were removed from the main `Button` component, we're going to need to get a little creative for using them in messaging components. This PR presents one possible direction. It's very rough, so you're going to have to use your imagination a bit.

Keep in mind that the variants we specifically need are:
- link variants in the banner text block
- outline buttons in the "action" prop
- icon buttons for the close button

Possible directions:
1) The styles are baked into the `Banner` CSS. We'll need some of the base `Button` styles, so those can be moved to mixins and reused here. (This is the direction shown in this PR.)
  - Pros: devs just pass in links and buttons and the styles are taken care of.
  - Cons: it's pretty messy, and it may be confusing that they pass in a purple or gray button/link and then the colors are overridden. Devs may waste time trying to figure out where to get these colored buttons from, not realizing that the component takes care of this for them. 
2) A new component (called something like `MessagingButton` because it will also be used in the `Toast component`). It would work like the old `Button` in that it can create all three variants and colors.
  - Pros: all the styles will be contained in this other component, it will be easy to reuse between the `Banner` and `Toast`, and it will be fairly intuitive because it's similar to the old version of the `Button`.
  - Cons: we're going to have to deal with the colors/variants API again (which may be especially confusing now because the main `Button` works very differently), and we're going to have to support rendering `button` and `a` tags because we don't actually know what the "action" functionality will be.
3) Change the "action" prop from being a react node to being an object that accepts attributes of the button, and we put the button together ourselves (it would look something like `actionProps={children: "Button children", onClick={() => console.log("clicked")!} `).
  - Pros: The API makes a little more sense because they're not passing in a purple or gray button and having the styles overridden – we can specify that we're going to create a button that matches the banner variant.
  - Cons: We're still going to have to handle the styling in some way, either by having a separate component or baking the styles into the `Banner`, and this especially doesn't help us when it comes to the close button or links in the text block.
4) Bring back the color x shape spread in a different component that the `Button` and `Link` rely on (possibly taking the place of `ClickableStyle` or be another component that `ClickableStyle` wraps, see https://github.com/chanzuckerberg/edu-design-system/pull/962 for what `ClickableStyle` is). The `Button` would still only produce the variants it does now (it's API wouldn't change), but there would be another component that only the "Messaging" components use. I'm thinking this would combine with option # 3 so devs wouldn't actually use this component.
  - Pros: effectively pushes the styles into another component to reduce messiness, and that component could be reused across all of these locations to reduce duplication
  - Cons: links would still need to baked into the messaging component styles though unless we do want to give devs access to this component which kind of defeats the purpose of reducing the variant options in `Button`, and the token to styles mapping would be really strange for the primary, secondary, tertiary, etc variants (we might want to just drop that token structure and have a set of button tokens that just includes the full color x shape spread)

Thoughts? Sorry if this is confusing; I feel like it requires a bit of context on the button variants and current tokens structures to understand why this is so complicated.

### Banner images (for reference on which buttons/links we need)
Figma: https://www.figma.com/file/P9VPyI821gLq832tuU2WeY/UI-Kit-Master-(Core)?node-id=3049%3A9285
the banner component in figma

Just FYI I did specifically ask design if we could get away with using the gray or purple outline buttons in the banner, but they were very clear that's not an option.

### Button variants
These are the variants that the `Button` component now supports. You used to be able to get any color and shape combo from this component, but it now only produces the variants used in the product outline of the `Banner` and `Toast` components.

new-variants

### Test Plan:
Chromatic and storybook.

### DIFF
--- a/src/components/Banner/Banner.module.css
+++ b/src/components/Banner/Banner.module.css
@@ -57,6 +57,10 @@
 
 .banner__textContent {
   gap: var(--eds-size-1);
+
+  a {
+    color: inherit;
+  }
 }
 
 .banner__action {
@@ -131,6 +135,7 @@
 /**
  * Banner neutral
  * 1) The thick left border matches the icon color
+ * 2) Links inside the Banner get their color from the banner variant styles
  */
 .banner--neutral {
   --border-light-color: var(--eds-theme-color-border-neutral-subtle);
@@ -138,11 +143,17 @@
 
   background: var(--eds-theme-color-background-neutral-subtle);
   color: var(--eds-theme-color-text-neutral-default);
+
+  a:hover,
+  a:focus {
+    color: var(--eds-theme-color-text-link-default-hover); /* 2 */
+  }
 }
 
 /**
  * Banner success
  * 1) The thick left border matches the icon color
+ * 2) Links inside the Banner get their color from the banner variant styles
  */
 .banner--success {
   /* TODO: move these styles to mixins or tokens so the Toast can reuse them */
@@ -151,11 +162,17 @@
 
   background: var(--eds-theme-color-background-utility-success);
   color: var(--eds-theme-color-text-utility-success);
+
+  a:hover,
+  a:focus {
+    color: var(--eds-color-other-mint-800); /* 2 */
+  }
 }
 
 /**
  * Banner warning
  * 1) The thick left border matches the icon color
+ * 2) Links inside the Banner get their color from the banner variant styles
  */
 .banner--warning {
   --border-light-color: var(--eds-theme-color-border-utility-warning-subtle);
@@ -163,11 +180,17 @@
 
   background: var(--eds-theme-color-background-utility-warning);
   color: var(--eds-theme-color-text-utility-warning);
+
+  a:hover,
+  a:focus {
+    color: var(--eds-color-other-orange-800); /* 2 */
+  }
 }
 
 /**
  * Banner alert
  * 1) The thick left border matches the icon color
+ * 2) Links inside the Banner get their color from the banner variant styles
  */
 .banner--alert {
   /* TODO: move these styles to mixins or tokens so the Toast can reuse them */
@@ -176,11 +199,17 @@
 
   background: var(--eds-theme-color-background-utility-error);
   color: var(--eds-theme-color-text-utility-error);
+
+  a:hover,
+  a:focus {
+    color: var(--eds-color-other-ruby-800); /* 2 */
+  }
 }
 
 /**
  * Banner brand
  * 1) The thick left border matches the icon color
+ * 2) Links inside the Banner get their color from the banner variant styles
  */
 .banner--brand {
   --border-light-color: var(--eds-theme-color-border-brand-primary-subtle);
@@ -188,4 +217,69 @@
 
   background: var(--eds-theme-color-background-brand-primary-default);
   color: var(--eds-theme-color-text-brand-primary);
+
+  a:hover,
+  a:focus {
+    color: var(--eds-color-brand-grape-800); /* 2 */
+  }
+}
+
+/*------------------------------------*\
+    # ACTION STYLES
+\*------------------------------------*/
+
+.banner__action > * {
+  /* TODO: move to mixin */
+  @mixin eds-theme-typography-button-label;
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  margin: 0; /* 3 */
+  gap: 0.75rem; /* 4 */
+  border-width: var(--eds-theme-border-width);
+  border-style: solid;
+  border-radius: var(--eds-theme-border-radius);
+  cursor: pointer;
+  transition: all var(--eds-anim-fade-quick) var(--eds-anim-ease);
+
+  &:focus {
+    @mixin focus;
+  }
+
+  &:disabled,
+  &:disabled:hover {
+    cursor: not-allowed;
+    border-color: var(--eds-theme-color-border-disabled);
+    background-color: transparent;
+    color: var(--eds-theme-color-text-disabled);
+  }
+
+  svg {
+    --icon-size-default: 2em; /* 5 */
+  }
+
+  @media screen and (prefers-reduced-motion) {
+    transition: none;
+  }
+}
+
+/* stylelint-disable-next-line no-descending-specificity */
+.banner--success .banner__action > * {
+  /* TODO: move these colors to tokens */
+  border-color: var(--eds-color-other-mint-700);
+  background-color: transparent;
+  color: var(--eds-color-other-mint-700);
+
+  &:hover,
+  &:focus {
+    border-color: var(--eds-color-other-mint-700);
+    background-color: var(--eds-color-other-mint-700);
+    color: white;
+  }
+
+  &:active {
+    border-color: var(--eds-color-other-mint-800);
+    background-color: var(--eds-color-other-mint-800);
+    color: white;
+  }
 }


src/components/Banner/Banner.module.css
```diff
@@ -57,6 +57,10 @@
 
 .banner__textContent {
   gap: var(--eds-size-1);
+
+  a {
+    color: inherit;
```
dierat, line 62: Might also want to add the

src/components/Banner/Banner.module.css
```diff
@@ -176,16 +199,87 @@
...
+    color: var(--eds-color-brand-grape-800); /* 2 */
+  }
+}
+
+/*------------------------------------*\
+    # ACTION STYLES
+\*------------------------------------*/
+
+.banner__action > * {
+  /* TODO: move to mixin */
```
dierat, line 232: This comment here is key – these styles are just copy/pasted over from the `Button` because I didn't want to put too much time into this direction in case we decide to do something else.

src/components/Banner/Banner.module.css
```diff
@@ -176,16 +199,87 @@
...
+    --icon-size-default: 2em; /* 5 */
+  }
+
+  @media screen and (prefers-reduced-motion) {
+    transition: none;
+  }
+}
+
+/* stylelint-disable-next-line no-descending-specificity */
+.banner--success .banner__action > * {
```
dierat, line 267: I only added the success action here as an example of what it would look like (again, I didn't want to dump too much time into this). I used `*` instead of `button` here because it's possible this button is actually going to be a link.

src/components/Banner/Banner.module.css
```diff
@@ -176,16 +199,87 @@
...
+  /* TODO: move these colors to tokens */
```
dierat, line 268: I'm imagining a tokens file similar to the button tokens file.

ifrost1:
@dierat Thanks for all of this thinking and for flagging this as something that's starting to feel a little off. What you suggested are a good range of options that get you most of the way there, but feel like duplicative work instead of just using the Button component. If this was just being used for the `Banner` I'd say we could just bake in an HTML `button` tag into that close button since it's always a close button and style it on the `Banner` component. 

But because we're starting to see that treatment span into multiple components (not sure how many of these light green backgrounds with darker button icons/text there will be), I suggest that we build some sort of `status="success"` or something to extend the existing buttons since you at least want to support `variant="secondary" status="success"` to allow for extension of `secondary` and `link` buttons at least. You could technically apply that to all existing styles, but I think the hestiancy around adding in these buttons into the `Button` was misuse by consumers of the system and having too many options. Do we support primary, success buttons? Technically you could add `variant="primary" status="success"`, but do you really want to surface that option to users if there's not a use-case for that? How can we provide value without giving too many options to a consumer? I think another reason we simplified was in case we do support multiple themes with this system (e.g. Along, which I know you are working on as well). Does something like Along need these kinds of buttons? If so, then those could be shared as status buttons across multiple themes without adding unnecessary variants in there. There are definitely times where we do just use regular, ol' HTML button tags (Accordion is one of those I believe) just so you can use a button, but aren't having to override all of the styles the `Button` component gives you.

If you do go this route, let's just make sure to document the crap out of those and make sure there are strict guidelines around when to use those. 

Hope this is helpful. Thanks for bringing this up!

csnizik:
I personally feel that button behaviors should be handled by the button component. A banner doesn't need to know about any buttons it may or may not contain. Single Responsibility and all that.

Question: `Button.module.css:270` has a comment that says "1) Set SVG icon fill to the current text color". The block of code that it's describing doesn't change the icon's color so maybe the comment is just a relic, but seeing that made me realize that the cleanest way I can think of to create the button variants in the Figma would be if both a) the button, and b) the icon could have a variant for "use current text color". (I don't like component A including a class that reaches into component B and alters/overrides its styles, which is why I suggest that both components have this same functionality.

csnizik:
And yes I realize that my suggested approach would facilitate misuse downstream. But locking things down has its own potential problems too, such as what happens when the button component gets updated and now these components that override the button get jacked up.

csnizik:
(I tend to write in a stream-of-consciousness style, and if I wait to hit "Submit" until I've completely thought through a thing, it would be sometime next week before I posted anything.) 

Seeing now that there's no need to do a current text color when what we're talking about are 1 of 5 colors: success, warning, error, brand, info.

csnizik:
@dierat your suggestion #2 looks interesting to me. A HOC that wraps <Button> and gives it extra flavors. Your wrapped component <MessagingButton /> is then only available within the components that it is designed for.

anniehu4 (commented):
I think ... our Banner design is too complicated 😂 

A critical question imo is your point #3 -- should consumers pass a react node or object for the main `action` button? 
- If the former, I think we should surface some Button component (either amend the existing one or create a new, specific one) that can create the styles. Otherwise I agree it's confusing to pass in one style & suddenly have it overridden
- The latter is kind of nice though because we already want them to pass button as a specific `action` prop. We could also probably simplify the `textContent` API, since there are fewer button variants

Otherwise:
- I feel comfortable defining styles for the close button however we want, since that's fully controlled by the Banner
- Also feel more comfortable overriding styles for the link -- I'd suspect people don't usually think about adding color to their Links, and would appreciate not having to specify one

And given that we haven't been able to simplify the design much, I think it's totally fine to go back to the `variant x color` (or `variant x status`) handling in `ClickableStyle` -- **but** adding warnings about combos that we don't want to support. e.g. no need to even style a `primary x orange` combo

dierat:
>But because we're starting to see that treatment span into multiple components (not sure how many of these light green backgrounds with darker button icons/text there will be)

@ifrost1 are you referring to something other than the `Banner` and `Toast`? Just wondering if we're expecting to have to support these shape x color combos outside of "messaging" components (banners and toasts).

ifrost1:
@anniehu4 We just had a team over the summer with a very similar pattern for banner and we kind of shifted them away from the green on green on green treatment because it does add complexity around base tokens because those greens have to be accessible on top of one another. Here's an example: 
Screen Shot 2022-04-14 at 2 05 52 PM

I know you looked into how often these "status" buttons are being used and it's really not all that often if I remember correctly. So does a design shift help simplify that? Definitely have been in situations recently where teams are trying to support this, but then they need a `theme-color-text-utility-error` and a `theme-color-text-utility-error-dark` to allow for these lighter utility color backgrounds.

dierat:
>Does something like Along need these kinds of buttons?

@ifrost1 the Along button is pretty simple – there's only one color, and it comes in "primary", "secondary", "icon", and "link" flavors (very similar to the current v1 `Button` API except it doesn't have "tertiary" or "destructive").

ifrost1:
> @ifrost1 are you referring to something other than the Banner and Toast? Just wondering if we're expecting to have to support these shape x color combos outside of "messaging" components (banners and toasts).

@dierat That's one of those questions that every person who works on a design system wonders: What future states might use this? And I don't know. This button situation definitely feels like it's on the edge of "let's include this in the button" or let's make this component(s) specific. I don't see this green on green pattern used a lot in design systems I've worked on. But needing to support and maintain these buttons when they're only used in 2 places feels a bit strange. I think with the link stuff I am okay with general `a` tags overall as long as they are used sparingly throughout the DS. The action buttons is where it becomes complicated because those are just a slot to throw buttons, and therefore it puts the burden on the user which we don't want. How do we keep these button variants to a minimum without burdening the user of the system?

dierat:
I did notice that as design is exploring these components, there is a version in there that has banners with white backgrounds and neutral close buttons, though it still have the colored outline buttons.

https://czi-edu.slack.com/archives/C0310HJ9WEN/p1649776632240249
screenshot from loom video that shows a figma file with a version of the banner component that has white backgrounds and neutral close buttons but colored outline buttons

dierat:
> Question: `Button.module.css:270` has a comment that says "1) Set SVG icon fill to the current text color". The block of code that it's describing doesn't change the icon's color so maybe the comment is just a relic,

@csnizik that is a relic; the `Icon` component can now handle color via props and the default color is `currentColor`. https://github.com/chanzuckerberg/edu-design-system/blob/next/src/components/Icon/Icon.tsx#L96 Thanks for asking ❤️ 

> but seeing that made me realize that the cleanest way I can think of to create the button variants in the Figma would be if both a) the button, and b) the icon could have a variant for "use current text color". (I don't like component A including a class that reaches into component B and alters/overrides its styles, which is why I suggest that both components have this same functionality.

I'm just concerned about the hover/active styles. Some sort of "currentColor" variant would only know which color to use for its resting state, right? 🤔

ifrost1:
> I did notice that as design is exploring these components, there is a version in there that has banners with white backgrounds and neutral close buttons, though it still has the colored outline buttons.
> 
> https://czi-edu.slack.com/archives/C0310HJ9WEN/p1649776632240249 screenshot from loom video that shows a figma file with a version of the banner component that has white backgrounds and neutral close buttons but colored outline buttons

Would using your `tertiary` button work for those actions if there's another visual cue about status other than the icon was added (like the thick left border example I showed you)? Would be good to see this in a page to see how it looks overlaying the other content. To me, there's not much value in these examples that those status buttons present aside from blending nicely with the green background. Could we leverage our simpler button options to support the close button and action buttons in the examples you just shared? Just gut feeling without actually putting that into a Storybook page is that the TimelineNav box and this may butt heads a bit.

dierat:
>(I tend to write in a stream-of-consciousness style, and if I wait to hit "Submit" until I've completely thought through a thing, it would be sometime next week before I posted anything.)

lol no problem; I tend to do the same thing, especially on slack.

ifrost1:
Also, do `Toast` and `Banner` need to be separate components? Sorry I'm asking a lot of higher level questions.

dierat:
> Also, do `Toast` and `Banner` need to be separate components? Sorry I'm asking a lot of higher level questions.

Functionally yes but there could be a component they both consume that handles these styles. I think we decided to keep them separate to avoid a bunch of overriding for the little details, like the `Toast` has the close button vertically centered and it has a max width.

dierat:
> > Also, do `Toast` and `Banner` need to be separate components?
> 
> Functionally yes but there could be a component they both consume that handles these styles. I think we decided to keep them separate to avoid a bunch of overriding for the little details, like the `Toast` has the close button vertically centered and it has a max width.

> Sorry I'm asking a lot of higher level questions.

Np, ask away ❤️

dierat:
> Would using your `tertiary` button work for those actions if there's another visual cue about status other than the icon was added (like the thick left border example I showed you)? Would be good to see this in a page to see how it looks overlaying the other content. To me, there's not much value in these examples that those status buttons present aside from blending nicely with the green background. Could we leverage our simpler button options to support the close button and action buttons in the examples you just shared? Just gut feeling without actually putting that into a Storybook page is that the TimelineNav box and this may butt heads a bit.

@ifrost1 we also have the thick left border on the `Banner`, so I could see it working. But how design feels about it.. 🤷  I wonder how far they'll actually go in trying to reduce the color overload in the product. I'm hesitant to push them in that direction because I'm afraid they'll realize later they do need these buttons and then it'll be a huge hassle to bring them back.

ifrost1:
@dierat Thanks for the info. Definitely difficult to build this stuff too while this is work that is currently being used and making sure that boat isn't rocked too much. I can see value in either adding the `variant x color` back to buttons, but if Along isn't planning on using these then why would they want to consume that stuff. That's why a recipe like `MessagingButton` and baking in the HTML button in as the close button is good from a "keeping this as a specific use-case" without adding in a bunch of variants/colors that might not get used. If you bake it into the button, then you do have those future use-cases covered if Along or another theme was to adopt those status buttons.

ifrost1:
> we also have the thick left border on the Banner, so I could see it working. But how design feels about it.. 🤷 I wonder how far they'll actually go in trying to reduce the color overload in the product. I'm hesitant to push them in that direction because I'm afraid they'll realize later they do need these buttons and then it'll be a huge hassle to bring them back.

@dierat I feel like this is why building pages in Storybook helps. Throw a `Banner` and a `Toast` with those styles in there (in an experimental branch) and show it to designers. It's definitely worth discussing with them to keep things simpler if that's possible.

dierat:
> I think ... our Banner design is too complicated 😂

lol 😂  honestly even without this Button convo it's already kinda painful.

> And given that we haven't been able to simplify the design much, I think it's totally fine to go back to the `variant x color` (or `variant x status`) handling in `ClickableStyle` -- **but** adding warnings about combos that we don't want to support. e.g. no need to even style a `primary x orange` combo

How do you feel about sectioning that off into a separate component so `Button`, `Link`, and maybe even `ClickableStyle` don't have access to those options because we only use them internally?

dierat:
> > we also have the thick left border on the Banner, so I could see it working. But how design feels about it.. 🤷 I wonder how far they'll actually go in trying to reduce the color overload in the product. I'm hesitant to push them in that direction because I'm afraid they'll realize later they do need these buttons and then it'll be a huge hassle to bring them back.
> 
> @dierat I feel like this is why building pages in Storybook helps. Throw a `Banner` and a `Toast` with those styles in there (in an experimental branch) and show it to designers. It's definitely worth discussing with them to keep things simpler if that's possible.

@ifrost1 one thing we've run into a few times is something is designed in a void and they only realize it doesn't work when we see it in the product. What would be really cool is if we could publish an alpha version from an experimental branch so you could actually tour the product with that version.

dierat:
Btw thanks for all the feedback; I really appreciate you all working through this conceptual knot with me ❤️

ifrost1:
> one thing we've run into a few times is something is designed in a void and they only realize it doesn't work when we see it in the product. What would be really cool is if we could publish an alpha version from an experimental branch so you could actually tour the product with that version.

@dierat This is the part that I love about Storybook and building pilot pages with it. We have a page. We have these components. Let's throw in Ryan's example (I think that's his example of banner) into project overview and make a few tweaks to include that thick left border, send a screenshot, and get feedback. Even if design doesn't agree that it's the perfect solution it may get them thinking a little more about that and they could provide a solution that works that is simpler if you feel the pain with these existing patterns.

That being said, how do we handle existing products using these components and making sure they feel alright with these changes? That's the part I'm a bit more scared about.

ifrost1:
> Btw thanks for all the feedback; I really appreciate you all working through this conceptual knot with me 

Appreciate you working through these tough decisions. They matter. Buttons are something to feel good about as the quintessential design system component. I think questioning this stuff is healthy and leads to a better system.

ifrost1:
I also wonder if creating a `MessagingButton` product-specific recipe would do the trick for now. That way it doesn't ship with the system but can be built at the product level to be reused and you can accomodate those use-cases without bloating the system. The way we tend to build recipes is in a way that may get inherited by the system if there are enough use-cases down the road. So building that recipe similar to button would allow for that enhancement on the `Button` component if that in fact needs to be a part of the `Button` component.

ifrost1:
@dierat @anniehu4 Kind of as I expected. Definitely feels a bit too plain huh?

Screen Shot 2022-04-14 at 3 31 31 PM

dierat:
> @dierat @anniehu4 Are `Banner` components absolutely positioned or are they always inline with the content? Or does it just depend?

I feel like they're usually inline but I'm not sure.

dierat:
> @dierat @anniehu4 Kind of as I expected. Definitely feels a bit too plain huh?
> 
> Screen Shot 2022-04-14 at 3 31 31 PM

Yeah doesn't really call attention to the content as much as we probably want.

dierat:
> I also wonder if creating a `MessagingButton` product-specific recipe would do the trick for now. That way it doesn't ship with the system but can be built at the product level to be reused and you can accomodate those use-cases without bloating the system. The way we tend to build recipes is in a way that may get inherited by the system if there are enough use-cases down the road. So building that recipe similar to button would allow for that enhancement on the `Button` component if that in fact needs to be a part of the `Button` component.

Yeah, right now the idea I like is sectioning these styles off into a different component that's not related to the `Button` at all (except they use the same base styles from a mixin). I'm thinking 2 components (`MessagingActionButton` and `MessagingCloseButton`) that each just product one shape (outline or icon) in the 5 colors for the `Banner`, `Toast`, and anything else that comes along that needs it. If we end up phasing out these buttons from banners – great, we'll just delete the buttons.

ifrost1:
Now I'm just riffing. I like the top border more with calling attention since the sidebar adds a lot of attention there and kinda matches that left border.

Screen Shot 2022-04-14 at 3 45 27 PM

ifrost1:
> Yeah, right now the idea I like is sectioning these styles off into a different component that's not related to the Button at all (except they use the same base styles from a mixin). I'm thinking 2 components (MessagingActionButton and MessagingCloseButton) that each just produce one shape (outline or icon) in the 5 colors for the Banner, Toast, and anything else that comes along that needs it. If we end up phasing out these buttons from banners – great, we'll just delete the buttons. The only overrides built into the Banner is the link colors. And the action prop will now accept an object that contains the button's props (except for variant stuff).

This may be a happy medium to keep the system clean. May only need 1 button with a `MessagingButton variant="icon"` and then `MessagingButton` default (or secondary) is the outlined button you can throw into the `actions` slot there. Is there a default button without needing a variant? Or does each button need that variant? We normally make your secondary button the default button since it's not as pronounced as the primary and used a little more (in our other projects). Then may be able to add `status` as a prop to change states. Or really limit it by just baking the dismissible close button into the component as a `<button>` and then just have duplicate styles in the Toast and Banner (although I know that doesn't sound ideal). Just throwing out ideas here. Just trying to stay away from creating too many extra recipes if we don't have to.

ifrost1:
I personally prefer using `status` over `color` simply because it gives more semantic meaning to the button rather than locking it into a color (although I know that utility colors tend to stay green, yellow/orange, red overall). And we also have `error` names for things like form field error states or `<Badge status="success" />`. Just throwing that out there. Definitely don't want to come in here and steamroll

anniehu4:
Lots of great convo going on, I'm good with the general direction we're heading!

dierat:
> This may be a happy medium to keep the system clean. May only need 1 button with a `MessagingButton variant="icon"` and then `MessagingButton` default (or secondary) is the outlined button you can throw into the `actions` slot there.

@ifrost1 That's what I was thinking initially, but I would prefer to keep each component simpler since, aside from possibly rendering a `button` tag, they're pretty different. But I'm not too opinionated at that level of the solution; I'm happy as long as this isn't all mashed into the `Banner` 😅 

>I personally prefer using status over color simply because it gives more semantic meaning to the button rather than locking it into a color (although I know that utility colors tend to stay green, yellow/orange, red overall). And we also have error names for things like form field error states or <Badge status="success" />. Just helps normalize conventions so consumers can intuitively make educated guesses without documentation.

That makes sense for "success", "error, "warning", and even "neutral", but `status="brand"` feels super weird to me. Making I'm taking the term "status" too literally. 🤔

dierat:
> Now I'm just riffing. I like the top border more with calling attention since the sidebar adds a lot of attention there and kinda matches that left border.
> 
> Screen Shot 2022-04-14 at 3 45 27 PM

That's nice! Also the banner's dark border moves to the top on vertical/mobile view, so it would be more consistent across the different orientations to always be at the top.

ifrost1:
> That's what I was thinking initially, but I would prefer to keep each component simpler since, aside from possibly rendering a button tag, they're pretty different. But I'm not too opinionated at that level of the solution; I'm happy as long as this isn't all mashed into the Banner 😅

I guess what I was thinking was to basically duplicate `Button` and match those conventions to allow for ease of conversion to the system if that is needed for `status` (whatever we want to call these) buttons. That way they can be consolidated if needed. 

> That makes sense for "success", "error, "warning", and even "neutral", but status="brand" feels super weird to me. Making I'm taking the term "status" too literally. 🤔 But, again, I'm not super opinionated about these details.

I hear you with this. Guess I would be under the assumption that the default would be `brand` for those or am I wrong about that? We've called things that are wishy washy `support` colors in the past and had some options there. Hard to think of something that works with a lot of different use-cases. Also just was trying to support both `variant` and `status` to allow for maximum flexibility without having to say `variant="success-link"`, `variant="success-secondary"`, etc. Haha. This is all good stuff.

ifrost1:
> > Now I'm just riffing. I like the top border more with calling attention since the sidebar adds a lot of attention there and kinda matches that left border.
> > Screen Shot 2022-04-14 at 3 45 27 PM
> 
> That's nice! Also the banner's dark border moves to the top on vertical/mobile view, so it would be more consistent across the different orientations to always be at the top.

Could always share with the designers. If they don't like it, then it was a few minutes down the drain, no biggie. Haha

ifrost1:
I think we're sticking with the green background for Toast and Banner. Just talked to Courtney a bit about it and she's thinking that status buttons should live in the system. I think Josh also mentioned that's probably a good semantic type of button in there. Do we create a status for all types of button variants and surface those or do you only create ones for the ones that exist? Do I have it right that `link`, `icon`, and `secondary` would be the variants that have these color treatments?

ifrost1:
Also rethinking the `recipe` part of the buttons simply because that stuff doesn't ship with the system but we'd need `MessagingButton` to ship to be a part of those components. This stuff is hard. Haha

dierat:
> I think we're sticking with the green background for Toast and Banner. Just talked to Courtney a bit about it and she's thinking that status buttons should live in the system. I think Josh also mentioned that's probably a good semantic type of button in there. Do we create a status for all types of button variants and surface those or do you only create ones for the ones that exist? Do I have it right that `link`, `icon`, and `secondary` would be the variants that have these color treatments?

Probably not `link` because those will be embedded in the content passed in by the consumer.

ifrost1:
@dierat I say we build that into the Buttons and live with it. And only support `icon` and `secondary` variants until there's a need for others. Does that sound like a good direction to go?

ifrost1:
@dierat @anniehu4 Just trying to keep consumers from using certain buttons just to use them (if you can't tell, I struggle with trusting product devs from teams we've worked with in the past. Haha)

dierat:
> @dierat I say we build that into the Buttons and live with it. And only support `icon` and `secondary` variants until there's a need for others. Does that sound like a good direction to go?

@ifrost1 To be clear, are we basically talking about reverting to the old API but renaming some things (ie: "flat" -> "primary", color -> "status") and dropping support for some combos (ex: primary green)? I just wanna make sure before I start tearing this thing apart lol

ifrost1:
> To be clear, are we basically talking about reverting to the old API but renaming some things (ie: "flat" -> "primary", color -status) and dropping support for some combos (ex: primary success)? I just wanna make sure before I start tearing this thing apart lol

It looks like you have `button--primary` in your CSS so we'd stick with that but you wouldn't add any styles around `button--primary button--success`. So we'll basically be adding this to CSS for the variant/status combo:

```
.button--secondary.button--success {
 // Success secondary button styles
}

.button--icon.button--success {
  // Success button icon styles
}
```

One thing I'm curious from a button standpoint: Is there a default button style? If so, what is it. If not, can we make one of the variants the default (maybe the button variant that is most used across the product). Like I said before, we use the outline treatment as our default button usually (your secondary button). Basically what happens when you include `<Button>Hello</Button>` without a variant? Just going off of what we've done in the past, but we're obviously flexible.

dierat:
> One thing I'm curious from a button standpoint: Is there a default button style? If so, what is it. If not, can we make one of the variants the default (maybe the button variant that is most used across the product). Like I said before, we use the outline treatment as our default button usually (your secondary button). Basically what happens when you include `<Button>Hello</Button>` without a variant? Just going off of what we've done in the past, but we're obviously flexible.

@ifrost1 Currently the default is a large primary brand button, but I'm down to make it secondary since we do want people to rely on the primary one less often.

dierat:
First pass at adding the `status` prop to the `Button` styles: https://github.com/chanzuckerberg/edu-design-system/pull/977

Closing this PR so the conversation can continue over there as needed.

PR #976, opened 2022-04-14
feat: move components/Icons to src/icons

anniehu4:
### Summary:

Move Icons folder so imports will now be from `@chanzuckerberg/eds/lib/icons`

### Test Plan:
- build passes
- `yarn build` publishes all icons to `lib/icons`


jinlee93 (approved):
Thanks for helping out with the publish

PR #977, opened 2022-04-15
feat(clickable-style): add 'status' prop, add full list of tokens

dierat:
### Summary:
First pass at updating the `Button` styles to be able to support all 5 colors for the secondary, icon, and link variants.

The main reason for this change is because we're going to need the secondary and icon buttons in the `Banner` and `Toast` components. I threw in the link variant as well since devs will need to pass text links into the content of these components and I think it would be confusing to pass in a purple link and expect it to turn orange or whatever. I was fine before with just overriding it in the component styles, but if we're going to support more button variant + color combos, I figured we might as well cover this case too.

Outstanding issues that I'll handle in separate PRs:
- make secondary variant the default instead of primary
- fix the props not showing up properly in the docs tab (I broke this in a different PR)

https://user-images.githubusercontent.com/7761701/163631657-ecbd8b46-284d-4626-8b3f-305a301a2d0c.mp4

### Test Plan:
Storybook + chromatic.


ifrost1:
@dierat This all feels great by the way. Thank you. I think it's a good idea for `default` buttons to be secondary if you are trying to get users to shy away from using primary buttons. There would need to be some sort of deprecation message though that says "This has changed to this other thing. Please update these props to get those stylles." Or something along those lines. Then comes the fun part of, "Does tertiary become secondary or how is that handled?" Shuffling a bunch of stuff around. But just things I'm thinking of. I think the console error makes sense around `primary + status`

src/components/Button/Button.module.css
```diff
@@ -16,20 +16,22 @@
 
 /**
  * Primary disabled button styles
+ * 1) Class is doubled to increase specificity
```
dierat, line 19: This is now required because the `ClickableStyle` is pairing the variant and status classes, which bumped its specificity.

src/components/Button/Button.stories.tsx
```diff
@@ -83,43 +83,60 @@ SecondaryMedium.args = { children: 'Button', variant: 'secondary', size: 'md' };
 export const SecondarySmall = Template.bind({});
 SecondarySmall.args = { children: 'Button', variant: 'secondary', size: 'sm' };
 
-export const Tertiary = Template.bind({});
-Tertiary.args = { children: 'Button', variant: 'tertiary' };
+export const SecondaryNeutral = Template.bind({});
```
dierat, line 86: I changed the story name from "Tertiary" to "SecondaryNeutral" so it's consistent with the props, but now it's at odds with the UI Kit. I can change it back if anyone feels strongly that it should match the UI Kit instead.

src/components/Button/Button.stories.tsx
```diff
@@ -213,25 +230,23 @@ LinkRightIcon.args = {
   variant: 'link',
 };
 
-export const Destructive = Template.bind({});
-Destructive.args = { children: 'Button', variant: 'destructive' };
-
-export const DestructiveDisabled = Template.bind({});
-DestructiveDisabled.args = {
+export const PrimaryAlert = Template.bind({});
```
dierat, line 233: Same as "Tertiary" – I changed the story name from "Destructive" to "Primary Alert" so it's consistent with the props, but now it's at odds with the UI Kit. I can change it back if anyone feels strongly that it should match the UI Kit instead.

src/components/Button/Button.stories.tsx
```diff
@@ -243,3 +258,100 @@ Loading.args = {
   loading: true,
   disabled: true,
 };
+
+export const SecondarySuccess = Template.bind({});
```
dierat, line 262: I tossed all the extra combos in at the end so people can see that they are available but they have lower visibility than the preferred combos above.

src/components/ClickableStyle/ClickableStyle.tsx
```diff
@@ -29,15 +29,13 @@ export type ClickableStyleProps<IComponent extends React.ElementType> = {
    */
   size?: 'sm' | 'md' | 'lg';
   /**
-   * Available _stylistic_ variations available for the ClickableStyle component
+   * Available _color_ variations available for the ClickableStyle component
```
dierat, line 32: I changed the comments to say "color" and "shape" because they're both stylistic variations. Lemme know if you want it changed back.

src/components/ClickableStyle/ClickableStyle.tsx
```diff
@@ -29,15 +29,13 @@ export type ClickableStyleProps<IComponent extends React.ElementType> = {
...
    */
-  variant?:
-    | 'primary'
-    | 'secondary'
-    | 'tertiary'
-    | 'icon'
-    | 'link'
-    | 'destructive';
+  status?: 'brand' | 'neutral' | 'success' | 'warning' | 'alert';
```
dierat, line 34: I'm not sure whether we should be using "alert" or "error". I know the utility tokens use "error" but some of the components are using "alert". I'd like for us to align on this and change the name everywhere, whichever direction is chosen. I'll open a ticket.

src/components/Button/Button.stories.tsx
anniehu4: hmmm I think it's helpful for designers when story names match the UI Kit (and eng can click "Show Code"). we should probably sync or pair with designers on naming though, because "Secondary" now has a different meaning in code than in the UI Kit

src/components/ClickableStyle/ClickableStyle.tsx
```diff
@@ -75,10 +87,14 @@ export const ClickableStyle = React.forwardRef(
       // Variants
       variant === 'primary' && styles['clickable-style--primary'],
       variant === 'secondary' && styles['clickable-style--secondary'],
-      variant === 'tertiary' && styles['clickable-style--tertiary'],
       variant === 'icon' && styles['clickable-style--icon'],
       variant === 'link' && styles['clickable-style--link'],
-      variant === 'destructive' && styles['clickable-style--destructive'],
+      // Colors
+      status === 'brand' && styles['clickable-style--brand'],
```
anniehu4, line 93: sorry if you already had this convo -- for default values for `variant` and `status`, could we put those styles directly in the main `.clickable-style` class & cut down the className size for most use cases?

to be fair we're probably not being consistent about this across the board, so could file a ticket & come back to it

src/components/Button/Button.stories.tsx
anniehu4: it'd probably be easiest to reference these in a Grid-layout story (like the current "All Variants" one or something)

anniehu4 (approved):

src/components/Button/Button.stories.tsx
dierat: I'll add a ticket to the backlog for syncing the names. I don't think it's time-sensitive but I feel like we keep coming back to this inconsistent naming issue.

If our concern is the code terms "primary" and "secondary" are conflicting with the UI Kit terms "primary" and "secondary", well, we could always go back to "flat" and "outline" in the code 🤷 

For now, I'll change this one to "Tertiary" and primary error to "Destructive", thanks!

PR #979, opened 2022-04-15
feat: remove lerna and use standard-version to publish v0.10

jinlee93:
[sc-187216]
[sc-189730]
### Summary:
- removes `lerna`
- adds `standard-version`
- adds [publishing doc  from `next`](https://github.com/chanzuckerberg/edu-design-system/blob/0d8155f1ac192f8a9ef12ae525339e2a3195b0dd/docs/PUBLISHING.md#releasing-a-new-version-of-eds) with a few step adds
- [`v0.10.0-alpha.0`](https://www.npmjs.com/package/@chanzuckerberg/eds/v/0.10.0-alpha.0) published to test 

### Test Plan:
- publishes (see above bullet about alpha release)
- usage in LP: https://github.com/FB-PLP/traject/pull/10725
  - as of writing right now, couldn't deploy on `czi-summit-test` as it is reserved

### Next steps:
 - After merge, publish actual v0.10?
 - Deprecate `eds-components` and `eds-tokens` in npm


scripts/updateYarnLockForLerna.sh
```diff
@@ -1,20 +0,0 @@
...
-# To work around that, we manually install deps so that the lock file gets updated. This should run
-# in a Lerna "version" lifecycle hook. See https://github.com/lerna/lerna/tree/a47fc294393a3e9507a8207a5a2f07648a524722/commands/version#lifecycle-scripts.
-#
-# Another workaround would be to use Yarn workspace ranges to specify dependencies between
-# packages. See https://yarnpkg.com/features/workspaces/#workspace-ranges-workspace.
-# Unfortunately, Lerna also doesn't support those - https://github.com/lerna/lerna/issues/2564.
-#
-# If either of these Lerna issues are resolved, we should remove this hack.
-yarn install
-git stage yarn.lock
```
jinlee93, line 20: Not sure if this file is needed anymore since Lerna not used

docs/publishing.md
```diff
@@ -0,0 +1,48 @@
...
+
+- set up Two Factor Authentication for your npm account
+- run `npm login` in your terminal to generate an access token for publishing
+
+#### Publishing steps
+
+1. Confirm that all checks are green on CI.
+2. Run `git checkout main`.
+3. Run `yarn` to ensure all dependencies are upto date.
+4. Run `yarn build` to build EDS.
```
jinlee93, line 32: Should we just have `yarn build` in the release scripts?

docs/publishing.md
```diff
@@ -0,0 +1,48 @@
+## Versioning
```
jinlee93, line 1: Not sure where to place this doc other than here.
Could consider displaying it in storybook like BM

anniehu4: this works! I think convention is to capitalize docs file names -> `docs/PUBLISHING.md`

docs/publishing.md
anniehu4: yeah that's a good idea 👍  `lerna` might have been doing that automatically

anniehu4 (approved):
sweet, thanks!

PR #981, opened 2022-04-15
build: use standard-version, make one package, move folders

jinlee93:
### Summary:
- merge work that's been done on `publish-next` to `main` so we can move forward with publishes as one package;
  - remove workspaces
  - use separate tsconfig for build
  - move icons to src/icons

### Test Plan:
- publish works via https://github.com/FB-PLP/traject/pull/10725

### Next Steps:
- publish v0.10.0 from `main`
- deprecate `eds-components` and `eds-tokens` in npm


docs/publishing.md
```diff
@@ -0,0 +1,58 @@
+## Versioning
```
anniehu4, line 1: let's rename this file to `docs/PUBLISHING.md`

anniehu4 (approved):
🚢

docs/publishing.md
jinlee93: 🤦  I swore I did that in the other pr but a wrinkle in my brain must have smoothened itself, will do

PR #983, opened 2022-04-18
chore(banner, toast): use mixins to share variant colors between messaging components

dierat:
### Summary:
The `Banner` and `Toast` have overlapping styles, and there are a couple more components in the UI Kit like them. This PR adds mixins to reuse the color-related styles (text, background, and border colors) for the 5 possible variants for messaging components. This will help us keep them in sync so it's easier to maintain and keep them consistent. (This PR does have a slight change to the component styles because it causes the `Toast` to now use the exact same tokens as the `Banner`.)

One thing to note is that the `Toast` only supports 2 of these variants, so it's possible this is overkill and the complexity isn't worth the deduplication in this case. I'm more inclined to add this because it could be reused in other future messaging components.

UI Kit: https://www.figma.com/file/P9VPyI821gLq832tuU2WeY/UI-Kit-Master-(Core)?node-id=2525%3A662

### Test Plan:
Verify there are no changes to the `Banner` or `Toast` stories (in storybook and chromatic) aside from the left border for the `Toast` being a little darker and the other borders for the `Toast` being a little lighter.

### DIFF
--- a/src/components/Toast/Toast.module.css
+++ b/src/components/Toast/Toast.module.css
@@ -27,21 +27,18 @@
   border-left-width: var(--eds-border-width-lg); /* 3 */
   border-radius: var(--eds-size-half);
 
+  border-color: var(--border-light-color);
+  border-left-color: var(--border-dark-color);
+
   box-shadow: var(--eds-box-shadow-sm);
 }
 
 .toast--success {
-  color: var(--eds-theme-color-text-utility-success);
-  background-color: var(--eds-theme-color-background-utility-success);
-  border-color: var(--eds-theme-color-border-utility-success-default);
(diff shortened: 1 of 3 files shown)


src/components/Toast/Toast.module.css
```diff
@@ -27,21 +27,18 @@
   border-left-width: var(--eds-border-width-lg); /* 3 */
   border-radius: var(--eds-size-half);
 
+  border-color: var(--border-light-color);
+  border-left-color: var(--border-dark-color);
```
anniehu4, line 31: nit: maybe leave a comment here that these are defined by the messaging mixins

anniehu4 (approved):
I dig it!

PR #984, opened 2022-04-18
feat: use 'error' instead of 'alert' for red variants

dierat:
### Summary:
I noticed we're sometimes using the word "error" for red variants and sometimes using "alert" (but mostly using "error"). Design confirmed that we should be using "error" everywhere. This PR updates the `Banner`, `Toast`, `Text`, and `Heading` components to use "error".

### Test Plan:
Verify there are no changes to these components in storybook aside from the story name changes.

### DIFF
--- a/src/components/Banner/Banner.module.css
+++ b/src/components/Banner/Banner.module.css
@@ -149,10 +149,10 @@
 }
 
 /**
- * Banner alert
+ * Banner error
  */
-.banner--alert {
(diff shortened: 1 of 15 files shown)


dierat:
https://app.shortcut.com/czi-edu/story/193227/align-on-using-alert-vs-error-for-red-variants

jinlee93 (approved):
LGTM, thanks for confirming with design and changing

PR #985, opened 2022-04-18
docs(clickable-style, button, link): add argTypes to stories

dierat:
### Summary:
I'm having a bit of trouble with the `ClickableStyle`/`Button`/`Link` stories because the props aren't showing up by default. I don't want to get stuck in the weeds on this component anymore than I already have, so this PR just adds `argTypes` to the stories to get the docs working in storybook. (Unfortunately, we don't get the prop comments, which I really like, but, oh well.) I can make a backlog ticket to look into this further if we think it's worth coming back for.

#### Before this change
button component in storybook before this change

#### After this change
button component in storybook after this change

### Test Plan:
Verify the `ClickableStyle`, `Button`, and `Link` components now have their main configurable props listed on the docs tab in storybook and you can click around in them to change the component styling.

### DIFF
--- a/src/components/Button/Button.stories.tsx
+++ b/src/components/Button/Button.stories.tsx
@@ -7,6 +7,40 @@ import Icon from '../Icon';
 export default {
   title: 'Molecules/Buttons/Button',
   component: Button,
+  args: {
+    children: 'Button',
+    variant: 'primary',
+    status: 'brand',
+    fullWidth: false,
+    size: 'lg',
+    loading: false,
+  },
+  argTypes: {
+    variant: {
+      control: {
+        type: 'select',
+        options: ['primary', 'secondary', 'icon', 'link'],
+      },
+    },
+    status: {
+      control: {
+        type: 'select',
+        options: ['brand', 'neutral', 'success', 'warning', 'error'],
+      },
+    },
+    size: {
+      control: {
+        type: 'select',
+        options: ['sm', 'md', 'lg'],
+      },
+    },
+    fullWidth: {
+      control: 'boolean',
+    },
+    loading: {
+      control: 'boolean',
+    },
+  },
 } as Meta;
 
 type Args = React.ComponentProps<typeof Button>;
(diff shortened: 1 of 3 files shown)


src/components/Button/Button.stories.tsx
```diff
@@ -7,17 +7,50 @@ import Icon from '../Icon';
...
+    },
+  },
 } as Meta;
 
 type Args = React.ComponentProps<typeof Button>;
 
 const Template: Story<Args> = (args) => <Button {...args} />;
 
 export const Primary = Template.bind({});
-Primary.args = { children: 'Button' };
```
dierat, line 17: I also remove the children props that are the same as the default value (all stories without an icon).

jinlee93:
Looks good to me, but was wondering if we should be moving towards CSF 3.0 format for story files?

jinlee93 (approved):

dierat:
> Looks good to me, but was wondering if we should be moving towards CSF 3.0 format for story files?

We talked about this in slack but for context for other reviewers – we're having trouble getting the CSF 3.0 format working for these components specifically. We don't know what's going on, but it also doesn't feel critical enough to stop and figure it out now, so we're gonna skip it.

anniehu4:
> I'm having a bit of trouble with the ClickableStyle/Button/Link stories because the props aren't showing up by default.

🥲  the IComponent stuff is tricky. probs worth filing a ticket b/c it's annoying to keep the story argTypes updated

dierat:
> > I'm having a bit of trouble with the ClickableStyle/Button/Link stories because the props aren't showing up by default.
> 
> 🥲 the IComponent stuff is tricky. probs worth filing a ticket b/c it's annoying to keep the story argTypes updated

Here you go! https://app.shortcut.com/czi-edu/story/193509/fix-button-etc-props-not-showing-on-docs-tab-by-default

dierat:
Also here's the ticket for the CSF 3.0 format issue https://app.shortcut.com/czi-edu/story/193401/use-csf-3-0-format-in-button-etc-stories

PR #987, opened 2022-04-19
Ifrost/menu button hover state

ifrost1:
### Summary:
- Added hover/focus states to the menu button for alignment with nav items below on small screen global header

### Test Plan:
`yarn test` runs with success (deprecation warning as usual)
`yarn types` runs without issues
`yarn lint` runs without issues
`yarn start` works


.storybook/recipes/GlobalHeader/GlobalHeader.module.css
```diff
@@ -39,13 +42,18 @@
  *    can be exposed
  */
 .global-header__menu-button.global-header__menu-button {
-  padding: 0;
+  /*padding: 0;*/
```
dierat, line 45: Do we still need this line?

src/components/Header/Header.module.css
```diff
@@ -10,7 +10,10 @@
 .header {
   display: flex;
   align-items: center;
-  padding: var(--eds-size-1) var(--eds-size-2);
+  padding-top: var(--eds-size-1);
+  padding-right: var(--eds-size-2);
+  padding-bottom: var(--eds-size-1);
+  padding-left: var(--eds-size-2);
```
dierat, line 16: Just curious – why was this changed?

dierat (approved):
Looks gorgeous; I love it ❤️

src/components/Header/Header.module.css
frostyweather: @dierat I changed the syntax to match the media query's. Also I was going to change this to have less padding on the left side to accommodate the menu button hover state but figured if another theme was going to pick it up later on (like Along), they might want equal heights on both sides instead of having to override that. Just felt a little weird changing it globally if say the menu button was on the right for Along for some reason.

dierat: Awesome, thanks!

ifrost1:
This was ported into #997 to better merge with the new `next` branch.

PR #988, opened 2022-04-19
feat(clickable-style): make secondary variant default

dierat:
### Summary:
@ifrost1 suggested we make the default variant "secondary" for `Button` etc. The reasoning here is that, ideally, there should really only be one primary button (the Most Important Action based on what the user is doing), which means we should be using more secondary buttons than primary, so making that the default prop value should reduce the number of times devs need to specify the variant and also possibly nudge them to use that variant.

### Test Plan:
No big chromatic changes (aside from some story name updates and some story buttons that I want to use the default variant regardless of what that is, like full width and loading button stories)

### DIFF
--- a/src/components/Button/Button.stories.tsx
+++ b/src/components/Button/Button.stories.tsx
@@ -9,7 +9,7 @@ export default {
   component: Button,
   args: {
     children: 'Button',
-    variant: 'primary',
+    variant: 'secondary',
     status: 'brand',
     fullWidth: false,
     size: 'lg',
@@ -47,10 +47,17 @@ type Args = React.ComponentProps<typeof Button>;
 
 const Template: Story<Args> = (args) => <Button {...args} />;
 
+export const Default = Template.bind({});
+Default.args = { variant: 'secondary' };
+
 export const Primary = Template.bind({});
+Primary.args = { variant: 'primary' };
 
 export const PrimaryDisabled = Template.bind({});
-PrimaryDisabled.args = { disabled: true };
+PrimaryDisabled.args = {
+  disabled: true,
+  variant: 'primary',
+};
 
 export const PrimaryLeftIcon = Template.bind({});
 PrimaryLeftIcon.args = {
@@ -60,6 +67,7 @@ PrimaryLeftIcon.args = {
       Button
     </>
   ),
+  variant: 'primary',
 };
 
 export const PrimaryRightIcon = Template.bind({});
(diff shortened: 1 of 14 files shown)


src/components/Button/Button.stories.tsx
```diff
@@ -47,10 +47,17 @@ type Args = React.ComponentProps<typeof Button>;
 
 const Template: Story<Args> = (args) => <Button {...args} />;
 
+export const Default = Template.bind({});
```
dierat, line 50: I moved the "secondary" story to the top and renamed it "default" so it makes a little more sense that when you drop a button in somewhere, it defaults to this style.

src/components/Button/Button.stories.tsx
```diff
@@ -117,52 +120,45 @@ SecondarySmall.args = { variant: 'secondary', size: 'sm' };
 
 export const Tertiary = Template.bind({});
 Tertiary.args = {
-  children: 'Button',
```
dierat, line 120: The `children` prop is defined in the `args` at the top of the file so I deleted this while I was in here.

src/components/Button/Button.stories.tsx
```diff
@@ -117,52 +120,45 @@ SecondarySmall.args = { variant: 'secondary', size: 'sm' };
...
-  variant: 'secondary',
   status: 'neutral',
 };
 
-export const SecondaryNeutralDisabled = Template.bind({});
```
dierat, line 125: I had renamed "SecondaryNeutral" to "Tertiary" but forgot to update the others, so I cleaned that up in here as well.

src/components/Button/Button.stories.tsx
```diff
@@ -261,13 +257,12 @@ LinkRightIcon.args = {
 
 export const Destructive = Template.bind({});
 Destructive.args = {
-  children: 'Button',
   status: 'error',
   variant: 'primary',
 };
 
-export const PrimaryErrorLeftIcon = Template.bind({});
```
dierat, line 269: Same as the "Tertiary"; I missed this one before.

src/components/Popover/PopoverExample.tsx
```diff
@@ -68,7 +68,7 @@ export const PopoverExample: React.FC<Props> = ({
       }}
       {...other}
     >
-      <Button onClick={openPopover} ref={popoverButton}>
+      <Button onClick={openPopover} ref={popoverButton} variant="primary">
```
dierat, line 71: I didn't really need to update the examples/stories, but.. 🤷

dierat:
https://app.shortcut.com/czi-edu/story/187502/move-button-to-new-architecture

anniehu4 (approved):

.browserslistrc
```diff
@@ -0,0 +1,9 @@
+# See our browser support policy at https://go.czi.team/edu-browser-support.
```
ahuth, line 1: FYI, I'm updating this in https://github.com/FB-PLP/traject/pull/9663

ahuth: Oh wait, this diff is probably because this branch is out of date with `next` or something. Never mind.

PR #991, opened 2022-04-19
chore(clickable-style): remove stories file

dierat:
### Summary:
We don't expose the `ClickableStyle` component on storybook on `main`. It also doesn't exist in zeroheight; it's a figment that's only referenced in the `Button` and `Link` docstrings and in the flow types file. This PR removes the stories file for this component to be consistent with `main`.

We might want to consider making this component more visible so people are aware it exists and how to use it. We made it available in `traject` because it can be used to style other components, like `react-router`'s `NavLink` and `Link` components if needed. It hasn't been used yet, but it's hard to know if that's because there was never a situation where it could have been useful or because no one knew it existed. If we do decide to expose it, we'll want to add the stories back and also add a lot of guidance in zeroheight. Ticket for considering adding this documentation: https://app.shortcut.com/czi-edu/story/193660/consider-adding-clickablestyle-to-storybook-and-zeroheight

### Test Plan:
Verify the component no longer exists in storybook.


jinlee93 (approved):
> If we do decide to expose it, we'll want to add the stories back and also add a lot of guidance in zeroheight.
Sounds good to me

PR #993, opened 2022-04-20
chore(button): capitalize main file

dierat:
### Summary:
There is this old annoying issue with the `Button` component on main where its files were originally lowercased. When `main` was merged into `next`, which has capitalized files, some weirdness happened where we sometimes got the capitalized file and sometimes the lowercase one.

This PR capitalizes the main `button.tsx` file.

### Test Plan:
Verify all the files (except the `index.ts` file of course) in the `Button` component folder are now capitalized.


anniehu4 (approved):
ugh good catch

PR #994, opened 2022-04-20
chore: convert double quotes to single in .storybook/main.js

dierat:
### Summary:
Just pretty-in' some stuff up after the merge of `main` into `next`.

### Test Plan:
Verify there are no lint errors.


jinlee93 (approved):
Not sure why the storybook is breaking, but thanks for the lint changes

anniehu4:
storybook should be fixed in https://github.com/chanzuckerberg/edu-design-system/pull/995

PR #996, opened 2022-04-20
chore(text, heading): match text and heading apis to next

jinlee93:
BREAKING CHANGE: "color" prop renamed to "variant" and "alert" color renamed to "error"
"info" color variant is deprecated, "yellow" color variant unused and removed
"weight" and "spacing" props removed
removed some "size" options

### Summary:
- changes the text and heading component to match `next`
- removes `Typography` from `main`
[sc-192414]
[sc-192415]

### Test Plan:
- unit/snap tests
- Chromatic/visual regression


src/components/Banner/Banner.tsx
```diff
@@ -166,7 +166,7 @@ type MessageProps = {
  */
 const BannerMessage: React.FC<MessageProps> = (props: MessageProps) => {
   return props.children ? (
-    <Text color={"inherit"} size="body">
+    <Text color="inherit" size="body">
```
dierat, line 169: Should this be
```suggestion
    <Text variant="inherit" size="body">
```
?

src/components/Banner/Banner.stories.module.css
```diff
@@ -0,0 +1,3 @@
+.headingBottomSpacing {
+  @apply mb-4;
```
dierat, line 2: Should we maybe be using regular CSS since tailwind is being removed later? Then we won't have to come back and change this.

src/components/Banner/Banner.tsx
jinlee93: Yes, shit, lost confidence in my codemod, thanks for catching

src/components/Heading/Heading.tsx
```diff
@@ -13,40 +27,53 @@ type Props = {
...
-  ) => (
-    <Typography as={as || size} ref={ref} size={size} {...rest}>
-      {children}
-    </Typography>
-  ),
+    ref: React.ForwardedRef<HTMLHeadingElement>,
+  ) => {
+    if (variant === "info" && process.env.NODE_ENV !== "production") {
+      console.warn(
+        "Info variant is deprecated and will be removed in an upcoming release. Please use the consider another variant instead.",
```
dierat, line 61: Maybe we should add test for these deprecation warnings.

src/components/Banner/Banner.stories.module.css
jinlee93: That would make sense for this file since it's just a story file, but for file stylesheets, would have to be changed for tokens anyways. What are your thoughts for this process?

src/components/Heading/Heading.tsx
jinlee93: Sounds like a good idea, will do

src/components/Banner/Banner.stories.module.css
dierat: For the file stylesheets, could we just leave the existing ones for now and clean it up later when the tokens are in?

jinlee93: That sounds good. I also think looking for `@apply` is a good identifier so clean up will be relatively easy

src/components/Heading/Heading.stories.tsx
```diff
@@ -52,10 +53,47 @@ export const Heading1AsHeading4: StoryObj<Args> = {
...
+      disabledRules: ["color-contrast"] /* 1 */,
+    },
+  },
+};
+
+export const Sizes: StoryObj<Args> = {
+  render: () => {
+    const sizes = ["h1", "h2", "h3", "h4", "h5", "h6"] as const;
+    const combinations = sizes.map((as: typeof sizes[number]) => {
+      const headings = sizes.map((size: typeof sizes[number]) => (
```
anniehu4, line 90: super nit: are the type annotations in `map`? I'd expect TS to be able to infer it. If not, maybe we can use the `HeadingElement` type instead (more readable)

src/components/Heading/Heading.tsx
```diff
@@ -13,40 +27,53 @@ type Props = {
...
+  size: HeadingElement;
   tabIndex?: number;
-  weight?: TypographyProps<HeadingElement>["weight"];
+  variant?: Variant;
 } & React.HTMLAttributes<HTMLHeadingElement>;
 
 /**
  * ```ts
- * import {Heading} from "@chanzuckerberg/eds";
+ * import {Heading} from "@chanzuckerberg/eds-components";
```
anniehu4, line 39:
```suggestion
 * import {Heading} from "@chanzuckerberg/eds";
```

src/components/Tag/Tag.tsx
```diff
@@ -62,9 +62,8 @@ function Tag({ color, children, className, icon, variant = "flat" }: Props) {
         color && stylesByColor[color],
         variant === "outline" && styles.outline,
       )}
-      color={color}
       size="sm"
-      spacing="none"
+      variant={color === "yellow" ? undefined : color}
```
anniehu4, line 66: 🤔  can we just get rid of the yellow variant? this feels like sneaky / unexpected behavior

src/components/Text/Text.stories.tsx
```diff
@@ -1,10 +1,13 @@
...
 import React from "react";
 
-import Text from "./Text";
+import { Text } from "./Text";
+import type { Variant } from "./Text";
+import styles from "./Text.stories.module.css";
 
 export default {
-  title: "Text",
+  title: "Atoms/Text/Text",
```
anniehu4, line 10:
```suggestion
  title: "Text",
```
?

src/components/Heading/Heading.stories.tsx
```diff
@@ -52,10 +53,47 @@ export const Heading1AsHeading4: StoryObj<Args> = {
...
+    });
+    return (
+      <div className={styles["variant__table"]}>
+        <div className={styles["variant--white"]}>{headings}</div>
+        <div className={styles["variant--light"]}>{headings}</div>
+        <div className={styles["variant--dark"]}>{headings}</div>
+      </div>
+    );
+  },
+  parameters: {
```
anniehu4, line 79: should we skip snapshot testing for this one also? since it's skipped for `Text`

anniehu4 (approved):

src/components/Heading/Heading.stories.tsx
jinlee93: Yea think we can, thanks for the catch

src/components/Tag/Tag.tsx
jinlee93: Good q, I'm not exactly sure. In here and `main`, the color of the `Tag` itself is passed to `Text` to style the `Text` color. However, in `next`, `Text` is styled via class styling passed from within `Tag`. (not sure if this word salad is making any sense).

Also in `next`, `yellow` does not exist but `yield` and `warning` which are two yellow `Tag`s, but design work is still ongoing this and even deprecating green/go, yellow/yield/warning, red/error Tag variants.

As a result, with regards to `Text` here in `Tag` in `main`,

I'll remove passing `variant` to `Text` and just inherit the color that's already being applied in `Tag`, like in `main` to preserve this `yellow` text for now and to be more similar to `next`. The `Tag` [ticket](https://app.shortcut.com/czi-edu/story/193791/update-tag-api-in-traject) itself can be used to fully migrate `Tag` later once decisions have been completed

src/components/Heading/Heading.stories.tsx
jinlee93: Sounds good, will do

src/components/Text/Text.stories.tsx
jinlee93: Yea idk how that happened, will change back

src/components/Tag/Tag.tsx
anniehu4: ahh right right, because we're removing the `yellow` variant from Text. Thanks for explaining, that solution sgtm 👍

src/components/Text/Text.tsx
```diff
@@ -1,49 +1,79 @@
...
  */
-const Text = forwardRef<HTMLElement, Props>(
+export const Text = forwardRef(
   (
     {
-      as,
+      as = "p",
       children,
+      className,
+      variant,
```
anniehu4, line 45: ah, another breaking change we should note is that previously the typography default would set `color = "base"`, but now the default is essentially `inherit`. doubt that'll break anything, but good to note anyways

anniehu4 (approved):

dierat (approved):

src/components/Text/Text.tsx
jinlee93: I was thinking about this, should I be adding `variant="base"` to all the changes in LP?

anniehu4: I don't think `variant="base"` was really that necessary 🤔  there's already a base color set on `body`, and I bet most text falls back to that. we can do a sanity check of the app to see if text colors change though

jinlee93:
So I was working with this and realized that the tailwind classname approach to increase spacing failed to override https://github.com/chanzuckerberg/edu-design-system/pull/996/files#diff-cfc7133a117f686b7db054299ae27c31ab8291f5f20f5b2a45a4605ba0851726R, need to consider some other approaches for spacing and maybe font weights

src/components/Heading/Heading.module.css
```diff
@@ -0,0 +1,67 @@
+/* sizes */
+
+.heading {
+  @apply mt-0 mb-0;
+}
```
jinlee93, line 5: with size being required, don't need to set text-body and font-normal, should always be bold weight

PR #1002, opened 2022-04-22
chore: re-export compound components from parent

anniehu4:
### Summary:

we can keep compound components as separate files for better maintainability, but still export them as sub-components from the parent for ease-of-use.

main motivation was thinking about the `Dropdown` component and how annoying it'd be to have to import 4 components instead of one -- but turns out that one was already a special snowflake

### Test Plan:
should be no visual changes


src/components/Checkbox/Checkbox.tsx
```diff
@@ -34,6 +34,12 @@ export type CheckboxProps = Omit<CheckboxInputProps, 'id'> & {
  * Checkbox control indicating if something is selected or unselected.
  *
  * Requires either a visible label or an accessible name.
+ *
+ * NOTE: usually, we would re-export subcomponents with
+ *   Checkbox.Input = CheckboxInput;
+ *   Checkbox.Label = CheckboxLabel;
+ * but this does not compile with Typescript since Checkbox itself is a
+ * forwarded ref component. Thus Input and Label must be imported separately.
```
anniehu4, line 42: 😭 probably could figure out this type issue but I remember it being a pain in the past

src/components/Breadcrumbs/Breadcrumbs.stories.tsx
```diff
@@ -12,9 +12,9 @@ export default {
```
dierat, line 12: Do we want that `subcomponents` part from line 10 in the other components too?

src/components/Checkbox/Checkbox.tsx
ahuth: Nothing a little `@ts-ignore` couldn't solve (probably)

anniehu4: I tried `@ts-ignore` and `@ts-expect-error` and restarting VSCode but sadly no dice :(
Screen Shot 2022-04-22 at 1 09 12 PM

down to pair on it at some point though

src/components/Breadcrumbs/Breadcrumbs.stories.tsx
anniehu4: ooh yeah that documentation is nice. will update

src/components/Checkbox/Checkbox.tsx
ahuth: Okay, let's pair on it sometime.

Looks like the ts-ignore lets us assign to the Checkbox function, but TS still doesn't know about what we added when we try to use them.

Probably we need a type annotation like `const Checkbox: SuperAmazingForwardRef = React.forwardRef` to tell TS about it.

ahuth (approved):

PR #1003, opened 2022-04-22
feat(list detail): adjustments

csnizik:
### Summary:
Shortcut: https://app.shortcut.com/czi-edu/story/190151/timeline-nav-build-responsive-behavior

This PR adds new list item variants to the list items: complete and incomplete*, plus the ability to have a bulleted and non-numbered first item in the list, with the second list item getting index of 1. Also, for list items using the "number" variant, the number and its circle border both get the neutral default text color when the list item is selected/active.

*NOTE: currently the Icon component does not have a variant for "star-outline"; this needs to be ticketed separately. 

From Figma:
Complete (green solid star icon):
Screen Shot 2022-04-22 at 11 22 26 AM

Incomplete (neutral subtle star outline icon* (see note above)):
Screen Shot 2022-04-22 at 11 16 17 AM

Active numbered list item with neutral default color on number and outline:
Screen Shot 2022-04-22 at 11 15 53 AM

First item "Overview" (not numbered in the list's counter, which starts with index 1 applied to the second <li>):
Screen Shot 2022-04-22 at 11 16 08 AM

### Test Plan:
Confirm in Storybook that each of the variants described above works as described/matches the Figma prototype.

### DIFF
--- a/src/components/ListDetail/ListDetail.module.css
+++ b/src/components/ListDetail/ListDetail.module.css
@@ -20,11 +20,11 @@
 }
 
 /**
- * List detail header
+ * List detail nav
  * 1) Contains the tab list
  * 2) Tab overflow behavior
  */
-.list-detail__header {
+.list-detail__nav {
   width: 100%;
   position: relative;
   padding: var(--eds-size-4) var(--eds-size-2);
@@ -89,9 +89,9 @@
     position: absolute;
     left: 0.25rem;
     top: 1.75rem; /* 2 */
-    height: 45%; /* 3 */ 
+    height: 45%; /* 3 */
     width: 0.0625rem;
-    background:  var(--eds-theme-color-border-neutral-subtle); /* 4 */
+    background: var(--eds-theme-color-border-neutral-subtle); /* 4 */
     border-radius: var(--eds-theme-border-radius);
   }
 
@@ -146,6 +146,7 @@
    * 1) Set to 9px to center the border
    * 2) This is the same color used for the vertical line when list style is "ordered"
    * 3) Using a percentage here instead of --eds-theme-border-radius because this is a cosmetic element (a cicular bullet), not a typical bordered object
+   * 4) Make the pseudoelement bullet dark when li is active
    */
   &:before {
     content: '';
@@ -158,6 +159,11 @@
     width: 9px; /* 1 */
     background: var(--eds-theme-color-border-neutral-subtle); /* 2 */
     border-radius: 50%; /* 3 */
+
+    /* stylelint-disable-next-line max-nesting-depth */
+    .list-detail__item.eds-is-active & {
+      background: var(--eds-theme-color-text-neutral-default); /* 4 */
+    }
   }
 
   /**
@@ -165,7 +171,7 @@
   * 1) Remove the bullet since the icon will take its place
   */
   li[class*='list-detail__item--'] & {
-    /* stylelint-disable-next-line max-nesting-depth */
+    /* stylelint-disable-next-line */
     &:before {
       content: none;
     }
@@ -177,7 +183,8 @@
 * 1) Used on list items using number variant; provides a cosmetic cicular border
 * 2) Using flex to keep the number centered
 * 3) Since this will display as a cirlce, keep height and width equal; should be wide enough to fit a two-digit number
-* 4) Using a percentage here instead of --eds-theme-border-radius because this is not a typical bordered object
+* 4) Using a percentage here instead of --eds-theme-border-radius because this must always render as a circle
+* 5) Make the border dark when li is active
 */
 .list-detail__number {
   position: relative;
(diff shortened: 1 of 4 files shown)


src/components/ListDetail/ListDetail.module.css
```diff
@@ -158,14 +159,19 @@
...
+      background: var(--eds-theme-color-text-neutral-default); /* 4 */
+    }
   }
 
   /**
   * List detail item variants
   * 1) Remove the bullet since the icon will take its place
   */
   li[class*='list-detail__item--'] & {
-    /* stylelint-disable-next-line max-nesting-depth */
```
dierat, line 168: Why was `max-nesting-depth` removed from this line?

src/components/ListDetail/ListDetail.tsx
```diff
@@ -295,9 +298,21 @@ export const ListDetail = ({
...
-                      <span className={styles['list-detail__number']}>
-                        {i + 1}
-                      </span>
+                      <span className={styles['list-detail__number']}>{i}</span>
+                    ) : itemVariant === 'incomplete' ? (
+                      <Icon
+                        className={styles['list-detail__icon']}
+                        color={EdsThemeColorBorderNeutralSubtle}
+                        name="star" /* TODO: need to add star-outline variant to Icon */
+                        purpose="decorative"
```
dierat, line 307: Should these icons instead be "informative" and have a `title`? Are the states they communicate also stated in text somewhere so a screen reader user knows that there's an error or warning or it's successfully completed, etc?

src/components/ListDetailPanel/ListDetailPanel.tsx
```diff
@@ -36,14 +47,15 @@ export const ListDetailPanel = ({
...
+  completed,
   id,
   variant,
   title,
   ...other
 }: Props) => {
   const componentClassName = clsx(styles['list-detail__panel'], className, {});
   return (
-    <div
+    <nav
```
dierat, line 58: Why is this element now a nav? I can see why the `tablist` would be a `nav` element but not the `tabpanel`.

src/components/ListDetail/ListDetail.module.css
csnizik: Because I was getting two linting errors on the same line; one for max-nesting-depth and then a second one for the order of the selectors. I tried re-ordering a few different ways and wasn't able to find a combination that the linter liked so I threw in the towel on the entire line.

src/components/ListDetail/ListDetail.tsx
```diff
@@ -295,9 +298,21 @@ export const ListDetail = ({
...
+                      />
+                    ) : itemVariant === 'complete' ? (
+                      <Icon
+                        className={styles['list-detail__icon']}
+                        color={EdsThemeColorBackgroundGradeCompleteDefault}
+                        name="star"
+                        purpose="decorative"
+                      />
```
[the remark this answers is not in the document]
csnizik: Good call, refactoring this.

src/components/ListDetailPanel/ListDetailPanel.tsx
csnizik: This will be revisited in the a11y iteration so I probably should have just left it alone for now. Reverting to <div>.

src/components/ListDetail/ListDetail.tsx
dierat: Sounds good, thanks! 👍

src/components/ListDetail/ListDetail.module.css
dierat: Can you try `/* stylelint-disable-next-line max-nesting-depth no-descending-specificity */`? I'm not 100% sure `no-descending-specificity` is the right eslint rule but that's the one I've been seeing lately. If that doesn't work, maybe just leave another comment above it stating which lint errors you're overriding so it's easier to fix it later if we want.

src/components/ListDetail/ListDetail.tsx
```diff
@@ -222,13 +223,66 @@ export const ListDetail = ({
       return child;
     },
   );
-  const TagName = variant === 'ordered' ? 'ol' : 'ul';
   const componentClassName = clsx(styles['list-detail'], className, {});
 
+  const iconVariant = (itemVariant, i) => {
```
dierat, line 228: Much better, thanks!

src/components/ListDetailPanel/ListDetailPanel.tsx
dierat: Actually, looking at it again, I think you reverted the other `nav` but left this one.

anniehu4 (approved):
> *NOTE: currently the Icon component does not have a variant for "star-outline"; this needs to be ticketed separately.

is there already a design ticket for this?

src/components/ListDetailPanel/ListDetailPanel.tsx
dierat: This one is still a `nav` element btw.

PR #1004, opened 2022-04-22
feat: replace nanoid with react-uid

anniehu4:
### Summary:
[ch192055]

Benefits of `react-uid`:
- stable ID generator (seems like nanoid is not), which allows for potential SSR in the future
- size difference is minimal (1kB gzipped vs. 500 bytes for nanoid)
- traject already uses it
- hook syntax is close to [React's native `useId` hook](https://reactjs.org/blog/2022/03/29/react-v18.html#useid) coming in React 18 (will upgrade when traject upgrades)

### Test Plan:
- sanity check that snapshot IDs still make sense

### DIFF
--- a/src/components/AccordionPanel/AccordionPanel.tsx
+++ b/src/components/AccordionPanel/AccordionPanel.tsx
@@ -1,5 +1,4 @@
 import clsx from 'clsx';
-import { nanoid } from 'nanoid';
 import React, {
   useCallback,
   useEffect,
@@ -8,6 +7,7 @@ import React, {
   MutableRefObject,
   ReactNode,
 } from 'react';
+import { useUID } from 'react-uid';
 import styles from './AccordionPanel.module.css';
 import { EdsThemeColorIconNeutralStrong } from '../../tokens-dist/ts/colors';
 import Icon from '../Icon';
@@ -62,8 +62,12 @@ export const AccordionPanel = ({
   const panelRef = useRef() as MutableRefObject<HTMLDivElement>;
   const [heightVar, setHeight] = useState('0');
   const [isActiveVar, setIsActive] = useState(isActive ? true : false);
-  const [ariaControlsVar, setAriaControls] = useState();
-  const [ariaLabelledByVar, setAriaLabelledBy] = useState();
+
+  const generatedAriaControlsId = useUID();
+  const ariaControlsVar = ariaControls || generatedAriaControlsId;
+
+  const generatedAriaLabelledById = useUID();
+  const ariaLabelledByVar = ariaLabelledBy || generatedAriaLabelledById;
 
   /**
    * Set panel height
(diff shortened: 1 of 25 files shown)


src/components/AccordionPanel/AccordionPanel.tsx
```diff
@@ -62,8 +62,12 @@ export const AccordionPanel = ({
   const panelRef = useRef() as MutableRefObject<HTMLDivElement>;
   const [heightVar, setHeight] = useState('0');
   const [isActiveVar, setIsActive] = useState(isActive ? true : false);
-  const [ariaControlsVar, setAriaControls] = useState();
-  const [ariaLabelledByVar, setAriaLabelledBy] = useState();
+
+  const generatedAriaControlsId = useUID();
+  const ariaControlsVar = ariaControls || generatedAriaControlsId;
```
anniehu4, line 67: `useUID` cannot be used in conditionals or hooks (to ensure consistent ID generation)

src/components/AccordionPanel/AccordionPanel.tsx
```diff
@@ -82,14 +86,12 @@ export const AccordionPanel = ({
   }, [isActiveVar]);
 
   useEffect(() => {
-    setAriaControls(ariaControls || nanoid());
-    setAriaLabelledBy(ariaLabelledBy || nanoid());
```
anniehu4, line 86: I guess this was to ensure `nanoid` only ran once? seems like a funky structure & the hook syntax makes more sense to me

src/components/Fieldset/__snapshots__/Fieldset.test.ts.snap
```diff
@@ -178,7 +178,7 @@ exports[`<Fieldset /> CheckboxField story renders snapshot 1`] = `
   <div
     aria-live="assertive"
     class="field-note checkbox-field__note"
-    id="mock-id"
+    id="6"
```
anniehu4, line 181: matches `aria-describedby` on the legend above 👍  and `input / label` associations look right 👍

src/components/Tabs/Tabs.tsx
```diff
@@ -159,16 +161,20 @@ export const Tabs = ({
...
     setAriaLabelledBy(
       tabs().map((tab) =>
-        tab.props.ariaLabelledBy ? tab.props.ariaLabelledBy : nanoid(),
+        tab.props.ariaLabelledBy
+          ? tab.props.ariaLabelledBy
+          : uidSeed(`${tab}-aria-labelledby`),
       ),
     );
-  }, [tabs]);
+  }, [tabs, uidSeed]);
```
anniehu4, line 177: documentation for `useUIDSeed`: https://github.com/thearnica/react-uid#hooks-168

this is 1 case where `nanoid` feels a little cleaner, but logic this complicated should be rare

ahuth (approved):
LGTM

jinlee93 (approved):
> stable ID generator (seems like nanoid is not), which allows for potential SSR in the future
> size difference is minimal (1kB gzipped vs. 500 bytes for nanoid)
> traject already uses it
> hook syntax is close to [React's native useId hook](https://reactjs.org/blog/2022/03/29/react-v18.html#useid) coming in React 18 (will upgrade when traject upgrades)

To add `react-uid` is [not compatible with Strict or Concurent mode.](https://github.com/thearnica/react-uid#uid) but `nanoid` is [not recommended to use as `key` prop](https://github.com/ai/nanoid#react).

I believe we're not using `React.StrictMode` anywhere in EDS or LP and Concurrent mode is a React 18 feature (where we should be using `useId` anyways) so my vote is also for `react-uid`

PR #1005, opened 2022-04-25
chore(heading, text): lower specificity for margin for app overrides

jinlee93:
### Summary:
- reduces specificity for app overrides in text and heading
### Test Plan:
- manual testing
  - build and use in app, see if class override for margins work
- chromatic


dierat (approved):
Nice!

anniehu4 (approved):

PR #1007, opened 2022-04-25
feat(banner): remove internal dismiss logic

dierat:
### Summary:
There's one thing about our `Banner` in the `next` branch that's very different than the one on `main` and that's the dismiss logic/API.

In the `main` `Banner`, you pass in an `onDismiss` prop to give the `Banner` a close button in the top right. You handle the dismiss logic yourself.

In the `next` `Banner`, the component has some internal dismiss logic that it tracks with component state. To make the `Banner` dismissable, you pass in `dismissable={true}`. The `onDismiss` prop is only needed if there's some logic you need to run at the time that the `Banner` is dismissed.

Now that I'm updating the API in `traject`, I'm wondering if we want to keep that logic or not.

### DIFF
--- a/src/components/Banner/Banner.stories.tsx
+++ b/src/components/Banner/Banner.stories.tsx
@@ -41,6 +41,10 @@ const getDescription = (status?: Variant) => (
   </>
 );
 
+const dismissMethod = () => {
+  console.log('dismissing~');
+};
+
 export const Brand: StoryObj<Args> = {
   render: ({ variant, ...other }) => {
     return (
@@ -137,7 +141,7 @@ export const ErrorWithAction: StoryObj<Args> = {
 export const BrandDismissable: StoryObj<Args> = {
   ...Brand,
   args: {
-    dismissable: true,
+    onDismiss: dismissMethod,
   },
 };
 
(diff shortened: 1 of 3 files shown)


src/components/Banner/Banner.stories.tsx
```diff
@@ -137,55 +141,59 @@ export const ErrorWithAction: StoryObj<Args> = {
 export const BrandDismissable: StoryObj<Args> = {
   ...Brand,
   args: {
-    dismissable: true,
+    onDismiss: dismissMethod,
```
anniehu4, line 144: nit: would [Storybook's actions addon](https://storybook.js.org/docs/react/essentials/actions) work?

```
{
  ...Brand,
  argTypes: { onDismiss: { action: 'dismissed' } },
};
```

if so, it's a little nicer since it surfaces messages in a storybook panel

anniehu4 (approved):

src/components/Banner/Banner.stories.tsx
dierat: It works in storybook but the close button is missing from the snapshots.. 🤔

dierat: https://github.com/chanzuckerberg/edu-design-system/compare/diedra/banner/dismiss-logic...diedra/banner/dismiss-logic-with-dismiss-actions

anniehu4: ahh shoot, I think that is a limitation of the addon 😞  oh well, thanks for trying it out

PR #1011, opened 2022-04-26
feat(tag): update API on main to match next

dierat:
#### High level summary
NOTE: We're planning to merge these API update PRs in a sequence. The sequence will go:
- for each PR, merge in the EDS API update PR
- publish a new package version
- update the package in traject in a PR that has already been reviewed and is waiting to update the component API based on the EDS PR and merge that in
- move on to the next component

To prepare for migrating the `main` branch to use the v1 component versions on `next`, I'm updating the `Tag` component to match the one on `next`. To keep things simple, I only updated the API as necessary and left the internal logic as much as possible. The rest of the component will be updated later.

After landing this, I'll publish a new npm package version, update the EDS package in `traject` and update the component usage to match these changes. That way when we merge the full `next` branch in and deploy, we won't have to update the tags in `traject` and all the other components at the same time. (This component may be published with one or more other components since it's not used as much).

#### API changes for this component:
- `flat` | `outline` variant changed to boolean `hasOutline`
- `color` => `variant`
- "alert" `variant` => "error"
- `children` => `text`

### Test Plan:
- no type errors
- no big chromatic changes for actual components (just story name changes and I extended the icon story to include all the color variants for more thorough testing)


src/components/Tag/Tag.stories.tsx
```diff
@@ -46,25 +46,40 @@ export const OutlineVariants: StoryObj<Args> = {
...
     <div className={styles.tagList}>
       {colorOptions.map((color) => {
-        return <Tag key={color} {...args} color={color} variant="outline" />;
+        return <Tag key={color} {...args} hasOutline={true} variant={color} />;
       })}
     </div>
   ),
 };
 
 export const WithIcon: StoryObj<Args> = {
```
dierat, line 55: I changed this story to include all the colors just for more thorough testing

src/components/Tag/Tag.tsx
```diff
@@ -22,16 +22,12 @@ export const stylesByColor = {
 
 type Props = {
   /**
-   * The color of the tag. New colors should be supported in the flow type,
-   * in the stylesByColor object, and by its separate style in CSS file.
+   * The color variant of the tag. New variants should be supported in the flow type,
+   * in the VARIANTS array, and by its separate style in CSS file.
```
anniehu4, line 26: nit: I know you're minimizing change here, but can we bring in the `VARIANTS` array now & remove "flow type" from the comment
```suggestion
   * The color variant of the tag. New variants should be supported
   * in the VARIANTS array and by its separate style in CSS file.
```

anniehu4 (approved):

src/components/Tag/Tag.tsx
dierat: Thanks! This was copied over from `next` so I'll update it there too.

PR #1015, opened 2022-04-27
feat(HorizontalStepper): first pass for HorizontalStepper component

jinlee93:
### Summary:
[sc-194255]
- introduces the HorizontalStepper component
  - uses the HorizontalStep subcomponent
  - for display only, non navigational

- I don't think there's much accessibility needed for this because it is currently non-interactable and for pagination. Screen reading for pagination should be part of <title> or the navigation buttons  

- not sure if it should be a list

- used for reference: https://mui.com/material-ui/react-stepper/

### Test Plan:
- create snap tests and stories

https://user-images.githubusercontent.com/86632227/165569086-0c2bf04f-2ba0-4b80-8170-1541d163c4f1.mov

Screen Shot 2022-04-27 at 9 40 11 AM
Screen Shot 2022-04-27 at 9 40 13 AM
Screen Shot 2022-04-27 at 9 40 15 AM
Screen Shot 2022-04-27 at 9 40 17 AM
Screen Shot 2022-04-27 at 9 40 30 AM
Screen Shot 2022-04-27 at 9 40 35 AM


src/components/HorizontalStep/HorizontalStep.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  const componentClassName = clsx(
+    styles['horizontal-step'],
+    styles[`horizontal-step--${variant}`],
+    className,
+  );
+  return (
+    <div className={componentClassName}>
+      {icon}
+      <Text as="span" size="sm" variant="inherit" weight="bold">
```
jinlee93, line 95: Can add styled class here to make Text hidden/visible according to breakpoint

src/components/HorizontalStepper/HorizontalStepper.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const stepComponents: React.ReactNode[] = [];
+  steps.forEach((step, index) => {
+    /* 1 */
+    if (index > 0)
+      stepComponents.push(
+        <hr
+          className={styles['horizontal-stepper__line']}
+          key={`horziontal-stepper__line-${index}`}
+        />,
+      );
```
jinlee93, line 50: Didn't use existing `<Hr>` component because it seemed to have unnecessary styling

src/components/HorizontalStep/HorizontalStep.module.css
```diff
@@ -0,0 +1,59 @@
+/*------------------------------------*\
+    #HORIZONTALSTEP
```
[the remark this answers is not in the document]
jinlee93: AH! thanks for the catch, readability is good

src/components/HorizontalStep/HorizontalStep.module.css
```diff
@@ -0,0 +1,59 @@
...
+}
+.horizontal-step--active {
+  color: var(--eds-theme-color-text-neutral-strong);
+}
+.horizontal-step--incomplete {
+  color: var(--eds-theme-color-text-neutral-subtle);
+}
+
+/**
+ * Horizontal Step check circle icon
```
dierat, line 28: Nit – I would avoid referencing the exact icon unless it's needed to understand the styling. These kinds of details in the design sometimes shift around and then the comments go stale and become confusing.

jinlee93: Makes sense to me, how about `Horizontal Step Complete`?

src/components/HorizontalStep/HorizontalStep.tsx
```diff
@@ -0,0 +1,100 @@
...
+   */
+  variant: 'complete' | 'active' | 'incomplete';
+}
+
+/**
+ * ```ts
+ * import {HorizontalStep} from "@chanzuckerberg/eds";
+ * ```
+ *
+ * A step sub component for the <HorizontalStepper>.
```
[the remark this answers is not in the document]
jinlee93: I was trying to say that the `HorizontalStep` is a sub component of the `HorizontalStepper`... How can I change it to make that more clear?

jinlee93:
@csnizik and I synced on this and its similarities to the `ListDetail` component ([mock](https://www.figma.com/file/uq098c88MZh036TLKYJn2X/EDS-Pilot%3A-Project-Overview?node-id=75%3A13806)) and made a choice to go in this direction of making a separate `HorizontalStepper` component due to:
1. No tabbing panel / navigation / interaction logic in the `HorizontalStepper`
  - This also means less accessibility logic needed 
  - `HorizontalStepper` is controlled externally
2. `HorizontalStepper` being, well, horizontal
  - Different responsive considerations
  - Separator logic different

ListDetail story: https://61313967cde49b003ae2a860-iamqqidqot.chromatic.com/?path=/story/example-listdetail--ordered
image

src/components/HorizontalStep/HorizontalStep.tsx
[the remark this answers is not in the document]
dierat: Oooh I see, I was reading it too fast. Those components are really similarly named. I say leave it unless other people find it confusing.

src/components/HorizontalStepper/HorizontalStepper.module.css
```diff
@@ -0,0 +1,26 @@
...
+ * The decorative line between steps.
+ * 1) Used to grow and shrink lines before wrapping step text.
+ */
+.horizontal-stepper__line {
+  border: var(--eds-theme-border-width) solid
+    var(--eds-theme-color-border-neutral-default);
+  margin-left: var(--eds-size-1);
+  margin-right: var(--eds-size-1);
+
+  flex-grow: 1; /* 1 */
```
dierat, line 25: Awesome!

src/components/HorizontalStepper/HorizontalStepper.module.css
```diff
@@ -0,0 +1,26 @@
+/*------------------------------------*\
+    # HORIZONTAL STEPPER
+\*------------------------------------*/
+
+/**
+ * Horizontal Stepper wraps the stepper and adds lines between steps.
```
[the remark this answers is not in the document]
jinlee93: Yea it means it wraps multiple `HorizontalStep` instances, think it was a late night typo, will clarify

src/components/HorizontalStepper/HorizontalStepper.stories.tsx
```diff
@@ -0,0 +1,118 @@
...
+  },
+};
+
+export const CompletedAllSteps: StoryObj<Args> = {
+  args: {
+    activeIndex: 5,
+  },
+};
+
+const InteractiveHorizontalStepper = ({ steps }) => {
```
dierat, line 52: Awesome story 👍

src/components/HorizontalStepper/HorizontalStepper.tsx
```diff
@@ -0,0 +1,73 @@
...
+        ? 'complete'
+        : index === activeIndex
+        ? 'active'
+        : 'incomplete';
+
+    /* 3 */
+    stepComponents.push(
+      <HorizontalStep
+        key={`horizontal-stepper__step-${index}`}
+        stepNumber={index === activeIndex ? index + 1 : undefined}
```
dierat, line 64: I'm finding the `stepNumber` logic a little confusing. Could we instead always pass in the number and determine whether or not to show it based on the variant?

jinlee93: Good call, we can always pass it in, will change

src/components/HorizontalStepper/HorizontalStepper.tsx
```diff
@@ -0,0 +1,73 @@
...
+   */
+  activeIndex: number;
+  /**
+   * CSS class names that can be appended to the component.
+   */
+  className?: string;
+  /**
+   * Ordered list of step texts.
+   */
+  steps: string[];
```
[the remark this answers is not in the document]
dierat: > Empty strings would work but would it be intuitive?

Not sure, but I also don't want to make the API convoluted for a case that we may not even need. This hasn't come up in the designs aside for mobile view, right?

src/components/HorizontalStep/HorizontalStep.module.css
```diff
@@ -0,0 +1,59 @@
...
+.horizontal-step__complete-icon {
+  fill: var(--eds-theme-color-icon-utility-success);
+}
+
+/**
+ * Horizontal Step Incomplete
+ */
+.horizontal-step__incomplete-icon {
+  margin-left: var(--eds-size-1);
+  margin-right: var(--eds-size-1);
```
csnizik, line 39: These left and right margins look like they're causing the gaps to be too wide

src/components/HorizontalStepper/HorizontalStepper.tsx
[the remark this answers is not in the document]
jinlee93: Yes, and for mobile view we can hide text in the subcomponent rather than here. Agree with not making the API more complex than necessary

src/components/HorizontalStepper/HorizontalStepper.tsx
```diff
@@ -0,0 +1,73 @@
+import clsx from 'clsx';
+import React from 'react';
+import styles from './HorizontalStepper.module.css';
+import HorizontalStep from '../HorizontalStep';
+
+export interface Props {
+  /**
+   * Identifies which index is the active step.
+   */
+  activeIndex: number;
```
dierat, line 10: Should we add a check to make sure that the `activeIndex` is within the length of `steps`? Ex: if there are 3 steps and they pass in `actvieIndex={3}`, throw a console warning.

dierat: Maybe also mention in the prop comment that it's using a 0-based index just in case; I get thrown off by that all the time.

src/components/HorizontalStep/HorizontalStep.module.css
jinlee93: Good callout, this was done intentionally to keep the `incomplete-icon` width the same as the other icons, so there's no jitter when navigating. Of course would have to pass it by design first, but was waiting until the actual pilot launch for that.

Without the margins, jitter:

https://user-images.githubusercontent.com/86632227/165593826-476df1dd-c43f-4129-b0f7-b18503e31547.mov

src/components/HorizontalStepper/HorizontalStepper.tsx
jinlee93: Good suggestions. For the check, all I know is the `console.warn` method so I can do that (unless other suggestions). Will definitely add that it's a 0-based index, thanks!

src/components/HorizontalStepper/HorizontalStepper.tsx
```diff
@@ -0,0 +1,73 @@
...
+   * 3) Add step to stepComponents.
+   */
+  const stepComponents: React.ReactNode[] = [];
+  steps.forEach((step, index) => {
+    /* 1 */
+    if (index > 0)
+      stepComponents.push(
+        <hr
+          className={styles['horizontal-stepper__line']}
+          key={`horziontal-stepper__line-${index}`}
```
csnizik, line 48: typo

jinlee93: Ah! Will fix

src/components/HorizontalStepper/HorizontalStepper.tsx
```diff
@@ -0,0 +1,82 @@
...
+ * Example usage:
+ *
+ * ```tsx
+ * <HorizontalStepper activeIndex={0} steps={['Step 1', 'Step 2', 'Step 3']} />
+ * ```
+ */
+export const HorizontalStepper = ({ activeIndex, className, steps }: Props) => {
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    !(activeIndex >= 0 && activeIndex <= steps.length)
```
jinlee93, line 37: Conditional written as such to also account for `NaN` values which pass the Typescript type check for `number`

src/components/HorizontalStep/HorizontalStep.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+export interface Props {
+  /**
+   * CSS class names that can be appended to the component.
+   */
+  className?: string;
+  /**
+   * Indicates which number the step is.
+   */
+  stepNumber?: number;
```
dierat, line 15: Can we make this required now since you're always passing it in on line 73 in `HorizontalStepper.tsx`? And then we could also remove the check on 49 of this file?

jinlee93: I initially made this optional in case there was use outside of the `HorizontalStepper` but I agree with making this required and crossing the other use cases when we get to it

src/components/HorizontalStepper/HorizontalStepper.tsx
dierat, line 37: Could we use an or operator to avoid the negation? The negatives makes it a littler harder to reason about the logic.
```suggestion
    (activeIndex < 0 || activeIndex > steps.length)
```

dierat: Also, maybe add a test?

jinlee93: Will add test for this and the other `console.warn` in the subcomponent.
Logic wise, I did the negative to account for `NaN` which pass the type check for number, but if you think that's low value, can change to your suggestion
image

src/components/HorizontalStepper/HorizontalStepper.tsx
```diff
@@ -0,0 +1,82 @@
+import clsx from 'clsx';
+import React from 'react';
+import styles from './HorizontalStepper.module.css';
+import HorizontalStep from '../HorizontalStep';
+
+export interface Props {
+  /**
+   * Zero-based index that identifies which index is the active step.
+   */
+  activeIndex: number;
```
dierat, line 10: I didn't realize until I read the warning code below that if all the steps are completed, the `activeIndex` is actually a step that doesn't exist, one beyond the last step – is that right? If there are 5 steps and you've completed all of them, the `activeIndex` would be 5, a non-existent step? That seems a bit odd to me.. I wonder if `activeIndex` should be `null` if the steps are completed since there is no active step.

jinlee93: I do agree that it's a little awkward to have the `activeIndex` actually point to a nonexisting index on the `steps` array, but I went with this since I see the parent logic being written like:
```
  const onClickBack = () => {
    if (activeIndex > 0) setActiveIndex(activeIndex - 1);
  };
  const onClickNext = () => {
    if (activeIndex < steps.length) setActiveIndex(activeIndex + 1);
  };
```
How do you feel about this?

dierat:
> I don't think there's much accessibility needed for this because it is currently non-interactable and for pagination. Screen reading for pagination should be part of <title> or the navigation buttons

Are you saying this widget should basically be hidden because the information should be accounted for somewhere else? I don't think that's true because this widget doesn't just say what step the user is on; it tells them how many steps they've completed, how many are coming up, and what those steps are called.

src/components/HorizontalStepper/HorizontalStepper.tsx
dierat: Sorry for the confusion – I figured it a little bit after leaving my comment. Follow-up though – does the type check not catch it when `activeNumber` is `NaN` since it's now a required `number`?

jinlee93: Type check doesn't catch it since `typeof NaN === 'number'`
TS playground showing no error on NaN:
https://www.typescriptlang.org/play?#code/ATBQBcE8AcFNgHIFcC2AjWAnA8mgVrAMbjAC8wA3qCCAIYBcwAdqhpgNygC+oohA9kwDOJZOiy48jMW0lES5KjWANgARgA03XgOGjaCOcWmsJ+eWUrUaqhAa08+gkYiQAbN0fAnxOc8UslG0YWD20gA

jinlee93:
> Are you saying this widget should basically be hidden because the information should be accounted for somewhere else? I don't think that's true because this widget doesn't just say what step the user is on; it tells them how many steps they've completed, how many are coming up, and what those steps are called.

Very fair point. I did not consider the info of the texts of the upcoming steps since I was thinking something like Approach 1 or Approach 2 of  https://www.w3.org/WAI/tutorials/forms/multi-page/.

However even in Approach 4 of that example, except for the current step `<a>` I don't think there's aria information about upcoming and past steps. This would be similar to having a <title> or <heading> with current step and using an external Back or Next button to communicate past and future steps, which are external to this component.
Unless formatting in a list as in the Approach 4 example communicates the information.

dierat:
>Very fair point. I did not consider the info of the texts of the upcoming steps since I was thinking something like Approach 1 or Approach 2 of https://www.w3.org/WAI/tutorials/forms/multi-page/.
>
>However even in Approach 4 of that example, except for the current step <a> I don't think there's aria information about upcoming and past steps. This would be similar to having a <title> or <heading> with current step and using an external Back or Next button to communicate past and future steps, which are external to this component.
Unless formatting in a list as in the Approach 4 example communicates the information.
>
>And in the [MUI stepper](https://mui.com/material-ui/react-stepper/) the widget is basically hidden?

My personal feeling is that all this really needs to be accessible is:
- make it a list
- add alternative text to the icons/number in a circle to say the step is completed, current, or future (not sure about future)

The w3.org example #4 uses hidden spans for the alternative text. They don't indicate the future steps so maybe we don't need to put anything there, I dunno. 

```
<div class="tlwrapper">
	<ol class="timeline">
		<li class="timeline-past">
				<span class="visuallyhidden">Completed: </span>
				<a href="billing.html">Billing Address</a>
		</li>
		<li class="timeline-current">
			<span class="visuallyhidden">Current: </span>
			<span>Shipping Address</span>
		</li>
		<li><span>Review Order</span></li>
		<li><span>Payment</span></li>
		<li><span>Finish Purchase</span></li>
	</ol>
</div>
```

src/components/HorizontalStepper/HorizontalStepper.tsx
dierat: Cool, thanks for checking. Could you put that comment in the code for future devs?

src/components/HorizontalStepper/HorizontalStepper.tsx
dierat: Yeah I'm not sure. I can see how it's adding extra logic for them to check if it's the last step and if so set it to `null`, but I also feel like the one-step-past-the-edge API is odd. Maybe just add a comment to the prop explaining what the `activeNumber` should be if all the steps are completed.

src/components/HorizontalStepper/HorizontalStepper.tsx
jinlee93: Yes will do

src/components/HorizontalStepper/HorizontalStepper.tsx
jinlee93: Sounds good!

jinlee93:
> My personal feeling is that all this really needs to be accessible is:
> 
> * make it a list
> * add alternative text to the icons/number in a circle to say the step is completed, current, or future (not sure about future)
> 
> The w3.org example #4 uses hidden spans for the alternative text. They don't indicate the future steps so maybe we don't need to put anything there, I dunno.

Will apply suggestions, thanks!

src/components/HorizontalStepper/HorizontalStepper.tsx
```diff
@@ -0,0 +1,87 @@
...
+ * Example usage:
+ *
+ * ```tsx
+ * <HorizontalStepper activeIndex={0} steps={['Step 1', 'Step 2', 'Step 3']} />
+ * ```
+ */
+export const HorizontalStepper = ({ activeIndex, className, steps }: Props) => {
+  /**
+   * Warns dev if the activeIndex is invalid
+   * 1) Negative conditional to account for 'NaN' values which pass the number type check since "typeof NaN === 'number'"
```
dierat, line 38: This is great, thanks!

jinlee93:
In the EDS Sync @ifrost1 @csnizik @dierat the number icon could possibly be extracted for reuse

src/components/HorizontalStep/HorizontalStep.tsx
```diff
@@ -0,0 +1,92 @@
...
+        purpose="informative"
+        size="1.5rem"
+        title={`completed step ${text}`}
+      />
+    ) : variant === 'active' ? (
+      <Text
+        aria-label={`current step ${stepNumber}: ${text}`}
+        as="span"
+        className={styles['horizontal-step__number']}
+        role="img"
```
jinlee93, line 63: Felt weird about this but think it should be `role="img"` because it's like an icon (svg) but using a span element for the text and borders for the circle. This also allows adding of the `aria-label` as there are reports of screen readers not reading the `aria-label` of `<span>`.
@dierat  thoughts?

jinlee93:
Synced with Ryan and Louie about the stepper:

When break
- Todo according to Courtney’s grid (possibly 375px or 320px) but not part of this case
Length of items
- More than 3 or 4 items would be corner case
Text Wrapping
- Should break and hide text and hence text wrapping should not be a consideration
Should have as for link or not 
- Not navigable for the use case of CAD
Which links to disable
- Not considered yet since navigable cases are not determined yet

jinlee93:
Will be extracting number icon in a future PR

src/components/HorizontalStep/HorizontalStep.module.css
jinlee93: Confirmed with design that we don't want jitter and would rather have extra spacing

PR #1016, opened 2022-04-27
feat(dropdown-button): rename children prop to text

dierat:
#### High level summary
NOTE: We're planning to merge these API update PRs in a sequence. The sequence will go:
- for each PR, merge in the EDS API update PR
- publish a new package version
- update the package in traject in a PR that has already been reviewed and is waiting to update the component API based on the EDS PR and merge that in
- move on to the next component

To prepare for migrating the `main` branch to use the v1 component versions on `next`, I'm updating the `DropdownButton` component to match the one on `next`. To keep things simple, I only updated the API as necessary and left the internal logic as much as possible. The rest of the component will be updated later.

After landing this, I'll publish a new npm package version, update the EDS package in `traject` and update the component usage to match these changes. That way when we merge the full `next` branch in and deploy, we won't have to update the tags in `traject` and all the other components at the same time. (This component may be published with one or more other components since it's not used as much).

#### API changes for this component:
- `children` => `text`

NOTE: I did not make any changes to the `Dropdown` API itself because the `Dropdown.Button` can actually take a function instead of text. I'm also wondering if we want this API change at all since `Button` uses `children`. It feels a little out-of-sync 🤔 

For an example of a `Dropdown` using a function for children of the `Dropdown.Button`, take a look at the `InteractiveExampleUsingFunctionChildren` story: https://github.com/chanzuckerberg/edu-design-system/blob/d8e6f690611a8b8bdccdaa5b8fe7bcbc7cee4ed6/src/components/Dropdown/Dropdown.stories.tsx#L109

### Test Plan:
- no type errors
- no chromatic changes
- `Dropdown` component stories work as expected

### DIFF
--- a/src/components/DropdownButton/DropdownButton.tsx
+++ b/src/components/DropdownButton/DropdownButton.tsx
@@ -5,7 +5,7 @@ import styles from "./DropdownButton.module.css";
 
 type Props = {
   className?: string;
-  children?: ReactNode;
+  text?: ReactNode;
 };
 
 /**
@@ -16,16 +16,16 @@ type Props = {
  * A styled button with an expand icon to be used in triggering Popovers, Dropdowns, etc.
  */
 const DropdownButton = forwardRef<HTMLButtonElement, Props>(
-  ({ children, className, ...rest }, ref) => {
+  ({ text, className, ...rest }, ref) => {
     return (
       <button
         className={clsx(styles.dropdownButton, className)}
         ref={ref}
         {...rest}
       >
-        {/* Wrapping span ensures that `children` and icon will be correctly pushed to
-            either side of the button even if `children` contains more than one element. */}
-        <span>{children}</span>
+        {/* Wrapping span ensures that `text` and icon will be correctly pushed to
+            either side of the button even if `text` contains more than one element. */}
+        <span>{text}</span>
         <Icon name="expand-more" purpose="decorative" size="1.25rem" />
       </button>
     );
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -390,9 +390,8 @@ const DropdownTrigger = function (
       ) : (
         <DropdownButton
           className={clsx(className, compact && styles.compactDropdownButton)}
-        >
-          {children}
-        </DropdownButton>
+          text={children}
+        />
       )}
     </Listbox.Button>
   );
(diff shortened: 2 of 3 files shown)


jinlee93 (approved):
LGTM

src/components/DropdownButton/DropdownButton.tsx
```diff
@@ -5,7 +5,7 @@ import styles from "./DropdownButton.module.css";
 
 type Props = {
   className?: string;
-  children?: ReactNode;
+  text?: ReactNode;
```
anniehu4, line 8: > I'm also wondering if we want this API change at all since Button uses children. It feels a little out-of-sync

hmmmm good question. could we use this heuristic?
- If it can be typed as a `string`, use `text` (we're always expecting simple input, so there's benefit to making the API match that expectation)
- If it has to be typed as `ReactNode`, use `children` (we're allowing potentially more complicated input, so "text" may not be an accurate prop name)

anniehu4: I think `DropdownButton` is mostly a utility component (only used in implementing our other components), so the API difference with Button isn't confusing.

Are we going to update the external-facing `Dropdown.Button` and `Dropdown.Label` API as well though? If not, less sure the benefit of this change

anniehu4 (changes requested):
upon re-reading realized this is only updating `DropdownButton` and not the sub-component APIs in `Dropdown` -- see new comment

src/components/DropdownButton/DropdownButton.tsx
dierat: >Are we going to update the external-facing Dropdown.Button and Dropdown.Label API as well though? If not, less sure the benefit of this change. I think those 3 should stay in sync (in whichever direction we choose)

That was actually where it started feeling really strange to me – `Dropdown.Button` has to be able to accept other types because it can be passed a function: https://github.com/chanzuckerberg/edu-design-system/blob/d8e6f690611a8b8bdccdaa5b8fe7bcbc7cee4ed6/src/components/Dropdown/Dropdown.stories.tsx#L109

anniehu4: oo shoot, right 😓  Feels easier to me to leave `children` then

dierat: Cool, I'll close this and the ticket then. Thanks!

PR #1017, opened 2022-04-27
fix: useUIDSeed usage and guidelines

anniehu4:
### Summary:
[ch194920]

Noticed in a Chromatic diff that the way I was using `useUIDSeed` isn't valid (in particular, doing something like `${object}-string` creates an output ID `[object Object]-string`). This corrects our usage & adds guidelines

### Test Plan:
- For `ListDetail`, confirmed valid IDs were generated, IDs were different for each item, and `id` and `aria-labelledby` were distinct for each item
Screen Shot 2022-04-27 at 1 48 46 PM

- Confirmed aria is hooked up correctly using the Accessibility Tree
Kapture 2022-04-27 at 14 11 06

### DIFF
--- a/docs/CODE_GUIDELINES.md
+++ b/docs/CODE_GUIDELINES.md
@@ -481,6 +481,38 @@ EDS adheres to the following API naming conventions:
 
 # Accessibility <a name="accessibility"></a>
 
+## Generating IDs
+
+ID attributes used for accessibility (e.g. associating `<label>` and `<input>` elements) should be unique and stable.
+
+We currently use [react-uid](https://www.npmjs.com/package/react-uid) hooks for ID generation. To ensure stable results, they cannot be invoked within conditionals or callbacks.
+
+- `useUID()` is the most common usage.
+
+```tsx
+const generatedId = useUID();
+```
+
+- `useUIDSeed()` generates a stable seed generator for use in iterators.
+
+```tsx
+const getUID = useUIDSeed();
+// you should either pass an object to getUID:
+items.forEach((item) => {
+  const generatedId = getUID(item);
+});
+
+// or pass a constructed string:
+items.forEach((item, index) => {
+  const generatedId = getUID(`item-${index}-aria-labelledby`);
+});
+
+// interpolating an object into a string will NOT work:
+// items.forEach((item) => {
+//   const generatedId = getUID(`${item}-id`);
+// });
+```
+
 ## Tools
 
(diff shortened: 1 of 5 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -164,17 +163,13 @@ export const Tabs = ({
    * Autogenerate ids on tabs if not defined.
    */
   useEffect(() => {
-    setId(
-      tabs().map((tab) => (tab.props.id ? tab.props.id : uidSeed(`${tab}-id`))),
-    );
+    setId(tabs().map((tab) => (tab.props.id ? tab.props.id : getUID(tab))));
```
ahuth, line 166: Would a better type on `idVar` / `setId` have prevented this issue?

I'm also surprised TS let us use `setId` with `any[]` as the implicit type 🤔

anniehu4: I think the typing improvement should be made on the `getUID` side instead (or better docs on their side) -- `getUID` was still returning a string (that happened to be something like `"1val-[object Object]-id"`, which is pretty meaningless), so better typing of `setId` wouldn't have caught it

ahuth: Oh, it's because `strict` mode is off on the `next` branch.

@anniehu4, do you know why that is? My recommendation is to turn that on. Probably would've prevented this issue.

anniehu4: although yes, agree that there are an unfortunate number of `any` types in this file 😬

anniehu4: yup, there are a couple of things we had to turn off in `next` because the existing BigMedium code had a bunch of violations, and it wasn't clear which components we needed to touch yet. We've been tracking this particular issue: https://app.shortcut.com/czi-edu/story/187478/re-enable-strict-true-in-tsconfig-json?cf_workflow=500033078&ct_workflow=all&groupBy=workflowState

ahuth: Ah gotcha, so either way it returns a string. We still probably always want `strict: true`.

ahuth (approved):
LGTM 🤘

docs/CODE_GUIDELINES.md
```diff
@@ -481,6 +481,38 @@ EDS adheres to the following API naming conventions:
...
+});
+
+// or pass a constructed string:
+items.forEach((item, index) => {
+  const generatedId = getUID(`item-${index}-aria-labelledby`);
+});
+
+// interpolating an object into a string will NOT work:
+// items.forEach((item) => {
+//   const generatedId =  getUID(`${item}-id`);
```
dierat, line 512: Super nit
```suggestion
//   const generatedId = getUID(`${item}-id`);
```

dierat (approved):
Awesome! Thanks for adding the documentation ❤️

docs/CODE_GUIDELINES.md
anniehu4: oop haha good catch

PR #1018, opened 2022-04-27
docs: make Tooltip snapshots not flaky

anniehu4:
### Summary:
[ch194410]

Looking into the newly flaky Tooltip snapshots ... some things I noticed:
- For some reason, the animation duration _looks_ longer on `next` (even though it's the same value as `main`). can't tell if this is me being crazy 😩 
- `@tippyjs/react` versions: `main` is on `^4.2.6` while `next` is on `4.2.5`
- setting `animation: false` doesn't work in stories because it also sets `data-state="hidden"`. we target the latter to set `opacity: 0` to enable the fade animation

So the two things I did in this PR are:
- upgrade `@tippyjs` to match `main`
- set `duration: 0` in the always-visible tooltip stories to ensure flakiness isn't because of the animation

If this still doesn't work, I can look into storybook `preview` styling differences between the two branches && possibly adding a delay to the snapshots

### Test Plan:
monitor chromatic? after this lands

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -7,7 +7,6 @@ module.exports = {
   moduleNameMapper: {
     '\\.css$': 'identity-obj-proxy',
     'react-portal': 'identity-obj-proxy',
-    shortid: 'identity-obj-proxy',
     '.*\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
       '<rootDir>/jest/mocks/fileMock.js',
   },
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
   },
   "dependencies": {
     "@headlessui/react": "^1.5.0",
-    "@tippyjs/react": "4.2.5",
+    "@tippyjs/react": "^4.2.6",
     "clsx": "^1.1.1",
     "downshift": "^6.1.7",
     "react-children-by-type": "^1.1.0",
--- a/src/components/Tooltip/Tooltip.stories.tsx
+++ b/src/components/Tooltip/Tooltip.stories.tsx
@@ -4,8 +4,6 @@ import { within, userEvent } from '@storybook/testing-library';
 import clsx from 'clsx';
 import React from 'react';
 import { Tooltip } from './Tooltip';
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore story files not type-checked
 import styles from './Tooltip.stories.module.css';
 import { Button } from '../Button/Button';
 
@@ -23,6 +21,9 @@ const defaultArgs = {
     </Button>
   ),
   align: 'right',
+  // most stories show a visible, non-interactive tooltip.
+  // this turns animation off to ensure stable visual snapshots
+  duration: 0,
   visible: true,
 };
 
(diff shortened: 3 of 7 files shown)


jest.config.js
```diff
@@ -7,7 +7,6 @@ module.exports = {
   moduleNameMapper: {
     '\\.css$': 'identity-obj-proxy',
     'react-portal': 'identity-obj-proxy',
-    shortid: 'identity-obj-proxy',
```
anniehu4, line 10: this isn't used anywhere

package.json
```diff
@@ -65,7 +65,7 @@
   },
   "dependencies": {
     "@headlessui/react": "^1.5.0",
-    "@tippyjs/react": "4.2.5",
```
anniehu4, line 68: @jinlee93 do you remember why you locked to a specific version when moving the Tooltip over?

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -124,14 +123,11 @@ export const LongButtonText: StoryObj<Args> = {
...
   },
-  parameters: {
-    // Is flaky on chromatic, disabling for now
-    chromatic: { disableSnapshot: true },
-  },
 };
 
 export const Interactive: StoryObj<Args> = {
   args: {
+    duration: undefined,
```
ahuth, line 130: Does the duration end up being `undefined` or the default of 200 here?

ahuth (approved):
LGTM 🤘

src/components/Tooltip/Tooltip.stories.tsx
anniehu4: oh great question -- looks like Tippy is still getting `200`
Screen Shot 2022-04-27 at 3 48 31 PM

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -23,6 +21,7 @@ const defaultArgs = {
     </Button>
   ),
   align: 'right',
+  duration: 0,
```
dierat, line 24: Maybe add a comment in the code explaining this line.

src/components/Tooltip/Tooltip.tsx
```diff
@@ -15,7 +15,9 @@ type TooltipProps = {
   /**
    * Behavior of the tooltip transition, defaults to an opacity "fade".
    * Animation guidelines are provided in https://atomiks.github.io/tippyjs/v5/animations/.
-   * A false value will disable animations.
+   *
+   * Set animation={false} to disable animations **in tests only**.
+   * For some reason this also hides the Tooltip visually.
```
dierat, line 20: Is this still accurate? Since you're changing the `duration` in this PR.

dierat (approved):
Thanks!

src/components/Tooltip/Tooltip.stories.tsx
ahuth: Makes sense. What's the intent of setting this to `undefined` here? Should we remove the line?

anniehu4: oh -- I set `duration : 0` in `defaultArgs` earlier (line 24), so thought it needed to be explicitly reset here the same way `visible` is being reset

src/components/Tooltip/Tooltip.tsx
anniehu4: hmm there was one tooltip test (the escape key one) passing `animation={false}`. lemme see if `duration` works instead

anniehu4: duration works! will update comment

PR #1019, opened 2022-04-28
chore(text): reintroduce spacing variant in Text

jinlee93:
### Summary:
- global styles in LP set margin bottom for `p` overriding the `:where(.text)` margin reset
- published alpha, tests passing in text/heading use with alpha in https://github.com/FB-PLP/traject/pull/10779
- once merged, will be publishing patch build `0.12.1` off `jlee/v0.12.1` and updating aforementioned pr to it 

### Test Plan:
- ci tests pass

### Next steps:
- merge v0.12.1 into v0.13.0 and release v0.13.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [0.12.1-alpha.0](https://github.com/chanzuckerberg/edu-design-system/compare/v0.12.0...v0.12.1-alpha.0) (2022-04-27)
+
+* chore(text): reintroduce spacing variant for text
+
 ## [0.12.0](https://github.com/chanzuckerberg/edu-design-system/compare/v0.10.0...v0.12.0) (2022-04-27)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,6 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
-## [0.12.1-alpha.0](https://github.com/chanzuckerberg/edu-design-system/compare/v0.12.0...v0.12.1-alpha.0) (2022-04-27)
-
-* chore(text): reintroduce spacing variant for text
 
 ## [0.12.0](https://github.com/chanzuckerberg/edu-design-system/compare/v0.10.0...v0.12.0) (2022-04-27)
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [0.12.1-alpha.0](https://github.com/chanzuckerberg/edu-design-system/compare/v0.12.0...v0.12.1-alpha.0) (2022-04-27)
+
+* chore(text): reintroduce spacing variant for text
+
```
dierat, line 8: Do we want the alpha versions in here? I don't think we've been adding them in the past.

dierat (approved):
Nice!

Questions:
- Are we thinking of phasing this prop out?
  - If so, should we add deprecation warnings and make a ticket to remove it at some point?
  - If not, should we add this prop to the stories?
- Are you planning to make this change on `next` or shall I? (I just don't want to get in your way or duplicate work.)

src/components/Text/Text.module.css
```diff
@@ -6,7 +6,7 @@
  * 1) Needs `:where` pseudoclass to reduce specificity so Tailwind and other class overrides 
  *    for margin can happen where used.
  */
-:where(.text) { /* 1 */
+.text { /* 1 */
```
anniehu4, line 9: nit: let's either remove or update the comment here

CHANGELOG.md
jinlee93: I think we could be keeping these, and historically are on npm. What are your thoughts?

jinlee93:
> Are we thinking of phasing this prop out?
> * If so, should we add deprecation warnings and make a ticket to remove it at some point?
> * If not, should we add this prop to the stories?

Good question, I will add stories for the spacing variants because I am not sure what the outcome of the CSS untangling would be, and would have to keep this prop for longer than hoped for.

> Are you planning to make this change on next or shall I? (I just don't want to get in your way or duplicate work.)

Would appreciate if you could, think I would need to focus some time on the CAD pilot

CHANGELOG.md
dierat: Personally I don't want to feel like I have to add every alpha version when we're just using them for testing. Seems like a hassle, and I don't see the benefit of having it logged here.

jinlee93: Ahh makes sense, will remove

src/components/Text/Text.stories.tsx
```diff
@@ -79,6 +79,23 @@ export const Overline: StoryObj<Args> = {
   },
 };
 
+export const Spacing: StoryObj<Args> = {
+  render: () => (
+    <div>
```
jinlee93, line 84: initially had this as fragment, and the snap would only take the first `<Text>` child

dierat (approved):

PR #1021, opened 2022-04-28
docs(banner): update component comment to use updated API

dierat:
### Summary:
The `Banner` component comment has an API example that's stale. The example is representative of the first version of the API I used on `next`. The API was changed during the PR process, but I forgot to update the example.

### Test Plan:
Verify the updated example appears in storybook.


anniehu4 (approved):
thanks!

PR #1024, opened 2022-04-29
chore(icon): add viewBox prop

dierat:
### Summary:
In updating the `Icon` in `traject` to use the new API, I discovered we still need the `viewBox` prop because custom icons (ie the `SummitLogoIcon`) may need it to render properly. I also added a story to test this use case.

Without `viewBox` prop:
the with svg children story in storybook without the viewbox prop. on the top left of the summit logo icon is visible; the rest is cut off

With `viewBox` prop:
the with svg children story in storybook with the viewbox prop. the summit logo icon is fully visible

### Test Plan:
Verify the icon renders as expected in the `WithSvgChildren` story and there are no other changes detected by chromatic.


anniehu4 (approved):
oop good catch

dierat:
Thanks for the quick review!

PR #1039, opened 2022-05-05
chore(banner): stop passing 'other' props down to the underlying article element

dierat:
### Summary:
Due to the `NuxMegaphone` component in `traject` being annoying, I removed the `other` props being passed down to the underlying `article` element in the `Banner` component in https://github.com/chanzuckerberg/edu-design-system/pull/1032. This PR brings that change to `next` as well.

### Test Plan:
There should be no changes in the `Banner` component.

### DIFF
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -106,7 +106,7 @@ const variantToIconAssetsMap: {
  * <Banner
  *   onDismiss={handleDismiss}
  *   title="Some Title"
- *   description={<>Some description, possibly with a <Link href="www.some-other-resource.com">link to some other resource</Link>.</>}
+ *   description={<>Some description, possibly with a <Link href="https://go.czi.team/eds">link to some other resource</Link>.</>}
  *   action={<Button onClick={handleAction}>Action</Button>}
  * />
  * ```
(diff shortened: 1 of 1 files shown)


src/components/Banner/Banner.tsx
```diff
@@ -119,10 +119,9 @@ export const Banner = ({
   isFlat,
   onDismiss,
   orientation,
-  variant = 'brand', // TODO: verify brand is the default variant and not neutral
```
dierat, line 122: I feel confident "brand" is the default `variant` option.

src/components/Banner/Banner.tsx
```diff
@@ -106,7 +106,7 @@ const variantToIconAssetsMap: {
  * <Banner
  *   onDismiss={handleDismiss}
  *   title="Some Title"
- *   description={<>Some description, possibly with a <Link href="www.some-other-resource.com">link to some other resource</Link>.</>}
+ *   description={<>Some description, possibly with a <Link href="https://go.czi.team/eds">link to some other resource</Link>.</>}
```
dierat, line 109: Made this change because this is the `href` we used in the button stories on `main`.

anniehu4 (approved):

PR #1041, opened 2022-05-05
test(databar): assign roles for accessibility

jinlee93:
### Summary:
- adds roles to the databar and its elements for accessibility
- this mirrors the horizontal bar in LP
### Test Plan:
- accessibility tests pass
- no visual diff
- unit tests pass

### DIFF
--- a/src/components/DataBar/DataBar.stories.tsx
+++ b/src/components/DataBar/DataBar.stories.tsx
@@ -130,5 +130,8 @@ export const InteractiveExample: StoryObj<Args> = {
     snapshot: {
       skip: true,
     },
+    axe: {
+      disabledRules: ['label'],
+    },
   },
 };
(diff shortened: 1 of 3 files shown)


jinlee93:
Not sure where the [visual diffs](https://www.chromatic.com/build?appId=61313967cde49b003ae2a860&number=1403) for the tokens are coming from

src/components/DataBar/DataBar.stories.tsx
```diff
@@ -130,5 +130,8 @@ export const InteractiveExample: StoryObj<Args> = {
     snapshot: {
       skip: true,
     },
+    axe: {
+      disabledRules: ['label'],
+    },
```
dierat, line 135: Can you explain this change?

jinlee93: The axe error associates with the textinput fields used to identify how much to change the data bar by (lines 103 and 115 in this file). I will change so that the associated text is a label and link them rather than disabling the rule. Thanks for pointing this out!

dierat (approved):
I'm not familiar with the "grid" role, but if the `HorizontalBar` is using it, that's probably right.

PR #1042, opened 2022-05-05
build: remove descending-specificty rule and allow nesting depth of 2

anniehu4:
### Summary:

- remove [no-descending-specificity](https://stylelint.io/user-guide/rules/list/no-descending-specificity/) rule, as it's been more confusing than valuable
- increase `max-nesting-depth` allowed to 2 -- [we agreed awhile ago](https://czi-edu.slack.com/archives/C0310HJ9WEN/p1648490911423979?thread_ts=1648488617.379909&cid=C0310HJ9WEN) that this was fine given other code guidelines, but never got around to it

### Test Plan:
build passes

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,19 +1,15 @@
 {
-  "extends": [
-    "stylelint-config-recommended"
-  ],
+  "extends": ["stylelint-config-recommended"],
   "rules": {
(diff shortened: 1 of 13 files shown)


jinlee93 (approved):
Lgtm

PR #1043, opened 2022-05-06
feat(numbericon): add number icon

jinlee93:
### Summary:
- add number icon
  - text number enclosed in a circle
- size variants: default 'lg' and 'sm'
  - Louie advised no 'md' variant, should default to 'lg'
- color variants: default 'base' and 'success'
- uses `span` but set `role='img'` and requires `aria-label` or `aria-labelledby`
  - No use case where a number icon would be purely decorative?

### Test Plan:
- snap tests and unit test written
- use in HorizontalStepper creates no visual diff

base-lg:
base-lg
base-sm:
base-sm
success-lg:
success-lg
success-sm:
success-sm
double-digits:
image

### DIFF
--- a/src/components/NumberIcon/NumberIcon.stories.tsx
+++ b/src/components/NumberIcon/NumberIcon.stories.tsx
@@ -1,0 +1,84 @@
+import { StoryObj } from '@storybook/react';
+import React from 'react';
+
+import { NumberIcon } from './NumberIcon';
+
+export default {
+  title: 'Atoms/Icons/NumberIcon',
+  component: NumberIcon,
+  args: {
+    'aria-label': '1',
+    number: '1',
+  },
+};
+
+type Args = React.ComponentProps<typeof NumberIcon>;
+
+export const Default: StoryObj<Args> = {};
+
+export const Small: StoryObj<Args> = {
+  args: {
+    size: 'sm',
+  },
+};
+
+export const SuccessLarge: StoryObj<Args> = {
+  args: {
+    size: 'lg',
+    variant: 'success',
+  },
+};
+
+export const SuccessSmall: StoryObj<Args> = {
+  args: {
+    size: 'sm',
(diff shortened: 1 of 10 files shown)


src/components/NumberIcon/NumberIcon.tsx
```diff
@@ -0,0 +1,82 @@
...
+  /**
+   * 1) Set as 'span' since icon use is more inline than block, but no effect since display is 'flex'.
+   */
+  return (
+    <Text
+      as="span" /* 1 */
+      className={componentClassName}
+      role="img"
+      size="sm"
+      variant={variant === 'base' ? 'base' : 'white'}
```
jinlee93, line 76: If `success` variant, want text to be `white`

src/components/NumberIcon/NumberIcon.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+    },
+    'aria-labelledby': {
+      table: {
+        disable: true,
+      },
+    },
+  },
+  render: (args) => (
+    <div>
+      <NumberIcon {...args} aria-label="0" number={0} />
```
dierat, line 63: The `aria-label` values in this story make my wonder if we should just have the number given as the default value.

jinlee93: I kept it required without default since I wanted meaningful labels and a simple `1` wouldn't carry meaning?

dierat: Then I think the examples in storybook and in the component docstring should demonstrate what kind of label we're looking for.

jinlee93: Got it, will say `Step 1` and so forth since I imagine that to be the general use case

dierat (approved):
Nice!

PR #1044, opened 2022-05-06
Enable TS strict mode in `next`

anniehu4:
### Summary:
[ch187478]

Turn on TS `strict: true`  for next branch. pre-req for merging into `main`

I also did a scan through existing `any` types and fixed any easy ones, left TODO comments for others

### Test Plan:
build passes, only expected changes in Chromatic

### DIFF
--- a/.storybook/components/Docs/Documentation.tsx
+++ b/.storybook/components/Docs/Documentation.tsx
@@ -1,4 +1,5 @@
 import React, { ReactNode, useEffect } from 'react';
+// @ts-expect-error prism.js must be in JS
 import Prism from './prism.js';
 import { LayoutContainer } from '../../../src/components/LayoutContainer/LayoutContainer';
 import { LayoutLinelengthContainer } from '../../../src/components/LayoutLinelengthContainer/LayoutLinelengthContainer';
--- a/src/components/Banner/Banner.stories.tsx
+++ b/src/components/Banner/Banner.stories.tsx
@@ -1,5 +1,5 @@
 import type { StoryObj } from '@storybook/react';
-import React, { ReactNode } from 'react';
+import React from 'react';
 import { Banner, Variant } from './Banner';
 import Button from '../Button';
 import Heading from '../Heading';
@@ -13,11 +13,7 @@ export default {
   },
 };
 
-type Args = React.ComponentProps<typeof Banner> & {
-  title: string;
-  description: string;
-  action: ReactNode;
-};
+type Args = React.ComponentProps<typeof Banner>;
 
 const getAction = (status?: Variant) => (
   <Button status={status} variant="secondary">
@@ -31,7 +27,7 @@ const getDescription = (status?: Variant) => (
     curriculum. To see the updates,{' '}
     <Button
       href="/"
-      onClick={(event) => event.preventDefault()}
+      onClick={(event: any) => event.preventDefault()}
       status={status}
       variant="link"
     >
@@ -190,10 +186,10 @@ export const DismissableBelowContent: StoryObj<Args> = {
     <>
       <Heading size="h1">Page Title</Heading>
       <Banner
+        {...args}
         description={getDescription()}
         onDismiss={dismissMethod}
         titleAs="h2"
-        {...args}
       />
     </>
   ),
--- a/src/components/Checkbox/Checkbox.stories.tsx
+++ b/src/components/Checkbox/Checkbox.stories.tsx
@@ -135,12 +135,14 @@ export const LongLabels = {
   },
 };
 
+type LabelArgs = React.ComponentProps<typeof CheckboxLabel>;
+
 // For visual regression testing
 export const LabelsOnly = {
   args: {
     htmlFor: 'id',
   },
-  render: (args) => (
+  render: (args: LabelArgs) => (
     <>
       <CheckboxLabel {...args} text="Medium label" />
       <br />
--- a/src/components/DataBar/DataBar.stories.tsx
+++ b/src/components/DataBar/DataBar.stories.tsx
@@ -1,4 +1,4 @@
-import { StoryObj, Meta } from '@storybook/react';
+import { StoryObj, Story } from '@storybook/react';
 import React from 'react';
 
 import { DataBar } from './DataBar';
@@ -18,7 +18,7 @@ export default {
     ],
   },
   decorators: [
-    (Story) => (
+    (Story: Story) => (
       <div
         style={{
           margin: '0.25rem',
@@ -28,7 +28,7 @@ export default {
       </div>
     ),
   ],
-} as Meta<Args>;
+};
 
 type Args = React.ComponentProps<typeof DataBar>;
 
@@ -68,9 +68,6 @@ export const NoSegments: StoryObj<Args> = {
 
 const Interactive = () => {
   const [max, setMax] = React.useState(100);
-  const handleChange = (e) => {
-    setMax(Number(e?.target?.value) || 100);
-  };
   const [segments, setSegments] = React.useState([
     { value: 90, text: 'Segment value 90' },
     { value: 1, text: 'Segment value 1' },
@@ -78,9 +75,6 @@ const Interactive = () => {
   ]);
 
   const [segmentValue, setSegmentValue] = React.useState(1);
-  const handleSegmentValueChange = (e: React.ChangeEvent<HTMLInputElement>) => {
-    setSegmentValue(Number(e?.target?.value));
-  };
 
   const onPush = () => {
     segments.push({
@@ -102,7 +96,9 @@ const Interactive = () => {
       <label htmlFor="segment-value-input">Value add: </label>
       <input
         id="segment-value-input"
-        onChange={handleSegmentValueChange}
+        onChange={(e) => {
+          setSegmentValue(Number(e?.target?.value));
+        }}
         value={segmentValue}
       ></input>
       <br />
@@ -116,7 +112,13 @@ const Interactive = () => {
       </ButtonGroup>
       <br />
       <label htmlFor="max-value-input">Max: </label>
-      <input id="max-value-input" onChange={handleChange} value={max}></input>
+      <input
+        id="max-value-input"
+        onChange={(e) => {
+          setMax(Number(e?.target?.value) || 100);
+        }}
+        value={max}
+      ></input>
     </div>
   );
 };
--- a/src/components/DataBar/DataBar.stories.tsx
+++ b/src/components/DataBar/DataBar.stories.tsx
@@ -1,4 +1,4 @@
-import { StoryObj, Story } from '@storybook/react';
+import { StoryObj, Meta } from '@storybook/react';
 import React from 'react';
 
 import { DataBar } from './DataBar';
@@ -18,7 +18,7 @@ export default {
     ],
   },
   decorators: [
-    (Story: Story) => (
+    (Story) => (
       <div
         style={{
           margin: '0.25rem',
@@ -28,7 +28,7 @@ export default {
       </div>
     ),
   ],
-};
+} as Meta<Args>;
 
 type Args = React.ComponentProps<typeof DataBar>;
 
--- a/src/components/Heading/Heading.stories.tsx
+++ b/src/components/Heading/Heading.stories.tsx
@@ -58,7 +58,7 @@ export const Heading6: StoryObj<Args> = {
 export const Variants: StoryObj<Args> = {
   ...Heading1,
   render: (args) => {
-    const headings = [];
+    const headings: Array<JSX.Element> = [];
     VARIANTS.forEach((variant) => {
       if (variant !== 'info')
         headings.push(
--- a/src/components/HorizontalStepper/HorizontalStepper.stories.tsx
+++ b/src/components/HorizontalStepper/HorizontalStepper.stories.tsx
@@ -1,4 +1,4 @@
-import { StoryObj, Story } from '@storybook/react';
+import { StoryObj, Meta } from '@storybook/react';
 import React from 'react';
 
 import { HorizontalStepper } from './HorizontalStepper';
@@ -15,7 +15,7 @@ export default {
     steps: ['Step 1', 'Step 2', 'Step 3', 'Step 4', 'Step 5'],
   },
   decorators: [
-    (Story: Story) => (
+    (Story) => (
       <div
         style={{
           margin: '1rem', // Pushes contents away from storybook borders.
@@ -25,7 +25,7 @@ export default {
       </div>
     ),
   ],
-};
+} as Meta<Args>;
 
 type Args = React.ComponentProps<typeof HorizontalStepper>;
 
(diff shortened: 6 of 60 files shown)


src/components/AccordionPanel/AccordionPanel.tsx
```diff
@@ -16,7 +16,7 @@ export interface Props {
   /**
    * Thinks the aria-control attribute in the accordion panel header to the id in the accordion panel body
    */
-  ariaControls?: any;
+  ariaControls?: string;
```
anniehu4, line 19: updated these aria id-related across the board to `string`

src/components/Banner/Banner.stories.tsx
```diff
@@ -31,7 +31,7 @@ const getDescription = (status?: Variant) => (
     curriculum. To see the updates,{' '}
     <Button
       href="/"
-      onClick={(event) => event.preventDefault()}
+      onClick={(event: any) => event.preventDefault()}
```
anniehu4, line 34: choosing to mostly ignore `any` types in stories

src/components/Feature/Feature.stories.tsx
```diff
@@ -27,5 +27,4 @@ Default.args = {
   imgSrc: 'https://via.placeholder.com/450x500.png',
   imgAlt: 'Placeholder Text',
   title: 'Feature Title',
-  text: 'This is a button',
```
anniehu4, line 30: was unused

src/components/Fieldset/Fieldset.stories.tsx
```diff
@@ -45,12 +45,22 @@ export const FieldsetLegends: StoryObj<FieldsetLegendProps> = {
   ),
 };
 
-export const CheckboxField = CheckboxFieldStory;
-CheckboxField.args.fieldNote =
-  'Checkbox fieldset usage. See CheckboxField Stories for more examples.';
-CheckboxField.parameters = { chromatic: { disableSnapshot: true } }; // Disabled since already being snapped in original story.
+export const CheckboxField = {
```
anniehu4, line 48: decided to convert this to CSF v3.0 while here (slightly easier to read) -- so also had to change the `CheckboxField` and `RadioField` stories to CSF v3

src/components/FileUploadField/FileUploadField.tsx
```diff
@@ -11,6 +11,13 @@ import TextInput from '../TextInput';
 import TextList from '../TextList';
 import TextListItem from '../TextListItem';
 
+// Remove native properties from TS interface (https://github.com/microsoft/TypeScript/blob/c388222b29228a572b4e3c9ca907863d0b1945c6/lib/lib.dom.d.ts#L2537)
+// that don't need to be passed or saved
+type FileInfo = Omit<
```
anniehu4, line 16: i'm not 100% sure this is the right thing to do but we're not actively working on this component yet anyways 😅

src/components/NotificationList/NotificationList.stories.tsx
```diff
@@ -1,36 +1,41 @@
...
+        source="Unit 3: Measuring Circles - Identifying Circumference and Diameter"
+        title="Math Teacher gave you feedback"
+      />
+    </NotificationList>
+  ),
+};
 
-export const MarkedAsRead = Template.bind({});
-MarkedAsRead.args = {
-  markedAsRead: true,
```
anniehu4, line 35: this is only a prop on `NotificationListItem`, so the story wasn't working correctly before (reflected in updated snapshot)

tsconfig.json
```diff
@@ -1,17 +1,9 @@
 {
-  "include": [
-    "src",
-    ".storybook/**/*",
-  ],
-  "exclude": [
-    "node_modules",
-    "typings"
-  ],
-  "typeRoots": ["./types", "./node_modules/@types"],
```
anniehu4, line 10: other small updates to make this match `main`

src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -135,12 +135,14 @@ export const LongLabels = {
...
 
+type LabelArgs = React.ComponentProps<typeof CheckboxLabel>;
+
 // For visual regression testing
 export const LabelsOnly = {
   args: {
     htmlFor: 'id',
   },
-  render: (args) => (
+  render: (args: LabelArgs) => (
```
jinlee93, line 145: Oh I assumed that `args` would be automatically inferred here. Did it error?

src/components/Banner/Banner.stories.tsx
jinlee93: Sounds good. For this case I'm wondering why the `onClick` isn't getting the typing from `ClickableStyle` (event: MouseEventHandler)

src/components/DataBar/DataBar.stories.tsx
```diff
@@ -28,7 +28,7 @@ export default {
       </div>
     ),
   ],
-} as Meta<Args>;
```
jinlee93, line 31: Interesting that there's a type error if using this over `(Story: Story)` in line 21, but doesn't seem to type error for `Checkbox` or `Tooltip`

jinlee93 (approved):
LGTM, just had a couple of questions for my own curiosity

src/components/Banner/Banner.stories.tsx
ahuth: Agreed, the type should definitely be coming from Button.

Looking at Button component (and its stories), I think we're actually getting `any` for all of it's props.

Not exactly sure why, but it looks like at least `ClickableStyleProps` is resulting in `any`. But something's going on there for sure 🤔

src/components/Banner/Banner.stories.tsx
```diff
@@ -49,6 +49,7 @@ export const Brand: StoryObj<Args> = {
   render: ({ variant, ...other }) => {
     return (
       <Banner
+        // @ts-expect-error description is expected to be overridden by story args
```
ahuth, line 52: If we're requiring `description` to be present (which the `Args` type indicates), I wouldn't expect to include it here at all (since stories are presumably always supplying there own).

Or if stories don't necessarily supply their own, should `description` be an optional type on `Args`?

src/components/CheckboxFieldItem/CheckboxFieldItem.tsx
```diff
@@ -79,8 +85,8 @@ export const CheckboxFieldItem = ({
   /**
    * Check for previous props/states
    */
-  function usePrevious(checkboxChecked: any) {
-    const ref = useRef();
+  function usePrevious(checkboxChecked: boolean) {
+    const ref = useRef() as MutableRefObject<boolean>;
```
ahuth, line 89:
```suggestion
    const ref = useRef<boolean>();
```

Where possible, it's better to use a type param, to prevent mistakes/shenanigans like

```ts
const ref = useRef('hi') as MutableRefObject<boolean>; // TS is fine with this

const ref = useRef<boolean>('hi'); // ERROR! TS saves us from ourselves, once again
```

src/components/Counter/Counter.tsx
```diff
@@ -192,7 +193,7 @@ export const Counter = ({
         <Button
           className={styles['counter__btn']}
           disabled={disabled || (max !== undefined && count === max)}
-          onClick={(e) => onIncrease(e)}
+          onClick={onIncrease}
```
ahuth, line 196: 🤘

src/components/DataBar/DataBar.stories.tsx
ahuth: What type error are we seeing with this? Locally I don't have any type errors with typing this whole thing as `Meta<Args>`.

src/components/DataBar/DataBar.stories.tsx
```diff
@@ -68,7 +68,7 @@ export const NoSegments: StoryObj<Args> = {
 
 const Interactive = () => {
   const [max, setMax] = React.useState(100);
-  const handleChange = (e) => {
+  const handleChange = (e: any) => {
```
ahuth, line 71: Nit - I'd put this directly on the element

```tsx
<input
  onChange={(e) => { ... }}
/>
```

and then you get the types for free, and without writing any types. Probably better than using `any`, in any case (which sort of defeats the purpose)

src/components/Heading/Heading.stories.tsx
```diff
@@ -58,7 +58,8 @@ export const Heading6: StoryObj<Args> = {
 export const Variants: StoryObj<Args> = {
   ...Heading1,
   render: (args) => {
-    const headings = [];
+    // `typeof Heading` doesn't work, but `any` typing is fine here
+    const headings: Array<any> = [];
```
ahuth, line 62: I'm guessing `typeof Heading` gets us a Component type, not a rendered element/JSX type.

Something like `const headings: Array<JSX.Element> = []` should work.

Can also see that by hovering over `headings.push` and seeing what type TS thinks is getting pushed onto the array.

src/components/HorizontalStepper/HorizontalStepper.stories.tsx
```diff
@@ -25,7 +25,7 @@ export default {
       </div>
     ),
   ],
-} as Meta<Args>;
```
ahuth, line 28: Did you see an issue with this? Not seeing anything locally

src/components/ListDetail/ListDetail.tsx
```diff
@@ -287,7 +296,8 @@ export const ListDetail = ({
           })}
           role="tablist"
         >
-          {listDetailItems().map((tab: any, i: any) => {
+          {/* TODO: improve `any` type */}
+          {listDetailItems().map((tab: any, i: number) => {
```
ahuth, line 300: Dang, must be that `listDetailItems` returns `any`

tsconfig.json
```diff
@@ -23,6 +15,7 @@
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "sourceMap": true,
+    "strict": true,
```
ahuth, line 18: 🤘

src/hooks/useMergedRefs.ts
```diff
@@ -26,7 +26,7 @@ export function assignRef<RefValueType = any>(
   ref: AssignableRef<RefValueType> | null | undefined,
   value: any,
 ) {
-  if (ref === null) return;
+  if (ref === null || ref === undefined) return;
```
ahuth, line 29:
```suggestion
  if (ref == null) return;
```

Shortcut that TS is (probably) aware of

.storybook/components/Docs/Documentation.tsx
```diff
@@ -1,4 +1,6 @@
 import React, { ReactNode, useEffect } from 'react';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore prism.js must be in JS
```
ahuth, line 3: @jeremiah-clothier has me convinced that we should always use `@ts-expect-error`. That way we never end up with unnecessary ts comments (since it will error if there is **_no_** error).

It's also nice because for some reason the lint rule doesn't complain about ts-expect-error, as long as there's a comment following is (e.g. the `prism.js must be in JS` here).

src/components/CheckboxField/CheckboxField.stories.tsx
```diff
@@ -8,27 +8,7 @@ export default {
...
-    <CheckboxField.Item
-      name="checkbox-example"
-      text="Checkbox 3"
-      value="checkbox3"
-    />
-  </CheckboxField>
-);
+};
 
 const DisabledTemplate: Story<Props> = (args) => (
```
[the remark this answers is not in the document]
jinlee93: Think all the ones brought over from BM were in CSF v2 so might be a little out of scope for this pr?

src/components/DataBar/DataBar.stories.tsx
anniehu4: +1 to Jin, it wasn't a type error but that snapshot issue

src/components/CheckboxFieldItem/CheckboxFieldItem.tsx
anniehu4: ah TIL, thanks. I was following the [guidelines BM added](https://github.com/chanzuckerberg/edu-design-system/blob/next/docs/CODE_GUIDELINES.md#useref) 😅  so will update those as well

src/components/FileUploadField/FileUploadField.tsx
```diff
@@ -20,7 +27,7 @@ export interface Props {
   /**
    * Aria-describedby id string
    */
-  ariaDescribedBy?: any;
+  ariaDescribedBy?: string;
```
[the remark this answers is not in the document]
anniehu4: this is old code from BM 🥲  we can update when these components are actually worked on

.storybook/components/Docs/Documentation.tsx
anniehu4: sgtm!

src/components/CheckboxFieldItem/CheckboxFieldItem.tsx
ahuth: Actually, technically it should be

```ts
const ref = useRef<boolean | undefiend>();
```

since we're using an initial value of `undefined`.

OR, use `false` as the initial value

```suggestion
    const ref = useRef(false);
```

and then there's no need to even specify the type (since it will be inferred)/

src/components/DataBar/DataBar.stories.tsx
ahuth: Alright, installing [@chanzuckerberg/story-utils@^3.0.5](https://github.com/chanzuckerberg/frontend-libs/blob/c6f6de297ddf12fa8dfd4480265a23b472925159/packages/story-utils/CHANGELOG.md#305-2022-05-09) resolves this

src/components/Banner/Banner.stories.tsx
anniehu4: ohh I didn't even notice we were typing it, will make optional

src/components/CheckboxField/CheckboxField.stories.tsx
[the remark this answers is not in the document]
anniehu4: I hit my timebox limit here 🤓  but I can finish off just those 2 files

src/components/Heading/Heading.stories.tsx
anniehu4: sweet, thanks!

tsconfig.json
```diff
@@ -1,17 +1,9 @@
...
-    "src",
-    ".storybook/**/*",
-  ],
-  "exclude": [
-    "node_modules",
-    "typings"
-  ],
-  "typeRoots": ["./types", "./node_modules/@types"],
+  "include": ["src", ".storybook/**/*"],
+  "exclude": ["node_modules", "lib"],
```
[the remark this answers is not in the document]
anniehu4: oh great point -- will update in `main` as well

src/components/DataBar/DataBar.stories.tsx
anniehu4: thanks!!

src/components/Banner/Banner.stories.tsx
anniehu4: ohhhh shoot excellent catch -- we were typing Button with `ClicakbleStyleProps<React.Element>` 😮

anniehu4:
gonna wait to merge until https://github.com/chanzuckerberg/edu-design-system/pull/1047 lands since I don't wanna cause merge issues with that one

PR #1046, opened 2022-05-09
chore(checkbox): use checkmark icon from icon set

jinlee93:
### Summary:
- uses the checkmark icon from the icon set
  - was originally reverted in https://github.com/chanzuckerberg/edu-design-system/pull/1008/commits/838aa68206434a5e7925cc526dbbee6b2216b014 for confirm
  - confirmed with Sean
### Test Plan:
- unit tests pass
- visual regression only for the checkmark

[sc-194626]
[sc-196703]


dierat (approved):
Nice! I prefer this icon to the custom path anyway.

PR #1048, opened 2022-05-09
Ifrost/1945970 cad page scaffolding

ifrost1:
### Summary:
- Scaffolded out CAD page
- Adjusted `Layout` pattern to accommodate wider sidebar
- Added `squared` and `flush` variants of the `Panel` component
- Created dynamic `OverflowList` component that shows and hides shadows to help with accessibility by calculating the width of the items inside. Could be reused with other elements.
- Made page header more dynamic with right slot that is more responsive.

### Test Plan:

`yarn lint` runs without errors
`yarn test` runs without errors
`yarn types` runs without errors.


src/components/PageHeader/PageHeader.tsx
```diff
@@ -57,14 +63,18 @@ export const PageHeader = ({
     <header className={componentClassName} {...other}>
       <div className={styles['page-header__left']}>
```
jinlee93, line 64: Looking at https://www.figma.com/file/0jYcObnngcbsiBOsOPUmtR/Implementation-H1-'22?node-id=2457%3A118577 and https://www.figma.com/file/yLYL19XZI0SWhbQJYuKVso/EDS-Pilot%3ACAD?node-id=1190%3A17197 I think the left and right sections are supposed to take up 50% width each for desktop?

jinlee93: And the button group would be aligned to the right

ifrost1: Let's get all of the content (like your stepper) in there first and then we can adjust if needed. One of those things that is hard to scope out without everything in there. Totally hear ya that we may need a 50/50 approach there. But let's get something in there and then we can adjust if that's alright

ifrost1: @jinlee93  Let's get all of the content (like your stepper) in there first and then we can adjust if needed. One of those things that is hard to scope out without everything in there. Totally hear ya that we may need a 50/50 approach there. But let's get something in there and then we can adjust if that's alright

jinlee93: That sounds good to me

PR #1049, opened 2022-05-10
feat(cadstep1): build page cad step 1

jinlee93:
### Summary:
- builds out cad step 1 according to [mock](https://www.figma.com/file/0jYcObnngcbsiBOsOPUmtR/Implementation-H1-'22?node-id=2457%3A118574)
- adds some responsiveness to `HorizontalStepper`
  - specifics will be fleshed out by designer 
- adds 50-50 variant to layout
- re-enabled axe testing for layout
### Test Plan:
- axe tests pass
- no visual regression for layout/horizontalstepper
[sc-194996]
[sc-195978]

image
image


jinlee93:
Branched from [ifrost/1945970-CAD-page-scaffolding](https://github.com/chanzuckerberg/edu-design-system/tree/ifrost/1945970-CAD-page-scaffolding) for some dependencies on `Layout` and `Panel`. Will probably point to `next` once that has been merged.

Also rebased off next after branching off Ian's branch, which would explain the large unrelated diffs

.storybook/pages/CadStep1/CadStep1.tsx
```diff
@@ -0,0 +1,124 @@
...
+                aria-describedby="minutes-per-class-description"
+                id="minutes-per-class-input"
+                placeholder="Enter"
+                type="number"
+              />
+              <Text as="span" id="minutes-per-class-description">
+                times per week
+              </Text>
+            </div>
+          </div>
```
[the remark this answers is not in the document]
jinlee93: Advise if these should be fpo blocks instead

.storybook/pages/CadStep1/CadStep1.tsx
```diff
@@ -0,0 +1,124 @@
...
+            >
+              How many times per week do you have class?
+            </label>
+            <div className={styles['right-section__input-container']}>
+              <input
+                aria-describedby="class-per-week-description"
+                id="class-per-week-input"
+                placeholder="Enter"
+                type="number"
+              />
```
dierat, line 95: Nice markup 👍

dierat (approved):
friggin' sweet yo

.storybook/pages/CadStep1/CadStep1.module.css
```diff
@@ -0,0 +1,70 @@
+/*------------------------------------*\
+    # CAD Step 1
+\*------------------------------------*/
+
+/**
+ * Page Layout
```
ifrost1, line 6: I think something overall that we try to do is limit any sort of page-specific CSS for the most part so that users of the system don't have to think much about that. One thing that I've been curious about from a utility class standpoint is not making those `Spacing.module.css` so that users could apply utility classes to space items out if need be. That's generally how we tend to build out pages is with a lot of margin bottom utilities to make it easier on consuming developers. From a design standpoint, there is usually a convention if there needs to be responsive behavior in between items (e.g. we've heard this called `major-section-break`), we can create a utility class around that so that users building out pages can use that between those sections to get `2rem` of spacing between sections on small screens and then `4rem` or spacing on large screens. I'd be curious as to what @dierat @anniehu4  @csnizik think of this or if we should actually have a discussion about this. Thanks!

.storybook/pages/CadStep1/CadStep1.module.css
```diff
@@ -0,0 +1,70 @@
...
+  padding: var(--eds-size-2);
+  @media all and (min-width: $eds-bp-md) {
+    padding: var(--eds-size-10);
+  }
+}
+
+/**
+ * Page Layout Left Section
+ */
+.cad-step-1__left-section {
```
ifrost1, line 18: I feel like we should either wrap this `Layout` in a `LayoutContainer` and align it with the content above or create a `variant="narrow"` version of this. Let's align with design on this.

.storybook/pages/CadStep1/CadStep1.module.css
```diff
@@ -0,0 +1,70 @@
...
+ */
+.left-section__heading {
+  margin-bottom: var(--eds-size-6);
+}
+
+/**
+ * Page Layout Left Section Texts
+ */
+.left-section__text {
+  font-weight: var(--eds-font-weight-light);
```
ifrost1, line 35: Let's remove this and flag for conversation. If this needs to be represented as a typography preset, then we need to apply that

.storybook/pages/CadStep1/CadStep1.tsx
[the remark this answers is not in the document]
ifrost1: Let's use our `TextField` component for this. Within the `text-field__body` we may want to add a prop for a slot `inputAfter` to allow for text like this to be added in there. There likely needs to be a discussion around form field design in general (unless you all have already normalized that).

src/components/HorizontalStep/HorizontalStep.module.css
```diff
@@ -4,11 +4,14 @@
 
 /**
  * Horizontal Step displays the step text and associated decorative icon.
+ * 1) Required to prevent overflowing outside container.
  */
 .horizontal-step {
```
ifrost1, line 9: Another general comment: Curious how this works when there are more than a few steps and how that breaks down. I think what you have right now works fine, but just thinking about future states if needed.

src/components/PageHeader/PageHeader.module.css
```diff
@@ -37,6 +37,7 @@
   flex: 1; /* 1 */
 
   @media all and (min-width: $eds-bp-md) {
+    max-width: 50%;
```
ifrost1, line 40: I see what you did here 😉 .

ifrost1 (changes requested):
This is a great start with this page. Some of the main things that require some discussion/refinement:
- Discussing utility classes and whether those should be pulled out of `module.css` and used as regular `className="u-margin-bottom-lg"`. 
- Discussing with designers if they want to align the content in the page header panel and the main content below that or if that should be a variant of `LayoutContainer`. Also discussing `Stepper` in the future to make sure it is future friendly.
- Pulling in `TextField` to accommodate the needs of those inputs.
- Figuring out an easier way to space things out properly rather than adding page level CSS.

.storybook/pages/CadStep1/CadStep1.module.css
anniehu4: yep, I totally agree with pulling utilities out of CSS modules! I think these got mass converted all at the beginning 🙂

.storybook/pages/CadStep1/CadStep1.tsx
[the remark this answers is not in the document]
jinlee93: There is already an `InputField` component in a candidates folder we have for use in the platform.
Will go forward with using `TextField` for this with it being replaceable in this page in mind

.storybook/pages/CadStep1/CadStep1.module.css
jinlee93: Sounds good to me, this was mostly done to match the mocks directly, but consuming product team will probably change as needed if we don't represent as a typography preset.

.storybook/pages/CadStep1/CadStep1.module.css
jinlee93: I agree with removing the page specific styling, and with it gone, the left/right section styling should also be gone. I will sync with designers on dividing the header to match the body.

src/components/HorizontalStep/HorizontalStep.module.css
jinlee93: The text starts to truncate with ellipsis and eventually reaches a breakpoint where it is hidden.
The dash dividers is supposed to have a minimum width but have not heard back from design yet.
Here is an example with 5 steps and long text:

https://user-images.githubusercontent.com/86632227/167727792-57031e05-9e74-4cc5-bfcf-0b955ecbf263.mov

.storybook/pages/CadStep1/CadStep1.module.css
jinlee93: Should I convert all the stylesheets in `src/components/Utilties` into regular CSS files and change the consuming files appropriately?

jinlee93:
> This is a great start with this page. Some of the main things that require some discussion/refinement:

> Discussing utility classes and whether those should be pulled out of module.css and used as regular className="u-margin-bottom-lg".
- I can convert `Spacing.module.css` into a regular css file and use that. In cases where the spacing utility classes are not available, should I add them? Can I add `eds-size-12` which doesn't exist yet but is used in `u-margin-bottom-xxl`?

> Discussing with designers if they want to align the content in the page header panel and the main content below that or if that should be a variant of LayoutContainer. Also discussing Stepper in the future to make sure it is future friendly.
- I'll contact design how they want to align the header content and the main content. As for the stepper, I attached a [video above, and a short explanation](https://github.com/chanzuckerberg/edu-design-system/pull/1049#discussion_r869707192), please advise how you feel about that.

> Pulling in TextField to accommodate the needs of those inputs.
- Also explanation [above ](https://github.com/chanzuckerberg/edu-design-system/pull/1049#discussion_r869703708) but in summary, will use `TextField`

> Figuring out an easier way to space things out properly rather than adding page level CSS.
- Can I add utility classes and size tokens as needed?

> Let's add an img tag to put a placeholder image in with the dimensions in the comp to see how that feels responsively
- Sounds good, is there an example use case where a placeholder img is used? 

@ifrost1

.storybook/pages/CadStep1/CadStep1.module.css
jinlee93: And as for responsive spacing, it would be unideal to create utility classes for various permutations of spacings, how is that usually decided?

src/upcoming-components/TextField/TextField.module.css
```diff
@@ -16,6 +16,17 @@
   }
 }
 
+.text-field--inline {
```
jinlee93, line 19: Cases where the text is long and/or the screen size is tiny cause this to break. Made for the purposes of usage in `CADStep1`. Not sure if creating and owning a pressure tested `TextField` component with design is possible within the scope of CADStep1?

.storybook/pages/CadStep1/CadStep1.tsx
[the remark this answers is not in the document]
ifrost1: I think that all makes sense. I know some systems have an `input` for every type of input in the sun. `TextField` can handle things like `type="number"` and those basics but we separate that out from `TextareaField`, `SelectField`, `RadioField`, `CheckboxField`, `InlineCheckbox` although they have the same structure and accessibility built into those form fields without users of the system having to wire up the `for` attribute of the `label` to the `input` and all of that necessary stuff.

jinlee93:
Synced with design for this page, and lower screen size views have not been developed yet, so thinking we can go forward with desktop view for now and without breaking smaller screen sizes, but finalizing smaller sizes will come later.

.storybook/data/tokens.json
```diff
@@ -80,6 +80,7 @@
   "eds-size-8": "4rem",
   "eds-size-9": "4.5rem",
   "eds-size-10": "5rem",
+  "eds-size-12": "6rem",
```
ifrost1, line 83: This feels weird and I think requires some discussion with design on if we can compromise on spacing between the top `Panel` and the content below.

.storybook/pages/CadStep1/CadStep1.tsx
```diff
@@ -0,0 +1,98 @@
...
+                Next <Icon name="arrow-narrow-right" purpose="decorative" />
+              </Button>
+            </ButtonGroup>
+          }
+          title="Course Planner: History 6"
+        />
+      </LayoutContainer>
+    </Panel>
+    <LayoutContainer>
+      <Layout className="u-padding-xl" variant="50-50">
```
ifrost1, line 51: This doesn't work well responsively. I think we need to talk with design about how to accommodate this in more of a methodical manner. Let's remove this for now and then add a `u-margin-bottom-lg` to the `Panel` for now.

src/upcoming-components/TextField/TextField.module.css
ifrost1: I think this will work perfectly fine once we get rid of all of the padding in the utility class.

ifrost1 (changes requested):
Just some additional comments (not doing anything wrong, just some foundational discussions that should happen):
- Remove padding utility on the `Layout`.
- We try to stick with `margin-bottom` as the go to for spacing items out. Let's do `margin-bottom-lg` for now and then have a conversation with design around this.

.storybook/pages/CadStep1/CadStep1.tsx
jinlee93: Will do

src/upcoming-components/TextField/TextField.module.css
jinlee93: Got rid of the padding and here's how it looks with 320px (which I think is our smallest view width) [Skip to content](https://github.com/chanzuckerberg/edu-design-system/pull/1049#start-of-content)
image

.storybook/data/tokens.json
jinlee93: Sounds good, will remove for now

src/upcoming-components/TextField/TextField.module.css
jinlee93: After our discussion about legacy `InputField` and using the existing designs, removed the inline variant so now it looks like:
Screen Shot 2022-05-12 at 10 51 59 AM
Screen Shot 2022-05-12 at 10 52 12 AM

jinlee93:
Closed for #1073 due to large rebase conflicts

PR #1052, opened 2022-05-10
feat(tokens): add new tier-1 color tokens

anniehu4:
### Summary:
[ch192417]

Brings new tier-1 color tokens into `main`. To ensure this is not a breaking change:
- new tokens use the existing hex values
- old token names are preserved, but now point to the new tokens

After this and the tier-1 typography tokens are merged into `main`, I'll:
1. create a branch that updates both to the new hex values & type sizes (for the future)
2. update the values in `next` as well to ensure a clean merge

### Test Plan:
- Inspect `colors.ts` and see old color names are all there with same values + new ones are added
- No changes to color tokens story


dierat (approved):
Makes sense to me 👍

PR #1053, opened 2022-05-10
feat(tokens): add tier-1 and tier-2 typography presets

anniehu4:
BREAKING CHANGE: size-font-base renamed

### Summary:
[ch192417]

Brings in new type presets & updates Heading component accordingly. 

Next PRs:
- Bring in all other tokens to `main`
- Create branch off `main` that uses new color hex values
- Use old hex values in `next`

### Test Plan:
- No changes in storybook
- Legacy font tokens are still exported


src/components/Heading/Heading.tsx
```diff
@@ -17,6 +17,16 @@ export const VARIANTS = [
 ] as const;
 export type Variant = typeof VARIANTS[number];
 export type HeadingElement = "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
+// For now, "h1"-"h6" sizes point to the old type ramp, while
+// design token sizes point to the new type ramp.
+// These will be brought in sync with the next major release.
```
anniehu4, line 22: Here's a ticket for how we'll bring them in sync: https://app.shortcut.com/czi-edu/story/196525/test-out-new-type-ramp-on-platform (I will make this ticket better before then)

Essentially, we're going to fully switch to the new type ramp and codemod traject headings to maintain their smaller size (size "h1" becomes "h3" in the new ramp, etc). However, as we merge in changes from `next`, I want to make sure it's clear to us within the EDS repo which size we want. I'm thinking EDS will internally use the design token names, but we'll still let app developers use "h1" - "h6" for convenience (this will be well documented)

lib/tokens/css/variables.css
```diff
@@ -135,14 +133,13 @@
   --legacy-color-purple-200: #7686D3;
   --legacy-color-purple-300: #5761BA;
   --legacy-color-purple-400: #32417C;
-  --legacy-size-font-h1: 1.5rem; /* Font size for Heading 1 sized text */
-  --legacy-size-font-h2: 1.125rem; /* Font size for Heading 2 sized text */
-  --legacy-size-font-base: 16px; /* Base font size for design system in px */
```
anniehu4, line 140: this is the same value as `eds-font-size-base`, we don't need both around

anniehu4:
There are some UI Test changes, but to my untrained eye the stories all look the same ... we are now setting `font-family` as part of typography presets, so I think previously storybook wasn't using "Arimo" as the font fanily, and now it is

src/design-tokens/tier-1-definitions/typography.json
```diff
@@ -0,0 +1,95 @@
...
+        "value": "400"
+      },
+      "medium": {
+        "value": "400",
+        "comment": "Arimo weight, should be updated to 500"
+      },
+      "bold": {
+        "value": "700",
+        "comment": "Arimo weight, should be updated to 600"
+      }
```
anniehu4, line 52: The new ramp is designed for Graphik, which is more visually distinct between different font weights. This preserves existing appearances

.storybook/preview-body.html
```diff
@@ -1,9 +1,9 @@
 <style>
   html {
-    font-size: var(--legacy-size-font-base);
+    font-size: var(--eds-font-size-base);
   }
 
   body {
-    font-family: Arimo, sans-serif;
```
dierat, line 7: Was this not working before?

lib/tokens/css/variables.css
```diff
@@ -99,23 +99,21 @@
...
+  --eds-font-size-12: 0.75rem;
+  --eds-font-size-14: 0.875rem;
+  --eds-font-size-16: 1rem;
+  --eds-font-size-18: 1.125rem;
+  --eds-font-size-24: 1.5rem;
+  --eds-font-size-32: 2rem;
+  --eds-font-size-40: 2.5rem;
+  --eds-font-size-base: 16px; /* Base font size for design system in px */
+  --eds-font-weight-light: 400;
+  --eds-font-weight-medium: 400; /* Arimo weight, should be updated to 500 */
```
dierat, line 114: Nit: could you add "when we switch to Graphik" at the end of this comment and the one on the next line?

src/components/Heading/Heading.tsx
dierat: Could you specify in the comment that when you say "design token", you're referring to the "headline-" and "title-" tokens? This has wording has been throwing me off a little bit because I think of the "h[size]" ones as tokens too, and it's not always clear to me what the "design token" is.

anniehu4: ohh yes, thanks!

.storybook/preview-body.html
anniehu4: oh I meant to swap in the new token 🤦‍♀️

PR #1054, opened 2022-05-10
feat: merge all new tokens

anniehu4:
### Summary:
[ch192417]

Adds remaining tokens, in preparation for merging in `next`. Also builds tokens to `src/tokens-dist/` to match `next`

After this, the last steps are:
- Move out CAD components from `next`
- Update token values in `next` to match existing usage (hex values, use explicit heading token names)
- The final big rebase and merge it all in

### Test Plan:
- Nothing should change in Storybook
- Confirm `tokens` folder is successfully built with `yarn build`


.eslintrc
```diff
@@ -67,7 +67,7 @@
       }
     },
     {
-      "files": ["config.js", "*.config.js", "main.js", "plopfile.js"],
+      "files": ["config.js", "*.config.js", "main.js", "plopfile.js", "src/**/*.js"],
```
anniehu4, line 70: specifically to target `tier-1-definitions/breakpoints.js`. we could be more specific, but I kinda figure any `.js` file we're putting in `src/` is gonna be for some special purpose / will want this config

style-dictionary.config.js
```diff
@@ -28,16 +28,14 @@ const EDSStyleDictionary = StyleDictionary.extend({
         },
         {
           format: "json/nested-css-variables",
+          // useful for tailwind configs in consuming apps
```
dierat, line 31: 👍

PR #1057, opened 2022-05-11
chore: add import examples to component docstrings

dierat:
### Summary:
We have a preferred pattern of adding an import example for each component to the component's docstring. Most of the new components don't have that yet. In this PR, I'm adding those comments to components in the `components` directory to make them all consistent, and I also added it to the plop template file.

#### In follow-up PRs, I will:
- utilize storybook badges to add a secondary method for communicating that these components are in beta: https://github.com/chanzuckerberg/edu-design-system/pull/1064
- also make these changes to components in the `upcoming-components` directory (I'm focusing on the components in `components` for right now since those are the ones that will appear in storybook)

### Test Plan:
Verify these comments are appearing in storybook as expected.


plop-templates/Component/Component.tsx.hbs
```diff
@@ -12,6 +12,10 @@ export interface Props {
 /**
  * BETA: This component is still a work in progress and is subject to change.
  *
+ * ```ts
+ * import { {{pascalCase name}} } from "@chanzuckerberg/eds";
```
dierat, line 16: I added a space between the outer brackets because otherwise the brackets didn't show up in the newly created file at all.

plopfile.js
```diff
@@ -1,6 +1,10 @@
+const pascalCase = require('pascal-case');
+
 module.exports = (plop) => {
   plop.setHelper('upperCase', (txt) => txt.toUpperCase());
 
+  plop.setHelper('inBraces', (txt) => `{${pascalCase.pascalCase(txt)}}`);
```
dierat, line 6: This is necessary to get the braces around the component name without having spaces around the name (ex: `{SuperAwesomeComponent}` vs `{ SuperAwesomeComponent}` – the second version doesn't need anything extra but it would be out of sync with how we actually write this line).

package.json
```diff
@@ -68,6 +68,7 @@
     "@tippyjs/react": "^4.2.6",
     "clsx": "^1.1.1",
     "downshift": "^6.1.7",
+    "pascal-case": "^3.1.2",
```
dierat, line 71: I tried to add this in dev dependencies but it complained 😕

plopfile.js
dierat: There's probably an easier way to do this but it's the only way I could think of.

package.json
anniehu4: oo was it a lint error? within `.eslintrc`, in the `"import/no-extraneous-dependencies"` rule specifically, you may need to add `plopfile.js` to the `"devDependencies" array.

that error is making sure we don't accidentally use a dev dependency in something that will be published, so just need to make sure we've told lint all the files that won't ever be published

plopfile.js
anniehu4: oh weird 🤔 could you leave that as a code comment?

package.json
dierat: Awesome, thanks!

dierat:
https://app.shortcut.com/czi-edu/story/192418/prep-pilot-components-for-release

anniehu4 (approved):

PR #1060, opened 2022-05-12
Csnizik/drag drop handles

csnizik:
### Summary:
Adds `useHandles` variant to DragDrop component. When set to true, draggable items can only be grabbed by clicking on the drag handle.

### Test Plan:
New panel is added to Storybook: Organisms/Interactive/Drag and Drop/Drag Using Handles. All other drag/drop functionality should be the same as the default configuration.


src/components/DragDrop/DragDrop.tsx
```diff
@@ -42,15 +46,17 @@ export const DragDrop = ({
...
   /**
-   * If either multipleContainers or unstyledItems is set, apply the corresponding class
+   * If either multipleContainers, unstyledItems or useHandles is set, apply the corresponding class
    */
   const componentClassName = clsx(
     styles['drag-drop'],
     className,
     multipleContainers && styles['drag-drop--multiple'],
     unstyledItems && styles['drag-drop--unstyled'],
+    useHandles && styles['drag-drop--handles'],
```
csnizik, line 59: The `drag-drop--handles` class enables the style variants

src/components/DragDropContainer/DragDropContainer.tsx
```diff
@@ -18,12 +18,21 @@ export interface Props {
    * Prop that will be an array of items
    */
   items: ItemType[];
+  /**
+   * Prop that gets passed to DragDropItem(s)
+   */
+  useHandles: boolean;
```
csnizik, line 24: Prop drilling seemed like a necessary evil here.

PR #1061, opened 2022-05-13
v0.21.0-rc (aka merge all of `next` in for the first time)

anniehu4:
### Summary:

The big merge 😱 

### Test Plan:
- build passes
- verify UI Tests in chromatic
- publish test release for LP: https://github.com/FB-PLP/traject/pull/11008

Manually tested:
- Sanity checked storybook stories, ensured Legacy color & type stories were still there
- Looked through each modified file from `main` to make sure we kept the changes we wanted from `main` 😅


src/components/Toast/Toast.tsx
```diff
@@ -45,21 +45,20 @@ export const Toast = ({
...
     <div className={componentClassName} {...other}>
-      <div className={styles["toast__content"]}>
+      <div className={styles['toast__content']}>
         <Icon
-          color="var(--icon-color)"
-          name={variant === "success" ? "check-circle" : "cancel"}
+          name={variant === 'success' ? 'check-circle' : 'warning'}
           purpose="informative"
-          size="1.25rem"
+          size="1.5rem"
```
dierat, line 58: I wonder why this was changed.. 🤔

src/components/Banner/Banner.tsx
```diff
@@ -1,31 +1,30 @@
...
   /**
    * A button or link that's placed in the banner separately from the main content.
    */
-  action?: React.ReactNode;
+  action?: ReactNode;
   /**
    * CSS class names that can be appended to the component.
    */
-  className?: string;
+  className?: ReactNode;
```
dierat, line 19: `ReactNode` here seems weird. I think this is the only component doing this. I'll change this on a separate PR.

dierat (approved):
LGTM. I skimmed the beginning and then spot-checked the components already on `main`. I mostly just saw stuff that I want to look into on `next` that are outside the scope of this merge. Thanks for handling all this! ❤️

src/components/Toast/Toast.tsx
dierat: Now it's the same size as in `Banner` but they're different colors. I'll just leave it for now because we will probably have to make some other small design changes to these components as part of the core component work.

anniehu4: oh funky... thanks for noticing that 🧐

src/components/Banner/Banner.tsx
anniehu4: good catch! thank you!

PR #967, opened 2022-04-11
chore: rm mismatched prop types

csnizik:
### Summary:
 https://app.shortcut.com/czi-edu/story/190821/styling-prop-cleanup

### Test Plan:
`yarn types` should return no errors or warnings

PR #1062, opened 2022-05-13
chore(banner, hr): fix types and remove comment

dierat:
### Summary:
Fixing a couple of small issues I noticed when reviewing https://github.com/chanzuckerberg/edu-design-system/pull/1061

### Test Plan:
`yarn types` produces no errors.

### DIFF
--- a/src/components/Banner/Banner.module.css
+++ b/src/components/Banner/Banner.module.css
@@ -73,7 +73,6 @@
   position: absolute;
   top: var(--eds-size-half);
   right: var(--eds-size-half);
-  /* TODO: handle hover styles */
   color: inherit; /* 2 */
 }
 
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -16,7 +16,7 @@ export type BannerProps = {
   /**
    * CSS class names that can be appended to the component.
    */
-  className?: ReactNode;
+  className?: string;
   /**
    * The description/body text of the banner
    */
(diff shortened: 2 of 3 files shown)


src/components/Hr/Hr.tsx
```diff
@@ -12,7 +12,7 @@ export interface Props {
    * Size variations:
    * - **lg** yields a thicker horizontal rule
    */
-  size?: 'lg' | 'brand';
```
dierat, line 15: This was fixed in https://github.com/chanzuckerberg/edu-design-system/pull/967 but that PR was long forgotten and buried beneath centuries of commits, so I figured it's just easier to fix it in a new PR.

src/components/Banner/Banner.tsx
```diff
@@ -16,7 +16,7 @@ export type BannerProps = {
   /**
    * CSS class names that can be appended to the component.
    */
-  className?: ReactNode;
```
dierat, line 19: This is the only component using the `ReactNode` type for the `className` prop; all the other components use `string`. I think this is my mistake and I don't remember how it got in there.

src/components/Banner/Banner.module.css
```diff
@@ -73,7 +73,6 @@
   position: absolute;
   top: var(--eds-size-half);
   right: var(--eds-size-half);
-  /* TODO: handle hover styles */
```
dierat, line 76: The hover styles are being handled by the `Button` component now; this was originally using a raw `button` with no hover styling.

ahuth (approved):

PR #1067, opened 2022-05-13
docs: update docstrings and add beta badges to upcoming components directory

dierat:
### Summary:
This is a follow-up to several PRs in which I updated the docstrings and added beta badges to components in the `src/components/` directory. This PR updates the components in `src/upcoming-components/` to match, and I also put the beta badge change in the plop stories template file (I already updated the docstrings in that file but I forgot this change).

Another thing we should do is update the story files to use the updated pattern.. but it's not pressing. Not sure if there's a file for this specifically but I found one closely related: https://app.shortcut.com/czi-edu/story/187754/re-enable-plugin-chanzuckerberg-stories-strict-convert-stories-to-csf-v3

### Test Plan:
Verify the updates docstrings and badges are appearing in storybook.


plop-templates/Component/Component.stories.tsx.hbs
```diff
@@ -7,10 +7,13 @@ import { {{pascalCase name}}, Props } from './{{pascalCase name}}';
 export default {
 	title: 'Example/{{pascalCase name}}',
 	component: {{pascalCase name}},
+	parameters: {
+    badges: [BADGE.BETA],
+  },
```
dierat, line 12: I don't know why the indentation in these files look so weird in github PRs but this file looks fine in VSCode.

dierat: That was it! Thanks! ❤️

PR #1069, opened 2022-05-16
chore(button): update styles to match new designs

dierat:
### Summary:
Please note this only changes the "button" variants – "primary", "secondary", and "icon". The "link" styles will be updated separately.

Changes:
- secondary variant's border in inactive state drops one level
  - the text stays the same so the border is now slightly lighter than the text
  - neutral buttons are exempt (I'm going to follow up with design to verify that this is correct and not an oversight in the design)
- icons inside secondary and icon variants drop 2 levels
  - so now the icon is 500, border is 600, and text is 700
  - neutral buttons are exempt (I'm going to follow up with design to verify that this is correct and not an oversight in the design)
- icon variant's background on hover bumped from 100 to 200
- on all variants, the hover style is decoupled from the focus style
  - so now the only difference between the default and focus state is the focus outline (unless the user is also hovering over the button)
  - icon variant is exempt; we still get the level 200 background in the focus state

### Test Plan:
This is a little hard to test because the hex values in the code are using the old tokens and the figma designs are using the new tokens, so I mostly double-checked the tokens and then looked at storybook to verify the general patterns but didn't check the actual colors as I knew those would be wrong.


src/components/ClickableStyle/ClickableStyle.module.css
```diff
@@ -131,8 +132,11 @@
   background-color: var(--eds-theme-color-button-secondary-neutral-background);
   color: var(--eds-theme-color-button-secondary-neutral-text);
 
-  &:hover,
-  &:focus {
+  svg {
+    color: var(--eds-theme-color-button-icon-neutral-icon);
```
anniehu4, line 136: this naming is a bit strange with the double "icon" 🤔  could it be
```suggestion
    color: var(--eds-theme-color-button-icon-neutral);
```
?

dierat: You got it! I had also forgotten to actually use the hover/focus tokens, so that's in there too now.

PR #1063, opened 2022-05-13
chore(plop): test by creating new component for example purposes

dierat:
### Summary:
This is the output of `npx plop` right now. The newly-created files look fine, but the `src/index.ts` file is a bit, um, odd.

PR #1071, opened 2022-05-17
chore(plop): fix sorting logic

dierat:
### Summary:
The plop generator creates files for a new component, and it also adds the new component to the `index.src` file and sorts the file so the new export is placed correctly in the file. Recently some components were commented out in the `index.src` file, which has thrown off the sorting logic – it wants to sort the commented-out components with the rest of them, which breaks the comment blocks and generally messes up the file. You can see the output of the `plop` logic in this PR: https://github.com/chanzuckerberg/edu-design-system/pull/1063

This PR updates the sort logic to use @csnizik 's code change suggestion, which only sorts the lines that start with "export". The code suggestion is posted in this comment: https://github.com/chanzuckerberg/edu-design-system/pull/1063#issuecomment-1126290554

### Test Plan:
Run `npx plop` and verify the `index/src` file is still intact (it just moves one component, `NumberIcon`, to the right place in the list).

### DIFF
--- a/plopfile.js
+++ b/plopfile.js
@@ -54,12 +54,23 @@ module.exports = (plop) => {
         if (fs.existsSync(indexFile)) {
           // Split the index file into lines.
           const lines = fs.readFileSync(indexFile, 'utf8').split('\n');
+          // Only sort lines that begin with "export {" in order to ignore
+          // comments and exported types.
+          const sortableLines = lines.filter((line) =>
+            /^export {.*/g.test(line),
+          );
+          // Ignore all other lines and turn them back into a string.
+          const ignoreLines = lines
+            .filter((line) => !/^export {.*/g.test(line))
+            .join('\n');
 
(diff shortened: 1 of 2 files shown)


plopfile.js
```diff
@@ -54,12 +54,23 @@ module.exports = (plop) => {
         if (fs.existsSync(indexFile)) {
           // Split the index file into lines.
           const lines = fs.readFileSync(indexFile, 'utf8').split('\n');
+          // Only sort lines that begin with "export {" in order to ignore
+          // comments and exported types.
+          const sortableLines = lines.filter((line) =>
+            /^export {.*/g.test(line),
```
dierat, line 60: The only change I made to the logic is now it only sorts lines that start with `export {` (as opposed to just `export`) to avoid moving the type import lines around.

anniehu4 (approved):

PR #1072, opened 2022-05-17
chore(tooltip): delay chromatic snapshots to see if it helps with flakiness

dierat:
### Summary:
Just what it says. @anniehu4 already turned off the animation for the tooltip stories, but these still keep coming up in chromatic. The chromatic snapshots show the tooltip as having moved ever to slightly, maybe 1 or 2 pixels, left or right. The only thing I know to try is delaying the snapshot in case there's some weirdness around the tooltip is still rendering when the snapshot is being taken. The button triggering the tooltip is not appearing changed; only the tooltip itself.

### Test Plan:
Wait and see if chromatic continues to flag changes in these stories on other PRs after this PR has merged.


anniehu4 (approved):
🥲

PR #1075, opened 2022-05-18
feat(heading): use new heading tokens in old heading size props

dierat:
### Summary:
Our new type ramp is 2 levels higher than the old one. The size prop values that don't use h-num format (`headline-lg`, `title-lg`, etc) are already using the new levels, but the h-num sizes (`h1`, `h2`, etc) are using the old levels because the `traject` codebase is still using those. In this PR, we're using the h-num size values to also use the new ramp. (After this change is published, we'll use a codemod to update the `Heading` usage in `traject` to use lower levels so the text will stay the same size as before.)

comparison of the 2 typeramps

One extra level has been added at the bottom to accommodate the smallest headings in `traject`, and we're marking that and the level above it as deprecated. I named this lowest level `h7` to keep it consistent with the other levels even though there is no `h7` element – it's just for visual size, not element size. `Heading`s with the `h7` size will default to using an `h6` element.

#### Before
typography ramp story in storybook before this change

#### After
typography ramp story in storybook after this change

### Test Plan:
Check out the typography ramp story and verify all the headings have been embiggened.


.storybook/components/DesignTokens/Legacy/LegacyTypography.stories.mdx
```diff
@@ -19,30 +19,42 @@ import styles from './TypographyTable.module.css';
       <td>
         <Heading size="h1">Heading 1</Heading>
       </td>
-      <td>24 / 32</td>
```
dierat, line 22: I'm wondering if we should just remove these legacy token stories as the tokens themselves are changed.

ahuth: Makes sense to me

ahuth (approved):

.storybook/components/DesignTokens/Legacy/LegacyTypography.stories.mdx
anniehu4: yeah that sounds good to me! or maybe we can just show the 2 deprecated legacy sizes?

.storybook/components/DesignTokens/Tier2/TypographyUsage.js
```diff
@@ -6,15 +6,18 @@ import { PRESET_SIZE_MAP } from '../Tier1/TypographyPresets';
...
 export class Tier2TypographyUsage extends Component {
   render() {
-    const renderTypeToken = (size, preset) => {
+    const renderTypeToken = (size, preset, legacySize) => {
       // remove variant suffix
       const normalizedPreset = preset.split('-')[0];
       const { fontSize, lineHeight } = PRESET_SIZE_MAP[normalizedPreset];
 
+      const name =
+        `eds-theme-typography-${size}` + (legacySize ? `/ ${legacySize}` : '');
```
anniehu4, line 15: super nit, adding a space for readability:
```suggestion
        `eds-theme-typography-${size}` + (legacySize ? ` / ${legacySize}` : '');
```

src/components/Heading/Heading.tsx
```diff
@@ -46,6 +47,14 @@ type Props = {
   variant?: Variant;
 } & React.HTMLAttributes<HTMLHeadingElement>;
 
+const getComputedSize = (size: HeadingSize) => {
```
anniehu4, line 50: oh now that I'm reading the code again, this should probably be named `getComputedAs` 🤔 my b

src/design-tokens/tier-2-usage/typography-usage.css
```diff
@@ -20,6 +20,14 @@
   @mixin eds-typography-preset-005-bold $important;
 }
 
+@define-mixin eds-theme-typography-body-sm $important {
+  @mixin eds-typography-preset-006-bold $important;
+}
+
+@define-mixin eds-theme-typography-body-xs $important {
+  @mixin eds-typography-preset-008-bold $important;
+}
```
anniehu4, line 29: aren't there existing `eds-theme-typography-body-text-sm-bold` and `eds-theme-typography-body-text-xs-bold` presets (I think at the bottom of the file)? Could we move those up & make sure they're pointing to the right presets

anniehu4 (changes requested):
req. changes to use existing body bold presets

.storybook/components/DesignTokens/Legacy/LegacyTypography.stories.mdx
dierat: >or maybe we can rename the story title to "Legacy Typography Sizes" and only show the 2 deprecated small sizes?

Oo good idea, especially since I left these two out of the main ramp doc because they're deprecated and I don't want to encourage their usage.

src/design-tokens/tier-2-usage/typography-usage.css
anniehu4: ahh yeah weird, I think that never got updated 🤔  since `body-text-xs` is pointing to 008 now

.storybook/components/DesignTokens/Tier2/TypographyUsage.js
dierat: Updated, thanks!

.storybook/components/DesignTokens/Legacy/LegacyTypography.stories.mdx
```diff
@@ -17,67 +17,13 @@ import styles from './TypographyTable.module.css';
   <tbody>
     <tr>
       <td>
-        <Heading size="h1">Heading 1</Heading>
+        <Heading size="h6">Heading 6</Heading>
       </td>
-      <td>24 / 32</td>
+      <td>14 / 22</td>
```
anniehu4, line 22: nit: think preset 006 is 14 / 24

src/design-tokens/tier-2-usage/typography-usage.css
```diff
@@ -90,27 +90,31 @@
 
 /* Old / legacy heading sizes */
```
dierat, line 91: Here's the mixin updates (adding this comment so I can link to it somewhere else).

anniehu4:
:shipit:

PR #1076, opened 2022-05-19
feat(inlinenotification): add inline notification

jinlee93:
### Summary:
- adds inline notification component
  - mocks:
    - https://www.figma.com/file/Ff7hJoSvXX9hkAmIklnsT2/EDS---UI-Components-(Main)?node-id=1439%3A11245
    - https://www.figma.com/file/Ff7hJoSvXX9hkAmIklnsT2/EDS---UI-Components-(Main)?node-id=997%3A8500
    - https://www.figma.com/file/Ff7hJoSvXX9hkAmIklnsT2/EDS---UI-Components-(Main)?node-id=997%3A8201
    - https://www.figma.com/file/yLYL19XZI0SWhbQJYuKVso/EDS-Pilot%3ACAD?node-id=855%3A14061
- full width variant and inactive mainly used for cad card
- [sc-198526]

### Test Plan:
 - adds snap tests
 - visual regression
 - full width usage in card works

Screen Shot 2022-05-19 at 12 50 42 AM
image
Screen Shot 2022-05-19 at 3 11 53 PM
Screen Shot 2022-05-19 at 3 11 56 PM
Screen Shot 2022-05-19 at 12 51 11 AM
Screen Shot 2022-05-19 at 12 51 15 AM
Screen Shot 2022-05-19 at 3 12 00 PM
Screen Shot 2022-05-19 at 3 12 03 PM


src/components/InlineNotification/InlineNotification.tsx
```diff
@@ -0,0 +1,136 @@
...
+            {' '}
+            Score:
+          </Text>
+        )}
+        {score && (
+          <Text as="span" variant="base" weight="bold">
+            {' '}
+            {score}
+          </Text>
+        )}
```
jinlee93, line 132: Please advise on a better way to add conditional score text

src/components/InlineNotification/InlineNotification.tsx
```diff
@@ -0,0 +1,129 @@
...
+  isFullWidth,
+  isStrong,
+  score,
+  text,
+  variant,
+  ...other
+}: Props) => {
+  if (!isFullWidth && inactive && process.env.NODE_ENV !== 'production') {
+    throw new Error('inactive prop must be used with isFullWidth prop.');
+  }
```
jinlee93, line 89: I believe score will only be used with !isFullWidth and success/yield variants. Should I add that here as well to throw error?

src/components/InlineNotification/InlineNotification.tsx
```diff
@@ -0,0 +1,129 @@
...
+        title={variantToIconAssetsMap[variant].title}
+      />
+      <Text
+        as="span"
+        className={textClassName}
+        variant={variant === 'yield' ? 'base' : variant}
+      >
+        {text}
+        {score && (
+          <Text as="span" className={textClassName} variant="base">
```
jinlee93, line 122: Not sure if extra element is necessary to make score 'base' variant but not sure how otherwise

jinlee93:
For first pass of reviews, please focus on the default/not full width variant since that is subject to change with more designer input and usage in cards

src/components/InlineNotification/InlineNotification.module.css
```diff
@@ -0,0 +1,104 @@
...
+ */
+.inline-notification {
+  display: inline-flex;
+  padding: var(--eds-size-half);
+  border: var(--eds-theme-border-width) solid;
+  border-radius: var(--eds-theme-border-radius);
+
+  background-color: var(--eds-theme-color-background-neutral-default);
+
+  @mixin eds-theme-typography-body-text-lg;
```
anniehu4, line 18: what is this styling for if the notification text has its own styling?

src/components/InlineNotification/InlineNotification.tsx
anniehu4: hmmm I'm not sure we should handle `score` in this component. `InlineNotification` should be pretty general-purpose / can accept anything in its container. If people want to pass in something like "Score", that should either:
- Be a recipe (i.e. they always do it themselves)
- Be a separate, more specific component (i.e. `ScoreNotification`) that can call `InlineNotification` and basically implement the recipe

src/components/InlineNotification/InlineNotification.tsx
```diff
@@ -0,0 +1,129 @@
...
+type Props = {
+  /**
+   * CSS class names that can be appended to the component for styling.
+   */
+  className?: string;
+  /**
+   * Indicates inactive state for the full width variant. Overrides variant prop.
+   * Only to be used with isFullWidth.
+   */
+  inactive?: boolean;
```
anniehu4, line 43: what's the use case for this?

src/components/InlineNotification/InlineNotification.tsx
jinlee93: That makes sense to me, will remove `score` from the scope of the `InlineNotification` and use it as a recipe since the text allows for ReactNodes

src/components/InlineNotification/InlineNotification.module.css
jinlee93: Ack, remnants of a previous commit, will remove

src/components/InlineNotification/InlineNotification.tsx
```diff
@@ -0,0 +1,129 @@
...
+   */
+  isStrong?: boolean;
+  /**
+   * The text contents of the tag, nested inside the component, in addition to the icon.
+   */
+  text: React.ReactNode;
+  /**
+   * The score as a string to be appended after the text.
+   */
+  score?: string;
```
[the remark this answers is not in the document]
jinlee93: Yup will drop `score`

src/components/InlineNotification/InlineNotification.tsx
jinlee93: This is for the `inactive` state of the `fullWidth` variant, most right mock in screenshot: 
image

src/components/InlineNotification/InlineNotification.module.css
```diff
@@ -0,0 +1,104 @@
...
+.inline-notification__icon--inactive {
+  display: none;
+}
+.inline-notification__text--inactive {
+  padding-left: var(--eds-size-3); /* 1 */
+  color: var(--eds-theme-color-text-neutral-subtle);
+}
+
+/**
+ * Color variants
```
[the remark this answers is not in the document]
jinlee93: Great callout, will explore using the messaging mixins, and will likely be used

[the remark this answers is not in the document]
jinlee93: Seems like the mixins have different colors, both borders and backgrounds, than the inline notification mocks. Think upcoming banner mocks align on border color, but not some of the background color. How should I proceed regarding?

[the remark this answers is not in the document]
dierat: Is this component the most mature one of the 3 (banner, toast, and inline)? From the figma file, it kinda looks like banner and toast are still in progress. If that's true, you could update the mixin so they stay in sync for now (and leave a note in the slack channel so they're aware), and we can get into the weeds when the banner and toast styles have hardened.

src/components/InlineNotification/InlineNotification.tsx
dierat: So it's kinda like a disabled state where it's gone extra subtle (no icon and lighter text)? Maybe you could be more explicit in the description for the prop and explain that visual change because this is a pretty unique situation.

src/components/InlineNotification/InlineNotification.tsx
jinlee93: Removing `score` so no longer relevant

src/components/InlineNotification/InlineNotification.module.css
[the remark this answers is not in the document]
jinlee93: That makes sense to me, will update the mixins with the tokens used in `InlineNotifications`

src/components/InlineNotification/InlineNotification.tsx
jinlee93: Sounds good to me, will elucidate in the prop description

src/components/InlineNotification/InlineNotification.module.css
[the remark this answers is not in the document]
jinlee93: Updated the mixins except the `error` mixin (think design wants it that red/pink) and instead added a `revise` mixin for use here

src/util/allIcons.ts
```diff
@@ -118,4 +118,5 @@ export const ALL_ICONS = [
   'view-headline',
   'view-module',
   'warning',
+  'yield',
```
jinlee93, line 121: Called the `error` icon with the dark exclamation mark `yield` to match tokens

jinlee93:
Synced with Louie, seems `fullWidth` variants and `isFullWidth` may likely be removed in favor of yet another Banner variant/component for these kinds of use cases, awaiting design direction for this component
@anniehu4 @dierat

src/util/allIcons.ts
dierat: Which token is that? I was wondering if we should name it something with "error" so it's easier for people to find, like `error-inverted` or something.

src/components/InlineNotification/InlineNotification.module.css
```diff
@@ -0,0 +1,88 @@
...
+}
+.inline-notification--success {
+  @mixin messagingSuccess;
+}
+.inline-notification--warning {
+  @mixin messagingWarning;
+}
+.inline-notification--yield {
+  @mixin messagingRevise;
+}
```
dierat, line 47: I find the "yield" vs "revise" bit here a little confusing. I feel like if the variant is called "yield", that's what the mixin should use even though the color token it's using is called "revise".

dierat: Sorry to get all annoying about these naming details 😓

src/components/InlineNotification/InlineNotification.tsx
```diff
@@ -0,0 +1,125 @@
...
+    name: 'check-circle',
+    title: 'success',
+  },
+  warning: {
+    name: 'warning',
+    title: 'alert',
+  },
+  yield: {
+    name: 'yield',
+    title: 'review',
```
dierat, line 28: Where did this title come from? I'm thinking this copy feels very specific when this component may be used in many different places. The other titles are more describing the level of importance the message has, and I feel like "review" is inconsistent.

src/util/allIcons.ts
jinlee93: I didn't want to confuse `border-utility-error` and `text-utility-error` so in hindsight `revise` might have been better but the variant naming is I think synced with `yield` (hopefully).

Do agree to call it `error-inverted` tho

src/components/InlineNotification/InlineNotification.module.css
jinlee93: It's so confusing will rename to `messagingYield`

src/components/InlineNotification/InlineNotification.tsx
jinlee93: Sounds good, will title it `yield` instead

dierat (approved):
Nice!

PR #1078, opened 2022-05-19
chore(DragDrop): add header and OverflowList/OverflowListItem

csnizik:


src/components/OverflowList/OverflowList.tsx
```diff
@@ -19,6 +19,7 @@ export interface Props {
    * CSS class names that can be appended to the component.
    */
   className?: string;
+  ref?: any;
```
ifrost1, line 22: Do we still need this?

ifrost1 (changes requested):
One last thing. Then let's merge it. Thank you!

src/components/OverflowList/OverflowList.tsx
csnizik: nope. gone.

ifrost1 (approved):

PR #1080, opened 2022-05-20
[do not merge] feat: use Graphik as primary font family

anniehu4:
### Summary:
[ch198079]

Updates primary font family to `Graphik` instead of `Arimo`.

Usage in traject:
- See test publish: https://github.com/FB-PLP/traject/compare/eds/font-graphik
- Essentially, consuming apps only need to import the `fonts.css` file with `@font-face` definitions. [Following BigMedium's recommendation](https://czi-edu.slack.com/archives/C0310HJ9WEN/p1653078194587639?thread_ts=1653077773.236189&cid=C0310HJ9WEN)

### Test Plan:
Storybook is gonna show a loooot of font changes, but there should only be font changes


package.json
```diff
@@ -29,7 +29,8 @@
     "build:styles": "postcss \"src/components/**/*.css\" --dir lib/ --base src/ --verbose",
     "chromatic": "chromatic",
     "commit": "./node_modules/.bin/cz",
-    "copy-to-lib": "yarn copy-icons-to-lib && yarn copy-tokens-to-lib",
+    "copy-to-lib": "yarn copy-icons-to-lib && yarn copy-tokens-to-lib && yarn copy-fonts-to-lib",
+    "copy-fonts-to-lib": "copyfiles -u 3 src/design-tokens/tier-1-definitions/fonts.css src/design-tokens/tier-1-definitions/fonts/**/* lib/tokens",
```
anniehu4, line 33: little bit of finagling so that traject can import from `@chanzuckerberg/eds/lib/tokens/font.css`. It's only imported in like 2 spots though, so we can drop this if we think it's too gross

package.json
```diff
@@ -50,7 +51,7 @@
     "release:patch": "yarn build && standard-version --release-as patch",
     "release:alpha": "yarn build && standard-version --prerelease alpha",
     "start": "yarn build:tokens && yarn storybook",
-    "storybook": "start-storybook -p 9009 -s public",
```
anniehu4, line 53: `-s` stands for `--static-dir` -- I don't believe we're serving any static assets (besides the fonts, which seem to already be handled), and the `public/` folder doesn't even exist

src/design-tokens/tier-1-definitions/fonts.css
```diff
@@ -0,0 +1,17 @@
+@font-face {
+  font-family: 'Graphik';
+  font-weight: 400;
+  src: url('./fonts/Graphik-Regular-Web.woff2') format('woff2');
```
anniehu4, line 4: According to [caniuse](https://caniuse.com/?search=woff2), all our supported browsers should support `'woff2'`. we could add `'woff'` lines if we want to be extra safe

.storybook/main.js
```diff
@@ -33,6 +33,7 @@ module.exports = {
       },
     },
   ],
+  staticDirs: ['../src/design-tokens/tier-1-definitions'],
```
anniehu4, line 36: hmmm I see something in traject that says:
```
/**
 * Although `--static-dir` is marked as deprecated. The The Chromatic CLI
 * currently pulls the staticDir from the build script, but does not support
 * the staticDirs configuration option.
 *
 * We should refrain from using the staticDirs option in this configuration until
 * https://github.com/chromaui/chromatic-cli/issues/462 is resolved.
 */
```
gonna guess this is also true here 😞

dierat (approved):
Makes sense to me!

PR #1083, opened 2022-05-23
chore(text): consolidate textpassage into text

jinlee93:
### Summary:
- consolidate `TextPassage` into `Text`
  - accept `as='div'`
    - also accepting new prop `capLinelength` which comes from `TextPassage`
  - transfer `TextPassage` styling
- moves `TextPassage` stories over to `Text`
- change `TextPassage` use to `Text`
- remove `TextPassage` component

[sc-199143]

### Test Plan:
- no significant visual regressions
- unit tests pass

default text passage:
Screen Shot 2022-05-23 at 2 28 43 PM

sm:
Screen Shot 2022-05-23 at 2 28 44 PM

lg:
Screen Shot 2022-05-23 at 2 28 47 PM


jinlee93:
Visual regressions: [overview: student view](https://www.chromatic.com/test?appId=61313967cde49b003ae2a860&id=628bf9f903d19f004afbcec8)

src/components/Text/Text.tsx
```diff
@@ -22,7 +23,8 @@ export type Props = {
   /**
    * Controls whether to render text inline (defaults to "p");
    */
-  as?: 'p' | 'span';
+  as?: 'p' | 'span' | 'div';
```
jinlee93, line 26: `as='div'` indicates use of a text passage

src/components/Text/Text.tsx
```diff
@@ -22,7 +23,8 @@ export type Props = {
...
+  capLinelength?: boolean;
```
jinlee93, line 27: This prop is from the previous `TextPassage`, kept it for sake of existing use cases, let me know how y'all feel

src/components/Text/Text.stories.tsx
```diff
@@ -173,3 +184,90 @@ export const TextVariantInherit: StoryObj<Args> = {
     variant: 'inherit',
   },
 };
+
+const TextPassageTemplate = (args: Args) => (
```
jinlee93, line 188: Replaced `TextPassage` with `Text`, surprisingly only one visual diff (noted above), but let me know if the `TextPassage` component shouldn't be deleted

src/components/Text/Text.module.css
```diff
@@ -92,3 +92,156 @@
...
+    margin-top: var(--eds-size-5);
+  }
+
+  /**
+	 * Sixth-level heading within text passage
+	 */
+  h6 {
+    @mixin eds-theme-typography-body-text-sm-bold;
+    margin-top: var(--eds-size-5);
+  }
```
jinlee93, line 178: These mixins are not the same as the mixins used for `h1` - `h6` in the `Heading` component, which mixins should be used?

src/components/Text/Text.module.css
```diff
@@ -92,3 +92,156 @@
...
 }
+
+/**
+ * 1) A passage of text, including various components (i.e. article, blog post)
+ */
+.text-passage {
+  /**
+	 * Paragraph within text passage
+	 */
+  p {
```
jinlee93, line 103: Noticed the original `TextPassage` component was not removing the browser spacing styles for `p` and `h1` - `h6` (not using @mixin reset). Is this intentional?

src/components/Text/Text.module.css
dierat: The mixins are being updated in https://github.com/chanzuckerberg/edu-design-system/pull/1075/files#r881562418 so the ones here and in `Heading` will produce the same styles.

src/components/Text/Text.tsx
dierat: Maybe add a comment about that, something about using a `div` if you have `p` tags in the text?

src/components/Toast/Toast.tsx
```diff
@@ -12,7 +12,7 @@ export type Props = {
    */
   className?: string;
   /**
-   * The child node(s) contains the toast message. Note: the toast message is displayed inside a TextPassage, so children can contain raw HTML
+   * The child node(s) contains the toast message. Note: the toast message is displayed inside a Text, so children can contain raw HTML
```
dierat, line 15: This actually isn't true anymore, so we could remove this comment or add `Text` around `children` to make it come true.

src/components/Text/Text.tsx
dierat: I'm just wondering if it should be noted in the component docstring since it's set to `true` by default.

dierat: I can't leave a comment on the docstring so I'll put it here – I think we should also add in some explanation about how the component works since there's a lot of magic under the hood now.

src/components/Text/Text.tsx
```diff
@@ -22,7 +23,8 @@ export type Props = {
...
   children: React.ReactNode;
   className?: string;
   variant?: Variant;
```
dierat, line 30: I can't leave a comment on the `spacing` prop line, but I wonder if we should consider renaming that to `spacingBottom` or something since `spacing` could now be misinterpreted to mean the spacing between sections inside the text passage.

src/components/Text/Text.tsx
jinlee93: Sounds good, will do

src/components/Text/Text.module.css
jinlee93: Thanks!

.storybook/pages/ProjectOverview/ProjectOverview.tsx
```diff
@@ -147,62 +147,62 @@ export const ProjectOverview = () => {
               />
             }
           >
-            <TextPassage>
+            <Text as="div">
```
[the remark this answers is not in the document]
jinlee93: Yea, there is some `<a>` styling within `.text-passage` which is attached with `as='div'`

src/components/Text/Text.tsx
jinlee93: That does make sense... will rename to `spacingBottom`

src/components/Text/Text.module.css
```diff
@@ -81,14 +81,167 @@
...
+.text--2x-bottom-spacing {
   margin-bottom: var(--eds-size-2);
 }
+
+/**
+ * 1) A passage of text, including various components (i.e. article, blog post)
+ */
+.text-passage {
+  /**
+	 * Paragraph within text passage
```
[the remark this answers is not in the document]
jinlee93: Oh definitely, thanks for catching this

src/components/Text/Text.tsx
```diff
@@ -21,13 +22,20 @@ export type Variant =
...
+   * Flags if the length of the text passage should be capped.
+   * Used only with as="div" and defaults to true
+   */
+  capLinelength?: boolean;
   children: React.ReactNode;
   className?: string;
   variant?: Variant;
   size?: Size;
-  spacing?: 'half' | '1x' | '2x';
+  bottomSpacing?: 'half' | '1x' | '2x';
```
dierat, line 38: @anniehu4 how do you feel about this change? Discussion's here: https://github.com/chanzuckerberg/edu-design-system/pull/1083#discussion_r881572437

src/components/Text/Text.tsx
```diff
@@ -36,16 +44,31 @@ export type Props = {
  * ```ts
  * import {Text} from "@chanzuckerberg/eds";
  * ```
+ *
+ * There are two perceived used cases for the text component.
```
dierat, line 48:
```suggestion
 * There are two perceived use cases for the text component.
```

src/components/Text/Text.tsx
jinlee93: `spacingBottom` also makes more sense in the text-passage use case

dierat (approved):
Looks nice! Tbh I only skimmed the stuff you copied from `TextPassage` since it was already in the codebase, just in a different component. And I would get @anniehu4 's take on the `spacingBottom` part since we're going to have to codemod that in `traject`.

jinlee93:
> And I would get \[Annie's\] take on the spacingBottom part since we're going to have to codemod that in traject.

aw crap forgot about the codemod part, good looking out, thanks for the review

src/components/Text/Text.tsx
anniehu4: Is there a different prop that controls spacing between sections? If not, as long as we document clearly I don't think people will be too confused. If we'd prefer more clarity (and don't mind the codemod -- looks like there are only ~49 usages) maybe we fully rename it to `marginBottom` for max clarity?

jinlee93: There is no prop that controls spacing between sections, spacing between sections is specific to the child element. Personally I would like to keep it as `spacing` for prop consistency but wouldn't mind `spacingBottom` or `marginBottom`

anniehu4: haha ya sorry sneaky edit, I decided that i'm indifferent between `spacingBottom` and `marginBottom` if we rename

src/components/Text/Text.tsx
anniehu4: thanks for the 👀 ! I'm good either way, just added some consideration to the thread

src/components/Text/Text.tsx
jinlee93: Do you mind if I revert it back to `spacing` and also add a comment such as `Adds margin bottom spacing to the <Text> component`

anniehu4: consistency is certainly valuable 🙂 I may err on the side of not introducing a breaking change unless we have proof people don't like the old way. hopefully with CSS untangling we can remove this prop altogether 🤞 but ultimately up to you!

jinlee93: Made back to `spacing` for consistency and not introducing breaking change, thanks all

PR #1085, opened 2022-05-24
chore(numbericon): remove aria labelledby prop for aria label req

jinlee93:
### Summary:
- removes `aria-labelledby` prop
- requires `aria-label` prop
- type checking makes previous console warns redundant hence removing
### Test Plan:
- unit tests pass
- no visual regressions


dierat (approved):
Nice! Thanks for updating ❤️

PR #1095, opened 2022-05-26
Ifrost/187997 create logo

ifrost1:
### Summary:
- Added the proper logo svg for LP to the `GlobalHeader` recipe since that is specific to LP
- Removed the `LogoImage` since that'll be handled at the application level
- Made the logo more flexible by taking in children (to accept either `img` tag or inline SVG`
- Added `ariaLabel` prop to the `a` tag inside `Logo` for accessibility

### Test Plan:
- `yarn test` runs without error
- `yarn lint` runs without error`
- `yarn types` runs without error


.storybook/recipes/GlobalHeader/GlobalHeader.tsx
```diff
@@ -64,7 +64,27 @@ export const GlobalHeader = ({ className, ...other }: Props) => {
   });
   return (
     <Header behavior="sticky" className={componentClassName} {...other}>
-      <Logo className={styles['global-header__logo']} href="#" />
+      <Logo
+        ariaLabel="Link to Learning Platform Homepage"
```
dierat, line 68: "Link to" seems redundant since the screen reader will announce that it's a link. Maybe just "Learning Platform Homepage"? I think I've heard from accessibility trainers before that it can just be the app name, so just "Learning Platform".

src/components/Logo/Logo.tsx
```diff
@@ -35,12 +42,27 @@ export interface Props {
...
   return (
     <div className={componentClassName} {...other}>
-      <a className={styles['logo__link']} href={href} rel="home" title={title}>
-        <LogoImage alt={alt} src={src} />
+      <a
+        aria-label={ariaLabel}
+        className={styles['logo__link']}
+        href={href}
+        rel="home"
+        title={title}
```
dierat, line 63: Do we need both an `ariaLabel` and a `title`?

ifrost1: title probably wouldn't hurt so a user could hover and get a message there. I don't know what other value that would add

dierat (approved):
Awesome thanks!

src/components/Logo/Logo.stories.tsx
```diff
@@ -12,9 +12,30 @@ export default {
...
+        />
+      </svg>
+    </Logo>
+  </div>
+);
 
 export const Default = Template.bind({});
 Default.args = {
+  href: '#',
+  ariaLabel: 'Link to learning platform',
```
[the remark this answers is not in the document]
ifrost1: Done. Thanks for the close eye

PR #1096, opened 2022-05-26
chore(messaging): darken icons and dark borders in mixins

dierat:
### Summary:
The icons and darker borders in the "warning", "success", and "brand" variants of the messaging components (`Banner`, `Toast`, InlineNotification`) are a little too light. This PR darkens them to be consistent with the mocks.

The messaging components are in a slightly awkward place right now because the border around the `InlineNotification` is darker than the border around the `Banner` and `Toast` but lighter than the dark border on the side of `Banner` and `Toast`. So we have 3 border colors (2 of which are being used in `Banner` and `Toast` and 1 that's being used in `InlineNotification`). From the mocks, it looks like the `InlineNotification` border will be used in the `Banner`, but when used in combination with the drop shadow that's on the `Banner` and `Toast`, that border looks really dark.

So as much as I wanted to keep these components in sync, I'm just going to accept that the `InlineNotification` is going to have its own border color that's separate from `Toast` and `Banner`.  We can rejigger this when the `Banner` and `Toast` are updated (or when the `Banner` is updated because I think `Toast` might be staying the same? in which case that will be the odd one out).

I'm also renaming the messaging mixin css variables to be a little more specific:
- `--dark-border-color` is now `--messaging-icon-color` because it's being used for the icon in all of the components (and the `Banner` and `Toast` just happen to also use it as the dark border color)
- `--light-border-color` is now `--messaging-border-color` since the other border color was renamed to `icon`

##### Before
Screen Shot 2022-05-27 at 9 41 06 AM

##### After
Screen Shot 2022-05-27 at 9 40 42 AM

##### Before
Screen Shot 2022-05-27 at 9 41 12 AM

##### After
Screen Shot 2022-05-27 at 9 40 49 AM

##### Before
Screen Shot 2022-05-27 at 9 41 14 AM

##### After
Screen Shot 2022-05-27 at 9 40 53 AM

### Test Plan:
- brand, warning, and success Banner and Toast icons are darker and match the mocks
- the light borders on the `Banner` and `Toast`s are unchanged
- `InlineNotification` is unchanged

### DIFF
--- a/src/components/Banner/Banner.module.css
+++ b/src/components/Banner/Banner.module.css
@@ -24,14 +24,14 @@
   border-style: solid;
   border-radius: var(--eds-theme-border-radius);
   border-width: var(--eds-theme-border-width);
-  border-color: var(--border-light-color);
+  border-color: var(--messaging-border-color);
 
   /* Vertical-specific styles */
   padding: var(--eds-size-4);
   text-align: center;
   justify-items: center;
   border-top-width: var(--eds-border-width-lg);
-  border-top-color: var(--border-dark-color); /* 3 */
+  border-top-color: var(--messaging-icon-color); /* 3 */
 }
 
 /**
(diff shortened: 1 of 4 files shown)


dierat:
https://app.shortcut.com/czi-edu/story/199697/edit-course-discrepancy-teacher-course-courseid-edit

anniehu4 (approved):

PR #1104, opened 2022-06-01
Ifrost/material sharp icons

ifrost1:
### Summary:
- Add material sharp icons
- Add custom properties to `status-` icons that allow a user to pass in another color if absolutely needed. But these should be reserved for statuses overall.

### Test Plan:
`yarn test` works
`yarn types` works
`yarn lint` works
`yarn start` works.


anniehu4 (commented):
overall looks great, thanks!

I was looking at the [visual icon grid changes](https://www.chromatic.com/test?appId=61313967cde49b003ae2a860&id=6297bfd2f6788d004a690469) and a few of these icons are now inverted 🤔 in particular, these 3 are currently used in LP:
- account-circle
- sentiment-very-satisfied
- supervised-user-circle
Screen Shot 2022-06-01 at 3 51 46 PM

Do you think we should worry about that? or would current uses still look fine

anniehu4 (approved):

PR #1105, opened 2022-06-01
feat(modal): add core modal

jinlee93:
### Summary:
- Brings in EDS Candidates modal and reconfigures based on [mock](https://www.figma.com/file/Ff7hJoSvXX9hkAmIklnsT2/EDS---UI-Components-(Main)?node-id=2764%3A14674)
- Adds modal stepper in footer
- [sc-194859]
- There is no js/react overflow logic. There will be documentation for teams to develop mobile first to gauge fit and control scrollable externally or handle overflow by themselves with `Modal.Body` children
- Brand modal header title font is unique
  - This was confirmed by designer
- There exists placeholder blocks from brand team for some assets, but they can be placed within children of `Modal.Header` as they please

### Test Plan:
- unit tests
- visual regression

### Concerns:
- Confirmed with designer: no modal size variants. Meaning only one responsive size.
- Fixed height
- Lots of spacing without tokenizing

Screen Shot 2022-06-03 at 12 12 28 PM
Screen Shot 2022-06-03 at 12 12 32 PM
Screen Shot 2022-06-03 at 12 12 48 PM
Screen Shot 2022-06-03 at 12 13 00 PM
Screen Shot 2022-06-03 at 12 13 02 PM
Screen Shot 2022-06-03 at 12 13 09 PM
Screen Shot 2022-06-03 at 12 13 20 PM
Screen Shot 2022-06-03 at 12 13 22 PM
Screen Shot 2022-06-03 at 12 13 28 PM
Screen Shot 2022-06-03 at 12 14 20 PM
Screen Shot 2022-06-03 at 12 14 31 PM


src/components/Modal/Modal.module.css
```diff
@@ -0,0 +1,140 @@
...
+  }
+}
+
+/**
+ * Allows scrolling of the modal content except for sticky footer.
+ * This functionality is our intended scroll behavior but consuming teams can 
+ * style the body content as they wish to handle overflow in various ways.
+ */
+.modal__content--scrollable {
+  overflow: scroll;
```
dierat, line 108: Does this work? I tried putting a bunch of content in a modal story and I can't get it to scroll.

dierat: Ah, I see, you need to pass a prop to get it to scroll.

I'm a bit concerned about that because some modals will have enough content to need to be scrollable on mobile but not enough content to be scrollable on desktop. So devs won't think to add the prop and it will be unusable on mobile...

src/components/Modal/Modal.module.css
```diff
@@ -0,0 +1,140 @@
...
+ * 1) This transparent border is for Window High Contrast Mode, which removes all
+    background colors but makes borders 100% opacity black. Without this, the
+    modal would have no clear boundary.
+ *  2) We use flexbox to handle height and width automatically. 
+    The body can scroll when content goes beyond the component,
+    while the footer remains static (if scrollable).
+ */
+.modal__content {
+  position: relative;
+  height: 43.125rem;
```
dierat, line 79: So the thing about the fixed height is, separate from the content scrolling part, if the browser window is too short to show the whole modal, you can't scroll it to see the whole thing, which means the header and footer might be unreachable to a user. For a specific use case, image a user is on a chromebook and has to zoom in to read the text. They can't zoom in and also interact with the modal.

src/components/ModalBody/ModalBody.tsx
```diff
@@ -0,0 +1,24 @@
...
+  /**
+   * CSS class names that can be appended to the component.
+   */
+  className?: string;
+};
+
+export const ModalBody = (props: Props) => {
+  return (
+    // This element is tabbable to allow keyboard users to scroll long content.
+    // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
```
dierat, line 19: Is it though?

src/components/ModalFooter/ModalFooter.module.css
```diff
@@ -0,0 +1,26 @@
...
+    # MODAL FOOTER
+\*------------------------------------*/
+
+/**
+ * Footer for the modal.
+ */
+.modal-footer {
+  padding-top: var(--eds-size-2-and-half);
+  padding-bottom: var(--eds-size-2-and-half);
+  padding-left: var(--modal-horizontal-padding, var(--eds-size-3));
```
dierat, line 11: Why are you using two values here?

dierat:
I added @pomfy113 as a reviewer because he worked on the LP version a lot and has used the modal in the product as well.

src/components/Modal/Modal.module.css
jinlee93: I definitely agree with your sentiment and spoke with the designer about this. But he advised that there will be notes to do mobile design first to handle overflow and then pass the prop as necessary. Also body content itself can be `overflow: scroll`.

One alternative is also to attach the prop based on overflow content, but that would make the component heavy

src/components/Modal/Modal.module.css
```diff
@@ -0,0 +1,140 @@
...
+ * style the body content as they wish to handle overflow in various ways.
+ */
+.modal__content--scrollable {
+  overflow: scroll;
+}
+
+/**
+ * The modal close button
+ */
+.modal__close-button {
```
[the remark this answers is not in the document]
jinlee93: Thanks, I'll try that!

src/components/ModalBody/ModalBody.tsx
jinlee93: It is not 😅, will remove

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,233 @@
...
+      typeof child.type !== 'string' &&
+      child.type?.name === 'ModalTitle'
+    ) {
+      return true;
+    } else if ('props' in child && child.props.children) {
+      return childrenHaveModalTitle(child.props.children);
+    }
+    return false;
+  });
+}
```
pomfy113, line 103: I'm not sure how common it is, but I've seen at least one instance of a Heading + Subheading in the modal title. If I'm reading this right, this allows for that so long as at least one of the children is a Modal.Title, right? Purely for example, something like
```tsx
<div>
  <Modal.Title>Heading 1</Modal.Title>
  <Heading as="h2">Heading 2</Heading>
</div>

File: src/components/Modal/Modal.module.css (hunk 1 above)
Comment (dierat) [reply src/components/Modal/Modal.module.css:108]: So passing the prop will be based on whether or not the content is scrollable at the mobile level? If so, maybe add that to the prop comment.

File: src/components/Modal/Modal.module.css (hunk 2 above)
Comment (dierat) [reply src/components/Modal/Modal.module.css:79]: My recommendation is this is a max height where the minimum is something like 90% of the viewport.

File: src/components/Modal/Modal.module.css (hunk 1 above)
Comment (jinlee93) [reply src/components/Modal/Modal.module.css:108]: That makes sense to me, will add to prop comment

File: src/components/ModalBody/ModalBody.tsx (hunk 3 above)
Comment (dierat) [reply src/components/ModalBody/ModalBody.tsx]: But should it be? My question is actually that I think you're missing the tab index here. That could be set based on the value of `isScrollable`.

Comment (jinlee93) [reply src/components/ModalBody/ModalBody.tsx]: The modal scrolls without tabindex with up and down arrow keys as long as there's a focusable element, but I guess we run into the case with no focusable elements (not even the close button if hidden). Wasn't a fan of allowing focus on the body but I will bring that back since I don't see an alternative

File: src/components/Modal/Modal.module.css (hunk 5 above)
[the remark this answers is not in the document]
Comment (jinlee93) [reply src/components/Modal/Modal.module.css:115]: Updated as suggested

**Comment by ifrost1** [comment]
@jinlee93 @dierat Just taking a look at here and have a few questions:
- I'm wondering where this modal came from. I'm assuming it's from the EDS candidates Storybook? Just wondering since we have the `Modal` we brought over from our boilerplate in the upcoming-components folder.
- What's the reasoning behind having a `Modal.Title` rather than simply using a `Heading` component inside of that `Modal.Header`?
- Seems like `isScrollable` could be removed in favor of the `overflow: auto` and `max-height` properties that we could add to the modal. Don't want to lock people into non-scrollable modals since you can't guess every device that a user will use
- What's the benefit of using headless-ui over the Modal we brought in? Not trying to say our Modal is better, but I'm curious around the benefits of it are or if there's a way to remove that third party dependency?

Just all questions. I'm not trying to say "don't use headless UI" but I'm wondering if we necessarily need that or can use the Modal in `upcoming-components`. Feel like we are moving quickly and I want to make sure that this is a modal component that folks can easily use and flexible enough. Just always get nervous around 3rd party libraries (although headless ui is one of the better ones). Thanks!

**Comment by jinlee93** [comment]
> I'm wondering where this modal came from. I'm assuming it's from the EDS candidates Storybook? Just wondering since we have the `Modal` we brought over from our boilerplate in the upcoming-components folder.

This modal is mostly from the EDSCandidates folder we had in LP. It was being used in some places across the platform and I updated it with new requirements from the ui kit. I recognize that there is a `Modal` existing already in the `upcoming-components` folder but used the one from EDSCandidates since it has been around and the product team should be more or less familiar with it; however, I did not remove the one in `upcoming-components` in case there was value in it.

> * What's the reasoning behind having a `Modal.Title` rather than simply using a `Heading` component inside of that `Modal.Header`?

The `Modal.Title` is essentially HeadlessUI's `Dialog.Title` component, which comes with some specific accessibility features.

> * Seems like `isScrollable` could be removed in favor of the `overflow: auto` and `max-height` properties that we could add to the modal. Don't want to lock people into non-scrollable modals since you can't guess every device that a user will use

I agree with the overflow / max-height styling, but design specifically wanted fixed height modal with a nonsticky footer without elevation if not scrollable, and a sticky footer with elevation if scrollable. I don't think there is value in incorporating overflow logic within this component?

> * What's the benefit of using headless-ui over the Modal we brought in? Not trying to say our Modal is better, but I'm curious around the benefits of it are or if there's a way to remove that third party dependency?

As I mostly brought in the Modal from EDSCandidates, I think @dierat(?) might be able to answer this question better.

> Just all questions. I'm not trying to say "don't use headless UI" but I'm wondering if we necessarily need that or can use the Modal in `upcoming-components`. Feel like we are moving quickly and I want to make sure that this is a modal component that folks can easily use and flexible enough. Just always get nervous around 3rd party libraries (although headless ui is one of the better ones). Thanks!

I do agree with moving away from HeadlessUI... but I wasn't involved in the original eng decision to use it... Think the others might have a better answer

File: src/components/Modal/Modal.module.css (hunk 1 above)
Comment (dierat) [reply src/components/Modal/Modal.module.css:108]: I've also thought in the past about trying to do something fancy here with waiting for the component to render then comparing the client height of the body container with the scroll height, then automatically add the scroll css and tab index to the body container if the scroll height is taller than the client height. I never pursued that idea because I was afraid it might be fragile with the timing of the measurement logic, but you could consider trying that out if you're interested.

File: src/components/ModalBody/ModalBody.tsx (hunk 3 above)
Comment (dierat) [reply src/components/ModalBody/ModalBody.tsx:19]: >Wasn't a fan of allowing focus on the body but I will bring that back since I don't see an alternative

Yeah it's kinda annoying.. Make sure the focus outline is fully visible too; I ran into issues with that in the LP component.

**Comment by dierat** [comment]
@ifrost1 @jinlee93 what I like about the `HeadlessUI` component is that it has some nice logic in there to automatically focus on the first interactive element inside the modal. If there's no close button, it will find the first button in the content (possible the footer buttons) and lock onto that. We've been told by auditors in the past that's important for the focus to be *inside* the modal, not on the modal itself, which is really annoying to write from scratch.

File: src/components/ModalBody/ModalBody.tsx (hunk 3 above)
Comment (jinlee93) [reply src/components/ModalBody/ModalBody.tsx:19]: Good call, offset the outline by negative px to make it visible 
image

File: src/components/Modal/Modal.module.css (hunk 1 above)
Comment (jinlee93) [reply src/components/Modal/Modal.module.css:108]: I've had to do work like that previously and I don't really think it's worth supporting due to high complexity and possible race conditions

Comment (dierat) [reply src/components/Modal/Modal.module.css:108]: Yeah that's what I was thinking; thanks for verifying! 👍

Comment (dierat) [reply src/components/Modal/Modal.module.css:109]: Can we change this from `overflow: scroll;` to `overflow: auto;`? That way the scroll bars will only appear when needed (on operating systems where they appear on all scrollable content – Mac is sneaky and hides it as if the style was `overflow: auto;`). See this PR for more details: https://github.com/FB-PLP/traject/pull/10197

Comment (jinlee93) [reply src/components/Modal/Modal.module.css:109]: Yup sounds good to me

File: src/components/ModalStepper/ModalStepper.tsx [hunk 7]
```diff
@@ -0,0 +1,64 @@
...
+  className,
+  totalSteps,
+  ...other
+}: Props) => {
+  const componentClassName = clsx(styles['modal-stepper'], className);
+  if (
+    process.env.NODE_ENV !== 'production' &&
+    (totalSteps < 1 || activeStep < 1 || totalSteps < activeStep)
+  ) {
+    throw new Error('Invalid step numbers.');
```
dierat, line 42: Could you make this error message more specific about which number is invalid?

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,234 @@
...
+    className,
+    hideCloseButton = false,
+    isScrollable,
+    onClose,
+    variant,
+  } = props;
+
+  const componentClassName = clsx(
+    styles['modal__content'],
+    isScrollable && styles['modal__content--scrollable'],
```
dierat, line 123: Actually, should this be the element that has the tab index instead of the body? In the last version of the modal, the header was sticky, so it made sense for the tab index to be one level down.

dierat: And then also we could only put the tab index on there if the content is scrollable.

src/components/ModalStepper/ModalStepper.tsx
jinlee93: Yup will update

src/components/Modal/Modal.tsx
jinlee93: Oh that would make sense. Let me explore this a bit and also see how the focus indicator looks

jinlee93: Having it on `modal__content` causes contrast issue with the brand header and is hidden by the footer

Screen Shot 2022-06-08 at 4 34 36 PM
Screen Shot 2022-06-08 at 4 35 41 PM

src/components/Modal/Modal.test.tsx
```diff
@@ -0,0 +1,139 @@
...
+const { DefaultInteractive } = composeStories(stories);
+
+// Required because the modal uses react portal under the hood.
+require('intersection-observer');
+
+describe('Modal', () => {
+  generateSnapshots(stories, {
+    getElement: async () => {
+      const nonInteractiveModal = screen.queryByTestId('non-interactive');
+      if (nonInteractiveModal) return nonInteractiveModal;
```
jinlee93, line 18: This way don't have to selectively extract stories

jinlee93:
More mobile screenshots after landscape orientation updates:

Screen Shot 2022-06-08 at 5 01 48 PM
Screen Shot 2022-06-08 at 5 01 54 PM
Screen Shot 2022-06-08 at 5 01 58 PM

jinlee93:
> Just something to think about moving forward. Always hard to "slow down" to think about this stuff when being pressed for deliverables. Don't want to rock the boat too hard, but do want to be mindful if HeadlessUI doesn't need to be relied on in the future. That's all. Thanks for reading my novel here.

@ifrost1 Thanks for really digging into the modal considerations. I agree with the sentiment on having less reliance on third party libraries and have run into api rigidity issues with other headless components myself and hence also think that relying less on headless would be great for future proofing. However the EDSCandidates modal that relies on Headless (which is what this modal is built off of) underwent many iterations when Diedra and Ferdinand worked on it together for their product team. Made that exploring non headless modal a separate [ticket](https://app.shortcut.com/czi-edu/story/201241) for this.

src/components/Modal/Modal.tsx
dierat: Could you add a comment explaining why the scrollable class is here while the tab index is somewhere else?

src/components/ModalStepper/ModalStepper.tsx
```diff
@@ -0,0 +1,70 @@
...
+  if (process.env.NODE_ENV !== 'production') {
+    if (totalSteps < 1) {
+      throw new Error('Must have more than one step in totalSteps.');
+    }
+    if (activeStep < 1) {
+      throw new Error('activeStep must be one or more.');
+    }
+    if (totalSteps < activeStep) {
+      throw new Error('activeStep cannot exceed totalSteps');
+    }
```
dierat, line 47: Thanks!

src/components/ModalBody/ModalBody.tsx
```diff
@@ -0,0 +1,24 @@
...
+    <div className={clsx(styles['modal-body'], props.className)} tabIndex={0}>
```
dierat, line 20: Can we only add this `tabIndex` when `isScrollable` is truthy?

src/components/Modal/Modal.tsx
jinlee93: Yup will do

src/components/ModalFooter/ModalFooter.tsx
```diff
@@ -0,0 +1,40 @@
+import clsx from 'clsx';
+import React, { ReactNode } from 'react';
+import styles from './ModalFooter.module.css';
+
+export type Props = {
+  /**
+   * Child node(s) that can be nested inside component. `ModalHeader`, `ModalBody`, and `ModelFooter` are the only permissible children of the Modal
```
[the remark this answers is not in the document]
jinlee93: ACK no, will fix

src/components/ModalBody/ModalBody.tsx
jinlee93: Lmk how that looks

src/components/ModalHeader/ModalHeader.tsx
```diff
@@ -0,0 +1,44 @@
+import clsx from 'clsx';
+import React, { ReactNode } from 'react';
+import styles from './ModalHeader.module.css';
+
+export type Props = {
+  /**
+   * Child node(s) that can be nested inside component. `ModalHeader`, `ModalBody`, and `ModelFooter` are the only permissible children of the Modal
```
dierat, line 7: This comment I pointed out before is on the `ModalHeader` too btw.

src/components/ModalBody/ModalBody.tsx
dierat: Yep that's just what I was thinking 👍 Thanks!

dierat (approved):
Looks great! Thanks for all the changes. There's just that one last comment that may have been accidentally copy/pasted on the `ModalHeader`.

src/components/ModalHeader/ModalHeader.tsx
jinlee93: God dang it thanks for all the catches

PR #1107, opened 2022-06-02
chore(clickable-style): update link button styles for v1

dierat:
### Summary:
This PR updates the link button styles to match the new mocks.

This includes:
- making the underline 2px all the time
- offsetting the underline by 2px (so it's a little further away from the text)
- giving the underline a different color than the text on default and hover states
- making the focus state a flat solid background color while the text and underline become white (the "brand" and "neutral" statuses have different background colors)
- adding a visited state style
- to support testing the visited and unvisited states, add randomized urls in storybook
- remove the "success", "warning", and "error" statuses
- remove the link button color tokens (because the designs are using tier 2 tokens now; we don't need custom tier 3 tokens now)
- consolidate the styles for the link button and the `textLink` mixin (so links in the `Text` component will look and act like the `Link` and `Button` link styles)

### Figma
Link: https://www.figma.com/file/Ff7hJoSvXX9hkAmIklnsT2/EDS-Paper-v1.0-%7C-UI-Components?node-id=912%3A4614
the link mocks in figma

### Video
https://user-images.githubusercontent.com/7761701/171753227-91ad558e-d5a9-4a06-9941-c92758397d7b.mp4

### Bugs / needs fixed
There is one things that's a little off in this change: the "neutral" status's underline still changes color on hover in the "visited" state. It's annoying me, but it's so small that I also kinda don't care.

link story in chromatic

### Test Plan:
Manually tested the link styles in storybook
- in the `Link`, `Button`, and `Text` components
- with and without reduced motion accessibility OS setting turned on
- for visited and unvisited states
and verified there were no snapshot changes.


dierat:
https://app.shortcut.com/czi-edu/story/198023/align-link-with-updated-tokens

.storybook/data/tokens.json
```diff
@@ -353,21 +353,6 @@
...
-  "eds-theme-color-button-link-neutral-active": "#383C43",
-  "eds-theme-color-button-link-success": "#007249",
-  "eds-theme-color-button-link-success-hover": "#005939",
-  "eds-theme-color-button-link-success-active": "#005939",
-  "eds-theme-color-button-link-warning": "#AC3400",
-  "eds-theme-color-button-link-warning-hover": "#842800",
-  "eds-theme-color-button-link-warning-active": "#842800",
-  "eds-theme-color-button-link-error": "#BD0044",
-  "eds-theme-color-button-link-error-hover": "#8F0134",
-  "eds-theme-color-button-link-error-active": "#8F0134",
```
dierat, line 370: I originally added the link token colors because the colors being used weren't all in the second tier tokens. That's no longer the case, so I just removed the tokens entirely.

src/components/Button/Button.stories.tsx
```diff
@@ -344,27 +350,3 @@ IconError.args = {
...
-LinkWarning.args = {
-  status: 'warning',
-  variant: 'link',
-};
-
-export const LinkError = Template.bind({});
-LinkError.args = {
-  status: 'error',
-  variant: 'link',
-};
```
dierat, line 370: These variants no longer exist.

src/components/ClickableStyle/ClickableStyle.module.css
```diff
@@ -457,91 +457,24 @@
...
-
-  &:active {
-    color: var(--eds-theme-color-button-link-brand-active);
-  }
 }
 
 /**
  * Link neutral clickable style
  */
 .clickable-style--link.clickable-style--neutral {
```
dierat, line 468: This component only has neutral link styles because the mixin can only make the brand status link.

src/components/PrimaryNavItem/PrimaryNavItem.module.css
```diff
@@ -10,7 +10,7 @@
  */
 .primary-nav__link {
   display: flex;
-  @mixin TextLinkInverted;
+  color: var(--eds-theme-color-text-neutral-default-inverse);
```
dierat, line 13: This was the only styling this link used from the mixin.

jest/helpers/removeRandomizedUrlJestSerializer.ts
```diff
@@ -0,0 +1,19 @@
...
+  test: (value) => {
+    return (
+      typeof value === 'string' &&
+      value.includes('https://www.google.com/search?q=')
+    );
+  },
+  print() {
+    return `"/"`;
+  },
+});
```
jinlee93, line 19: This looks pretty similar to the Modal serializer. Do you think it would be took awkward to a common utility that would receive `test` and `print` arguments or would that be too close to `addSnapshotSerializer`?

src/components/ClickableStyle/ClickableStyle.module.css
```diff
@@ -457,91 +457,24 @@
...
-  &:active {
-    color: var(--eds-theme-color-button-link-error-active);
+    text-decoration-color: var(--eds-theme-color-border-neutral-strong);
   }
 }
 
-/* Style icons that may be in the clickable style children */
-.clickable-style--link svg {
-  --icon-size-default: 1.25em;
+.clickable-style--link.clickable-style--neutral::before {
```
jinlee93, line 476: Is this supposed to append to / override the `::before` pseudoelement in the `textLInk` mixin? Can you add a comment for this block?

src/components/ClickableStyle/ClickableStyle.tsx
```diff
@@ -2,6 +2,9 @@ import clsx from 'clsx';
 import React, { MouseEventHandler, ReactNode } from 'react';
 import styles from './ClickableStyle.module.css';
 
+type Variant = 'primary' | 'secondary' | 'icon' | 'link';
+type Status = 'brand' | 'neutral' | 'success' | 'warning' | 'error';
```
jinlee93, line 6: nit: what do you feel about something like this:
```ts
export const VARIANTS = [
  'primary',
  'secondary',
  'icon',
  'link',
] as const;

export type Variant = typeof VARIANTS[number];
```
and similarly for `type Status`?

Not a change suggestion but wanted to get your thoughts on it since the exported `const VARIANTS` can be imported for use in the story files / other files but not sure how much use this will actually see and might just make the code look more complex

src/components/ClickableStyle/ClickableStyle.tsx
```diff
@@ -31,13 +34,45 @@ export type ClickableStyleProps<IComponent extends React.ElementType> = {
...
+      variant === 'primary' ? primaryStatuses : linkStatuses;
+
+    return `'${validStatuses[0]}' and '${validStatuses[1]}'`;
+  };
+
+  console.warn(
+    `*** Invalid prop combo warning ***:\n`,
+    `This component does not support using the '${status}' status with a '${variant}' variant.`,
+    `The '${variant}' variant can only be used with the ${getValidStatusesString} statuses.`,
+  );
```
jinlee93, line 73: So I been throwing errors and not `console.warn`ing. Not sure which is the better approach

src/util/getRandomUrl.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Helper method for generating a randomized valid url string.
+ *
+ * This is useful for testing links that have a distinct visited style. If the link always has
+ * the same href, you can only test the non-visited state once (unless you're using incognito
+ * browser windows) because once you've visited the site, the link now uses the visited style.
+ * And it needs to be a valid url in order to visit the site and trigger the visited state.
+ */
+const getRandomUrl = () => {
```
jinlee93, line 9: I like this lol. Think faker had a url generator but would rather not import it and keep it lighter like this. Also faker had that [corrupt package release](https://www.theverge.com/2022/1/9/22874949/developer-corrupts-open-source-libraries-projects-affected) so not keen on using it

src/components/Link/Link.stories.tsx
```diff
@@ -41,16 +42,22 @@ export default {
 
 const Template: Story<LinkProps> = (args) => (
```
jinlee93, line 43: Noticing the `Link` stories is not csf v3. Is this one of those story files where there are errors when made into csf v3?

src/design-tokens/mixins.css
```diff
@@ -6,34 +6,104 @@
...
+ * 2) Smaller gap between text and icon in link button style
+ * 3) 2px padding gives the text breathing room in the focus state
+ * 4) 2px margin compensates for the padding so the link doesn't have too much space between it
+ *    and surrounding text
+ * 5) Has position relative to support the background color pseudo element
+ * 6) We hide overflow so the pseudo element can be pushed out of site
+ * 7) Has smaller icons than other button styles
+ * 8) Make the focus outline invisible but don't remove it. High contrast mode will remove the background 
+      color, but it will also make borders 100% opacity black
+ */
```
jinlee93, line 21: Great comments!

jest/helpers/removeRandomizedUrlJestSerializer.ts
dierat: Yeah, it's so simple that I kinda feel like we'd just be making an alias for `addSnapshotSerializer`. Though I do find the `addSnapshotSerializer` API really weird so just being able to rename these props might be worth it 😅

src/components/ClickableStyle/ClickableStyle.tsx
dierat: I think it makes sense to throw an error if they're using something that is not supported/invalid and log a warning if it's something that's going to be deprecated but currently does work (because they're probably blocked by design on this one). I'll update this one to throw an error. Thanks!

src/components/ClickableStyle/ClickableStyle.module.css
dierat: Yeah it overrides it for situations where the purple bg might not mesh well.

src/components/ClickableStyle/ClickableStyle.tsx
dierat: I think that especially makes sense here because the `Button` and `Link` components use those values in their story `argTypes`. Updated in this PR 👍

src/components/ClickableStyle/ClickableStyle.module.css
dierat: Added a comment, thanks for the suggestion!

src/design-tokens/mixins.css
```diff
@@ -6,34 +6,104 @@
...
+  text-decoration: underline;
+  text-decoration-color: var(--eds-theme-color-background-brand-primary-strong);
+  text-decoration-thickness: 2px;
+  text-underline-offset: 2px;
+
+  padding: 2px !important; /* 3 */
+  margin: -2px; /* 4 */
+  position: relative; /* 5 */
+  overflow: hidden; /* 6 */
+  z-index: var(--eds-z-index-0);
```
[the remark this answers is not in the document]
dierat: Yes – will add a comment.

jest/helpers/removeRandomizedUrlJestSerializer.ts
jinlee93: Yea makes sense that it's so simple it'd just be an alias, upto you but will resolve conversation

src/components/Link/Link.stories.tsx
dierat: Testing it out with a simple story and none of the extras, it seems okay. How about I investigate that in a separate PR?

jinlee93: That sounds good to me!

jinlee93 (approved):
Lgtm, thanks for the changes

PR #1111, opened 2022-06-06
chore(storybook): upgrade to 6.5 and use webpack 5

jinlee93:
### Summary:
- upgrades SB to 6.5
- uses webpack 5 to match LP
### Test Plan:
- storybook builds
- axe tests run
- chromatic no diffs


package.json
```diff
@@ -167,7 +169,8 @@
     "svg-spritemap-webpack-plugin": "^3.9.1",
     "ts-jest": "^27.1.3",
     "typescript": "^4.6.2",
-    "web-vitals": "^2.1.4"
+    "web-vitals": "^2.1.4",
+    "webpack": "^5.73.0"
```
jinlee93, line 173: Having webpack as dep is required for using webpack 5 to build storybook.

I added using webpack 5 to build storybook to match LP, but lmk if should go back to webpack 4

dierat (approved):
Makes sense to me!

package.json
ahuth: Seems fine.

Since Storybook comes with it's own version of Webpack (v4), I'd probably just use that. Because I'm lazy. And because this library doesn't actually need Webpack (so it's kind of annoying to have to include our own dep on it).

But it also doesn't hurt anything, and is perfectly reasonable to use 5.

PR #1112, opened 2022-06-07
chore(breadcrumbs): align breadcrumbs with design

jinlee93:
### Summary:
- match breadcrumbs to design
- [mocks](https://www.figma.com/file/Ff7hJoSvXX9hkAmIklnsT2/branch/ogT9Jn4eqHHSPxHyWuNc33/EDS-Paper-v1.0-%7C-UI-Components?node-id=72%3A355)
[sc-200363]
### Test Plan:
- unit tests pass
- consuming pages/recipes not broken
- no visual regression

Screen Shot 2022-06-15 at 3 10 45 PM
Screen Shot 2022-06-15 at 3 11 02 PM
Screen Shot 2022-06-15 at 3 11 10 PM


src/components/Breadcrumbs/Breadcrumbs.stories.tsx
```diff
@@ -1,26 +1,67 @@
 import { BADGE } from '@geometricpanda/storybook-addon-badges';
-import { Story, Meta } from '@storybook/react';
+import { StoryObj } from '@storybook/react';
```
dierat, line 2: Nice 👍

src/components/BreadcrumbsItem/BreadcrumbsItem.module.css
```diff
@@ -0,0 +1,50 @@
...
+  flex-wrap: nowrap;
+  align-items: center;
+}
+
+/**
+ * Breadcrumbs link
+ */
+.breadcrumbs__link {
+  color: var(--eds-theme-color-text-neutral-subtle);
+  text-decoration: none;
```
dierat, line 21: I'm super not a fan of this.. It's literally impossible to tell which of the bread crumbs are links without interacting with the text.

jinlee93:
Just synced with design for this component, there needs to be more design work done regarding responsive layouts, and long breadcrumb lengths, do not merge until further design changes have been implemented

src/components/BreadcrumbsItem/BreadcrumbsItem.module.css
jinlee93: I brought this up with design but apparently doesn't need to be the purple link color due to the context of its usage. I'll swing it by design again...

dierat: Looking at it in the product and the examples in EDS, at least it looks like they are all always links (I've seen it before where the one furthest right isn't a link because it's the current page). I'm still not happy about it because it's perfectly possible for bread crumbs to be purely informative and not interactive.. but I dunno how hard we should push back on it because maybe all users do expect bread crumbs to be interactive; I don't know.

dierat (approved):

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -41,21 +43,125 @@ export const Breadcrumbs = ({
...
+    }
+  };
+
+  /**
+   * Needs useLayoutEffect over useEffect since it needs to be run before paint.
+   * TODO: with React 18, might be able to use useEffect https://github.com/reactjs/reactjs.org/blob/d14cbdca2445cd676526c4c52e1e106342ff7bb3/content/docs/hooks-reference.md?plain=1#L155
+   */
+  React.useLayoutEffect(() => {
+    updateShouldTruncate();
+    window.addEventListener('resize', updateShouldTruncate);
```
dierat, line 70: Should we debounce `updateShouldTruncate`?

jinlee93: I was gonna ask the same thing. Seemed like the other window resize listeners in `Header`, `GlobalHeader` etc. were not debouncing and `lodash` isn't in our `package.json`. 

I do want to debounce this however, should I install `lodash.debounce` and do so?

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -41,21 +43,125 @@ export const Breadcrumbs = ({
...
+  /**
+   * Checks if breadcrumbs would overflow
+   */
+  const updateShouldTruncate = () => {
+    setShouldTruncate(false);
+    if (
+      ref &&
+      ref.current &&
+      ref.current.clientWidth < ref.current.scrollWidth
+    ) {
```
dierat, line 59: Can we do the chainyquestionythingy?
```suggestion
    if (ref?.current?.clientWidth < ref.current.scrollWidth) {
```

src/components/Breadcrumbs/Breadcrumbs.tsx
dierat: I think that's the right direction, yeah. I think in production it's weird to not have a debounce on something like this.

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -41,21 +43,125 @@ export const Breadcrumbs = ({
...
+      setShouldTruncate(true);
+    }
+  };
+
+  /**
+   * Needs useLayoutEffect over useEffect since it needs to be run before paint.
+   * TODO: with React 18, might be able to use useEffect https://github.com/reactjs/reactjs.org/blob/d14cbdca2445cd676526c4c52e1e106342ff7bb3/content/docs/hooks-reference.md?plain=1#L155
+   */
```
dierat, line 67: Great comment btw 👍

src/components/BreadcrumbsItem/BreadcrumbsItem.tsx
```diff
@@ -26,22 +39,92 @@ export interface Props {
...
+          </>
+        ) : (
+          <a className={styles['breadcrumbs__link']} href={href as string}>
+            {/* 2 */}
+            {variant === 'back' ? (
+              <Icon
+                className={styles['breadcrumbs__back-icon']}
+                name="chevron-left"
+                purpose="informative"
+                title="back"
```
jinlee93, line 115: Not sure if this is the appropriate title

src/util/flattenReactChildren.ts
```diff
@@ -0,0 +1,18 @@
...
+): ReactChildArray {
+  const childrenArray = React.Children.toArray(children);
+  return childrenArray.reduce((flatChildren: ReactChildArray, child) => {
+    if ((child as React.ReactElement<any>).type === React.Fragment) {
+      return flatChildren.concat(
+        flattenReactChildren((child as React.ReactElement<any>).props.children),
+      );
+    }
+    flatChildren.push(child);
+    return flatChildren;
```
dierat, line 16: Could you do
```suggestion
    return [...flatChildren, child];
```
?

src/components/BreadcrumbsItem/BreadcrumbsItem.tsx
dierat: Should the accessible title be the `text` then? Since that's where the link is going, it sounds like.

src/util/flattenReactChildren.ts
jinlee93: Actually I think it's better to keep it this way to prevent multiple iterations of destructuring `flatChildren`

src/components/BreadcrumbsItem/BreadcrumbsItem.tsx
jinlee93: Yes! Will change

src/components/Breadcrumbs/Breadcrumbs.tsx
jinlee93: Doesn't work :( 
image

src/components/BreadcrumbsItem/BreadcrumbsItem.tsx
```diff
@@ -26,22 +39,92 @@ export interface Props {
...
+            </button>
+            <DropdownMenu
+              className={dropdownMenuClassName}
+              isActive={isActive}
+              onBlur={handleBlur}
+            >
+              {dropdownMenuItems}
+            </DropdownMenu>
+          </>
+        ) : (
```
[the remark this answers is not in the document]
jinlee93: Let me know how that looks, may or may not need some more drying

src/components/Breadcrumbs/Breadcrumbs.tsx
dierat: Weird 🤔

dierat: I'm fine with leaving it as-is, but in case you're curious, this might explain why it's not working: https://stackoverflow.com/questions/66017941/object-is-possibly-undefined-error-on-optional-chaining-typescript

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -41,21 +44,126 @@ export const Breadcrumbs = ({
...
+    updateShouldTruncate();
+    window.addEventListener('resize', debouncedUpdateShouldTruncate);
+    return () => {
+      window.removeEventListener('resize', debouncedUpdateShouldTruncate);
+    };
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+
+  /**
+   * 1) Flattens breadcrumb items that may be wrapped in React.Fragment into an array so they can be manipulated easily
+   * 2) Finds the second last breadcrumb item for mobile Breadcrumbs back icon
```
dierat, line 80:
```suggestion
   * 2) Finds the second to last breadcrumb item for mobile Breadcrumbs back icon
```

src/components/Breadcrumbs/Breadcrumbs.tsx
jinlee93: Yea it's because the initial value is `null`. I changed it to empty object so the optional chaining works now, thanks!

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -41,21 +44,126 @@ export const Breadcrumbs = ({
...
+  const componentClassName = clsx(styles['breadcrumbs'], className);
   return (
     <nav
       aria-label={ariaLabel}
       className={componentClassName}
       id={breadcrumbsId}
       {...other}
     >
-      <ul className={styles['breadcrumbs__list']}>{children}</ul>
+      {/**
```
dierat, line 122: This is much easier to read, thanks!

src/components/Breadcrumbs/Breadcrumbs.tsx
jinlee93: Ah TIL about nullish colescing and optional chaining and TS lol. Changed to:
```tsx
  if ((ref?.current?.clientWidth ?? 0) < (ref?.current?.scrollWidth ?? 0)) {
```

src/components/BreadcrumbsItem/BreadcrumbsItem.tsx
```diff
@@ -26,22 +39,94 @@ export interface Props {
...
+      /* The collapsed variant is a button with ellipsis. Interaction spawns a dropdown containing the collapsed breadcrumb links. */
+      return (
+        <>
+          <button
+            className={ellipsisButtonClassName}
+            onClick={() => {
+              setIsActive(!isActive);
+            }}
+          >
+            ...
```
dierat, line 90: You're going to need an aria label on this. I just tested it out with voiceover and it just says "button" because a period is just a pause. It also doesn't have the `aria-expanded` info so a user knows it opens a dropdown. Should we be using the `ButtonDropdown` component here instead of using the `DropdownMenu` directly?

src/components/BreadcrumbsItem/BreadcrumbsItem.tsx
```diff
@@ -26,22 +39,94 @@ export interface Props {
...
-      <Icon
-        className={styles['breadcrumbs__icon']}
-        color={EdsThemeColorIconNeutralStrong}
-        name="chevron-right"
-        purpose="decorative"
-        size="0.75rem"
-      />
+      {getInteractionElement()}
+      {/* The decorative separator between breadcrumbs items. */}
+      <span aria-hidden className={styles['breadcrumbs__icon']}>
```
dierat, line 127: Maybe the class name here could be more self-explanatory so the comment isn't needed, like `breadcrumbs__separator`.

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -41,21 +44,126 @@ export const Breadcrumbs = ({
...
+  const shouldThrowError = (flattenedChildren as JSX.Element[]).some(
+    (child) => {
+      if (child.type === BreadcrumbsItem || child.type === Breadcrumbs.Item) {
+        return false;
+      }
+      return true;
+    },
+  );
+  if (process.env.NODE_ENV !== 'production' && shouldThrowError) {
+    throw 'Only <Breadcrumbs.Item>, <BreadcrumbsItem>, or React.Fragment of aforementioned components allowed';
```
dierat, line 164: I hate to give you more work, but should we have a test for this?

src/components/BreadcrumbsItem/BreadcrumbsItem.tsx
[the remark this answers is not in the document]
dierat: Looks good, thanks!

src/components/BreadcrumbsItem/BreadcrumbsItem.tsx
```diff
@@ -26,22 +39,94 @@ export interface Props {
...
+      );
+    } else if (variant === 'back') {
+      /* The back variant is a left pointing icon that usually links to the second last breadcrumb href. */
+      return (
+        <a className={styles['breadcrumbs__link']} href={href as string}>
+          <Icon
+            className={styles['breadcrumbs__back-icon']}
+            name="chevron-left"
+            purpose="informative"
+            title={href as string}
```
dierat, line 109: Could we use `text` here?

src/components/Breadcrumbs/Breadcrumbs.tsx
dierat: I think what you had originally is easier to read though, even if it takes up multiple lines with the `&&` operators.

jinlee93: Same, will change back

src/components/Breadcrumbs/Breadcrumbs.tsx
jinlee93: Thanks, fixed grammar

src/components/BreadcrumbsItem/BreadcrumbsItem.tsx
jinlee93: Yep, will use `ButtonDropdown`. Initially thought that it wouldn't work, but just tried it and worked. Also will add `aria-label="Show more breadcrumbs"` (same as the [atlassian design system](https://atlassian.design/components/breadcrumbs/examples#long-breadcrumbs) lol)

src/components/BreadcrumbsItem/BreadcrumbsItem.tsx
jinlee93: Changed as suggested and removed comment, thanks!

src/components/Breadcrumbs/Breadcrumbs.tsx
jinlee93: Nah, but good callout, will add test

src/components/BreadcrumbsItem/BreadcrumbsItem.tsx
jinlee93: Yes, that would make more sense, thanks!

dierat (approved):
Just tested the collapsed state with voiceover and it sounds much better, thanks!

PR #1114, opened 2022-06-07
chore(banner): update text styles to match figma

dierat:
### Summary:
We have brand new shiny `Banner` component styles, which are similar to the old banner but there are important changes. The component in those designs is intended to be used as a top level page banner whereas the existing `Banner` is used in the page, sometimes even in sidebars.

As such, I'm going to make a new component, called `PageLevelBanner` to serve that need. For the existing `Banner`, I'm going to make a couple of style changes so it matches the new `PageLevelBanner` a little bit, but I'm not going to change too much. I mostly just want to update the text to be gray because the links will now be gray, but I'm also updating the spacing a little bit as well. In the future, we actually want people to use the `InlineNotification` for these in-page banners (also known as "section banners"), so this component will be deprecated and slowly phased out.

Changes made in this PR:
- `description` text is now gray (the `title` is still variant-colored)
- close button icon is now gray
- remove gap between `title` and `description`
- `title` text is a little bigger
- `description` text is a little smaller
- add deprecation badge to the whole component, not just the `isFlat` style

Changes in the new design I'm *not* making in this component:
- more horizontal padding
  - since the `Banner` component is sometimes used in sidebars, I don't want to add more horizontal padding that might make existing instances more squished
- light border and box shadow removed; dark border moved to the bottom
  - these new borders make sense on the `PageLevelBanner` because it's at the top of the page and only needs the bottom border to separate it from the content below. since the `Banner` is used inside the page, it needs separated on all 4 sides

#### Before
error with action and brand dismissable banner stories in storybook, before this change

#### After
error with action and brand dismissable banner stories in storybook, after this change

#### PageLevelBanner in figma
page level banner designs in figma

### Test Plan:
Storybook + chromatic

### DIFF
--- a/src/components/Banner/Banner.stories.tsx
+++ b/src/components/Banner/Banner.stories.tsx
@@ -3,6 +3,7 @@ import type { StoryObj } from '@storybook/react';
 import React from 'react';
 
 import { Banner, Variant } from './Banner';
+import styles from './Banner.stories.module.css';
 import Button from '../Button';
 import Heading from '../Heading';
 
@@ -188,7 +189,9 @@ export const DismissableWithAction: StoryObj<Args> = {
 export const DismissableBelowContent: StoryObj<Args> = {
   render: (args) => (
     <>
-      <Heading size="h1">Page Title</Heading>
+      <Heading className={styles.headingBottomSpacing} size="h1">
+        Page Title
+      </Heading>
       <Banner
         {...args}
         description={getDescription()}
--- a/src/components/Banner/Banner.stories.tsx
+++ b/src/components/Banner/Banner.stories.tsx
@@ -189,7 +189,7 @@ export const DismissableWithAction: StoryObj<Args> = {
 export const DismissableBelowContent: StoryObj<Args> = {
   render: (args) => (
     <>
-      <Heading className={styles.headingBottomSpacing} size="h1">
+      <Heading className={styles['heading']} size="h1">
         Page Title
       </Heading>
       <Banner
--- a/src/components/Banner/Banner.stories.tsx
+++ b/src/components/Banner/Banner.stories.tsx
@@ -13,6 +13,9 @@ export default {
     title:
       'New curriculum updates are available for one or more of your courses.',
   },
+  parameters: {
+    badges: [BADGE.DEPRECATED],
+  },
 };
 
 type Args = React.ComponentProps<typeof Banner>;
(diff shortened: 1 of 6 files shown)


src/components/Banner/Banner.tsx
```diff
@@ -168,14 +168,14 @@ export const Banner = ({
       />
 
       <div className={clsx(styles['banner__textAndAction'])}>
-        <div className={clsx(styles['banner__textContent'])}>
+        <div>
           {title && (
-            <Heading as={titleAs} size="title-sm" variant="inherit">
+            <Heading as={titleAs} size="title-md" variant={variant}>
```
dierat, line 173: I changed the variant here because I realized we really don't need to be setting the text color in the styles when the text components support that through props. (The color is being defined in the mixin. I'll update the other messaging components before updating the mixin itself.)

jinlee93:
> In the future, we actually want people to use the InlineNotification for these in-page banners (also known as "section banners"), so this component will be deprecated and slowly phased out.

Hopefully there are design guidelines on which notification to use for specific cases

src/components/Banner/Banner.tsx
```diff
@@ -93,6 +89,10 @@ const variantToIconAssetsMap: {
 };
 
 /**
+ * @deprecated The Banner component is deprecated and will be removed in an upcoming release.
+ * If your banner appears at the top of the page, please use the PageLevelBanner instead.
+ * If the banner appears lower in the page, please use the InlineNoticication component.
```
jinlee93, line 94:
```suggestion
 * If the banner appears lower in the page, please use the InlineNotification component.
```

src/components/Banner/Banner.stories.tsx
```diff
@@ -185,7 +189,9 @@ export const DismissableWithAction: StoryObj<Args> = {
 export const DismissableBelowContent: StoryObj<Args> = {
   render: (args) => (
     <>
-      <Heading size="h1">Page Title</Heading>
+      <Heading className={styles.headingBottomSpacing} size="h1">
```
jinlee93, line 192: nit: do we still want to use bracket notation for styles in stories?

dierat, line 192: I happened to notice that we have a class name for adding spacing to this heading that we weren't actually using.

dierat: Good call, will update 👍

src/components/Banner/Banner.tsx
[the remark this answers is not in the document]
dierat: Maybe we should just link to zeroheight?

jinlee93 (approved):
Lgtm, pending nit and comment updates

PR #1115, opened 2022-06-08
chore(textfield): move text field to components folder

jinlee93:
### Summary:
- moves TextField upcoming component to components folder
### Test Plan:
- tests pass
- cadstep1 not broken
- no visual regression


src/components/TextField/TextField.tsx
```diff
@@ -2,10 +2,10 @@ import clsx from 'clsx';
 import React, { ChangeEventHandler, MouseEventHandler, ReactNode } from 'react';
 import { useUID } from 'react-uid';
 import styles from './TextField.module.css';
-import Icon, { IconName } from '../../components/Icon';
-import FieldNote from '../FieldNote';
-import Label from '../Label';
-import TextInput from '../TextInput';
+import FieldNote from '../../upcoming-components/FieldNote';
```
ifrost1, line 5: Let's add `FieldNote` to components as well so it ships with the system

src/components/TextField/TextField.tsx
```diff
@@ -2,10 +2,10 @@ import clsx from 'clsx';
...
+import Label from '../../upcoming-components/Label';
```
ifrost1, line 6: Let's add `Label` to components as well so this ships with the system

src/components/TextField/TextField.tsx
```diff
@@ -2,10 +2,10 @@ import clsx from 'clsx';
...
+import TextInput from '../../upcoming-components/TextInput';
```
ifrost1, line 7: Let's add `TextInput` to the components as well so it ships with the system

src/index.ts
```diff
@@ -50,6 +50,7 @@ export { default as Tab } from './components/Tab';
 export { default as Tabs } from './components/Tabs';
 export { default as Tag } from './components/Tag';
 export { default as Text } from './components/Text';
+export { default as TextField } from './components/TextField';
```
ifrost1, line 53: Let's add `Label`, `TextInput`, and `FieldNote` to this `index.ts` as well once those are moved into components

ifrost1 (changes requested):
Let's add Label, TextInput, and FieldNote to components as well. If it is teams we are working with directly at the moment I think it's okay for them to use them. Might be worth communicating that if someone else is trying to pick this up.

ifrost1:
> I was intending on minimizing shipping components that haven't been vetted by EDS yet. I'm hesitant on shipping those components since the more usage they get, the harder it will be to change if they do need any. And since TextField seems to be more of a wrapper to TextInput I think there will be less usage of it in case changes need to be made as well.

In order for `TextField` to work, we need all of those in there. Otherwise, those imports won't work for them since only the `components` ship with the package

jinlee93:
> In order for TextField to work, we need all of those in there. Otherwise, those imports won't work for them since only the components ship with the package. That's where communication comes into play that no other teams should be using this aside from CAD Step 1

Makes sense to me, added the dependent components to the components folder

src/components/TextField/TextField.tsx
jinlee93: Added

src/components/TextField/TextField.tsx
jinlee93: Added

src/components/TextField/TextField.tsx
jinlee93: Added

src/index.ts
jinlee93: ADded

src/components/TextField/TextField.tsx
jinlee93: Added

ifrost1 (changes requested):
This all looks good to me. I'd like for @dierat to take a look too just to double check that nothing is wrong since this is going into a release. Thanks!

dierat (approved):
Looks good! I would just clean up the changelog since we don't need that level of detail about the changes.

ifrost1 (approved):
Everything looks good pending Diedra's comment about the CHANGELOG

jinlee93:
> Looks good! I would just clean up the changelog since we don't need that level of detail about the changes.

Reverted changelog edits that came from alpha publish, thanks all

PR #1119, opened 2022-06-09
feat(calendarcard): add recipe

csnizik:
### Summary:
https://www.figma.com/file/Ff7hJoSvXX9hkAmIklnsT2/EDS-Paper-v1.0-%7C-UI-Components?node-id=968%3A5311

Recipe for a calendar card to represent an event within calendar UI

### Test Plan:
Storybook: Recipes/CalendarCard

### DIFF
--- a/.storybook/recipes/CalendarCard/CalendarCard.module.css
+++ b/.storybook/recipes/CalendarCard/CalendarCard.module.css
@@ -1,0 +1,50 @@
+@import '../../../src/design-tokens/mixins.css';
+
+/*------------------------------------*\
+    #CALENDARCARD
+\*------------------------------------*/
+
+/**
+ * 1) 
+ */
+.calendar-card {
+  width: 16rem;
+  border-radius: var(--eds-border-radius-md);
+  background-color: var(--eds-theme-color-background-neutral-medium);
+}
+
+.calendar-card--brand {
+  border-left: var(--eds-border-width-lg) solid
+  var(--eds-theme-color-border-brand-primary-strong);
+}
+
+.calendar-card--revise {
+  border-left: var(--eds-border-width-lg) solid var(--eds-theme-color-border-grade-revise-strong);
+}
+
+.calendar-card--success {
+  border-left: var(--eds-border-width-lg) solid var(--eds-theme-color-border-utility-success-strong);
+}
+
+.calendar-card__body {
+  flex-direction: row;
+  gap: var(--eds-);
+}
+
+.calendar-card__footer {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+}
+
+.calendar-card__meta {
+  @mixin eds-theme-typography-body-text-xs;
+  color: var(--eds-theme-color-text-neutral-subtle);
+}
+
+.calendar-card__meta-icon {
+  margin-right: var(--eds-size-1);
+  position: relative;
+  bottom: 1px;
+}
--- a/.storybook/recipes/CalendarCard/CalendarCard.module.css
+++ b/.storybook/recipes/CalendarCard/CalendarCard.module.css
@@ -6,33 +6,37 @@
 
 /**
  * Card recipe for displaying calendar data with a colored left border (color dependent on variant)
- * 1) Brand variant (default): brand colored left border
- * 2) Revise variant
- * 3) Success variant
  */
 .calendar-card {
-  width: 16rem;
   background-color: var(--eds-theme-color-background-neutral-default-hover);
 }
-
+/**
+* Brand variant (default): brand colored left border
+*/
 .calendar-card--brand {
-  /* 1 */
   border-left: var(--eds-border-width-lg) solid
     var(--eds-theme-color-border-brand-primary-strong);
 }
 
+/**
+* Revise variant
+*/
 .calendar-card--revise {
-  /* 2 */
   border-left: var(--eds-border-width-lg) solid
     var(--eds-theme-color-border-grade-revise-strong);
 }
 
+/** 
+* Success variant
+*/
 .calendar-card--success {
-  /* 3 */
   border-left: var(--eds-border-width-lg) solid
     var(--eds-theme-color-border-utility-success-strong);
 }
 
+/**
+* Stack the card's contents in a flex column.
+*/
 .calendar-card__body {
   display: flex;
   flex-direction: column;
@@ -40,18 +44,31 @@
   gap: var(--eds-size-1);
 }
 
+/**
+* Card title row which holds an optional icon and a text string.
+* 1) Keeps the icon positioned at the top of the cross-axis if the text string expands to use multiple lines.
+* 2) If icon is used, the gap setting will maintain spacing between icon and text.
+*/
 .calendar-card__title {
   @mixin eds-typography-preset-007-bold;
   display: flex;
   align-items: flex-start;
-  justify-content: flex-start;
-  gap: var(--eds-size-1);
+  justify-content: flex-start; /* 1 */
+  gap: var(--eds-size-1); /* 2 */
 }
 
+/** 
+* Span to hold optional icon.
+*/
 .calendar-card__title-icon {
   min-width: 1.2rem;
 }
 
+/**
+* Card calendar row which holds an icon and a combination of text strings.
+* 1) Keeps the icon positioned at the top of the cross-axis if the text strings expands to use multiple lines.
+* 2) The gap setting will maintain spacing between icon and text.
+*/
 .calendar-card__dates {
   display: flex;
   align-items: flex-start;
@@ -59,10 +76,16 @@
   gap: var(--eds-size-1);
 }
 
+/* 
+* Span to hold icon before calendar text
+*/
 .calendar-card__dates-icon {
   min-width: 1.2rem;
 }
 
+/*
+* Keeps tags evenly spaced in rows, wrapping when needed
+*/
 .calendar-card__tags {
   display: flex;
   flex-wrap: wrap;
--- a/.storybook/recipes/CalendarCard/CalendarCard.module.css
+++ b/.storybook/recipes/CalendarCard/CalendarCard.module.css
@@ -50,7 +50,6 @@
 * 2) If icon is used, the gap setting will maintain spacing between icon and text.
 */
 .calendar-card__title {
-  @mixin eds-typography-preset-007-bold;
   display: flex;
   align-items: flex-start;
   justify-content: flex-start; /* 1 */
--- a/.storybook/recipes/CalendarCard/CalendarCard.module.css
+++ b/.storybook/recipes/CalendarCard/CalendarCard.module.css
@@ -10,7 +10,6 @@
 .calendar-card {
   width: 16rem;
   border-radius: var(--eds-border-radius-md);
-  background-color: var(--eds-theme-color-background-neutral-medium);
 }
 
 .calendar-card--brand {
@@ -24,27 +23,4 @@
 
 .calendar-card--success {
   border-left: var(--eds-border-width-lg) solid var(--eds-theme-color-border-utility-success-strong);
-}
-
-.calendar-card__body {
-  flex-direction: row;
-  gap: var(--eds-);
-}
-
-.calendar-card__footer {
-  display: flex;
-  flex-direction: column;
-  align-items: center;
-  justify-content: center;
-}
-
-.calendar-card__meta {
-  @mixin eds-theme-typography-body-text-xs;
-  color: var(--eds-theme-color-text-neutral-subtle);
-}
-
-.calendar-card__meta-icon {
-  margin-right: var(--eds-size-1);
-  position: relative;
-  bottom: 1px;
 }
--- a/.storybook/recipes/CalendarCard/CalendarCard.module.css
+++ b/.storybook/recipes/CalendarCard/CalendarCard.module.css
@@ -1,11 +1,11 @@
 @import '../../../src/design-tokens/mixins.css';
 
 /*------------------------------------*\
-    #CALENDARCARD
+    #CALENDAR CARD
 \*------------------------------------*/
 
 /**
- * 1) 
+ * Card recipe for displaying calendar data
  */
 .calendar-card {
   width: 16rem;
--- a/.storybook/recipes/CalendarCard/CalendarCard.module.css
+++ b/.storybook/recipes/CalendarCard/CalendarCard.module.css
@@ -11,7 +11,6 @@
   width: 16rem;
   justify-content: space-between;
   gap: var(--eds-size-1);
-  padding: var(--eds-size-1);
   background-color: var(--eds-theme-color-background-neutral-default-hover);
 }
 
@@ -38,11 +37,19 @@
   gap: var(--eds-size-1);
 }
 
+.calendar-card__title--icon {
+  min-width: 1.2rem;
+}
+
 .calendar-card__dates {
   display: flex;
-  align-items: center;
+  align-items: flex-start;
   justify-content: flex-start;
   gap: var(--eds-size-1);
+}
+
+.calendar-card__dates--icon {
+  min-width: 1.2rem;
 }
 
 .calendar-card__tags {
--- a/.storybook/recipes/CalendarCard/CalendarCard.module.css
+++ b/.storybook/recipes/CalendarCard/CalendarCard.module.css
@@ -5,39 +5,50 @@
 \*------------------------------------*/
 
 /**
- * Card recipe for displaying calendar data
+ * Card recipe for displaying calendar data with a colored left border (color dependent on variant)
+ * 1) Brand variant (default): brand colored left border
+ * 2) Revise variant
+ * 3) Success variant
  */
 .calendar-card {
   width: 16rem;
-  justify-content: space-between;
-  gap: var(--eds-size-1);
   background-color: var(--eds-theme-color-background-neutral-default-hover);
 }
 
 .calendar-card--brand {
+  /* 1 */
   border-left: var(--eds-border-width-lg) solid
     var(--eds-theme-color-border-brand-primary-strong);
 }
 
 .calendar-card--revise {
+  /* 2 */
   border-left: var(--eds-border-width-lg) solid
     var(--eds-theme-color-border-grade-revise-strong);
 }
 
 .calendar-card--success {
+  /* 3 */
   border-left: var(--eds-border-width-lg) solid
     var(--eds-theme-color-border-utility-success-strong);
+}
+
+.calendar-card__body {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  gap: var(--eds-size-1);
 }
 
 .calendar-card__title {
   @mixin eds-typography-preset-007-bold;
   display: flex;
-  align-items: center;
+  align-items: flex-start;
   justify-content: flex-start;
   gap: var(--eds-size-1);
 }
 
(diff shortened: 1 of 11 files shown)


.storybook/recipes/CalendarCard/CalendarCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+  ...other
+}: Props) => {
+  const componentClassName = clsx(
+    styles['calendar-card'],
+    className,
+    styles[`calendar-card--${variant}`],
+  );
+  return (
+    <div className={componentClassName}>
+      <Card className={styles['calendar-card__card']} {...other}>
```
dierat, line 42: I don't see `calendar-card__card` being defined anywhere. Could we put `componentClassName` here and drop the wrapping `div`? The border looks a little detached right now because it's on the wrapper `div` around the card. And then you could drop the `border-radius` on this component so we don't have to worry about this getting out of sync with `Card`.

calendar card in storybook, zoomed in and cropped to only show the left side

.storybook/recipes/CalendarCard/CalendarCard.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+  args: {
+    children: (
+      <>
+        <CalendarCard.Body>
+          <Heading as="h3" size="body-sm">
+            <Icon
+              name="status-check-circle"
+              purpose="decorative"
+              size="1.57rem"
+            />
```
dierat, line 30: I'm wondering if the icon and heading text should be props so we can style the spacing and keep it consistent.

.storybook/recipes/CalendarCard/CalendarCard.module.css
```diff
@@ -0,0 +1,26 @@
+@import '../../../src/design-tokens/mixins.css';
+
+/*------------------------------------*\
+    #CALENDARCARD
+\*------------------------------------*/
+
+/**
+ * 1)
```
dierat, line 8: Did you mean to leave this here?

.storybook/recipes/CalendarCard/CalendarCard.module.css
```diff
@@ -0,0 +1,26 @@
+@import '../../../src/design-tokens/mixins.css';
+
+/*------------------------------------*\
+    #CALENDARCARD
+\*------------------------------------*/
+
+/**
+ * 1) 
+ */
+.calendar-card {
```
dierat, line 10: In figma I see that the card has a background color of `Background/neutral-default-hover`. Is there a specific reason why you left that out?

.storybook/recipes/CalendarCard/CalendarCard.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export const Default: StoryObj<Args> = {
+  args: {
+    children: (
+      <>
+        <CalendarCard.Body>
+          <Heading as="h3" size="body-sm">
+            <Icon
+              name="status-check-circle"
+              purpose="decorative"
```
jinlee93, line 28: Think the green check circle conveys a meaning and should be "informative" and have a title such as "success" (and similarly for either status icons that may be placed here)

.storybook/recipes/CalendarCard/CalendarCard.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+            />
+            Project name here
+          </Heading>
+          <Text as="p" size="xs">
+            <Icon name="event-note" purpose="decorative" size="1.2rem" />
+            Feb 2 - Apr 1, 2022
+          </Text>
+          <Tag text="Ancient Mayan History" variant="neutral" />
+        </CalendarCard.Body>
+      </>
```
jinlee93, line 39: Is there value in make this into a helper component so that the storybook args can be accepted and have different stories for the variants? Think `variant` arg is being passed already, so would just have to componentize and have the `Icon` update with `variant`

.storybook/recipes/CalendarCard/CalendarCard.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+    children: (
+      <>
+        <CalendarCard.Body>
+          <Heading as="h3" size="body-sm">
+            <Icon
+              name="status-check-circle"
+              purpose="decorative"
+              size="1.57rem"
+            />
+            Project name here
```
jinlee93, line 31: How long can this copy get and deal with overflow / wrap?

.storybook/recipes/CalendarCard/CalendarCard.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+          </Heading>
+          <Text as="p" size="xs">
+            <Icon name="event-note" purpose="decorative" size="1.2rem" />
+            Feb 2 - Apr 1, 2022
+          </Text>
+          <Tag text="Ancient Mayan History" variant="neutral" />
```
jinlee93, line 37: How long can this copy get and deal with overflow / wrap? Can there be multiple tags for the calendar card?

.storybook/recipes/CalendarCard/CalendarCard.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+            <Icon
+              name="status-check-circle"
+              purpose="decorative"
+              size="1.57rem"
+            />
+            Project name here
+          </Heading>
+          <Text as="p" size="xs">
+            <Icon name="event-note" purpose="decorative" size="1.2rem" />
+            Feb 2 - Apr 1, 2022
```
jinlee93, line 35: qq: do you think there is value in passing date isostrings for this or should we just allow any string like this?

.storybook/recipes/CalendarCard/CalendarCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+  ...other
+}: Props) => {
+  const componentClassName = clsx(
+    className,
+    styles['calendar-card'],
+    styles[`calendar-card--${variant}`],
+  );
+  return (
+    <Card className={componentClassName} {...other}>
+      <Heading
```
ifrost1, line 74: We always want to use at least `CardBody` as a child component. This allows the card to take up the entire space if put inside a grid. So let's wrap all of the content in `CardBody`. `CardHeader` and `CardFooter` are always optional, but `CardBody` should be required.

.storybook/recipes/CalendarCard/CalendarCard.module.css
```diff
@@ -0,0 +1,59 @@
...
+ * Card recipe for displaying calendar data
+ */
+.calendar-card {
+  width: 16rem;
+  justify-content: space-between;
+  gap: var(--eds-size-1);
+  background-color: var(--eds-theme-color-background-neutral-default-hover);
+}
+
+.calendar-card--brand {
```
ifrost1, line 17: Let's start getting into the habit of commenting our CSS. I need to follow these rules myself 😉

.storybook/recipes/CalendarCard/CalendarCard.module.css
```diff
@@ -0,0 +1,59 @@
...
+
+/*------------------------------------*\
+    #CALENDAR CARD
+\*------------------------------------*/
+
+/**
+ * Card recipe for displaying calendar data
+ */
+.calendar-card {
+  width: 16rem;
```
ifrost1, line 11: Will these cards always be 16rem or will these depend on the layout that they sit in. Just flagging this as something to think about moving forward so this can be reused in other contexts if needed.

.storybook/recipes/CalendarCard/CalendarCard.module.css
```diff
@@ -0,0 +1,59 @@
...
+}
+
+.calendar-card__dates {
+  display: flex;
+  align-items: flex-start;
+  justify-content: flex-start;
+  gap: var(--eds-size-1);
+}
+
+.calendar-card__dates--icon {
```
ifrost1, line 51: Let's call this `dates-icon` and not `dates--icon`. We don't use this syntax since it looks like a modifier when it should just be a child. So we always use a `[child]-[part-of-child]` if we need to be specific generally

.storybook/recipes/CalendarCard/CalendarCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+        as="h3"
+        className={styles['calendar-card__title']}
+        size="body-sm"
+      >
+        {(variant === 'success' || variant === 'revise') && (
+          <span className={styles['calendar-card__title--icon']}>
```
ifrost1, line 80: Let's call this `title-icon` and not `title--icon`. We don't use this syntax since it looks like a modifier when it should just be a child. So we always use a `[child]-[part-of-child]` if we need to be specific generally

.storybook/recipes/CalendarCard/CalendarCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+              purpose="informative"
+              size="1.2rem"
+              title={variant}
+            />
+          </span>
+        )}
+        <span>{title}</span>
+      </Heading>
+      <Text as="p" className={styles['calendar-card__dates']} size="xs">
+        <span className={styles['calendar-card__dates--icon']}>
```
ifrost1, line 94: Let's call this `dates-icon` and not `dates--icon`. We don't use this syntax since it looks like a modifier when it should just be a child. So we always use a `[child]-[part-of-child]` if we need to be specific generally

.storybook/recipes/CalendarCard/CalendarCard.module.css
```diff
@@ -0,0 +1,59 @@
...
+
+.calendar-card__title {
+  @mixin eds-typography-preset-007-bold;
+  display: flex;
+  align-items: center;
+  justify-content: flex-start;
+  gap: var(--eds-size-1);
+}
+
+.calendar-card__title--icon {
```
ifrost1, line 40: Let's call this `title-icon` and not `title--icon`. We don't use this syntax since it looks like a modifier when it should just be a child. So we always use a [child]-[part-of-child] if we need to be specific generally

.storybook/recipes/CalendarCard/CalendarCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+      </div>
+      {children}
+    </Card>
+  );
+};
+
+/**
+ * Re-exports subcomponents so consuming apps don't have to import the Card component or the individual subcomponents.
+ */
+CalendarCard.Body = Card.Body;
```
ifrost1, line 117: Not sure the reasoning behind this stuff. The `CardBody` should be baked into this component since it's a specific composition

ifrost1 (changes requested):
This looks good so far from a styling standpoint. Just some syntactical and structural things to address and CSS comments. Thanks!

.storybook/recipes/CalendarCard/CalendarCard.module.css
```diff
@@ -0,0 +1,59 @@
...
+
+.calendar-card--success {
+  border-left: var(--eds-border-width-lg) solid
+    var(--eds-theme-color-border-utility-success-strong);
+}
+
+.calendar-card__title {
+  @mixin eds-typography-preset-007-bold;
+  display: flex;
+  align-items: center;
```
dierat, line 35: I'm wondering if this should be `flex-start`, like in `calendar-card__dates`, so the icon always floats to the top when the text wraps.
long text strings calendar card story in storybook

.storybook/recipes/CalendarCard/CalendarCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+        <span>{title}</span>
+      </Heading>
+      <Text as="p" className={styles['calendar-card__dates']} size="xs">
+        <span className={styles['calendar-card__dates--icon']}>
+          <Icon name="event-note" purpose="decorative" size="1.2rem" />
+        </span>
+        {dateStart} &mdash; {dateEnd} {meta}
+      </Text>
+      <div className={styles['calendar-card__tags']}>
+        {tags?.map(({ hasOutline, text, variant }) => (
```
dierat, line 100: Can you do
```suggestion
        {tags?.map(({ hasOutline = true, text, variant }) => (
```
so line 102 doesn't need the `||` operator?

.storybook/recipes/CalendarCard/CalendarCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+          <Tag
+            hasOutline={hasOutline || true}
+            key={`${text}`}
+            text={text}
+            variant={variant ? variant : 'neutral'}
```
dierat, line 105: Same question as the `hadOutline` one above.

.storybook/recipes/CalendarCard/CalendarCard.tsx
```diff
@@ -0,0 +1,119 @@
...
+        <span className={styles['calendar-card__dates--icon']}>
+          <Icon name="event-note" purpose="decorative" size="1.2rem" />
+        </span>
+        {dateStart} &mdash; {dateEnd} {meta}
+      </Text>
+      <div className={styles['calendar-card__tags']}>
+        {tags?.map(({ hasOutline, text, variant }) => (
+          <Tag
+            hasOutline={hasOutline || true}
+            key={`${text}`}
```
dierat, line 103: Can this line be
```suggestion
            key={text}
```
?

.storybook/recipes/CalendarCard/CalendarCard.tsx
dierat: Do you mean the reasoning behind having the component in a separate directory, like why not just define it in this file?

.storybook/recipes/CalendarCard/CalendarCard.stories.tsx
csnizik: I added a storybook example to demonstrate long lines

.storybook/recipes/CalendarCard/CalendarCard.stories.tsx
csnizik: I added a storybook example to demonstrate long lines

.storybook/recipes/CalendarCard/CalendarCard.module.css
csnizik: if size variants are added to the base Card component then this CSS size can be rm'ed.

.storybook/recipes/CalendarCard/CalendarCard.module.css
csnizik: done

.storybook/recipes/CalendarCard/CalendarCard.tsx
csnizik: I took it out

.storybook/recipes/CalendarCard/CalendarCard.module.css
csnizik: done

.storybook/recipes/CalendarCard/CalendarCard.tsx
csnizik: elegant and tidy, thank you 💯

.storybook/recipes/CalendarCard/CalendarCard.tsx
csnizik: done

.storybook/recipes/CalendarCard/CalendarCard.tsx
csnizik: done

.storybook/recipes/CalendarCard/CalendarCard.stories.tsx
dierat: Thanks!

.storybook/recipes/CalendarCard/CalendarCard.module.css
dierat: Thank you!

dierat (approved):
Thanks for all the changes!

.storybook/recipes/CalendarCard/CalendarCard.module.css
ifrost1: I don't think we'd want to add size variants to the base `Card` component. This card is used within like a calendar horizontal timeline thing and I'd imagine that each one of those columns could differ and may not be `16rem` all the time.

.storybook/recipes/CalendarCard/CalendarCard.tsx
ifrost1: @dierat I think we should normalize how subcomponents are being used. Seeing a mix of dot notation vs. non-dot notation if that's what you are talking about

ifrost1 (changes requested):
Getting there! Here are the biggest things:
- Let's start adding CSS comments
- Remove width
- Let's not use tier 1 variables (outside of border and box shadow values) within the CSS

.storybook/recipes/CalendarCard/CalendarCard.module.css
```diff
@@ -0,0 +1,70 @@
...
+
+.calendar-card__body {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  gap: var(--eds-size-1);
+}
+
+.calendar-card__title {
+  @mixin eds-typography-preset-007-bold;
```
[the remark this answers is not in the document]
dierat: Sorry, I'm having trouble understanding the question. Why can't we use `@mixin eds-theme-typography-label-md;`?

>but we only have label-md as the 007-bold value

Isn't the text bold in both the design and the preset?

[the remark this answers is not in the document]
anniehu4: If we want all the tier 2 type presets available for use by the `Heading` component, we could add them to a `typography.module.css` file that both `Heading` and `Text` use. i.e.
```
// typography.module.css
.typography--label-md {
  @mixin eds-theme-typography-label-md;
}

// usage
<Heading as="h3" size="label-md" />
<Text size="label-md" />
```

is that what you're thinking @ifrost1 ?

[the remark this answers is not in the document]
ifrost1: @dierat @anniehu4 Thanks for this. 

> Why can't we use @mixin eds-theme-typography-label-md;

We can. The issue lies in other designers using this for future cases and this being baked into this specific recipe.

> is that what you're thinking @ifrost1 ?

That is what I'm thinking. In an ideal world, we'd never have to style a `Heading` outside of the `Heading` component's CSS. That being said, if you don't want that in the `Heading` component, I understand having it as a recipe-specific thing. That being said, who is stopping designers from using that in other patterns? It would be nice to have the `Heading` handle typography preset changes. I prefer Annie's suggestion just because it gives options if other designs come out with this treatment simply because designers will design and I'd imagine this may get used again at some point.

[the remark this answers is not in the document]
ifrost1: Another option is to create a `title-xs` token that would be more associated with headings. `label-md` is associated with form field labels and that may differ in terms of multiple themes using this stuff.

.storybook/recipes/CalendarCard/CalendarCard.tsx
dierat: We do have a preferred pattern for subcomponents. I know the `Checkbox` is a little weird though because there was something blocking us from using dot notation. https://github.com/chanzuckerberg/edu-design-system/blob/main/docs/CODE_GUIDELINES.md#conventions-for-compound-components

ifrost1 (approved):

PR #1121, opened 2022-06-10
chore: don't import from @storybook/addon-docs/blocks

dierat:
### Summary:
I noticed a console warning in storybook about 
>browser.js:39 Importing from '@storybook/addon-docs/blocks' is deprecated, import directly from '@storybook/addon-docs' instead:
>
>https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-scoped-blocks-imports

storybook with the warning in question circled in the browser dev tools

This PR follows the warning's advice and updates the one place where we were importing from "@storybook/addon-docs/blocks" to instead import from "@storybook/addon-docs".

### Test Plan:
Load up storybook and verify that warning is gone.


jinlee93 (approved):
Damn nice catch, thanks for the fix

PR #1123, opened 2022-06-10
chore: move options out of control in story argTypes

dierat:
### Summary:
I noticed a console warning about how we're using `options` in `argTypes` in our stories. It looks like we just need to move that property outside of the `control` object so they're siblings. This PR goes through and makes that change for all the relevant components.

the relevant console warning in the browser dev tools on the button story page

### Test Plan:
Navigate to all the changes components (and `Checkbox` because that also uses `options` but didn't have this issue) and verify none of the components have this warning in the console now.

### DIFF
--- a/src/components/Button/Button.stories.tsx
+++ b/src/components/Button/Button.stories.tsx
@@ -19,20 +19,20 @@ export default {
     variant: {
       control: {
         type: 'select',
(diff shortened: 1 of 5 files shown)


jinlee93 (approved):
Thanks for the catch, lgtm

PR #1124, opened 2022-06-10
chore: upgrade storybook to v6.5.8

dierat:
### Summary:
I noticed a buncha console errors in storybook about
>The pseudo class ":first-child" is potentially unsafe when doing server-side rendering. Try changing it to ":first-of-type".

It looks like this was fixed in storybook v6.5.8 just a couple days ago:
- https://github.com/storybookjs/storybook/releases/tag/v6.5.8
- https://github.com/storybookjs/storybook/issues/6998#issuecomment-1151876307

Unfortunately, upgrading to v6.5.8 in this repo does not completely remove the console error, but it *is* less annoying.

#### Before
Screen Shot 2022-06-10 at 3 31 09 PM

#### After
first-child-console-error

### Test Plan:
Load up storybook and verify that error is now collapsed.


jinlee93 (approved):
> Unfortunately, upgrading to v6.5.8 in this repo does not completely remove the console error, but it is less annoying.

I'll take less annoying, thanks!

PR #1125, opened 2022-06-10
chore(token-presets): add key to map

dierat:
### Summary:
I noticed a console warning in storybook about:
>Warning: Each child in a list should have a unique "key" prop.
>
>Check the render method of `Tier1TypographyPresets`.

In this PR, I'm adding a key to quell react's concerns.

colors tier 1 definitions story in storybook with the relevant error highlighted in the browser dev tools

### Test Plan:
Navigate to any of the token stories in storybook and verify this error is now gone.

### DIFF
--- a/.storybook/components/DesignTokens/Tier1/TypographyPresets.js
+++ b/.storybook/components/DesignTokens/Tier1/TypographyPresets.js
@@ -58,6 +58,7 @@ export class Tier1TypographyPresets extends Component {
   render() {
     const renderTypeToken = (
       preset,
+      index,
       { boldVariant = true, lightVariant = false } = {},
     ) => {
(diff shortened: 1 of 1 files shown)


.storybook/components/DesignTokens/Tier1/TypographyPresets.js
```diff
@@ -68,7 +69,7 @@ export class Tier1TypographyPresets extends Component {
       };
 
       return (
-        <>
+        <Grid key={`tier-1-typography-preset-${index}`}>
```
dierat, line 72: Needed to use a `Grid` instead of a `div` here because there are 3 tokens in here and grouping them removed the gap between them. I also tried putting the key on each `TokenSpecimen`, but that didn't remove the warning.

jinlee93 (approved):
Lgtm! 🚢

PR #1126, opened 2022-06-10
chore(token-specimen): suppress content editable warning

dierat:
### Summary:
I noticed a console error in storybook about:
>Warning: A component is `contentEditable` and contains `children` managed by React.

I think what it's concerned about is that a user can click into the typography token sample text, edit the text, and then something could cause react to rerender the component, which would wipe away the user's changes. This editing capability is just so people can see what different strings will look like with these presets, so I think it's fine to just suppress the warning, which I'm doing in this PR. (I don't expect people to actually use this, especially since there's nothing in the UI to communicate that it's even possible, so I would also be fine with just removing the content editable feature.)

colors tier 1 definitions story in storybook with the relevant error highlighted in the browser dev tools

### Test Plan:
Navigate to any of the token stories in storybook and verify this error is now gone.

### DIFF
--- a/.storybook/components/TokenSpecimen/TokenSpecimen.js
+++ b/.storybook/components/TokenSpecimen/TokenSpecimen.js
@@ -21,6 +21,7 @@ export class TokenSpecimen extends Component {
           }
           contentEditable
           style={this.props.inlineStyles}
+          suppressContentEditableWarning={true}
         >
           AaBbCcDdEeFfGg
         </div>
@@ -34,6 +35,7 @@ export class TokenSpecimen extends Component {
           }
           contentEditable
           style={this.props.inlineStyles}
+          suppressContentEditableWarning={true}
         >
           Almost before we knew it, we had left the ground.
         </div>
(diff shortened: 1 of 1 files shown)


.storybook/components/TokenSpecimen/TokenSpecimen.js
```diff
@@ -21,6 +21,7 @@ export class TokenSpecimen extends Component {
           }
           contentEditable
           style={this.props.inlineStyles}
+          suppressContentEditableWarning={true}
```
jinlee93, line 24: since `contentEditable` is implied truthy, nit:
```suggestion
          suppressContentEditableWarning
```

.storybook/components/TokenSpecimen/TokenSpecimen.js
```diff
@@ -34,6 +35,7 @@ export class TokenSpecimen extends Component {
...
```
jinlee93, line 38: same nit as above

jinlee93 (approved):
Thanks for going through all these storybook errors,
Lgtm pending your thoughts on the nits

.storybook/components/TokenSpecimen/TokenSpecimen.js
dierat: Sounds good, will do, thanks!

dierat:
I don't know what's going on with the test status on the PR here, but chromatic says there are no changes:
test statuses on this pr; all have passed except the ui tests which say 'waiting for status to be reported'
this pr in chromatic, which says there are no changes

PR #1127, opened 2022-06-13
chore(plop): add spaces in css file headline

dierat:
### Summary:
In the css modules files for components and recipes, we have a headline text that looks like `# A COMPONENT NAME`. Right now, the plop generator doesn't add spaces between the words (so it would come out like `# ACOMPONENTNAME`, which is really hard to read). In this PR, I updated the helper that formats that text to place spaces between the words.

### Test Plan:
Test out plop for both components and recipes using both a spaced component name (ex: "a component name") and pascal cased component name (ex: "AComponentName"), and verify that all of these cases produce an uppercase text headline in the css file with spaces between the words, as expected.

### DIFF
--- a/plopfile.js
+++ b/plopfile.js
@@ -1,7 +1,10 @@
 const pascalCase = require('pascal-case');
+const { snakeCase } = require('snake-case');
 
 module.exports = (plop) => {
-  plop.setHelper('upperCase', (txt) => txt.toUpperCase());
+  plop.setHelper('spacedUpperCase', (txt) =>
+    snakeCase(txt).split('_').join(' ').toUpperCase(),
+  );
 
   // This helper allows us to place a variable inside curly braces without spaces
   // between the text and the braces.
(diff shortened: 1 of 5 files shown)


plopfile.js
```diff
@@ -1,7 +1,10 @@
 const pascalCase = require('pascal-case');
+const { snakeCase } = require('snake-case');
 
 module.exports = (plop) => {
-  plop.setHelper('upperCase', (txt) => txt.toUpperCase());
+  plop.setHelper('spacedUpperCase', (txt) =>
+    snakeCase(txt).split('_').join(' ').toUpperCase(),
```
dierat, line 6: I used `snakeCase` so that it would correctly handle pascal cased component names, which would otherwise require me to go through and check each character to see if it was a lowercase letter or not.

dierat:
No component changes on chromatic.
Screen Shot 2022-06-13 at 10 45 17 AM

ahuth (approved):

PR #1128, opened 2022-06-13
chore: remove remaining tab spaces

dierat:
### Summary:
The big medium codebase we started with used tabs instead of spaces. I thought there were a bunch more, but they must've been automatically converted when components were moved around and such, so there are only a couple left. This PR converts the last remaining tabs to spaces.

### Test Plan:
Verify the indentation in these files still looks right.


jinlee93 (approved):
spaces > tabs

PR #1129, opened 2022-06-13
chore(plop): update story template files to use CSF 3.0 format

dierat:
### Summary:
We've been moving to using CSF 3.0 format for our stories files. The big medium codebase did not, so most of the components are using CSF 2.0. I'd like to write a codemod or something to go through and update those, but, for now, I'm just updating the plop templates so new components generated by plop will at least start off in the right direction.

### Test Plan:
Generate a component, a recipe, and a page using `npx plop`,
boot them all up in storybook,
verify they all show up as expected.


jinlee93 (approved):
Think this is exactly how the csf v3 conversion was done, thanks!

PR #1134, opened 2022-06-14
chore(release): 1.2.0

jinlee93:


jinlee93:
> Yikes, what's with all the test failures 🤔 They're failing at the install dependency step.

Reinstalled via yarn, rebuilt, and repushed and CI is continuing now, not sure what the problem was

dierat (approved):

PR #1137, opened 2022-06-15
fix(popover): close popover on escape

jinlee93:
### Summary:
- now closes popover on escape
  - tested in unit test
- also includes popover unit test snaps
### Test Plan:
- tests pass
- no visual regressions

[sc-201722]

### DIFF
--- a/src/components/Popover/Popover.test.tsx
+++ b/src/components/Popover/Popover.test.tsx
@@ -1,0 +1,36 @@
+import { generateSnapshots } from '@chanzuckerberg/story-utils';
+import { userEvent } from '@storybook/testing-library';
+import { composeStories } from '@storybook/testing-react';
+import { render, screen } from '@testing-library/react';
+import React from 'react';
+import * as PopoverStoryFile from './Popover.stories';
+
+const { Default } = composeStories(PopoverStoryFile);
+
+describe('<Popover />', () => {
+  generateSnapshots(PopoverStoryFile, {
+    // Need to open the popover for snapping
+    getElement: (wrapper) => {
+      userEvent.click(screen.getByRole('button'));
+      return wrapper.baseElement;
+    },
+  });
+
+  it('should close Popover via escape key', async () => {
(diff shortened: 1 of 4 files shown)


src/components/Popover/Popover.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+    render(<Default />);
+    userEvent.click(screen.getByRole('button'));
+    expect(screen.getByTestId('popover-test')).toBeInTheDocument();
+    expect(screen.getByTestId('popover-test')).toHaveAttribute(
+      'aria-hidden',
+      'false',
+    );
+    // Needed to focus inside the popover due to animation
+    userEvent.tab();
+    expect(screen.getByTestId('popover-test')).toHaveFocus();
```
jinlee93, line 29: Alternative could have `screen.getByTestId('popover-test').focus()` but I felt using `tab()` felt more user reflective

jinlee93: Fixed with `isActive` being `true` having Popover open to begin with

src/components/Popover/Popover.tsx
```diff
@@ -154,8 +154,8 @@ export const Popover = ({
    * Handle onKeyDown
    * 1) If escape button is struck, close the popover
```
dierat, line 155: "is struck" huh, never hear it described quite like that

src/components/Popover/Popover.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const { Default } = composeStories(PopoverStoryFile);
+PopoverStoryFile.default.args = {
+  isActive: true,
+};
+
+describe('<Popover />', () => {
+  generateSnapshots(PopoverStoryFile);
+
+  it('should close Popover via escape key', async () => {
```
dierat, line 16: Nice test, thanks!

dierat (approved):

PR #1118, opened 2022-06-08
feat(page-level-banner): add component

dierat:
### Summary:
Shortcut: https://app.shortcut.com/czi-edu/story/199935/banner-align-with-eds-in-figma
Figma: https://www.figma.com/file/Ff7hJoSvXX9hkAmIklnsT2/EDS---UI-Components-(Main)?node-id=2974%3A17980

Follow-up to https://github.com/chanzuckerberg/edu-design-system/pull/1114 In that PR, I'm updating the `Banner` styles to match the new designs a little more. As I explained there, I'm going to leave the `Banner` component to support existing usage (section banner inside the page) and make a new `PageLevelBanner` component for the new usage (top level banner only used at the top of the page). This PR adds that component.

This component is mostly copy/pasted from `Banner` with a few important changes:
- no `action` prop (this component does not support buttons, just links in the text)
- this banner has a border on the bottom but nothing on the other sides and no drop shadow (also has no bottom margin since there's no drop shadow to account for)
- no "neutral" variant
- no `orientation` prop (but I did keep the mobile design)
- no `isFlat` variant
- there's a little more horizontal padding
- there is no custom mobile view; I asked design if they wanted to keep the banner mobile view and they specifically said they want it to look like the desktop view

I chose to copy/paste the code instead of trying to find a way to share it between the 2 components because:
1) I think it would be annoyingly complicated
2) we're hoping to deprecate and remove the `Banner` entirely, and it would be nice to not have to detangle it from another component when that happens

#### Storybook
all the page level banner stories

#### Figma
the page level banner designs in figma

### Test Plan:
Storybook + chromatic.

PR #1138, opened 2022-06-15
chore(page-level-banner): clean up

dierat:
### Summary:
Follow-up to https://github.com/chanzuckerberg/edu-design-system/pull/1118 which had some feedback that I didn't address before merging.

### Test Plan:
No changes in storybook; this is all refactoring.

### DIFF
--- a/src/components/Banner/Banner.module.css
+++ b/src/components/Banner/Banner.module.css
@@ -45,7 +45,7 @@
 }
 
 .banner__textAndAction,
-.banner__textContent {
+.banner__text-content {
   display: flex;
   flex-direction: column;
   justify-content: space-between;
--- a/src/components/Banner/Banner.module.css
+++ b/src/components/Banner/Banner.module.css
@@ -44,14 +44,10 @@
   --icon-size-default: var(--eds-size-5);
 }
 
-.banner__textAndAction,
-.banner__text-content {
+.banner__text-and-action {
   display: flex;
   flex-direction: column;
   justify-content: space-between;
-}
-
-.banner__textAndAction {
   gap: var(--eds-size-2);
 }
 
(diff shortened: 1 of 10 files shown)


src/components/PageLevelBanner/__snapshots__/PageLevelBanner.test.tsx.snap
```diff
@@ -25,24 +25,11 @@ exports[`<PageLeveBanner /> Brand story renders snapshot 1`] = `
...
-      Summit Learning has a full-time team dedicated to constantly improving our curriculum. To see the updates,
-       
-      <button
-        class="clickable-style button button--link clickable-style--link clickable-style--brand"
-        type="button"
-      >
-        click into the course
-      </button>
-      .
-    </p>
```
dierat, line 40: Looks like the change in `.stories` is messing up the `description` prop value generation logic 🤔 cc @jinlee93

dierat: Actually we don't need that `getDescription` method anymore since the link will always has the "neutral" status.

src/components/Banner/Banner.module.css
```diff
@@ -44,14 +44,10 @@
   --icon-size-default: var(--eds-size-5);
 }
 
-.banner__textAndAction,
-.banner__textContent {
```
dierat, line 48: Turns out `banner__textContent` isn't being used anywhere anymore anyway.

src/components/PageLevelBanner/PageLevelBanner.stories.tsx
```diff
@@ -1,111 +1,94 @@
...
+        </Button>
+        .
+      </>
+    ),
   },
-};
-
-type Args = React.ComponentProps<typeof PageLeveBanner>;
+  render: ({ variant, ...other }) => {
+    return <PageLevelBanner variant={variant} {...other} />;
```
jinlee93, line 29: Since `getDescription` is no longer being used, I don't think we need this render method anymore

jinlee93 (approved):
Lgtm 🚢

PR #1139, opened 2022-06-15
Ifrost/feedback overview scaffolding

ifrost1:
### Summary:
- Rename `ListDetail` to `TimelineNav`
- Use `NumberIcon` in the `TimelineNav`
- Scaffold out Feedback Overview page.
- Added `behavior="stacked"` to the `Table` to create the card-like tables on small screens
- Made the `NumberIcon` inherit colors for the icon to take on hover and active states for components like the `TimelineNav`

### Test Plan:
`yarn test` runs fine
`yarn types` runs fine
`yarn lint` runs fine
`yarn start` spins up


src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -59,9 +60,9 @@ export interface Props {
    */
```
dierat, line 60: Could you address the todo comment on the `items` prop? Especially since it says the array has a specific shape; would be nice to know what it is.

src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -22,7 +22,8 @@ import {
   D_ARROW_KEYCODE,
 } from '../../util/keycodes';
 import Icon from '../Icon';
-import ListDetailPanel, { ListDetailPanelVariant } from '../ListDetailPanel';
+import NumberIcon from '../NumberIcon';
+import TimelineNavPanel, { TimelineNavPanelVariant } from '../TimelineNavPanel';
 
 export interface Props {
   /**
```
dierat, line 29: Could you change `ariaLabelledBy` to `aria-labelledby`? It's our preferred pattern to use the aria api names when possible.

src/components/TimelineNav/TimelineNav.stories.tsx
```diff
@@ -2,22 +2,22 @@ import { BADGE } from '@geometricpanda/storybook-addon-badges';
 import { StoryObj, Meta } from '@storybook/react';
```
dierat, line 2: Could you add a `.test.tsx` file for this component that generates snapshots? I just realized we haven't been doing that when moving components over. Example file: https://github.com/chanzuckerberg/edu-design-system/blob/main/src/components/Banner/Banner.test.tsx

src/components/TimelineNav/TimelineNav.stories.tsx
```diff
@@ -110,13 +110,13 @@ export default {
...
       </>
     ),
   },
 } as Meta;
 
-type Args = React.ComponentProps<typeof ListDetail>;
+type Args = React.ComponentProps<typeof TimelineNav>;
 
 export const Default: StoryObj<Args> = {
   parameters: {
```
dierat, line 122: Does this axe rule still need to be disabled?

src/components/TimelineNav/TimelineNav.stories.tsx
ifrost1: Done. Thanks!

src/components/TimelineNav/TimelineNav.stories.tsx
ifrost1: Nope. Thanks!

dierat (approved):
Thanks for all the cleanup in these files! Much appreciated ❤️

ifrost1:
Thanks for the feedback!

PR #1140, opened 2022-06-15
chore: delete duplicate components in upcoming-components

dierat:
### Summary:
I happened to notice there are a few components that are in both the `src/upcoming-components` and `src/components` directories; they were copied over to `components` but not deleted (or only partially deleted) in `upcoming-components`.

### Test Plan:
Verify the components being deleted still have a version in `src/components`.


jinlee93 (approved):
This makes sense to me, thanks for the cleanup

PR #1141, opened 2022-06-15
test: add test files for all components with a .stories file

dierat:
### Summary:
I noticed that when we move a component from `upcoming-components` to `components`, we're not always making a test file to generate snapshots. This PR adds a `.test.tsx` file with snapshot testing to every component that has a `.stories` file. (A lot of components in the system are actually subcomponents and don't have their own stories; they rely on the parent for testing coverage.)

I also added test files to components in `upcoming-components` but commented out the actual test so we have a bunch of garbage tests maintaining components we're not using and may never be added to the system. People may still forget to uncomment these tests, but I figure it's easier to remember that because the file will already be there.

### Test Plan:
All the components in `src/components` with `.stories` files should now have snapshots 🙌


dierat:
https://app.shortcut.com/czi-edu/story/201758/add-snapshot-tests-for-components-with-a-stories-file

src/upcoming-components/Accordion/Accordion.test.tsx
```diff
@@ -0,0 +1,8 @@
+// import { generateSnapshots } from '@chanzuckerberg/story-utils';
+// import * as stories from './Accordion.stories';
+
+describe('<Accordion />', () => {
+  // TODO: uncomment this and delete the skipped test when moving this component into the main src/components directory
+  // generateSnapshots(stories);
+  test.skip('Workaround', () => undefined);
```
ahuth, line 7: Any reason to not snapshot it yet? Is it incomplete or something?

ahuth (approved):

src/upcoming-components/Accordion/Accordion.test.tsx
dierat: We talked about this in a meeting but I'll post it here for posterity – the components in `upcoming-components` are very much in a WIP state and some may never be used, so it feels like bloat to me to snapshot all of them and have to accept the changes when components they depend on change. But since we will have to remember to uncomment-out these test files when we start serious work on any of these components (and we often forget stuff like that), what we're going to do instead is exclude this whole directory from the test runners. That way we can have the fully functional test files, but they just won't run until the component is moved into the active `components` directory.

jinlee93 (approved):
Thanks for adding tests!

PR #1142, opened 2022-06-16
fix(cardwithnotification): re-export subcomponents directly

jinlee93:
### Summary:
- CardWithNotification indirect export causes component naming issues:
image

- Direct importing and re-exporting fixes component naming issues:
image

### Test Plan:
- names correctly represented when built
- unit tests pass
- no visual regressions


dierat (approved):
Awesome, thanks!

PR #1143, opened 2022-06-16
chore(timeline-nav): remove console log from testing

dierat:
### Summary:
While running tests, I noticed there's a `console.log` in `TimelineNav` that I'm guessing was for testing. This PR removes that logging.
code terminal with a bunch of really long text being logged by the TimelineNav component

### Test Plan:
Run `yarn test` and verify you don't see any logging.


dierat:
That type error is on `next` too.

jinlee93 (approved):
Thanks for removing the console.log, don't think it was intentional

PR #1144, opened 2022-06-16
docs: add guidelines for component comment

dierat:
### Summary:
A while back I updated the component comments to match our preferred pattern, but I didn't update the docs to explain this. This PR updates the documentation and also updates the component I referenced. I added that component comments should include example usage, which we've sometimes been doing but not always, so I'm wondering how people feel about adding that to the guidelines. I think it's useful and I find that sort of thing really helpful when reading documentation for other systems, but I also understand if people feel like it's overkill for simple components.

ButtonGroup in storybook in the docs tab with the component comment displayed
the documentation with the new section visible

### Test Plan:
Verify the `ButtonGroup` comment is showing properly in storybook and that the new documentation section appears as expected.

### DIFF
--- a/src/components/ButtonGroup/ButtonGroup.tsx
+++ b/src/components/ButtonGroup/ButtonGroup.tsx
@@ -15,10 +15,12 @@ type Props = {
   className?: string;
   /**
    * How much space there should be between the buttons.
+   * 'max' does not work with 'vertical' orientation.
    */
   spacing?: 'none' | '1x' | 'max';
   /**
    * Whether the buttons should be laid out horizontally or stacked vertically.
+   * 'vertical' does not work with 'max' spacing.
    */
   orientation?: 'horizontal' | 'vertical';
 };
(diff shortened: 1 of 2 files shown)


src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -15,10 +15,12 @@ type Props = {
   className?: string;
   /**
    * How much space there should be between the buttons.
+   * 'max' does not work with 'vertical' orientation.
```
dierat, line 18: When thinking about the documentation for this component, I realized we're not surfacing this unexpected behavior anywhere.

dierat:
https://app.shortcut.com/czi-edu/story/194027/add-import-example-to-component-docstrings-and-better-starter-docstring-in-plopfile

jinlee93 (approved):
Oh man, didn't see this until now, but thanks for adding updating docs and comments

dierat:
Wow I totally forgot about this one too. Thanks for finding and reviewing it! 😄

PR #1145, opened 2022-06-16
fix(table-object): fix behavior type updated in Table

dierat:
### Summary:
I'm seeing some type errors in the `TableObject` component because a prop type was updated in `Table` but the `TableObject` stories are using the old type values.

### Test Plan:
Run `yarn types` and verify there are no errors.


jinlee93 (approved):
This make sense

PR #1151, opened 2022-06-21
chore(media-block, feature, stacked-block): use image in codebase instead of external url in stories

dierat:
### Summary:
We have a couple of stories in the `upcoming-components` directory that are using external urls for images in their stories. This is frowned upon because those image are more likely to be flaky than images that are readily available within the codebase.

We already have some recipes using internal images that are stored in `.storybook/static/`. This PR reuses one of those images in the stories in `src`. There was a lint rule advising against this, but the circumstances around that are no longer relevant, so I removed it instead. (See comments on this PR for more information.)

### Pix
feature story in storybook with the compass illustration

media block story in storybook with the compass illustration

stacked block story in storybook with the compass illustration

### Test Plan:
Temporarily move the components to the `components` directory, then verify the image shows up correctly in the stories for all 3 components.


dierat:
https://app.shortcut.com/czi-edu/story/191080/add-placeholder-image-file-for-stories-that-need-an-img-src-url

src/upcoming-components/Feature/Feature.stories.tsx
```diff
@@ -5,6 +5,7 @@ import React from 'react';
 import { Feature, Props } from './Feature';
 import Heading from '../../components/Heading';
 import Text from '../../components/Text';
+import PlaceholderImage from '../../static/compass-solo.svg';
```
ahuth, line 8: > So one question is – should the static folder:
> - be at the top level, above src and .storybook
> - in src and the recipes reach into this folder (just like they pull components out of src)
> - be duplicated like this in both places

I think it's fine to leave it in .storybook/static, with the other stuff in there. Or can we not import from there for some reason?

dierat: Looks like there's a lint check against that:
```
/Users/drater/edu-design-system/src/upcoming-components/Feature/Feature.stories.tsx
  6:1  error  '../../../.storybook/static/compass-solo.svg' import is restricted from being used by a pattern. Do not import files from `.storybook/` into `src`. If needed, move files into `src/storyUtils` instead  no-restricted-imports
```
🤔

dierat: PR: https://github.com/chanzuckerberg/edu-design-system/pull/716
Shortcut: https://app.shortcut.com/czi-edu/story/172777/add-lint-error-against-importing-from-storybook-or-docs-into-src

jinlee93: I think generally the assets should be left in `.storybook/` for use in recipes and pages and components can have fpo blocks (ideally), but for this case would make sense to have actual assets. 

I vote for importing from `.storybook/` and using a eslint ignore for this case

ahuth: Interesting. Looks like that was introduced in #716, but the Slack thread with context is lost to the sands of time.

@anniehu4, do you remember why we added the no-restricted-imports item for `.storybook/`?

anniehu4: sorry I'm so late to this 😓  @dierat @jinlee93 @ahuth I did some hunting and I believe it was related to https://github.com/chanzuckerberg/edu-design-system/pull/713. Essentially:
- the way the build used to be set up, `tsc` would compile everything in `src/` (including stories)
- then we'd remove the stories in a post-processing step
- because of that, importing anything from `.storybook` into `src` would inadvertently also compile `.storybook`, which broke the output structure

However, now we have a separate TS config for building the project vs. local development. We only build components listed in `src/index.ts`. I think it would be fine to:
- get rid of the lint rule now
- keep these images in `.storybook/static`
- verify that `yarn build` outputs the same structure into `lib` (since that was the original problem)

dierat: Thanks @anniehu4 ! I imported the image into a component in the `components` directory and didn't see any change in structure in the `lib` directory after running `yarn build`, so I guess we're in the clear?

anniehu4: sgtm!

dierat:
Oops I left this on the back burner so long it burnt up. All 3 of the components I was updating have been removed because we dropped the whole `src/upcoming-components` directory rendering this PR moot. We could still remove the lint rule. There isn't a need to do so right now, but it might be useful to remove it now while we have the context around why it was created in the first place. I'm going to close this PR and archive the ticket, but let me know if anyone wants the lint rule removed and I'll do so separately.

PR #1156, opened 2022-06-23
chore(tabs): match mocks

jinlee93:
### Summary:
- matches tabs [mocks](https://www.figma.com/file/DEm9VhynEYAlx66ogEuskp/Tab-Navigation---Component-Exploration-%2B-Creation?node-id=3870%3A10986)
  - adjusts tabs font
  - adjusts active indicator box size
  - adds scroll indicators

https://user-images.githubusercontent.com/86632227/175191559-76d83b5c-e32c-412f-9d15-c9b3f2d15883.mov

### Test Plan:
- unit tests pass
- no visual regression except those using tabs

[sc-202249]

### DIFF
--- a/src/components/Tabs/Tabs.module.css
+++ b/src/components/Tabs/Tabs.module.css
@@ -7,51 +7,6 @@
 /**
  * List of of links where each link toggles open associated information
  */
-.tabs {
-  position: relative;
-}
-
-/**
- * Fade scrollable indicators to display if there is scrollable area on either side.
- * 1) Indicators are non interactable and should be able to click tab under them.
- */
-.tabs--scrollable-left::before,
-.tabs--scrollable-right::after {
-  pointer-events: none; /* 1 */
-  content: '';
-  position: absolute;
-  top: 0;
-  width: var(--eds-size-8);
-
-  height: 2.75rem;
-  @media all and (max-width: $eds-bp-md) {
-    height: 2.5rem;
-  }
-}
-
-/**
- * Left fade scrollable indicator.
- */
-.tabs--scrollable-left::before {
-  left: 0;
-  background-image: linear-gradient(
-    to right,
-    var(--eds-theme-color-background-neutral-default),
-    transparent 90%
-  );
-}
-
-/**
- * Right fade scrollable indicator.
- */
-.tabs--scrollable-right::after {
-  right: 0;
-  background-image: linear-gradient(
-    to left,
-    var(--eds-theme-color-background-neutral-default),
-    transparent 90%
-  );
-}
 
 /**
  * Inverted variant.
@@ -66,14 +21,42 @@
  * 2) Tab overflow behavior
  */
 .tabs__header {
-  width: 100%;
-  overflow: hidden;
   overflow-x: auto;
   position: relative;
 
   .tabs--inverted & {
     background: none;
   }
+}
+
+/**
+ * Fade scrollable indicators to display if there is scrollable area on either side.
+ * 1) The color "white" is arbitrary and any non transparent color can be used here.
+ */
+.tabs--scrollable-left {
+  -webkit-mask-image: -webkit-linear-gradient(
+    left,
+    transparent,
+    white var(--eds-size-8) /* 1 */
+  );
+}
+
+.tabs--scrollable-right {
+  -webkit-mask-image: -webkit-linear-gradient(
+    right,
+    transparent,
+    white var(--eds-size-8) /* 1 */
+  );
+}
+
+.tabs--scrollable-left.tabs--scrollable-right {
+  -webkit-mask-image: -webkit-linear-gradient(
+    left,
+    transparent,
+    /* 1 */ white var(--eds-size-8),
+    white calc(100% - var(--eds-size-8)),
+    transparent 100%
+  );
 }
 
 /**
--- a/src/components/Tabs/Tabs.module.css
+++ b/src/components/Tabs/Tabs.module.css
@@ -40,7 +40,7 @@
   @mixin reset;
   list-style: none;
   display: flex; /* 2 */
-  @mixin eds-theme-typography-body-text-md;
+  @mixin eds-theme-typography-tab;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/components/Tabs/Tabs.module.css
```diff
@@ -5,12 +5,57 @@
...
+ */
+.tabs--scrollable-left::before,
+.tabs--scrollable-right::after {
+  pointer-events: none; /* 1 */
+  content: '';
+  position: absolute;
+  top: 0;
+  width: var(--eds-size-8);
+
+  height: 2.75rem;
```
jinlee93, line 26: Height needs to be fixed since `content: ''`. Alternative is to have a `calc(padding + text-size * line-height multiplier)` but 1) text sizes and line-height multipliers are not css variables/tokens anyways and 2) figured that would be more confusing than having fixed height.

jinlee93: Is it possible to have `--current-font-size` and `--current-line-height` declared within typography tokens for use in cases like this, or would that be too niche?

src/components/Tabs/Tabs.stories.tsx
```diff
@@ -40,7 +40,79 @@ export default {
...
+            <p>
+              Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
+              eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut
+              enim ad minim veniam, quis nostrud exercitation ullamco laboris
+              nisi ut aliquip ex{' '}
+            </p>
+          </Text>
+        </Tab>
+
+        <Tab title="Tab Title 9">
```
jinlee93, line 115: added more tab components in the tabs to display scroll functionality. Maybe looks cleaner with some array looping but this is more explicit especially with how the `Tab` component works

src/components/Tabs/Tabs.tsx
```diff
@@ -181,6 +185,48 @@ export const Tabs = ({
...
+  /**
+   * Listens for window resize to display scroll fade indicators.
+   */
+  useEffect(() => {
+    if (headerRef && headerRef.current) {
+      const resizeHandleTabs = () => {
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        handleTabsScroll(headerRef.current!);
+      };
+      resizeHandleTabs();
```
jinlee93, line 222: The event listener actually calls the callback once when initiated, but the event listener is not triggered with prop changes so this line is required.
Sure it may be called twice on initial paint but that's better than it not being called at all.

dierat:
How do I select another tab using a keyboard? I only see a focus indicator on the active tab; the arrow keys scrolls the container of tabs and I don't know where the focus goes after clicking the "tab" key 🤔

dierat:
>How do I select another tab using a keyboard? I only see a focus indicator on the active tab; the arrow keys scrolls the container of tabs and I don't know where the focus goes after clicking the "tab" key 🤔

Looks like this is an existing issue; I'm seeing the same behavior on `next`. I guess there's an error in `onKeyDown`, probably a similar one to the `Popover` keyboard bug. Once that's fixed, presumably you would scroll the tabs via keyboard by using the arrow keys to select tabs that are off-screen or something.

src/components/Tabs/Tabs.tsx
dierat: Maybe add something to that effect as a comment? Tracking down bugs associated with react lifecycle event triggers can be pretty annoying.

jinlee93: Makes sense to me, will add

jinlee93:
> > How do I select another tab using a keyboard? I only see a focus indicator on the active tab; the arrow keys scrolls the container of tabs and I don't know where the focus goes after clicking the "tab" key 🤔
> 
> Looks like this is an existing issue; I'm seeing the same behavior on `next`. I guess there's an error in `onKeyDown`, probably a similar one to the `Popover` keyboard bug. Once that's fixed, presumably you would scroll the tabs via keyboard by using the arrow keys to select tabs that are off-screen or something.

Yea think that was an existing issue, as I didn't touch functionality outside of the fade indicators. I'll fix the arrow key to select tabs bug in this PR as well

src/components/Tabs/Tabs.module.css
```diff
@@ -5,12 +5,57 @@
...
+/**
+ * Left fade scrollable indicator.
+ */
+.tabs--scrollable-left::before {
+  left: 0;
+  background-image: linear-gradient(
+    to right,
+    var(--eds-theme-color-background-neutral-default),
+    transparent 90%
+  );
```
dierat, line 41: Another thing you could try is a mask with a linear gradient on the `tabs__list` (or the header). That way, it would fade into the background no matter what the background color is behind the tabs, not just on white backgrounds. https://stackoverflow.com/questions/60996982/css-opacity-gradient-top-to-bottom-without-color (Feel free to ignore this comment or make a separate task if it feels like too much work right now; it's not blocking.)

jinlee93: I think using mask is a more valid approach. Will try it, thanks!

jinlee93:
> Looks like this is an existing issue; I'm seeing the same behavior on next. I guess there's an error in onKeyDown, probably a similar one to the Popover keyboard bug. Once that's fixed, presumably you would scroll the tabs via keyboard by using the arrow keys to select tabs that are off-screen or something.

> Yea think that was an existing issue, as I didn't touch functionality outside of the fade indicators. I'll fix the arrow key to select tabs bug in this PR as well

Fixed and added test for it

src/components/Tabs/Tabs.test.tsx
```diff
@@ -1,6 +1,35 @@
...
+import { composeStories } from '@storybook/testing-react';
+import { render, screen } from '@testing-library/react';
+import React from 'react';
 import * as stories from './Tabs.stories';
 
+const { Default } = composeStories(stories);
+
 describe('<Tabs />', () => {
   generateSnapshots(stories);
+  it('should focus and select with keyboard controls', () => {
```
dierat, line 12: This is great, thanks!

dierat (approved):
Thanks for fixing the functionality while you were in here!

src/components/Tabs/Tabs.module.css
jinlee93: Definitely looks a lot better, thanks!


https://user-images.githubusercontent.com/86632227/175438883-5eb81d9b-1c4d-4b9e-81c2-dfadc9f0f87f.mov


Now there is a visual bug with the mask covering the scroll bar as well, but I'm going to file a separate [backlog ticket](https://app.shortcut.com/czi-edu/story/202962) for it 

Screen Shot 2022-06-23 at 4 51 02 PM
Screen Shot 2022-06-23 at 5 47 57 PM

jinlee93: Yup will do!

PR #1157, opened 2022-06-23
feat(timeline nav): add responsive behavior

csnizik:
### Test Plan:
https://app.shortcut.com/czi-edu/story/190151/timeline-nav-build-responsive-behavior


src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -366,14 +379,38 @@ export const TimelineNav = ({
...
+      <div
+        className={clsx(
+          styles['timeline-nav__body'],
+          isActive && styles['eds-is-active'],
+        )}
+      >
+        <Button
+          className={clsx(styles['timeline-nav__back'])}
+          onClick={onClick}
+          size="sm"
```
dierat, line 407: The `link` variant is unaffected by the `size` prop – its size is inherited.

src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -366,14 +379,38 @@ export const TimelineNav = ({
...
-                  {tab.props.title}
+                  <div className={clsx(styles['timeline-nav__link-title'])}>
+                    {tab.props.title}
+                  </div>
+
+                  <Icon
+                    className={clsx(styles['timeline-nav__link-arrow'])}
+                    name="arrow-forward"
+                    purpose="decorative"
+                    size="20px"
```
dierat, line 390: Should we be using a size variable for this?

src/components/TimelineNav/TimelineNav.module.css
```diff
@@ -8,14 +8,33 @@
  * 1) List of of links where each link toggles open associated information
  */
 .timeline-nav {
+  display: flex;
+  gap: var(--eds-size-2);
+  width: 100%;
+  overflow: hidden;
+  transition: var(--eds-anim-fade-long) var(--eds-anim-ease);
```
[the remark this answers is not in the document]
dierat: Can you add
```
  @media screen and (prefers-reduced-motion) {
    transition: none;
  }
```
here please?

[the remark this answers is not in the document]
csnizik: Good catch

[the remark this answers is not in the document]
dierat: But whats being transitioned here? Can you list it out? I prefer it to be explicit.

src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -366,14 +379,37 @@ export const TimelineNav = ({
...
-                  {tab.props.title}
+                  <div className={clsx(styles['timeline-nav__link-title'])}>
+                    {tab.props.title}
+                  </div>
+
+                  <Icon
+                    className={clsx(styles['timeline-nav__link-arrow'])}
+                    name="arrow-forward"
+                    purpose="decorative"
+                    size="var(--eds-size-2-and-half)"
```
dierat, line 390: Nice!

src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -366,14 +379,37 @@ export const TimelineNav = ({
...
                   </div>
-                  {tab.props.title}
+                  <div className={clsx(styles['timeline-nav__link-title'])}>
+                    {tab.props.title}
+                  </div>
+
+                  <Icon
+                    className={clsx(styles['timeline-nav__link-arrow'])}
+                    name="arrow-forward"
+                    purpose="decorative"
```
dierat, line 389: Is this decorative? I feel like the arrows do have a meaning – that clicking on the row will switch the view to the main content panel.

src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -318,13 +326,19 @@ export const TimelineNav = ({
     }
   };
 
+  const onClick = () => setIsActive(false);
```
dierat, line 329: I think the keyboard focus is getting lost after switching views. We should manually place the focus on the first interactive element on the new view each time.

dierat (changes requested):
Mostly requesting changes for the keyboard accessibility piece.

src/components/TimelineNav/TimelineNav.tsx
csnizik: Good catch!

src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -247,7 +250,12 @@ export const TimelineNav = ({
       return child;
```
dierat, line 250: I can't leave a comment up there but can you change the `onKeyDown` method to use `e.key` instead of `e.code`? The keyboard navigation down the list isn't working and I think that's why (this has come up a couple times).

src/components/TimelineNav/TimelineNav.tsx
csnizik: good catch! focus now remains on the last selected nav item

src/components/TimelineNav/TimelineNav.stories.tsx
```diff
@@ -52,9 +52,17 @@ export default {
           </Text>
         </TimelineNavPanel>
 
-        <TimelineNavPanel title="TimelineNavPanel 3" variant="warning">
+        <TimelineNavPanel
```
dierat, line 55: Thanks for adding more robust stories!

src/components/TimelineNav/TimelineNav.tsx
dierat: We want it to go the other way as well – when you click on a timeline item, it switches to the panel and sets the focus on the back button.

src/components/TimelineNav/TimelineNav.tsx
csnizik: Done

src/components/TimelineNav/TimelineNav.tsx
dierat: This is still unresolved.

dierat (approved):
Thanks for all the improvements here! I've been testing it as you've been working on it, and it's looking really nice 🎉 We can come back to the back button focus logic separately if needed.

src/components/TimelineNav/TimelineNav.tsx
csnizik: This is done but could use TD to understand why it only works when it's within a delayed callback.

ifrost1 (changes requested):
Sorry to harp on this. A couple more things:
- Focus ring seems to be peeking its head in from the side when using keyboard on small screens
Screen Shot 2022-06-24 at 5 39 51 PM

- The focus should be on the panel (which acts like a `Modal`). Wondering if you can just focus on the first focusable element (the back button). Wondering if we need similar things to the `Modal` (we had a `FocusLock` library that we used for modals) but this acts slightly differently. As long as this reads out properly with voiceover (`aria-labelledby` on the title maybe) and the focus moves to that panel then I think we're okay once that happens. Seems like locking the focus in is a bit too much since this is an in-canvas panel and not an overlay.

ifrost1 (approved):

PR #1158, opened 2022-06-24
chore: override transition for prefers reduced motion everywhere

dierat:
### Summary:
Most of the animations in the repo right now don't have an override for animations when the user has the OS accessibility feature "prefers reduced motion" turned on, so I went ahead and added those overrides. I also added a section about this to the documentation.

preview of section being added to documentation

### Test Plan:
I spot-checked a couple of the components to verify that the animation runs when the setting is turned off and does not run when the setting is turned on.

### DIFF
--- a/src/components/ShowHide/ShowHide.module.css
+++ b/src/components/ShowHide/ShowHide.module.css
@@ -12,8 +12,8 @@
   display: none;
 
   /**
-     * 1) Button that toggles open and closes a panel
-     */
+   * 1) Button that toggles open and closes a panel
+   */
   .show-hide.eds-is-active & {
     margin-top: var(--eds-size-1);
     display: block;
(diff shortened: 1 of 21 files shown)


src/components/ShowHide/ShowHide.module.css
```diff
@@ -12,39 +12,30 @@
...
     margin-top: var(--eds-size-1);
     display: block;
   }
 }
 
 /**
-   * Button icon within show hide component
-   * 1) Add transition for icon flip when activated
-   */
-.show-hide__btn .button__icon {
```
dierat, line 27: The `show-hide__btn`, `button__icon`, and `show-hide--sm` classes don't actually exist in the component.

ahuth (approved):
LGTM 🤘 

Thanks for adding these!

PR #1159, opened 2022-06-24
chore(show-hide, drawer): use e.key instead of e.code in keyboard events

dierat:
### Summary:
`event.code` for some reason doesn't work in our environment, but `event.key` does. There are 4 components using `e.code`:
- `TimelineNav` (being updated in https://github.com/chanzuckerberg/edu-design-system/pull/1157)
- `ShowHide` (being updated in this PR)
- `Drawer` (being updated in this PR)

### Test Plan:
Verify that the `ShowHide` component now closes when clicking the escape key while the component is open and you're focused on the trigger. (`Drawer` is not in storybook because it's in the `upcoming-components` directory, so I just tested `ShowHide` for now.)

### DIFF
--- a/src/components/ShowHide/ShowHide.tsx
+++ b/src/components/ShowHide/ShowHide.tsx
@@ -100,7 +100,7 @@ export const ShowHide = ({
    * 1) If escape button is struck, close the show hide contents
    */
   function handleOnKeyDown(e: KeyboardEvent<HTMLElement>) {
-    if (e.code === ESCAPE_KEYCODE) {
+    if (e.key === ESCAPE_KEYCODE) {
       setIsActive(false); /* 1 */
     }
   }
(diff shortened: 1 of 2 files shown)


src/components/ShowHide/ShowHide.tsx
```diff
@@ -100,7 +100,7 @@ export const ShowHide = ({
    * 1) If escape button is struck, close the show hide contents
    */
   function handleOnKeyDown(e: KeyboardEvent<HTMLElement>) {
-    if (e.code === ESCAPE_KEYCODE) {
+    if (e.key === ESCAPE_KEYCODE) {
```
ahuth, line 103: > doesn't work in our environment

Does this mean it didn't work in tests (maybe jsdom doesn't implement `code`)? Or did it not work in the browser (maybe React synthetic events don't implement it)?

In any case, I think `event.key` may be what we want anyway. If I'm reading MDN right:
- `code` represents the physical key, not the resulting character that key represents.
- `key` represents the character

ahuth (approved):
LGTM 🤘

PR #1160, opened 2022-06-24
chore(ci): build storybook separately from chromatic

jinlee93:
### Summary:
- build storybook in a step separate from chromatic
- should fix [ci errors](https://github.com/chanzuckerberg/edu-design-system/runs/6941477884?check_suite_focus=true) when chromatic thinks there are no differences in code and hence never builds storybook for axe testing
### Test Plan:
https://github.com/chanzuckerberg/edu-design-system/runs/7047367707?check_suite_focus=true
- check that storybook is built in separate step
- check that chromatic does not build storybook
- check that axe tests work

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -52,14 +52,12 @@ jobs:
         run: yarn install --immutable
       - name: Build design tokens
         run: yarn build:tokens
-      - name: Build Storybook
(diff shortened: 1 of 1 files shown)


ahuth (approved):
Nice, LGTM 🤘

PR #1161, opened 2022-06-27
Mgong/fix disabled a11y rules

meishangong:
### Summary:

Fixed disabled `jsx-a11y` rules
please see detail in story 
[https://app.shortcut.com/czi-edu/story/192829/fix-disabled-jsx-a11y-rules](https://app.shortcut.com/czi-edu/story/192829/fix-disabled-jsx-a11y-rules) and comment

### DIFF
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -1,3 +1,7 @@
+/**
+ * we need to keep the escape key close logic on line 190 <article>
+ * so keep the lint suppression for now and fix it later on.
+ */
 /* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
 import clsx from 'clsx';
 import React, {
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
 import clsx from 'clsx';
 import React, {
   useState,
@@ -182,25 +181,28 @@ export const Popover = ({
 
   return (
     <FocusLock disabled={!activeFocus}>
-      <article
-        aria-describedby={ariaDescribedBy}
+      <div
         aria-hidden={!isActive}
-        aria-labelledby={ariaLabelledBy}
         className={componentClassName}
         onKeyDown={(e) => handleOnKeyDown(e)}
-        ref={ref}
-        role="dialog"
-        // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+        role="button"
         tabIndex={0}
         {...other}
       >
-        <div className={styles['popover__content']}>
-          {header}
-          {body}
-          {footer}
-        </div>
-        <div className={styles['popover__arrow']}></div>
-      </article>
+        <article
+          aria-describedby={ariaDescribedBy}
+          aria-labelledby={ariaLabelledBy}
+          ref={ref}
+          role="dialog"
+        >
+          <div className={styles['popover__content']}>
+            {header}
+            {body}
+            {footer}
+          </div>
+          <div className={styles['popover__arrow']}></div>
+        </article>
+      </div>
     </FocusLock>
   );
 };
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
 import clsx from 'clsx';
 import React, {
   useState,
@@ -181,28 +182,25 @@ export const Popover = ({
 
   return (
     <FocusLock disabled={!activeFocus}>
-      <div
+      <article
+        aria-describedby={ariaDescribedBy}
         aria-hidden={!isActive}
+        aria-labelledby={ariaLabelledBy}
         className={componentClassName}
         onKeyDown={(e) => handleOnKeyDown(e)}
-        role="button"
+        ref={ref}
+        role="dialog"
+        // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
         tabIndex={0}
         {...other}
       >
-        <article
-          aria-describedby={ariaDescribedBy}
-          aria-labelledby={ariaLabelledBy}
-          ref={ref}
-          role="dialog"
-        >
-          <div className={styles['popover__content']}>
-            {header}
-            {body}
-            {footer}
-          </div>
-          <div className={styles['popover__arrow']}></div>
-        </article>
-      </div>
+        <div className={styles['popover__content']}>
+          {header}
+          {body}
+          {footer}
+        </div>
+        <div className={styles['popover__arrow']}></div>
+      </article>
     </FocusLock>
   );
 };
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -1,6 +1,6 @@
 /**
- * we need to keep the escape key close logic on line 190 <article>
- * so keep the lint suppression for now and fix it later on.
+ * We're intentionally adding keyboard interaction to an article element for better accessibility
+ * so we're comfortable suppressing this rule
  */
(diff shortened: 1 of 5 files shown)


meishangong:
https://app.shortcut.com/czi-edu/story/192829/fix-disabled-jsx-a11y-rules

src/upcoming-components/Drawer/Drawer.tsx
```diff
@@ -216,6 +215,8 @@ export const Drawer = ({
           onClick={handleOnClickOutside}
           onKeyDown={handleOnKeyDown}
           ref={ref}
+          role="button"
+          tabIndex={0}
           {...other}
```
meishangong, line 220: error: 
1.Static HTML elements with event handlers require a role.
2.Elements with the 'button' interactive role must be tabbable.
solution:
Added role and tabIndex

src/upcoming-components/Drawer/Drawer.tsx
```diff
@@ -225,7 +226,6 @@ export const Drawer = ({
             className={styles['drawer__window']}
             ref={windowRef}
             role="dialog"
-            tabIndex={0} // eslint-disable-line jsx-a11y/no-noninteractive-tabindex
           >
```
meishangong, line 229: error: `tabIndex` should only be declared on interactive elements.

because <article> is static element, so removed tabIndex, added it to parent interactive element

src/components/Popover/Popover.tsx
```diff
@@ -182,25 +181,28 @@ export const Popover = ({
...
+        >
+          <div className={styles['popover__content']}>
+            {header}
+            {body}
+            {footer}
+          </div>
+          <div className={styles['popover__arrow']}></div>
+        </article>
+      </div>
     </FocusLock>
```
meishangong, line 206: error: Non-interactive elements should not be assigned mouse or keyboard event listeners.

Since <article> is Non-interactive element, added <div> as parent element also added event listener to parent element. 
Structure is similar as `drawer.tsx` line 212 - 234

src/components/ShowHide/__snapshots__/ShowHide.test.tsx.snap
```diff
@@ -6,6 +6,8 @@ exports[`<ShowHide /> Default story renders snapshot 1`] = `
 >
   <div
     class="show-hide"
+    role="button"
+    tabindex="0"
   >
     <button
```
dierat, line 12: Looks like you have a button in a button here. Testing out the component in storybook, it looks like the escape key logic isn't working, and I feel like it's generally not that useful even if it did work. Popovers, tooltips, and modals need escape key logic, but I don't think I've ever seen it in a show/hide component like this one. I would just remove the `onKeyDown` logic entirely for this component. (All the code you're interacting with in this PR wasn't written by us but inherited from a collection of template components brought by the Big Medium group, so we haven't really looked at this stuff before to determine if we want/need it.)

I'll need some time to look over and think about the other components.

src/components/Popover/__snapshots__/Popover.test.tsx.snap
```diff
@@ -19,247 +19,251 @@ exports[`<Popover /> BottomLeft story renders snapshot 1`] = `
...
-          class="popover__body"
-        >
-          <ul
-            class="notification-list"
+                Notifications (4)
+              </h3>
+              <span
+                class="popover__header-title-after"
+              >
+                <button
```
dierat, line 52: Here's another case of a button in a button. This one definitely needs the escape key close logic. I'm looking around the web for advice, and I'm starting to think we should just leave the lint suppression comments 😞 I was hoping there was some role that we were missing that would solve our problem, but it might just be that these lint rule were designed to make sure people aren't using `div`s to make buttons (and forgetting to add `role="button"`) but these cases are exceptions. 🤔

src/components/ShowHide/__snapshots__/ShowHide.test.tsx.snap
ahuth: Agreed that the button within a button is no bueno.

I haven't looked at the UI or analyzed this too deeply, but if removing `onKeyDown` makes sense, I say do it 👍.

src/components/Popover/Popover.tsx
```diff
@@ -182,25 +181,28 @@ export const Popover = ({
...
-      <article
-        aria-describedby={ariaDescribedBy}
+      <div
         aria-hidden={!isActive}
-        aria-labelledby={ariaLabelledBy}
         className={componentClassName}
         onKeyDown={(e) => handleOnKeyDown(e)}
-        ref={ref}
-        role="dialog"
-        // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
```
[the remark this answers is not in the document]
ahuth: It might be fine to suppress the lint error, as long as the keyboard interaction is what we expect.

Is the idea of this component that the popover is sort of like a modal and you can tab onto the container?

src/components/Popover/__snapshots__/Popover.test.tsx.snap
ahuth: The lint rules are a tool we should generally follow, but it's totally fine to suppress them when they don't apply to our context.

I'm sure this rule was made to prevent people from applying random click handlers to divs when they should've had button.

src/components/ShowHide/__snapshots__/ShowHide.test.tsx.snap
meishangong: moved onKeyDown attribute to triggerElement

src/components/Popover/Popover.tsx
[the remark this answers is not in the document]
meishangong: > It might be fine to suppress the lint error, as long as the keyboard interaction is what we expect.
> 
> Is the idea of this component that the popover is sort of like a modal and you can tab onto the container?

Yes I think so.. yes I think keep the suppress the lint error is makes sense. I have reversed `Popover.tsx` file back.

src/components/Popover/__snapshots__/Popover.test.tsx.snap
meishangong: I have reversed `Popover.tsx`and snapshot back. to just leave the lint suppression comments.

src/components/Popover/Popover.tsx
```diff
@@ -1,3 +1,7 @@
+/**
+ * we need to keep the escape key close logic on line 190 <article>
+ * so keep the lint suppression for now and fix it later on.
```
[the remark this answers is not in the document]
dierat: Would it work if you added the suppression comment in the jsx above the `article` element? Like
```
  return (
    <FocusLock disabled={!activeFocus}>
      {/**
       * We're intentionally adding keyboard interaction to an article element for better accessibility,
       * so we're comfortable suppressing this rule
       */}
      {/* eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions */}
      <article
```

src/upcoming-components/Drawer/Drawer.tsx
```diff
@@ -216,6 +215,8 @@ export const Drawer = ({
           onClick={handleOnClickOutside}
           onKeyDown={handleOnKeyDown}
           ref={ref}
+          role="button"
+          tabIndex={0}
```
dierat, line 219: I don't think I want `role="button"` here either because the drawer will probably have buttons inside. How about we do the same thing as before and just suppress the interaction with a comment? For the `tabIndex`, I tested the component in storybook without it at all, and I actually thought it worked better. So I would just remove both lines and add the comments.

dierat:
Sorry for all the back-and-forth on this one; it's a really confusing ticket 😬

meishangong:
> Sorry for all the back-and-forth on this one; it's a really confusing ticket 😬

never mind, it is truly confusing and thanks for all suggestions

src/upcoming-components/Drawer/Drawer.tsx
meishangong: have updated, and  added comment also

ahuth (approved):
LGTM

PR #1162, opened 2022-06-28
chore(text): use neutral-subtle for neutral variant

dierat:
### Summary:
We're using the `neutral-default` color token for the "neutral" `Text` `variant`, but it should be `neutral-subtle`. (Not sure how this got out of sync; oh well, it happens.)

### ScreenyMcShotPants
This is a really subtle change, but you can see in the second screenshot that the neutral text is now visible in the third column because it's lighter than the token used for the background.

#### Before
before

#### After
after

### Test Plan:
Verify the "neutral" text is now using the `subtle-neutral` token in the `variants` `Text` story: http://localhost:9009/?path=/story/atoms-text-text--variants

### DIFF
--- a/src/components/Text/Text.module.css
+++ b/src/components/Text/Text.module.css
@@ -57,7 +57,7 @@
   color: inherit;
 }
 .text--neutral {
-  color: var(--eds-theme-color-text-neutral-default);
+  color: var(--eds-theme-color-text-neutral-subtle);
 }
 .text--success {
   color: var(--eds-theme-color-text-utility-success);


dierat:
https://app.shortcut.com/czi-edu/story/203349/update-text-neutral-variant-to-use-neutral-subtle-instead-of-neutral-default

anniehu4 (approved):
ScreenyMcShotPants 😂 

Should we also check what token Heading neutral should use? idk if design has a guideline there either

jinlee93 (approved):
Thanks for the fix!

dierat:
>Should we also check what token Heading neutral should use? idk if design has a guideline there either

I don't see a figma file that clearly lays this out, so I went to the `Modal` component (since that has both `Text` and `Heading`) and it has `neutral-default` for body text and `neutral-strong` for headings (which is what is currently in the system). `neutral-default`, `neutral-subtle`, and `neutral-strong` are all listed as text tokens in the tokens list and in figma. So, I think after this PR goes through, we'll be all set.

https://www.figma.com/file/IfMW
text color tokens in figma
AtWqNrQprImlco4pi0/EDS-Paper-v1.0-%7C-UI-Foundations?node-id=0%3A259

avniprasad (commented):
Could this change affect jest snapshots? Did it affect chromatic tests?

dierat:
> Could this change affect jest snapshots? Did it affect chromatic tests?

Not jest snapshots but it will have some chromatic effects for components that are using this variant (they'll show up in the PR where we update the package).

dierat:
That reminds me that I should double-check whether the `Banner` and `PageLevelBanner` should still be using this variant – those are the only EDS components using it right now.

dierat:
This PR is currently blocked by design – I want to get more clarity on exactly which color tokens are used by which `Text` and `Heading` variants so we don't end up switching back and forth.

dierat:
I talked to design and we're going to change the variant names to make it clearer that this is a lighter version of base, but for now I'll merge this and I can update the variant names later. Thanks all!

PR #1163, opened 2022-06-28
chore(storybook): convert to csf v3 and enable lint

jinlee93:
### Summary:
- converts more stories into csf v3
- enables eslint rule for csf v3
[sc-187754]
[sc-193401]

### Test Plan:
- no test failures
- no relevant visual regression


jinlee93:
Will follow up with converting more files to ts

jinlee93:
Interestingly, chromatic is now capturing some pages, wonder why they weren't captured before

.eslintrc
```diff
@@ -19,7 +19,7 @@
     "plugin:react/recommended",
     "plugin:storybook/recommended",
     "plugin:testing-library/react",
-    "plugin:@chanzuckerberg/stories/recommended"
+    "plugin:@chanzuckerberg/stories/strict"
```
ahuth, line 22: Let me know if you see any oddities with the no-csf-v2 and csf-v3-types rules.

ahuth (approved):
LGTM

.eslintrc
jinlee93: Sounds good, will do

PR #1166, opened 2022-06-29
chore(clickable-style): add data-bootstrap-override attributes

dierat:
### Summary:
I tested the new link styles in `traject` using an alpha build, and I noticed that bootstrap styles are adding an underline for `a` tags on hover and focus for "primary" and "secondary" variants, and that's also reducing the text-decoration thickness for "link" variants on hover and focus.

This PR adds data-bootstrap-override attributes for the different `ClickableStyle` variants so I can override these styles in `traject`.

### Test Plan:
There should be no visible changes in chromatic or storybook.

### DIFF
--- a/src/components/ClickableStyle/ClickableStyle.tsx
+++ b/src/components/ClickableStyle/ClickableStyle.tsx
@@ -152,7 +152,16 @@ export const ClickableStyle = React.forwardRef(
       fullWidth && styles['clickable-style--full-width'],
     );
 
-    return <Component className={componentClassName} ref={ref} {...other} />;
+    const dataAttribute = `clickable-style-${variant}`;
+
+    return (
+      <Component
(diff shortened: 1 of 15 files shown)


src/components/ClickableStyle/ClickableStyle.tsx
```diff
@@ -142,7 +142,16 @@ export const ClickableStyle = React.forwardRef(
...
+    const dataAttribute = `clickable-style-${variant}`;
+
+    return (
+      <Component
+        className={componentClassName}
+        data-bootstrap-override={dataAttribute}
+        ref={ref}
+        {...other}
+      />
+    );
```
dierat, line 154: This is the important change.

jinlee93 (approved):
Lgtm!

PR #1170, opened 2022-06-29
fix(modal): render tooltips on top of modal

jinlee93:
### Summary:
- Tooltips have a default z-index of 9999
- Tooltips need to render on top of modal
- Modal z-index shrank to 1050 to match [bootstrap modal z-index](https://github.com/FB-PLP/traject/blob/fe07779f6bca7920df27be8750dabf7517b34a74/app/assets/stylesheets/v2/_overrides/_bootswatch_variables.scss#L263-L276)
- Tooltips props extended with Tippy props (the base component) to allow custom z-index if 9999 is too high
- Created story to align more z-indexes https://app.shortcut.com/czi-edu/story/203871

### Test Plan:
- Unit tests pass
- No visual regression
- Manual test tooltip spawning above modal:
  - In Modal stories, hover over button 1

### DIFF
--- a/src/components/Modal/Modal.module.css
+++ b/src/components/Modal/Modal.module.css
@@ -26,7 +26,8 @@
 
 /**
  * The modal container which positions the modal in the center of the screen.
- * 1) Ensures modal is above other components.
+ * 1) Ensures modal is above other components. This is not a design token for now since we need to align on 
+ *    z-indeces across the app and in the platform. https://app.shortcut.com/czi-edu/story/203871
  */
 .modal {
   display: flex;
--- a/src/components/Modal/Modal.module.css
+++ b/src/components/Modal/Modal.module.css
@@ -34,7 +34,7 @@
   justify-content: center;
   padding: var(--eds-size-2);
 
-  z-index: var(--eds-z-index-top); /* 1 */
+  z-index: 1050; /* 1 */
 }
 
 /**
--- a/src/components/Modal/Modal.stories.tsx
+++ b/src/components/Modal/Modal.stories.tsx
@@ -4,7 +4,7 @@ import React, { ReactNode } from 'react';
 import { useState } from 'react';
 import Modal, { ModalContent } from './Modal';
 import styles from './Modal.stories.module.css';
-import { Button, ButtonGroup, Heading, Text } from '../../';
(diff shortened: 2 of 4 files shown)


src/components/Modal/Modal.stories.tsx
```diff
@@ -46,14 +46,17 @@ const getChildren = (
...
-        <Button
-          onClick={
-            () => {} /* eslint-disable-line @typescript-eslint/no-empty-function */
-          }
-          status="neutral"
-        >
-          Button 1
-        </Button>
+        {/* This has to be manually tested since Tooltip tests are flaky in Chromatic */}
+        <Tooltip text="Tooltip should spawn on top of modal">
```
anniehu4, line 50: ooh out of scope of this ticket but just had a thought -- if instead of passing `visible: true` to the tooltip stories, we used a play function to focus on them, would that help the flakiness?

src/components/Modal/Modal.module.css
```diff
@@ -34,7 +34,7 @@
   justify-content: center;
   padding: var(--eds-size-2);
 
-  z-index: var(--eds-z-index-top); /* 1 */
+  z-index: 1050; /* 1 */
```
anniehu4, line 37: can you leave a comment explaining why this doesn't use a design token for now?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -79,6 +79,7 @@ type Plugin = Plugins[number];
  * ```
  *
  * A styled tooltip built on Tippy.js.
+ * If used within the EDS Modal component, pass zIndex greater than 99999.
```
anniehu4, line 82: we can remove this comment since you reduced the modal z-index right?

anniehu4 (approved):

src/components/Modal/Modal.stories.tsx
jinlee93: I initially tried using play function to hover and capture the snap with chromatic, but I wasn't having a successful time. Could spend more time later on it tho

src/components/Modal/Modal.module.css
jinlee93: Yes, will do!

PR #1174, opened 2022-07-06
feat(DataTable, Table)

csnizik:
### Summary:
Adds a DataTable component that extends Table component's functionality; filters, row selection, sorting, expandable rows were originally in the scope of this ticket; all but the filter button were removed

### Test Plan:

- yarn lint passes
- yarn test passes
- yarn types passes
- yarn start works


ifrost1 (changes requested):
Just a brief review for now. Once things are more settled I can get into the details:
- Add in the TableObject into the components repo
- Use that and the Toolbar  component in the TableObjectHeader to get that layout you have in DataTable
- Add the sortable TableHeaderCell variant to render buttons inside of those cells to allow for the arrows to show for sorting. The actual sorting doesn’t need to work
- Adjust the `Feedback Overview` table variant style to be a more lined approach and less of a card approach on small screens to match Figma a bit better

csnizik:
Closing this PR; putting up new PR #1183 for same ticket

PR #1175, opened 2022-07-06
chore(clickable-style): downgrade error to warning for invalid combos

dierat:
### Summary:
We're now throwing an error in `Button` and `Link` components if an unsupported `status`/`variant` prop combination is being used. Unfortunately, there's an instance in `traject` that is using an invalid combination and it will take a little work to fix it. (The instance is in `Token`, which is being used in a few different places with different styling overrides, so I don't want to update it without taking the time to work out the styles and verify there are no regressions.) Ticket to fix this: https://app.shortcut.com/czi-edu/story/204850/fix-variant-in-token-in-traject

This PR downgrades the error to a console warning so we're still alerting devs that they're using the wrong combination but it doesn't break every page with a token on it.

I also added a default `as` to `ClickableStyle` because it's really annoying to debug rendering issues caused by not passing one in.

### Test Plan:
All automated tests should pass,
and the `Button` and `Link` chromatic snapshots should be unchanged.

### DIFF
--- a/src/components/ClickableStyle/ClickableStyle.tsx
+++ b/src/components/ClickableStyle/ClickableStyle.tsx
@@ -78,7 +78,8 @@ const throwInvalidPropComboError = (variant: Variant, status: Status) => {
     return `'${validStatuses[0]}' and '${validStatuses[1]}'`;
   };
 
-  throw new Error(
+  // TODO: change to `throw new Error()` when invalid combos have been removed from downstream product
+  console.warn(
     // We have to add the strings and variables together because string interpolation doesn't work in console messages.
     "\n*** Invalid prop combo ***:\n\nThe `Button` and `Link` components do not support using the '" +
       variant +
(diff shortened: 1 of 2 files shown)


jinlee93 (approved):
Makes sense to me, thanks!

src/components/ClickableStyle/ClickableStyle.tsx
```diff
@@ -78,7 +78,8 @@ const throwInvalidPropComboError = (variant: Variant, status: Status) => {
     return `'${validStatuses[0]}' and '${validStatuses[1]}'`;
   };
 
-  throw new Error(
+  // TODO: change to `throw new Error()` when invalid combos have been removed from downstream product
+  console.warn(
```
ahuth, line 82: Random idea... not sure if it's a good one or not...

Could add a `loose` prop that warns instead of throwing, so that the one instance can be less strict.

ahuth (commented):
LGTM

src/components/ClickableStyle/ClickableStyle.tsx
dierat: Ooo interesting! It would be nice to make sure more invalid instances don't sneak in the meantime 🤔

PR #1177, opened 2022-07-06
chore(link): lower spacing specificity

dierat:
### Summary:
In testing the new link styles in LP, I noticed that the margin is for some reason not allowing local styles to override it. Here I'm adding `:where` around those specific styles, which the button margin style already has. 

### Test Plan:
No changes in `Button` and `Link` storybook.

### DIFF
--- a/src/components/Banner/__snapshots__/Banner.test.tsx.snap
+++ b/src/components/Banner/__snapshots__/Banner.test.tsx.snap
@@ -34,7 +34,7 @@ exports[`<Banner /> Brand story renders snapshot 1`] = `
         Summit Learning has a full-time team dedicated to constantly improving our curriculum. To see the updates,
          
         <button
-          class="clickable-style button button--link clickable-style--link clickable-style--neutral"
+          class="clickable-style clickable-style--link clickable-style--neutral button button--link"
           data-bootstrap-override="clickable-style-link"
           type="button"
         >
@@ -52,7 +52,7 @@ exports[`<Banner /> BrandDismissable story renders snapshot 1`] = `
   class="banner banner--brand banner--horizontal banner--dismissable"
 >
   <button
-    class="clickable-style button banner__close-btn button--icon clickable-style--lg clickable-style--icon clickable-style--neutral"
+    class="clickable-style clickable-style--lg clickable-style--icon clickable-style--neutral button banner__close-btn button--icon"
     data-bootstrap-override="clickable-style-icon"
     type="button"
   >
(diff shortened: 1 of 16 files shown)


src/design-tokens/mixins.css
```diff
@@ -10,14 +10,11 @@
...
- * 5) Has position relative to support the background color pseudo element
- * 6) We hide overflow so the pseudo element can be pushed out of site
- * 7) z-index 0 required for the background to appear correctly on focus
- * 8) Has smaller icons than other button styles
- * 9) Make the focus outline invisible but don't remove it. High contrast mode will remove the background 
+ * 3) Has position relative to support the background color pseudo element
+ * 4) We hide overflow so the pseudo element can be pushed out of site
+ * 5) z-index 0 required for the background to appear correctly on focus
+ * 6) Has smaller icons than other button styles
+ * 7) Make the focus outline invisible but don't remove it. High contrast mode will remove the background
```
dierat, line 17: Most of the changes in this file are just the numbers being updated since 3 and 4 were moved to the `ClickableStyle` file.

jinlee93:
> In testing the new link styles in LP, I noticed that the margin is for some reason not allowing local styles to override it. Here I'm adding :where around those specific styles, which the button margin style already has.

Can we also make append `className` last in the order of styling in `ClickableStyle.tsx`?

```tsx
  // Base styles
  styles['clickable-style'],
  // Sizes
  variant !== 'link' && [
    size === 'sm' && styles['clickable-style--sm'],
    size === 'md' && styles['clickable-style--md'],
    size === 'lg' && styles['clickable-style--lg'],
  ],
  // Variants
  variant === 'secondary' && styles['clickable-style--secondary'],
  variant === 'icon' && styles['clickable-style--icon'],
  variant === 'link' && styles['clickable-style--link'],
  // Colors
  status === 'brand' && styles['clickable-style--brand'],
  status === 'neutral' && styles['clickable-style--neutral'],
  status === 'success' && styles['clickable-style--success'],
  status === 'warning' && styles['clickable-style--warning'],
  status === 'error' && styles['clickable-style--error'],
  // Other options
  fullWidth && styles['clickable-style--full-width'],
  // Append last for higher specificity
  className,
);
```

dierat:
> Can we also make append `className` last in the order of styling in `ClickableStyle.tsx`?

Yeah that makes sense. I've noticed we're kind of inconsistent about that and am wondering if we should standardize it.

src/components/Banner/__snapshots__/Banner.test.tsx.snap
```diff
@@ -34,7 +34,7 @@ exports[`<Banner /> Brand story renders snapshot 1`] = `
         Summit Learning has a full-time team dedicated to constantly improving our curriculum. To see the updates,
          
         <button
-          class="clickable-style button button--link clickable-style--link clickable-style--neutral"
+          class="clickable-style clickable-style--link clickable-style--neutral button button--link"
```
dierat, line 37: These snapshot changes are from moving the `className` prop to the bottom of the class list.

dierat:
> > Yeah that makes sense. I've noticed we're kind of inconsistent about that and am wondering if we should standardize it.
> 
> Yea I think we would want TW and other parent class styling to generally override the component one.

Ticket: https://app.shortcut.com/czi-edu/story/204928/make-classname-prop-the-last-class-in-list-for-all-components

PR #1178, opened 2022-07-07
Diedra/theming/test

dierat:
### Summary:
Theming logic comes from this slack thread: https://czi-edu.slack.com/archives/C0310HJ9WEN/p1657047516904539

I split the tier 1 tokens into 2 directories:
- `learningPlatform` (same tokens we had, just moved)
- `along` (mostly the same tokens but I changed the colors to be more saturated and bumped the border radii just so we could tell one theme from another)


webpack.config.js
```diff
@@ -0,0 +1,132 @@
+/**
```
dierat, line 1: https://czi-edu.slack.com/archives/C0310HJ9WEN/p1657052534744609?thread_ts=1657047516.904539&cid=C0310HJ9WEN

src/design-tokens/tier-1-definitions/learningPlatform/breakpoints.js
```diff
@@ -0,0 +1,17 @@
+//------------------------------------//
```
dierat, line 1: Same as LP.

src/design-tokens/tier-1-definitions/learningPlatform/typography-presets.css
```diff
@@ -0,0 +1,219 @@
+/* Updates to font-size and line-height in this file
```
dierat, line 1: Same as LP.

src/design-tokens/tier-1-definitions/learningPlatform/z-index.json
```diff
@@ -0,0 +1,30 @@
+{
```
dierat, line 1: Same as LP.

webpack.config.js
ahuth: I don't see what this Webpack config has to do with theming 🤔 

Looks like it's a generic config to process
- Sass
- Icon.js
- SVGs

.storybook/preview.tsx
```diff
@@ -13,15 +13,57 @@ import './css/styleguide-only.css';
...
+
+  if (!document.getElementById('ds-styles')) {
+    const style = document.createElement('style');
+    style.id = 'ds-styles';
+    document.head.appendChild(style);
+  }
+
+  // Add the CSS in to the head
+  // @ts-ignore
+  document.getElementById('ds-styles').innerHTML = css;
```
ahuth, line 48: Looks like we're:
- Generating a different CSS file for each theme
- Swapping out the content of a style tag here when changing theme

While that works, I think a better approach would be to use CSS variables. Something along the lines of https://lukelowrey.com/css-variable-theme-switcher/, where:
- Use CSS variables for tokens we want to be themable
- Set the default values on `:root`, which we do anyway
- Now define new values for the variables anywhere we want to theme them

This has a couple benefits:
- Much simpler. No need to switch between multiple stylesheets.
- The application (such as LP) can define the theme. This repo doesn't necessarily need to be aware of what the themes are.
- Uses web standards

dierat:
We're no longer entertaining this direction for theming.

PR #1183, opened 2022-07-08
feat(Table)

csnizik:
### Test Plan:

- yarn lint passes
- yarn test passes
- yarn types passes
- yarn start works


.storybook/pages/FeedbackOverview/FeedbackOverview.tsx
```diff
@@ -51,7 +51,6 @@ export const FeedbackOverview: React.FC = () => (
               <div className="fpo u-margin-bottom-xl">Progress bar</div>
               <div className="fpo u-margin-bottom-lg">Filters</div>
               <Table
-                behavior="stacked"
```
ifrost1, line 54: I'm thinking of using `stacked` here but re-styling this to match closer to this. I know Sean and I were discussing potentially changing up their comps with this a bit: https://www.figma.com/file/hDfTdVzAAuf2DDb3hVSnFM/Notebooks%3A-Feedback-Overview?node-id=257%3A36353

Right now we have a card treatment for this. Let's change it to only an underline for that. The actual data with the table and all of that can be done after this gets merged in

src/components/TableHeaderCell/TableHeaderCell.tsx
```diff
@@ -82,7 +103,19 @@ export const TableHeaderCell = ({
       scope={scope}
       {...other}
     >
-      {children}
+      <div className={clsx(styles['table__header-cell--contents'])}>
+        {children}
+        {sortable && (
```
ifrost1, line 108: Let's wrap button around this whole table cell content. That way they have a bigger click target, etc.

```
{sortable ? (
<button>
 {children}
 <Icon>
 </button>
 ) : ( 
{children}
)
}
)}
```

src/components/TableHeaderCell/TableHeaderCell.tsx
```diff
@@ -65,12 +71,27 @@ export const TableHeaderCell = ({
...
+        styles['table__header-cell--descending'],
+      sortDirection && styles['table__header-cell--ascending']),
   );
 
+  const iconName =
+    sortDirection === 'ascending'
+      ? 'arrow-narrow-up'
+      : sortDirection === 'descending'
+      ? 'arrow-narrow-down'
+      : 'unfold-more';
```
ifrost1, line 86: I thought we decided against using this "unfold-more" icon for sortable table headers. Worth bringing up with the team

src/components/TableObject/TableObject.stories.tsx
```diff
@@ -0,0 +1,230 @@
...
+    axe: {
+      // TODO: re-enable when component is worked on
+      skip: true,
+    },
+    badges: [BADGE.BETA],
+  },
+} as Meta;
+
+const Template: Story<Props> = (args) => (
+  <TableObject>
```
ifrost1, line 34: Make sure onclick handlers are working on the button once that is in place. I think we want to keep some sort of pattern. Go from default -> ascending -> descending -> default when clicked. That establishes the pattern, and then the data would just be manipulated to take that form.

src/components/Table/Table.module.css
```diff
@@ -141,15 +141,46 @@
...
+      cursor: pointer; /* 3 */
+    }
+  }
+}
+
+/*
+* Table header cell contents
+* 1) A wrapper for the title and sort icon. Uses a flex row so that icon remains at the bottom when one or more titles breaks into multiple lines 
+*/
+.table__header-cell--contents {
```
ifrost1, line 180: Change this to `.table__header-cell-contents`. We only use the double dashes for modifiers.

src/components/Table/Table.module.css
```diff
@@ -141,15 +141,46 @@
...
 .table__header-cell {
   @mixin eds-theme-typography-label-sm;
   padding: var(--eds-size-1) var(--eds-size-1);
   text-align: left;
 
-  .table__body & {
-    padding: var(--eds-size-2) var(--eds-size-1);
-    @mixin eds-theme-typography-body-text-xs-bold;
+  /* 1 */
+  .table__header-cell--sort {
```
ifrost1, line 157: Change this to `table__header-cell-sort`. Although this will likely become the button around the text and the icon.

ifrost1 (changes requested):
This looks good overall. Few changes:
- Let's wrap the all of `children` in a `<button>` element with a class of `.table__header-cell-button` when this is a sortable table cell
- Let's create a consistent pattern for the arrows. Go from `default -> ascending -> descending` or whichever makes sense to them.
- Remove the double dashes for children's children. We use 1 dash for this since double dashes signify modifiers to the base class.
- Let's fix the shadows on the table object. We only want that when it is an overflow behavior on the table object. I can help with this since I wrote the logic around that stuff if you need it.

src/components/TableHeaderCell/TableHeaderCell.tsx
csnizik: I couldn't find anything in Figma discussing that. It can always be swapped out with a change order after deployment.

src/components/TableObject/TableObject.stories.tsx
csnizik: I added this logic, but right now the click handlers only affect the state of an individual `TableHeaderCell`. Downstream devs will need to implement all of the contextual logic, including resetting the sort statuses of all other `TableHeaderCell`s when a sort order is set on one of them. I didn't implement that here because it adds the complexity of wrapping components in a context and passing and manipulating state, etc.

src/components/Table/Table.module.css
csnizik: Done

src/components/TableHeaderCell/TableHeaderCell.tsx
```diff
@@ -82,7 +103,19 @@ export const TableHeaderCell = ({
       scope={scope}
       {...other}
     >
-      {children}
+      <div className={clsx(styles['table__header-cell--contents'])}>
```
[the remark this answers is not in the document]
csnizik: Done

src/components/TableHeaderCell/TableHeaderCell.tsx
csnizik: Also something I just noticed: when you click on the sortable table header and the icon cycles between up, down, and unfold more, it actually looks pretty tight, i.e. all 3 icons are using the same visible space

src/components/Table/Table.module.css
csnizik: Done

.storybook/pages/FeedbackOverview/FeedbackOverview.tsx
ifrost1: @csnizik Stacked is what I added for the small screen responsive "card" layout. We may want to use this for the layout in that Figma file that is more of `bottom-bordered` "table row" on small screens. So if you look at that Figma file it's just restyling these small screen styles to just have a bottom border for now to look more like that

src/components/TableObject/TableObject.stories.tsx
ifrost1: This is exactly the right thinking @csnizik .

src/components/TableHeaderCell/TableHeaderCell.tsx
ifrost1: Tight meaning "good" or tight meaning too close together? Let's tighten up styles by adding some space in between the icon and text as well as centering that icon.

csnizik: Tight as in good. Tight, tight tight!

Tight!

ifrost1 (approved):

PR #1184, opened 2022-07-08
Ifrost/204796 append classnames drag drop container

ifrost1:
### Summary:
- Adds `columnClassName` prop to the drag drop column to give teams flexibility around adding additional styles like large padding bottom on the column to accommodate bottom aligned dropdowns since the overflow is set to auto for scroll on small screens.
- Request from CAD team

### Test Plan:
- `yarn types` runs
- `yarn test` runs
- `yarn lint` runs without errors
- `yarn start` runs

### DIFF
--- a/src/components/DragDrop/DragDropTypes.ts
+++ b/src/components/DragDrop/DragDropTypes.ts
@@ -21,4 +21,5 @@ export interface ContainerType {
   itemIds: string[];
   header?: React.ReactNode;
   emptyContent?: React.ReactNode;
+  columnClassName?: string;
 }
--- a/src/components/DragDropContainer/DragDropContainer.tsx
+++ b/src/components/DragDropContainer/DragDropContainer.tsx
@@ -13,6 +13,11 @@ export interface Props {
    */
   className?: string;
   /**
+   * Column class names that can be appended to the component.
+   * 1) Used to add additional styles to the column
+   */
+  columnClassName?: string;
+  /**
    * Prop that will contain an id for each container and an array of itemIds that will be used on initial render
    */
   container: ContainerType;
@@ -34,6 +39,7 @@ export const DragDropContainer = ({
   container,
   items,
   emptyContent,
+  columnClassName,
 }: Props) => {
   const componentClassName = clsx(
     styles['drag-drop__container'],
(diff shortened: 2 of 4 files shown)


src/components/DragDrop/DragDropTypes.ts
```diff
@@ -21,4 +21,5 @@ export interface ContainerType {
   itemIds: string[];
   header?: React.ReactNode;
   emptyContent?: React.ReactNode;
+  columnClassName?: string;
```
csnizik, line 24: could we use string[] to allow for more than one to be added? (if so, the prop name might be changed to columnClassNames, since we use itemIds)

ifrost1: Good point. Thanks for this @csnizik

ifrost1: Made that change @csnizik . Thanks again!

csnizik (approved):

src/components/DragDropContainer/DragDropContainer.tsx
```diff
@@ -56,7 +62,9 @@ export const DragDropContainer = ({
         {(provided: DroppableProvided) =>
           items.length > 0 ? (
             <ol
-              className={styles['drag-drop__container-inner']}
+              className={`${styles['drag-drop__container-inner']} ${
+                container.columnClassName && container.columnClassName
+              }`}
```
jinlee93, line 66: nit: think we need to use `clsx` here and possibly extract the logic out into a variable for readability

ifrost1: Was on the fence about that. But if that's what you all want, I can do that

jinlee93: Yea otherwise, we get "undefined" as part of class name https://github.com/chanzuckerberg/edu-design-system/blob/264655eaef9ff57f70451f5ad02057080d34ff53/src/components/DragDrop/__snapshots__/DragDrop.test.tsx.snap#L32

ifrost1: Just pushed that up. Thanks!

jinlee93 (approved):
LGTM, thanks for the changes

PR #1190, opened 2022-07-14
chore(banner): clarify close button label

jinlee93:
### Summary:
- updates close button labels of banners to be clearer
### Test Plan:
- only banner snaps affected for aria label
- no visual regressions


jinlee93:
worked off `next` 🤦 will reset commits and re-raise to `next`

PR #1192, opened 2022-07-15
Csnizik/190341 clickable card recipe

csnizik:
### Summary:
https://app.shortcut.com/czi-edu/story/190341/base-card-recipe

Allows link text and link url to be passed to a card; the link text/url are visually hidden but still accessible to screen readers and via keyboard navigation. The entire card becomes clickable to follow the link, but users can still click on the card to select text or access shortcuts via right-clicking. 

### Test Plan:

- yarn lint passes
- yarn test passes
- yarn types passes
- yarn start works

### DIFF
--- a/.storybook/recipes/BaselineCard/BaselineCard.stories.tsx
+++ b/.storybook/recipes/BaselineCard/BaselineCard.stories.tsx
@@ -8,7 +8,7 @@ export default {
   component: BaselineCard,
   args: {
     label: 'Selection of Evidence',
-    body: "In this Focus Area, you explore the main question: How does the structures of organisms enable life's functions?",
+    body: `In this Focus Area, you explore the main question: How does the structures of organisms enable life's functions?`,
   },
 } as Meta<Args>;
 
@@ -17,15 +17,13 @@ type Args = React.ComponentProps<typeof BaselineCard>;
 export const Default: StoryObj<Args> = {};
 
 export const Clickable: StoryObj<Args> = {
-  args: {},
-  decorators: [
-    (Story) => (
-      // eslint-disable-next-line jsx-a11y/anchor-is-valid
-      <a href="#">
-        <Story />
-      </a>
-    ),
-  ],
+  args: {
+    label: 'Clickable card',
+    body: 'This card is linked and clickable. Clicking anywhere on the card will open a link in a new tab. Its link is visually hidden but is still focusable when tabbing. Text within the card can still be selected. Right-clicking within the card works.',
+    linkHref:
+      'https://css-tricks.com/block-links-the-search-for-a-perfect-solution/',
+    linkText: 'Block links article',
+  },
 };
 
 export const WithMetadata: StoryObj<Args> = {
--- a/.storybook/recipes/BaselineCard/BaselineCard.stories.tsx
+++ b/.storybook/recipes/BaselineCard/BaselineCard.stories.tsx
@@ -8,7 +8,7 @@ export default {
   component: BaselineCard,
   args: {
     label: 'Selection of Evidence',
-    body: `In this Focus Area, you explore the main question: How does the structures of organisms enable life's functions?`,
+    body: "In this Focus Area, you explore the main question: How does the structures of organisms enable life's functions?",
   },
 } as Meta<Args>;
 
@@ -20,9 +20,10 @@ export const Clickable: StoryObj<Args> = {
   args: {
     label: 'Clickable card',
     body: 'This card is linked and clickable. Clicking anywhere on the card will open a link in a new tab. Its link is visually hidden but is still focusable when tabbing. Text within the card can still be selected. Right-clicking within the card works.',
-    linkHref:
-      'https://css-tricks.com/block-links-the-search-for-a-perfect-solution/',
-    linkText: 'Block links article',
+    linkProps: {
+      href: 'https://css-tricks.com/block-links-the-search-for-a-perfect-solution/',
+      text: 'Block links article',
+    },
   },
 };
 
--- a/.storybook/recipes/BaselineCard/BaselineCard.stories.tsx
+++ b/.storybook/recipes/BaselineCard/BaselineCard.stories.tsx
@@ -16,6 +16,18 @@ type Args = React.ComponentProps<typeof BaselineCard>;
 
 export const Default: StoryObj<Args> = {};
 
+export const Clickable: StoryObj<Args> = {
+  args: {},
+  decorators: [
+    (Story) => (
+      // eslint-disable-next-line jsx-a11y/anchor-is-valid
+      <a href="#">
+        <Story />
+      </a>
+    ),
+  ],
+};
+
 export const WithMetadata: StoryObj<Args> = {
   args: {
     metadata: {
--- a/.storybook/recipes/BaselineCard/BaselineCard.tsx
+++ b/.storybook/recipes/BaselineCard/BaselineCard.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable jsx-a11y/anchor-is-valid */
 import clsx from 'clsx';
 import React, { MutableRefObject, useEffect, useRef } from 'react';
 import styles from './BaselineCard.module.css';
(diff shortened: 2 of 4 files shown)


.storybook/recipes/BaselineCard/BaselineCard.stories.tsx
```diff
@@ -8,14 +8,24 @@ export default {
   component: BaselineCard,
   args: {
     label: 'Selection of Evidence',
-    body: "In this Focus Area, you explore the main question: How does the structures of organisms enable life's functions?",
+    body: `In this Focus Area, you explore the main question: How does the structures of organisms enable life's functions?`,
```
dierat, line 11: Just curious – why did you make this change?

.storybook/recipes/BaselineCard/BaselineCard.tsx
```diff
@@ -49,64 +58,133 @@ export interface Props {
...
+   * 2) Attach click handler to the card.
+   * 3) Check to see if text is selected; if so, do not call .click() on the
+   * linkRef
+   * 4) Pass the click to the linkRef.
+   * 5) Cleanup.
+   *
+   * TODO: If the visible text of the card also contains links, these will
+   * need to have an event listener added to them to stop propagation of the
+   * click event; otherwise, users will not be able to click on these inner
+   * links.
```
dierat, line 88: Can you add a story with a link and button in it that shows an example of how to get it to work?

dierat: Or should we add a follow-up task for that?

.storybook/recipes/BaselineCard/BaselineCard.tsx
```diff
@@ -49,64 +58,133 @@ export interface Props {
...
+        <CardBody className={styles['baseline-card__body']}>
+          {body}
+          {` `}
+          {linkHref && (
+            <a
+              className={styles['baseline-card__link']}
+              href={linkHref}
+              ref={linkRef}
+              rel="noreferrer"
+              target="_blank"
```
dierat, line 135: Does the link always need to open in a new tab? Should that be based on props? Maybe instead of having `linkText` and `linkHref` props, you could have `linkProps` so devs can add in the `rel` and `target` if they want.

.storybook/recipes/BaselineCard/BaselineCard.tsx
```diff
@@ -49,64 +58,133 @@ export interface Props {
...
+
+    <div className={styles['baseline-card__container']} ref={cardRef}>
+      <Card className={componentClassName} {...other}>
+        <CardHeader className={styles['baseline-card__label']}>
+          {label}
+        </CardHeader>
+        <CardBody className={styles['baseline-card__body']}>
+          {body}
+          {` `}
+          {linkHref && (
```
dierat, line 129: Should this be
```suggestion
          {linkHref && linkText && (
```
? What happens currently if the user passes in `linkHref` without `linkText`?

.storybook/recipes/BaselineCard/BaselineCard.tsx
```diff
@@ -49,64 +58,133 @@ export interface Props {
...
+     */
+
+    <div className={styles['baseline-card__container']} ref={cardRef}>
+      <Card className={componentClassName} {...other}>
+        <CardHeader className={styles['baseline-card__label']}>
+          {label}
+        </CardHeader>
+        <CardBody className={styles['baseline-card__body']}>
+          {body}
+          {` `}
```
dierat, line 128: Should this be inside the check below? Do we still want this space even if there is no link?

.storybook/recipes/BaselineCard/BaselineCard.tsx
```diff
@@ -1,5 +1,6 @@
+/* eslint-disable jsx-a11y/anchor-is-valid */
```
dierat, line 1: Is it possible to associate this with a specific line of code or does it have to exempt the whole file?

.storybook/recipes/BaselineCard/BaselineCard.tsx
```diff
@@ -30,6 +31,14 @@ export interface Props {
    * CSS class names that can be appended to the component.
    */
   className?: string;
+  /**
+   * Clickable card: link URL
+   */
+  linkHref?: string;
+  /**
+   * Clickable card: link text
+   */
```
dierat, line 40: Maybe add that the link text will only be visible when focused by a keyboard.

dierat:
Really interesting solution. The way we did it in the previous `Card` component is we added an `onClick` handler to the wrapper and added guidance that there needed to be a link or button somewhere in the card children that had the equivalent effect for keyboard users. That one had much less code but I think it was harder to misuse since there was nothing stopping someone from passing in an `onClick` with no in-card equivalent behavior, whereas in your implementation the `linkHref` powers both the card click and the hidden link.
Based on this post: https://inclusive-components.design/cards/).
Implementation: https://github.com/FB-PLP/traject/blob/1c234866f58c1e410b35e81b1d1d8820e52d83f5/app/assets/javascripts/v2/core/EDSCandidates/Card/Card.tsx#L30.

dierat (changes requested):
Main things I'm concerned about is the possibility of `linkHref` without `linkText` and that the card always opens in a new tab.

.storybook/recipes/BaselineCard/BaselineCard.tsx
csnizik: I like your approach, thanks!

.storybook/recipes/BaselineCard/BaselineCard.tsx
csnizik: Good catch. Originally I had `linkText ? linkText : "Link"` but "Link" as link text is bad a11y... but I don't know what would make a good generic, catch-all default link text. Suggestions?

.storybook/recipes/BaselineCard/BaselineCard.tsx
csnizik: Good catch.

.storybook/recipes/BaselineCard/BaselineCard.tsx
csnizik: Good catch.

.storybook/recipes/BaselineCard/BaselineCard.tsx
csnizik: Nice, thank you.

.storybook/recipes/BaselineCard/BaselineCard.stories.tsx
csnizik: This was a relic from an earlier approach that I switched away from.

.storybook/recipes/BaselineCard/BaselineCard.tsx
csnizik: Do we know if it's a scenario that will likely occur (additional links within a clickable card)? I hesitated to add it b/c I felt like the recipe was getting a tad complex already.

dierat: I just looked around the Learning Platform product and the cards are all pretty simple, so maybe you could skip this and just put in the component comment that when the card itself is clickable and navigates to another page, it does not support links or buttons in the content of the card.

.storybook/recipes/BaselineCard/BaselineCard.tsx
dierat: I agree that's bad for accessibility and I think any default generic text would be bad – can we just make it required if there's a `linkHref`? And add that to the prop comments that both are required for the clickable version of the card to work.

.storybook/recipes/BaselineCard/BaselineCard.tsx
dierat: And we should also tell design to add guidance for other designers to not put links or buttons in a clickable card.

.storybook/recipes/BaselineCard/BaselineCard.tsx
```diff
@@ -44,69 +48,151 @@ export interface Props {
...
+/*
+ * Clickable card: link properties
+ * 1) Href and text are required attributes
+ * 2) Text will be used by screen readers; it will only be visible when the
+ * link is focused by a keyboard
+ * 3) Target attribute is optional
+ */
+export type LinkType = {
+  href: string /* 1 */;
+  text: string /* 2 */;
```
dierat, line 60: I'd like this to be a `React.Node` because if the link does open in a new tab, devs might want to add screen reader-only text like "opens in a new tab" and an icon or something like that. (Could also just put it in parenthesis but I feel like that's weird for sighted keyboard users.)

dierat (approved):

PR #1193, opened 2022-07-18
fix: optional meta in project card

ifrost1:
### Summary:
- Added conditional wrapper around `meta` in `ProjectCard` to make that optional for Emma
- Added story with this to show that this is possible

### Test Plan:
`yarn lint` runs through
`yarn test` runs through
`yarn types` runs through
`yarn start` runs


csnizik (approved):
Does exactly what it says on the tin.

dierat (approved):

PR #1194, opened 2022-07-18
Csnizik/baseline card link

csnizik:
### Summary: I missed a requested change in the previous PR for Clickable Card 
https://github.com/chanzuckerberg/edu-design-system/pull/1192#discussion_r923519379
> I'd like this to be a `React.Node` because if the link does open in a new tab, devs might want to add screen reader-only text like "opens in a new tab" and an icon or something like that. (Could also just put it in parenthesis but I feel like that's weird for sighted keyboard users.)


dierat (approved):
Awesome, thanks!

PR #1195, opened 2022-07-18
Csnizik/205790 timeline right slot

csnizik:
### Summary: adds a content slot to the right of the link title. 
**UPDATED** https://www.figma.com/file/hDfTdVzAAuf2DDb3hVSnFM/Notebooks%3A-Feedback-Overview?node-id=366%3A26162

Changed story to use dot icons (but any ReactNode could go there)

### Test Plan:
- yarn lint passes
- yarn test passes
- yarn types passes
- yarn start works

### DIFF
--- a/src/components/TimelineNav/TimelineNav.stories.tsx
+++ b/src/components/TimelineNav/TimelineNav.stories.tsx
@@ -34,7 +34,8 @@ export default {
           </Text>
         </TimelineNavPanel>
         <TimelineNavPanel
-          right={
+          title="TimelineNavPanel 1"
+          titleAfter={
             <Icon
               color={EdsThemeColorIconUtilitySuccess}
               name="circle-small"
@@ -43,7 +44,6 @@ export default {
               title="29 students have completed"
             />
           }
-          title="TimelineNavPanel 1"
           variant="number"
         >
           <Text as="div">
@@ -70,7 +70,10 @@ export default {
         </TimelineNavPanel>
 
         <TimelineNavPanel
-          right={
+          title="
+              Panel with a long name that breaks into multiple lines on smaller
+              viewports"
+          titleAfter={
             <Icon
               color={EdsThemeColorIconBrandPrimary}
               name="circle-small"
@@ -79,9 +82,6 @@ export default {
               title="2 students need feedback"
             />
           }
-          title="
-              Panel with a long name that breaks into multiple lines on smaller
-              viewports"
           variant="warning"
         >
           <Text as="div">
--- a/src/components/TimelineNav/TimelineNav.stories.tsx
+++ b/src/components/TimelineNav/TimelineNav.stories.tsx
@@ -3,6 +3,7 @@ import { StoryObj, Meta } from '@storybook/react';
 import React from 'react';
 
 import { TimelineNav } from './TimelineNav';
+import Tag from '../Tag';
 import Text from '../Text';
 import TimelineNavPanel from '../TimelineNavPanel';
 
@@ -28,7 +29,11 @@ export default {
             </p>
           </Text>
         </TimelineNavPanel>
-        <TimelineNavPanel title="TimelineNavPanel 1" variant="number">
+        <TimelineNavPanel
+          right={<Tag hasOutline text="Feedback" variant="brand" />}
+          title="TimelineNavPanel 1"
+          variant="number"
+        >
           <Text as="div">
             <h3>TimelineNavPanel 1</h3>
             <p>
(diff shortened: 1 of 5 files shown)


src/components/TimelineNavPanel/TimelineNavPanel.tsx
```diff
@@ -57,6 +61,7 @@ export const TimelineNavPanel = ({
   className,
   completed,
   id,
+  right,
```
dierat, line 64: Where is this being used?

src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -359,6 +360,9 @@ export const TimelineNav = ({
           {timelineNavItems().map((tab: TimelineNavItem, i: number) => {
             const isActive = activeIndexState === i;
             const itemVariant = variant && tab.props.variant;
+            {
+              console.log({ tab: tab });
+            }
```
dierat, line 365: Did you mean to leave this in?

src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -109,6 +109,7 @@ export const TimelineNav = ({
   overflow,
   onChange,
   required,
+  right,
```
dierat, line 112: Where is this being used?

src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -48,44 +52,40 @@ export interface Props {
...
-  items?: Array<any>;
+  overflow?: 'inverted';
   /**
    * Indicates that field is required for form to be successfully submitted
    */
   required?: boolean;
+  /**
+   * Right slot: content that will be positioned to the right of title
+   */
+  right?: ReactNode;
```
dierat, line 66: I'm wondering if there is a more specific prop name we could use here, like `rightElement` or something of that nature.

src/components/TimelineNav/TimelineNav.stories.tsx
```diff
@@ -28,7 +29,11 @@ export default {
             </p>
           </Text>
         </TimelineNavPanel>
-        <TimelineNavPanel title="TimelineNavPanel 1" variant="number">
+        <TimelineNavPanel
+          right={<Tag hasOutline text="Feedback" variant="brand" />}
```
dierat, line 33: Sounds like design landed on using a dot here: [figma spec](https://www.figma.com/file/hDfTdVzAAuf2DDb3hVSnFM/Notebooks%3A-Feedback-Overview?node-id=366%3A26162)

src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -31,6 +31,10 @@ export interface Props {
    * The aria-labelledby attribute creates a relationship between the tab list and the tab panels
    */
   'aria-labelledby'?: string;
+  /**
+   * Passed down to initially set the activeIndex state
+   */
+  activeIndex?: number;
```
dierat, line 37: Thanks for alphabetizing these! ❤️

src/components/TimelineNav/TimelineNav.tsx
csnizik: Oopsie lol

src/components/TimelineNavPanel/TimelineNavPanel.tsx
csnizik: Looks like all of `<TimelineNavPanel>`'s props get passed to `<TimelineNav>`. In `<TimelineNav>` the panels, props and children get collected by timelineNavItems(), which gets iterated over by a .map() method inside the return; the props then get applied to each <li>.

src/components/TimelineNav/TimelineNav.tsx
csnizik: See above.

src/components/TimelineNav/TimelineNav.tsx
csnizik: Good catch. The `<PageHeader>` component uses the same prop and calls it `titleAfter` so I changed this to match.

src/components/TimelineNavPanel/TimelineNavPanel.tsx
dierat: Could you add a quick comment, here and on the other one?

src/components/TimelineNav/TimelineNav.tsx
dierat: Awesome, thanks!

ifrost1 (approved):

PR #1198, opened 2022-07-20
fix: adjust table header cell padding table body

ifrost1:
### Summary:
- Change padding of table header cells within the table body (to align properly with regular table cells

### Test Plan:
`yarn lint` runs fine
`yarn test` runs fine
`yarn types` runs fine
`yarn start` runs fine


jinlee93 (approved):
Looks good to me

PR #1202, opened 2022-07-20
chore(modal): expose modal sizes

jinlee93:
### Summary:
- there have been a couple requests to make modal smaller
- this pr allows the setting of the modal size to not break larger
- keeps responsiveness
- [slack context](https://czi-edu.slack.com/archives/CTFV79JH4/p1658331665549089)
Screen Shot 2022-07-20 at 9 46 38 AM
Screen Shot 2022-07-20 at 9 46 40 AM
Screen Shot 2022-07-20 at 9 46 42 AM

### Test Plan:
- updated modal snaps
- no visual regressions on non affected chromatic tests
- new modal size stories captured in chromatic

### DIFF
--- a/src/components/Modal/Modal.module.css
+++ b/src/components/Modal/Modal.module.css
@@ -93,6 +93,10 @@
   --modal-horizontal-padding: var(--eds-size-4);
 }
 
+/**
+ * The medium modal size used for the md modal.
+ * Also used for the lg modal size for when the screen size is smaller than 75rem.
+ */
 .modal__content--md {
   @media all and (min-width: $eds-bp-md) {
     width: 42rem;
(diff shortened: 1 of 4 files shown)


src/components/Modal/Modal.module.css
```diff
@@ -91,10 +91,16 @@
 
   width: 22.5rem;
   --modal-horizontal-padding: var(--eds-size-4);
+}
+
+.modal__content--md {
```
dierat, line 96: For some reason I found the styles confusing because the large style includes medium but that's not obvious from the css module, so I would like a comment  here on the medium style that it's used for both medium and large, or maybe rename the class to make that clearer.

dierat:
Just verifying – did you mean to branch this off of `main`?

src/components/Modal/Modal.module.css
jinlee93: Mmm that makes sense, I'll leave a comment

dierat (approved):

PR #1208, opened 2022-07-22
feat(text, heading): allow all 3 neutral text colors via variant prop

dierat:
### Summary:
We have 3 different neutral color tokens for text:
- `neutral-subtle`
- `neutral-default`
- `neutral-strong`

Currently the `Text` component only supports `neutral-default` via the "base" variant and `neutral-subtle` via the "neutral" variant. The `Heading` component only supports `neutral-strong` via both the "base" and "neutral" variants.

I talked to design and they said all 3 tokens should be available to both components. I suggested that we replace the "base" and "neutral" variant options with 3 different options representing the 3 neutral tokens:
- "neutral-subtle"
- "neutral-medium"
- "neutral-strong"

I suggested using "medium" instead of "default" because the `neutral-strong` token is typically used for the `Heading` component, so it feels confusing to me to use "default" when "strong" is really the default.

#### Heading variants, before and after in storybook
Before version is on the left, after is on the right.
heading variants change in chromatic

#### Text variants, before and after in storybook
Before version is on the left, after is on the right.
text variants change in chromatic

### Test Plan:
Verify there are no chromatic changes outside of the `Text` and `Heading` components,
and verify the new `Text` and `Heading` variants appear as expected.

### DIFF
--- a/.storybook/pages/CoursePlannerEdit/CoursePlannerEdit.tsx
+++ b/.storybook/pages/CoursePlannerEdit/CoursePlannerEdit.tsx
@@ -340,7 +340,7 @@ export const CoursePlannerEdit = () => {
         <DragDropContainerHeader>
           <Toolbar className="u-margin-bottom-md" variant="bare">
             <ToolbarItem>
-              <Heading as="h2" size="title-sm" variant="base">
+              <Heading as="h2" size="title-sm" variant="neutralStrong">
                 Available projects
               </Heading>
             </ToolbarItem>
@@ -361,7 +361,7 @@ export const CoursePlannerEdit = () => {
         <DragDropContainerHeader>
           <Toolbar className="u-margin-bottom-md" variant="bare">
             <ToolbarItem>
-              <Heading as="h2" size="title-sm" variant="base">
+              <Heading as="h2" size="title-sm" variant="neutralStrong">
                 Planned projects
               </Heading>
             </ToolbarItem>
@@ -508,7 +508,7 @@ export const CoursePlannerEdit = () => {
                 as="h2"
                 className="u-margin-bottom-md"
                 size="headline-sm"
-                variant="base"
+                variant="neutralStrong"
               >
                 Select projects for your History 6 plan
               </Heading>
--- a/.storybook/pages/CoursePlannerEdit/CoursePlannerEdit.tsx
+++ b/.storybook/pages/CoursePlannerEdit/CoursePlannerEdit.tsx
@@ -340,7 +340,7 @@ export const CoursePlannerEdit = () => {
         <DragDropContainerHeader>
           <Toolbar className="u-margin-bottom-md" variant="bare">
             <ToolbarItem>
-              <Heading as="h2" size="title-sm" variant="neutralStrong">
+              <Heading as="h2" size="title-sm" variant="neutral-strong">
                 Available projects
               </Heading>
             </ToolbarItem>
@@ -361,7 +361,7 @@ export const CoursePlannerEdit = () => {
         <DragDropContainerHeader>
           <Toolbar className="u-margin-bottom-md" variant="bare">
             <ToolbarItem>
-              <Heading as="h2" size="title-sm" variant="neutralStrong">
+              <Heading as="h2" size="title-sm" variant="neutral-strong">
                 Planned projects
               </Heading>
             </ToolbarItem>
@@ -508,7 +508,7 @@ export const CoursePlannerEdit = () => {
                 as="h2"
                 className="u-margin-bottom-md"
                 size="headline-sm"
-                variant="neutralStrong"
+                variant="neutral-strong"
               >
                 Select projects for your History 6 plan
               </Heading>
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -139,6 +139,8 @@ export const Banner = ({
     isFlat && styles['banner--flat'],
   );
 
+  const variantComputed = variant === 'neutral' ? 'neutralStrong' : variant;
+
   return (
     <article className={componentClassName}>
       <Icon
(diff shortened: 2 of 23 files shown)


src/components/Heading/Heading.tsx
```diff
@@ -3,14 +3,15 @@ import React, { forwardRef } from 'react';
 import styles from './Heading.module.css';
 
 export const VARIANTS = [
-  'error',
-  'base',
+  'neutralSubtle',
+  'neutralMedium',
+  'neutralStrong',
```
dierat, line 8: I also reorganized the variants here and in the stories so they feel more organized.

.storybook/pages/CoursePlannerEdit/CoursePlannerEdit.tsx
```diff
@@ -340,7 +340,7 @@ export const CoursePlannerEdit = () => {
         <DragDropContainerHeader>
           <Toolbar className="u-margin-bottom-md" variant="bare">
             <ToolbarItem>
-              <Heading as="h2" size="title-sm" variant="base">
+              <Heading as="h2" size="title-sm" variant="neutralStrong">
```
dierat, line 343: The headings have changed from "base" to "neutralStrong".

src/components/Banner/Banner.tsx
```diff
@@ -166,12 +168,12 @@ export const Banner = ({
...
         <div>
           {title && (
-            <Heading as={titleAs} size="title-md" variant={variant}>
+            <Heading as={titleAs} size="title-md" variant={variantComputed}>
               {title}
             </Heading>
           )}
           {description && (
-            <Text as={descriptionAs} size="sm" variant="base">
+            <Text as={descriptionAs} size="sm" variant="neutralMedium">
```
dierat, line 176: Text has changed from "base" to "neutralMedium".

dierat:
https://app.shortcut.com/czi-edu/story/203349/update-text-neutral-variant-to-use-neutral-subtle-instead-of-neutral-default

src/components/PageHeader/PageHeader.tsx
```diff
@@ -42,10 +42,11 @@ export interface Props {
    */
   headingVariant?:
     | 'error'
-    | 'base'
     | 'brand'
     | 'inherit'
-    | 'neutral'
+    | 'neutralSubtle'
```
anniehu4, line 47: maybe outside scope of this PR, but could this import the `Variant` type from `Heading`?

src/components/Text/Text.tsx
```diff
@@ -14,14 +14,15 @@ export type Size =
   | 'callout';
 
 export type Variant =
-  | 'error'
-  | 'base'
+  | 'neutralSubtle'
```
anniehu4, line 17: could this use the `Variant` type from `Heading` also? (or define them in Text and import into Heading)

src/components/Heading/Heading.tsx
anniehu4: **blocking**: is it easier to keep these names kebab cased (e.g. `neutral-subtle`)? that's consistent with the multi-word size values and would mean we can get rid of the extra `variantComputed` code

dierat: Hmm I feel like we usually use camel case for props, like `className`, `onDismiss`, and `descriptionAs`. I feel like the sizes should be changed because I think that's the only place we use kebab in props.

dierat: I can't find any multi-word string prop values beyond the text sizes and kebab class names.

dierat: No wait I found some more:
- `position` for `ButtonDropdown`: 'top-left' | 'top-right' | 'bottom-right'
- `variant` for `Grid`: '2up' | 'side-by-side' | '1-3up' | '3up' | '4up' | '1-2-4up' | '1-2-1up'

That's a definite pattern. Will update. If this isn't in the documentation yet we should probably add it.

anniehu4: ooh nice digging, I didn't realize we had those. ya +1 adding to documentation 👍

src/components/PageHeader/PageHeader.tsx
dierat: Is it cool if I clean that up in a separate PR? I don't want to bloat this one with that change.

anniehu4: def!

src/components/Heading/Heading.tsx
```diff
@@ -92,12 +93,21 @@ export const Heading = forwardRef(
...
     }
 
+    let variantComputed: string | undefined = variant;
+    if (variant === 'neutral-subtle') {
+      variantComputed = 'neutral-subtle';
+    } else if (variant === 'neutral-medium') {
+      variantComputed = 'neutral-medium';
+    } else if (variant === 'neutral-strong') {
+      variantComputed = 'neutral-strong';
+    }
```
anniehu4, line 103: can we remove this and just use `variant`?

src/components/Text/Text.tsx
```diff
@@ -90,6 +91,15 @@ export const Text = forwardRef(
...
     ref: ForwardedRef<HTMLParagraphElement>, // Setting as HTMLParagraphElement to satisfy TS, but unit test covers both span and p cases for sanity
   ) => {
+    let variantComputed: string | undefined = variant;
+    if (variant === 'neutral-subtle') {
+      variantComputed = 'neutral-subtle';
+    } else if (variant === 'neutral-medium') {
+      variantComputed = 'neutral-medium';
+    } else if (variant === 'neutral-strong') {
+      variantComputed = 'neutral-strong';
+    }
```
anniehu4, line 101: ditto

anniehu4 (approved):

src/components/Heading/Heading.tsx
dierat: Oh yeah, thanks for pointing that out!

PR #1210, opened 2022-07-25
fix(banner): move close button to end of content

dierat:
### Summary:
The close button in our `Banner` and `PageLevelBanner` components comes before the content of the banners for screen reader users. That could be confusing because they don't know what the close button would actually be closing. The close button and content are contained in an `article` element, so they're semantically grouped, but screen reader users shouldn't have to read the content and then go *back* to the close button to remove the component. This PR moves the button to the end of the component, after the content. The appearance of the banner is unchanged.

#### Before
https://user-images.githubusercontent.com/7761701/180834055-f8a5b8fe-5942-4033-8aca-dd80aeb7eda5.mp4

#### After
https://user-images.githubusercontent.com/7761701/180834088-677469e6-a63e-4805-8156-f3f71ca01a8d.mp4

### Test Plan:
Verify there are no visual changes (except I moved the period in the banner content to be part of the link), and screen readers now hit the close button last, just before the end of the article.

### DIFF
--- a/src/components/Banner/Banner.stories.tsx
+++ b/src/components/Banner/Banner.stories.tsx
@@ -36,9 +36,8 @@ const getDescription = (status?: Variant) => (
       status="neutral"
       variant="link"
     >
-      click into the course
+      click into the course.
     </Button>
-    .
   </>
 );
 
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -141,21 +141,6 @@ export const Banner = ({
 
   return (
     <article className={componentClassName}>
-      {onDismiss && (
-        <Button
-          className={styles['banner__close-btn']}
-          onClick={onDismiss}
-          status="neutral"
-          variant="icon"
-        >
-          <Icon
-            name="close"
-            purpose="informative"
-            title="dismiss notification"
-          />
-        </Button>
-      )}
-
       <Icon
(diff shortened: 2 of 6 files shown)


src/components/Banner/Banner.stories.tsx
```diff
@@ -36,9 +36,8 @@ const getDescription = (status?: Variant) => (
       status="neutral"
       variant="link"
     >
-      click into the course
+      click into the course.
```
dierat, line 39: I noticed that `VoicerOver` was reading out the period as it's own thing because it was outside of the link content, so I moved it in.

src/components/Banner/Banner.tsx
```diff
@@ -181,6 +166,21 @@ export const Banner = ({
           <div className={clsx(styles['banner__action'])}>{action}</div>
         )}
       </div>
+
+      {onDismiss && (
```
dierat, line 170: Because the button is absolutely positioned, we can move it in the DOM order without affecting the visual positioning.

dierat:
https://app.shortcut.com/czi-edu/story/205122/meaningful-sequence-mentor-access-to-ssp-tool-screen-reader-does-not-announces-permissions-for-specialized-support-profiles

timzchang (approved):

ahuth (approved):
LGTM 🤘

PR #1216, opened 2022-07-28
Csnizik/208576 feedback punchlist

csnizik:
### Summary

The work in this PR covers https://app.shortcut.com/czi-edu/story/208576/feedback-overview-styling-punch-list

### Test Plan

All work can be viewed in Storybook: **Pages > Projects > Feedback > Overview**

---
1) **Hide the first row (legend) on mobile**

**Testing step:** on mobile view, select the first item in the Timeline, "Overview".

**Before:**
Screen Shot 2022-07-28 at 11 06 34 AM

**After:**
Screen Shot 2022-07-28 at 11 07 47 AM

**IMPORTANT:** On desktop view, the 'legend' row should still be visible. The border beneath it is changed to a heavier line due to this content being into the Table Header; this matches the Figma design.

Figma design:
Screen Shot 2022-07-28 at 11 32 53 AM

Desktop view after the work in this PR:
Screen Shot 2022-07-28 at 11 38 05 AM

---

2) **Back button link focus should only be visible on keyboard interaction, not mouse click**

**Testing step:** on mobile view, using a mouse, click on any item from the Timeline so that the "Back" button becomes visible.

**Before:**
Screen Shot 2022-07-28 at 11 14 20 AM

**After:**
Screen Shot 2022-07-28 at 11 17 04 AM

**IMPORTANT:** the Back button should still be active, with the active background animating in, when a Timeline nav item is selected using a keyboard.

---

3) **Update the FPO icons to `&nbsp;`**

**Testing step:** on either mobile or desktop view, click the first Timeline item, "Overview".

**Before:**
Screen Shot 2022-07-28 at 11 21 51 AM

**After:**
Screen Shot 2022-07-28 at 11 23 18 AM

---

4) **"More Options" button becomes icon-only on mobile; button remains in same flex row as page title text**

**Testing step 1:** In mobile view, check the display and placement of the 3 vertical dot icon to the right of the page title.

**Before:**
Screen Shot 2022-07-28 at 11 44 10 AM

**After:**
Screen Shot 2022-07-28 at 11 46 15 AM

**Testing step 2:** In desktop view, check that "More options" is visible and the button is at the right end of the row. (This should remain unaffected by the work in this PR.

Screen Shot 2022-07-28 at 11 47 59 AM

---

### Regression testing

The work in this PR touches other Storybook pages, which will need to be checked to make sure there are no regressions.

**What to look for**
In `PageHeader.module.css`, the style definition for `.page-header__right` was setting `margin-top: var(--eds-size-4);` on smaller breakpoints. This was having the unwanted effect of making the "More options" button lower than the page title (on smaller vp). To correct this, the `margin-top` property was removed from `.page-header__right`. 

Removing the `margin-top` property here had the unwanted side effect of removing `margin-top` from the header-right contents on 3 other Storybook pages. To compensate for this, a new utility class was created, `.u-margin-top-xl-mobile`, which adds `margin-top: var(--eds-size-4) !important;` on < md bp, and removes it on >= md bp with `margin-top: unset !important;`. 

**Potential regression #1 - Storybook > Pages > Course Planner > Edit > Default**
Desktop: the button group ("Back" and "Next") in the header-right slot should not move from their current position.

**Desktop** (red outline shows edges of `header.PageHeader-module__page-header`)

Screen Shot 2022-07-28 at 1 11 45 PM

Screen Shot 2022-07-28 at 1 12 43 PM

**Mobile** 

Screen Shot 2022-07-28 at 1 09 50 PM

Screen Shot 2022-07-28 at 1 07 44 PM

**Potential regression #2 - Storybook > Pages > Course Planner > Step 1 > Default**
(Same as #1 above; buttons "Return to Course" and "Next")

**Potential regression #3 - Storybook > Pages > Projects > Project > Student View**
(Same as #1 above; button "View plan")

### DIFF
--- a/.storybook/pages/FeedbackOverview/FeedbackOverview.tsx
+++ b/.storybook/pages/FeedbackOverview/FeedbackOverview.tsx
@@ -103,11 +103,9 @@ export const FeedbackOverview = ({ activeIndex = 0 }: Props) => (
                         <TableHeaderCell>Checkpoints Status</TableHeaderCell>
                         <TableHeaderCell>Final Product</TableHeaderCell>
                       </TableRow>
-                    </TableHeader>
-                    <TableBody>
                       <TableRow>
-                        <TableCell></TableCell>
-                        <TableCell>
+                        <TableHeaderCell></TableHeaderCell>
+                        <TableHeaderCell>
                           <NumberIconList>
                             <NumberIcon
                               aria-label="Item 1"
@@ -168,9 +166,11 @@ export const FeedbackOverview = ({ activeIndex = 0 }: Props) => (
                               size="sm"
                             />
                           </NumberIconList>
-                        </TableCell>
-                        <TableCell></TableCell>
+                        </TableHeaderCell>
+                        <TableHeaderCell></TableHeaderCell>
                       </TableRow>
+                    </TableHeader>
+                    <TableBody>
                       <TableRow>
                         <TableCell data-heading="Student">
                           Araya, Raquel
--- a/.storybook/pages/FeedbackOverview/FeedbackOverview.tsx
+++ b/.storybook/pages/FeedbackOverview/FeedbackOverview.tsx
@@ -194,7 +194,7 @@ export const FeedbackOverview = ({ activeIndex = 0 }: Props) => (
                                 flexShrink: 0,
                               }}
                             >
-                              1
+                              &nbsp;
                             </div>
                             <div
                               className="fpo"
@@ -206,7 +206,7 @@ export const FeedbackOverview = ({ activeIndex = 0 }: Props) => (
                                 flexShrink: 0,
                               }}
                             >
-                              2
+                              &nbsp;
                             </div>
                             <div
                               className="fpo"
@@ -218,7 +218,7 @@ export const FeedbackOverview = ({ activeIndex = 0 }: Props) => (
                                 flexShrink: 0,
                               }}
                             >
-                              3
+                              &nbsp;
                             </div>
                             <div
                               className="fpo"
@@ -230,7 +230,7 @@ export const FeedbackOverview = ({ activeIndex = 0 }: Props) => (
                                 flexShrink: 0,
                               }}
                             >
-                              4
+                              &nbsp;
                             </div>
                             <div
                               className="fpo"
@@ -242,7 +242,7 @@ export const FeedbackOverview = ({ activeIndex = 0 }: Props) => (
                                 flexShrink: 0,
                               }}
                             >
-                              5
+                              &nbsp;
                             </div>
                             <div
                               className="fpo"
@@ -254,7 +254,7 @@ export const FeedbackOverview = ({ activeIndex = 0 }: Props) => (
                                 flexShrink: 0,
                               }}
                             >
-                              6
+                              &nbsp;
                             </div>
                             <div
                               className="fpo"
@@ -266,7 +266,7 @@ export const FeedbackOverview = ({ activeIndex = 0 }: Props) => (
                                 flexShrink: 0,
                               }}
                             >
-                              7
+                              &nbsp;
                             </div>
                             <div
                               className="fpo"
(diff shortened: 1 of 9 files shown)


src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -237,6 +235,10 @@ export const TimelineNav = ({
     } else if ([L_ARROW_KEYCODE, U_ARROW_KEYCODE].includes(e.key)) {
       /* 4 */
       timelineNavItemRefs[prev].current.focus();
+    } else if (e.key === ENTER_KEYCODE) {
+      setTimeout(() => {
+        backRef.current?.focus(); /* 5 */
+      }, 500);
```
dierat, line 241: Oh nice solution! I think we also need spacebar though?

.storybook/pages/FeedbackOverview/FeedbackOverview.tsx
```diff
@@ -194,7 +194,7 @@ export const FeedbackOverview = ({ activeIndex = 0 }: Props) => (
                                 flexShrink: 0,
                               }}
                             >
-                              1
+                              &nbsp;
```
dierat, line 197: Could we maybe use `map` to render all of these instead of specifying each one? Kinda annoying that you had to go through and replace every number in a million little boxes.

src/components/Utilities/Spacing.css
```diff
@@ -84,6 +84,18 @@
...
+ * Margin top xl mobile
+ * 1) Forces margin top to 2rem (on smaller viewports only)
+ */
+.u-margin-top-xl-mobile {
+  margin-top: var(--eds-size-4) !important;
+
+  @media all and (min-width: $eds-bp-md) {
+    margin-top: unset !important;
+  }
+}
```
dierat, line 97: Could you help me understand this `unset` stuff? Why can't we just target the size that needs the margin? Like
```
.u-margin-top-xl-mobile {
  @media all and (max-width: $eds-bp-md) {
    margin-top: var(--eds-size-4) !important;
  }
}
```
Same for the `u-mobile-hidden` class.

dierat (changes requested):
Requesting changes for the focus logic being triggered by spacebar as well as the enter key.

src/components/Utilities/Spacing.css
csnizik: Great question! Because the style on `.page-header` changes `flex-direction` from `column` to `row` using the media query `min-width: $eds-bp-md`. So any styles that are affecting the same elements at the same breakpoint should use the same media query (i.e., all use min-width OR all use max-width). Otherwise there's a 1px "gap" where the min-width rule is being applied but the max-width rule isn't. 

https://user-images.githubusercontent.com/24812780/181619408-4bf7f135-9065-4a3c-801e-bf0ae28defd8.mov

In a perfect world, all media queries in a project would either be "max-width" or "min-width"; if so, there's never any gap between where one rule starts and another stops.

csnizik: BTW you can see the "gap" between media queries on that same page with the header size on "Select projects for your History 6 plan". Out of the scope of this ticket though, whew.

.storybook/pages/FeedbackOverview/FeedbackOverview.tsx
csnizik: regex find/replace to the rescue.  find `(?<=flexShrink: 0,\n\s*}}\n\s*>\n\s*)\d+(?=\n\s*<\/div>)` replace `&nbsp;` 🤣

src/components/TimelineNav/TimelineNav.tsx
csnizik: Thanks! Added.

src/components/Utilities/Spacing.css
dierat: Oo nice catch. Could you leave a small comment about needing to use the same media query as the `.page-header`?

dierat (approved):

PR #1217, opened 2022-07-28
chore(dropdown): ts error

csnizik:
### Summary:
Clears TS error, "No overload matches this call" in DropdownMenu.tsx

If this solution is accepted, it can also be applied to:
ButtonDropdown.tsx:172
Tabs.tsx:306
TimelineNav.tsx:252

### Test Plan:
`yarn types` should run without errors.

### DIFF
--- a/src/components/DropdownMenu/DropdownMenu.tsx
+++ b/src/components/DropdownMenu/DropdownMenu.tsx
@@ -117,18 +117,22 @@ export const DropdownMenu: React.FC<Props> = ({
       if (React.isValidElement(child)) {
         if (isReactFragment(child)) {
           const newChildren = child.props.children;
-          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
           return newChildren.map((item: ReactNode, i: number) => {
-            // @ts-expect-error TODO: fix "No overload matches this call" error
-            return React.cloneElement<Props>(item, {
-              ref: (el: HTMLLIElement) => (childRefs.current[i] = el) /* 1 */,
-            });
+            console.log({ item: item });
+            return React.cloneElement<Props>(
+              item as any,
+              {
+                ref: (el: HTMLLIElement) => (childRefs.current[i] = el) /* 1 */,
+              } as any,
+            );
           });
         } else {
-          // @ts-expect-error TODO: fix "No overload matches this call" error
-          return React.cloneElement<Props>(child, {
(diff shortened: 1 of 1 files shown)


src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -117,18 +117,21 @@ export const DropdownMenu: React.FC<Props> = ({
...
-            // @ts-expect-error TODO: fix "No overload matches this call" error
-            return React.cloneElement<Props>(item, {
-              ref: (el: HTMLLIElement) => (childRefs.current[i] = el) /* 1 */,
-            });
+            return React.cloneElement<Props>(
+              item as any,
+              {
+                ref: (el: HTMLLIElement) => (childRefs.current[i] = el) /* 1 */,
+              } as any,
+            );
```
csnizik, line 126: Using the type assertion `as any` here was the only way I found to quiet this error. In effect it's turning off type checking of the item to be cloned (line 122) and the value being passed in to the `ref` prop of the new element, which is going to be an `HTMLLIElement` (line 125). Aren't those two variables already type checked when they're passed in to the `cloneElement()` method?

anniehu4: > Since those two variables are already type checked when they're passed in to the cloneElement() method

Could you say more about this? I thought asserting as `any` would mean they're no longer type checked by `cloneElement`.

Either way, my understanding is people _should_ only be passing in `DropdownMenuItem` as children, which is typechecked through that component, so the typechecking isn't important here. Could you leave a comment here saying that to explain the `any` types?

csnizik: > > Since those two variables are already type checked when they're passed in to the cloneElement() method
> 
> Could you say more about this? I thought asserting as `any` would mean they're no longer type checked by `cloneElement`.

My logic was probably overly knotty, but what I meant was: on line 120 we typecheck `item`; we know it's a ReactNode. On line 124, we typecheck `el` as a HTMLLIElement.  And as you point out, these items have already been typechecked in their component. They're like an adult who gets carded at the bar every single time 🤣

csnizik: I added comments, plus I cleaned up the old comments that weren't relevant anymore.

anniehu4 (approved):
cool, thanks!

PR #1220, opened 2022-08-09
chore(clickable-style, button, link): set display to inline on link variant

dierat:
### Summary:
We got a comment about how links aren't wrapping well in paragraphs – if the link text is in a paragraph that wraps but the link text itself does not wrap it's fine; but if the link text is situated in such a way that the link text wraps, it will break away from the rest of the text and sit on its own line. This is because we have `display: inline-flex` on the link styles.

We were aware of this behavior, but it had been like that for a long time and no one had commented on it – *until now*. And since we're pushing for more mobile-friendly designs and less ultra-short link text (ie "here" and "learn more"), I think it's going to keep coming up.

There are 2 side effects from removing `display: inline-flex`:
- the animation for the focus state won't work (which is fine; that was just a fun little thing I wanted to throw in)
- the `gap` style we're using to add space between text and icons (is an icon is present) in links won't work anymore

**Also**, we're not able to fix this for `button` elements that are styled to look like links – the text can wrap within the button, but it doesn't look like (from my exploration) that it's possible to get the button text to wrap *with* paragraph text. So the `Button` component might as well keep `display: inline-flex` and the gap between the text and icon (if an icon is present). I'm going to remove the focus animation though because I think it's too messy from a code and UX point of view to have that only some of the time when it's not really important.

#### Before
the new link in paragraph context story but without the other changes from this pr

#### After
the new link in paragraph context story with the other changes from this pr

### Test Plan:
Verify that:
- buttons styled like links and all links (regardless of style) in the system are unchanged
- the focus state for buttons and links styled like links still works (but with a quick fade animation instead of a vertical slide)
- the link text wraps with the paragraph text in the new `LinkInParagraphContext` story

### DIFF
--- a/src/components/ClickableStyle/ClickableStyle.module.css
+++ b/src/components/ClickableStyle/ClickableStyle.module.css
@@ -454,21 +454,25 @@
 
 /**
  * Link button styles
+ * 1) Override default Clickable styles
+ * 2) Smaller gap between text and icon in link button style
  */
 .clickable-style--link {
   @mixin textLink;
 
-  border-radius: 0;
+  /* 1 */
   border: 0;
+  border-radius: 0;
   background-color: transparent;
+
+  gap: 0.25em; /* 2 */
 }
 
 /**
- * Margin 0 added becuase Safari has margin on buttons.
  * 1) Needs `:where` pseudoclass to reduce specificity so class overrides
  *    for margin can happen where used.
  * 2) 2px padding gives the text breathing room in the focus state
- * 3) 2px margin compensates for the padding so the link doesn't have too much space between it
+ * 3) -2px margin compensates for the padding so the link doesn't have too much space between it
  *    and surrounding text
  */
 :where(.clickable-style--link) {
@@ -478,6 +482,8 @@
 
 /**
  * Link neutral clickable style
+ * 1) Overrides the default background color on focus. Often used when the default
+ *    focus background color doesn't mesh well with the surrounding colors.
  */
 .clickable-style--link.clickable-style--neutral {
   text-decoration-color: var(--eds-theme-color-icon-neutral-strong);
@@ -485,15 +491,10 @@
   &:hover {
     text-decoration-color: var(--eds-theme-color-border-neutral-strong);
   }
-}
 
-/**
- * Link neutral clickable style pseudo element for the focus background state.
- * Overrides the default background color on focus. Often used when the default
- * focus background color doesn't mesh well with the surrounding colors.
- */
-.clickable-style--link.clickable-style--neutral::before {
-  background-color: var(--eds-theme-color-icon-neutral-strong);
+  &:focus {
+    background-color: var(--eds-theme-color-icon-neutral-strong); /* 1 */
+  }
 }
 
 /**
--- a/src/components/Link/Link.module.css
+++ b/src/components/Link/Link.module.css
@@ -1,0 +1,13 @@
+@import '../../design-tokens/mixins.css';
+
+/*------------------------------------*\
+    # LINK
+\*------------------------------------*/
+
+/**
+ * Link button with link variant styles
+ * 1) Override display: inline-flex from ClickableStyle
+ */
+.link--link {
+  display: inline; /* 1 */
+}
--- a/src/components/Link/Link.stories.tsx
+++ b/src/components/Link/Link.stories.tsx
@@ -347,3 +347,24 @@ export const IconError: StoryObj<Args> = {
     variant: 'icon',
   },
 };
+
+export const LinkInParagraphContext = {
+  render: () => (
+    <div>
+      Lorem ipsum dolor sit amet,{' '}
+      <Link href={getRandomUrl()}>consectetur adipiscing elit</Link>. Morbi
+      porta at ante quis molestie. Nam scelerisque id diam at iaculis. Nullam
+      sit amet iaculis erat. Nulla id tellus ante.{' '}
+      <Link href={getRandomUrl()}>
+        Aliquam pellentesque ipsum sagittis, commodo neque at, ornare est.
+        Maecenas a malesuada sem, vitae euismod erat. Nullam molestie nunc non
+        dui dignissim fermentum.
+      </Link>{' '}
+      Aliquam id volutpat nulla, sed auctor orci. Fusce cursus leo nisi. Fusce
+      vehicula vitae nisl nec ultricies. Cras ut enim nec magna semper egestas.
+      Sed sed quam id nisl bibendum convallis. Proin suscipit, odio{' '}
+      <Link href={getRandomUrl()}>vel pulvinar</Link> euismod, risus eros
+      ullamcorper lectus, non blandit nulla dui eget massa.
+    </div>
+  ),
+};
--- a/src/design-tokens/mixins.css
+++ b/src/design-tokens/mixins.css
@@ -16,11 +16,6 @@
 
 /**
  * Link button styles
- * 1) display: inline-flex allows us to manage icon placement and the focus background animation
- * 2) Smaller gap between text and icon in link button style
- * 3) Has position relative to support the background color pseudo element
- * 4) We hide overflow so the pseudo element can be pushed out of site
- * 5) z-index 0 required for the background to appear correctly on focus
  * 6) Has smaller icons than other button styles
  * 7) Make the focus outline invisible but don't remove it. High contrast mode will remove the background 
       color, but it will also make borders 100% opacity black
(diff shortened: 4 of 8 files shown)


src/components/ClickableStyle/ClickableStyle.module.css
```diff
@@ -454,21 +454,25 @@
...
 .clickable-style--link {
   @mixin textLink;
 
-  border-radius: 0;
+  /* 1 */
   border: 0;
+  border-radius: 0;
   background-color: transparent;
+
+  gap: 0.25em; /* 2 */
```
dierat, line 468: I moved this from the mixin because it will only have an effect here due to the `display: inline-flex` base style.

src/components/ClickableStyle/ClickableStyle.module.css
```diff
@@ -454,21 +454,25 @@
...
+ * 1) Override default Clickable styles
+ * 2) Smaller gap between text and icon in link button style
  */
 .clickable-style--link {
   @mixin textLink;
 
-  border-radius: 0;
+  /* 1 */
   border: 0;
+  border-radius: 0;
```
dierat, line 465: Moved this down one line just because it felt like a more natural order to me.

src/components/ClickableStyle/ClickableStyle.module.css
```diff
@@ -478,22 +482,19 @@
...
   }
-}
 
-/**
- * Link neutral clickable style pseudo element for the focus background state.
- * Overrides the default background color on focus. Often used when the default
- * focus background color doesn't mesh well with the surrounding colors.
- */
-.clickable-style--link.clickable-style--neutral::before {
-  background-color: var(--eds-theme-color-icon-neutral-strong);
```
dierat, line 496: This used to be on the `::before` pseudo element all the time because that element was only visible on focus. Now that that element is gone, we apply the style to the clickable element only on focus.

src/components/Link/Link.module.css
```diff
@@ -0,0 +1,13 @@
...
+    # LINK
+\*------------------------------------*/
+
+/**
+ * Link button with link variant styles
+ * 1) Override display: inline-flex from ClickableStyle
+ */
+.link--link {
+  display: inline; /* 1 */
+}
```
dierat, line 13: This is a little awkward but I only want to add a style to the `Link` component when it's using the "link" variant.

src/components/Link/Link.stories.tsx
```diff
@@ -347,3 +347,24 @@ export const IconError: StoryObj<Args> = {
     variant: 'icon',
   },
 };
+
+export const LinkInParagraphContext = {
```
dierat, line 351: This is for testing whether the links are wrapping as expected in a paragraph.

src/components/Link/Link.stories.tsx
```diff
@@ -347,3 +347,24 @@ export const IconError: StoryObj<Args> = {
...
+  render: () => (
+    <div>
```
dierat, line 353: This needs to be in a `div` because jest doesn't capture the whole story content if it's in a fragment.

src/design-tokens/mixins.css
```diff
@@ -29,26 +24,13 @@
   font-size: inherit;
   font-weight: var(--eds-font-weight-medium);
   line-height: inherit;
-  text-align: left;
```
dierat, line 32: I don't know why we had `text-align: left;` on this at all.

src/design-tokens/mixins.css
```diff
@@ -29,26 +24,13 @@
...
-  display: inline-flex; /* 1 */
```
dierat, line 33: Removing `display: inline-flex;` allows the link to wrap with paragraph text.

src/design-tokens/mixins.css
```diff
@@ -29,26 +24,13 @@
...
-  gap: 0.25em; /* 2 */
```
dierat, line 34: This doesn't do anything without `display: inline-flex` so I moved this to the `ClickableStyle` styles where `display: inline-flex` is still present.

src/design-tokens/mixins.css
```diff
@@ -29,26 +24,13 @@
...
   text-decoration: underline;
   text-decoration-color: var(--eds-theme-color-background-brand-primary-strong);
   text-decoration-thickness: 2px;
   text-underline-offset: 2px;
 
-  position: relative; /* 3 */
-  overflow: hidden; /* 4 */
-  z-index: var(--eds-z-index-0); /* 5 */
-  transition: color var(--eds-anim-move-medium) var(--eds-anim-ease),
-    text-decoration-color var(--eds-anim-move-medium) var(--eds-anim-ease);
```
dierat, line 46: This was all supporting the animation.

src/design-tokens/mixins.css
```diff
@@ -29,26 +24,13 @@
...
-
-  @media screen and (prefers-reduced-motion) {
-    transition: none;
-  }
```
dierat, line 50: I think this was redundant..?

src/design-tokens/mixins.css
```diff
@@ -64,52 +46,20 @@
...
-  &::before {
-    /* 1 */
-    content: '';
-    width: 100%;
-    height: 100%;
-    position: absolute;
-    top: 0;
-    left: 0;
-    z-index: var(--eds-z-index-bottom); /* 2 */
-    background-color: var(--eds-theme-color-background-brand-primary-strong);
```
dierat, line 92: The background-color change was moved up into the `&:focus` styles; the rest just supported the animation.

dierat:

dierat:
https://app.shortcut.com/czi-edu/story/210540/remove-display-inline-flex-from-link-variant

src/components/ClickableStyle/ClickableStyle.module.css
dierat: I'm also wondering if I should instead put `display: inline` on the mixin and move this to the `Button` styles. I feel like the gap behavior should be the default but 🤷

dierat:
I forgot to put reviewers on this.. again.. I wish Github showed a reminder to add reviewers..

src/components/ClickableStyle/ClickableStyle.module.css
ahuth, line 468: Would the comments be easier to make sense of if they were inline / next to the thing they're commenting?

I haven't looked at EDS code in a little while, but this style of referencing another point in this file feels harder for me to put understand.

Also may be less likely to update or remove comments if they're next to the commentee.

src/components/Link/Link.stories.tsx
ahuth: What's it capture?

ahuth (approved):
LGTM 🤘

src/components/ClickableStyle/ClickableStyle.module.css
dierat: I think they would be able to understand, yes. I've seen a few times where outdated comments hung around after the code below it had changed, and I do think that would be less likely if the comment were next to the code. This format came from Big Medium. cc @jinlee93 @anniehu4

src/components/Link/Link.stories.tsx
dierat: Before adding the `div`, it only picked up the first line of text before the first `Link` (`Lorem ipsum dolor sit amet,{' '}`).

This is a sidebar but kinda relevant – I also saw a PR on Along where jest wasn't picking up a big section of a component's children, and based on this PR I guessed that it was because a wrapper component was using a fragment, and changing it to a `div` fixed the issue. https://github.com/chanzuckerberg/along/pull/4924#discussion_r942493700

src/components/ClickableStyle/ClickableStyle.module.css
anniehu4: very pro ditching this commenting style and going back to having things right next to what they're referencing :')

anniehu4: https://app.shortcut.com/czi-edu/story/211537/move-css-comments-next-to-what-they-re-referencing

src/components/Link/Link.stories.tsx
ahuth: 😲 this is one of the weirdest things I've ever seen. Feels like Jest/React/TheUniverse shouldn't have any problems with it 🤔

src/components/ClickableStyle/ClickableStyle.module.css
ahuth: Next time we work with a contractor they gotta go through some training (roles and responsibilities, quality standards, expectations, etc).

dierat:
Tested this in `traject` with an alpha build and didn't see any changes to links in chromatic.

PR #1221, opened 2022-08-09
feat(filters): add filters component 1/n

jinlee93:
apologies in advance for the larger pr
### Summary:
[sc-202286]
- Motivation
  - As part of the feedback overview pilot, a filter component was needed to filter table categories. 
  - Conversations also included a block component with a grid of checkbox filters. However a block component would push the associated table down too far in a mobile view. A drawer component would be preferred for mobile. 
  - A disadvantage of the drawer component would be increased elevation complexity if the table were to be in components such as a Modal. Discussion landed that the Table component would be advised against in modals, and the filters component even more discouraged.
- Implementation
  - In this PR, the Filters component takes `checkboxFields` as a prop to place them within the Filters Drawer. When the drawer is closed, an object mapping the checkbox values to their checked status is passed to a callback prop.
    - In an upcoming pr, the `checkboxFields` prop will be replaced as children `<FiltersCheckboxFields>` housing children `<Checkbox>` components to allow increased flexibility regarding the filters
  - The Drawer component is moved over from `upcoming-components` to `components`.

### Test Plan:
- added unit tests
- no visual regressions on other components
- axe tests pass
Screen Shot 2022-08-09 at 3 52 44 PM
Screen Shot 2022-08-09 at 3 53 01 PM
Screen Shot 2022-08-09 at 3 51 52 PM
Screen Shot 2022-08-09 at 3 52 05 PM
Screen Shot 2022-08-09 at 3 52 14 PM
Screen Shot 2022-08-09 at 3 52 22 PM


src/components/Drawer/Drawer.module.css
```diff
@@ -11,34 +11,29 @@
  */
 .drawer {
   display: flex;
+  flex-direction: column;
+  align-items: flex-end;
   justify-content: flex-end;
   visibility: hidden;
-  width: 100%;
+  width: 28rem;
```
jinlee93, line 18: Moves sizing of the drawer to the drawer container. The housed drawer window will be a relative value

src/components/Drawer/Drawer.module.css
```diff
@@ -11,34 +11,29 @@
...
 
-  &:focus-visible {
-    @mixin focus;
-  }
+/**
+* Active modal
+*/
+.drawer--is-active {
+  visibility: visible;
+  z-index: var(--eds-z-index-400);
```
jinlee93, line 36: Just removing some nesting

src/components/Drawer/Drawer.tsx
```diff
@@ -166,16 +160,22 @@ export const Drawer = ({
...
+      ) {
+        handleOnClose(); /* 1 */
+      }
     }
-  }
+    document.addEventListener('click', handleOnClickOutside);
+    return () => {
+      document.removeEventListener('click', handleOnClickOutside);
+    };
+  }, [isActive, windowRef]); // eslint-disable-line react-hooks/exhaustive-deps
```
jinlee93, line 178: The previous `handleOnClickOutside` did not work (since there's it didn't detect outside clicks with no listener on the document)

src/components/Fieldset/Fieldset.module.css
```diff
@@ -10,5 +10,6 @@
  */
 .fieldset {
   @mixin fieldsetStyles; /* 1 */
+  --checkbox-svg-size: normal;
```
jinlee93, line 13: Removes some extra margin bottom on the checkbox. Scoping it here since changing it on the Checkbox component will affect Checkbox sizes across LP

src/components/Filters/Filters.test.tsx
```diff
@@ -0,0 +1,50 @@
+import { generateSnapshots } from '@chanzuckerberg/story-utils';
+import { composeStories } from '@storybook/testing-react';
+import { fireEvent, render, screen } from '@testing-library/react';
+import React from 'react';
+import * as stories from './Filters.stories';
+
+const { OverflowInteractive } = composeStories(stories);
+const consoleLogMock = jest.spyOn(console, 'log').mockImplementation();
```
jinlee93, line 8: Removes the console logs from the stories. Will also remove once the filters logic is redone as stated in the summary.

dierat:
>Discussion landed that the Table component would be advised against in modals, and the filters component even more discouraged.

Nyuh I foresee this being a real issue.. Scrum teams have limited resources and will continue to use modals as necessary despite our recommendations.. Oh well.

src/components/Drawer/__snapshots__/Drawer.test.tsx.snap
```diff
@@ -0,0 +1,16 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`<Drawer /> Default story renders snapshot 1`] = `
```
dierat, line 3: Can we get a snapshot where it's open?

src/components/Drawer/Drawer.tsx
```diff
@@ -45,15 +40,15 @@ export interface Props {
...
   /**
    * Handler to be called when the drawer is being closed (by ESCAPE / clicking X / clicking outside)
    */
   onClose?: (e?: any) => void;
-}
+  /**
+   * CSS class names that can be appended to the Drawer window component that the Drawer components reside in.
+   */
+  windowClassName?: string;
+};
```
dierat, line 51: I find the name and description here a bit confusing – I had to skip down and read the code to be sure I understood which class name went with which element. I'm actually wondering if the outer element should have the `windowClassName` or `drawerContainerClassName` or something like that (the `Modal` has `modalContainerClassName`, for example).

src/components/Filters/Filters.tsx
```diff
@@ -0,0 +1,101 @@
...
+   */
+  checkboxFields: CheckboxField[];
+  /**
+   * CSS class names that can be appended to the component.
+   */
+  className?: string;
+  /**
+   * Callback called when filters drawer is closed.
+   */
+  closeFilters: (checkedValues: { [key: string]: boolean }) => void;
```
dierat, line 19: Could this be called something like `onClose`?

src/components/Drawer/__snapshots__/Drawer.test.tsx.snap
jinlee93: Yea will add

src/components/Drawer/Drawer.tsx
jinlee93: Hmmm that makes sense, I'll rename to be less confusing

src/components/Filters/Filters.tsx
jinlee93: Yup, that would be more intuitive and follow other prop patterns

src/components/DrawerHeader/DrawerHeader.tsx
```diff
@@ -34,7 +35,7 @@ export interface Props {
  * import {DrawerHeader} from "@chanzuckerberg/eds";
  * ```
  *
- * TODO: update this comment with a description of the component.
+ * The header content of the Drawer component. Usually houses the heading and the close button.
```
dierat, line 38: Thanks for updating this! ❤️

src/components/Drawer/Drawer.module.css
```diff
@@ -11,34 +11,29 @@
...
-  }
+.drawer:focus-visible {
+  @mixin focus;
+}
 
-  &:focus-visible {
-    @mixin focus;
-  }
+/**
+* Active modal
```
dierat, line 32: Btw should we update these comments to say "Drawer" instead of "Modal"?

dierat:
Looks good to me but I'm generally not as opinionated about how form components work.. @anniehu4 could you take this one?

src/components/Drawer/Drawer.module.css
jinlee93: Yup will do

src/components/Filters/Filters.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  /**
+   * Manages the active state of the filters drawer.
+   */
+  const [isActive, setIsActive] = useState(false);
+
+  function closeFiltersDrawer(checkedValues: { [key: string]: boolean }) {
+    setTimeout(() => {
+      filtersButton?.current?.focus();
+    }, 1);
```
anniehu4, line 59: what's the significance of waiting 1 ms?

src/components/Filters/__snapshots__/Filters.test.tsx.snap
```diff
@@ -0,0 +1,55 @@
...
+      />
+    </svg>
+    Filters 
+  </button>
+</div>
+`;
+
+exports[`<Filters /> FiltersDrawerComponent story renders snapshot 1`] = `null`;
+
+exports[`<Filters /> FiltersDrawerComponentNoLegend story renders snapshot 1`] = `null`;
```
anniehu4, line 30: what's up w these?

src/components/Filters/Filters.stories.tsx
```diff
@@ -0,0 +1,150 @@
...
+    checkboxFields: overflowCheckboxFields,
+    onClose: (checkedValues) => {
+      console.log(checkedValues);
+    },
+  },
+};
+
+type FiltersDrawerArgs = React.ComponentProps<typeof FiltersDrawer>;
+/**
+ * 1) Axe testing throws error due to Drawer component not being controlled, but this story is mostly for visual regression testing as the interactive stories would only capture the toggle button.
```
anniehu4, line 103: should we add a play function so the stories show the open filter?

anniehu4 (approved):
general structure lgtm

jinlee93:
> could you leave that as a comment? but also kinda confused b/c the point of this variable is supposed to keep the height of the SVG the same as the line height

Yup, will do

src/components/Filters/Filters.tsx
jinlee93: Good question, removed the `setTimeout`, and works, thanks

src/components/Fieldset/Fieldset.module.css
jinlee93: Ah great question, turns out I had to move where the style was being placed in `Checkbox`, thanks for the catch

src/components/Filters/__snapshots__/Filters.test.tsx.snap
jinlee93: Let me investigate

src/components/Filters/Filters.stories.tsx
jinlee93: The purpose of the these stories are to snap the open filter (non interactive), and the interactive stories I think should start not closed

src/components/Filters/Filters.tsx
anniehu4: another question mark from BM lol ... we may want to move it to the end of this function so focus is returned after `setIsActive` runs?

src/components/Filters/__snapshots__/Filters.test.tsx.snap
jinlee93: Turns out this is happening due to the usage of the `react-portal` library to append the drawer on the `document`. There is a solution for snaps to happen, which I will add for now, but since there is a native [React Portal](https://reactjs.org/docs/portals.html) I will move the native solution in the upcoming pr

src/components/Filters/Filters.tsx
jinlee93: I think the intention of the `setTimeout` was to have the `focus()` occur after `setIsActive` since `setState` is async. Since it's async, don't think moving the `focus()` to after `setIsActive` would change any functionality right now

anniehu4: ahhh ok that makes sense! so should we keep the setTimeout? 🤔  and maybe leave that comment

jinlee93: I don't think it's necessary to keep the `setTimeout` unless you think there is going to be some some crazy async stuff in the `onClose` that has to run before the focus occurs (such as a navigate away? or data fetching?)

anniehu4: In my head it makes sense that the order of events should be close the drawer -> then return focus to the Filters button, and seems like removing the `setTimeout` would make it the reverse? It's up to you though (and lmk if I'm wrong)

jinlee93: Yea that makes sense to me, will keep the `setTimeout`

src/components/Drawer/__snapshots__/Drawer.test.tsx.snap
```diff
@@ -0,0 +1,17 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`<Drawer /> Default story renders snapshot 1`] = `null`;
```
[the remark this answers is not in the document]
jinlee93: As stated below, story snap should not be null, fixed

src/components/Filters/__snapshots__/Filters.test.tsx.snap
jinlee93: Did some pairing with Andrew, cleaned up Drawer component and removed some unnecessary states and snaps work now. Still going to work on removing `react-portal` tho

src/components/Drawer/Drawer.tsx
```diff
@@ -62,80 +57,66 @@ export interface Props {
...
+  const isMountedRef = useRef(true);
 
   /**
    * Use effect
    * 1) Set prevIsActive to previous isActive prop
    * 2) If prevIsActive is defined and previous isActive prop is not equal
    * to current isActive prop, toggle state
    */
-  const prevIsActive = usePrevious(isActive); /* 1 */
+  // const prevIsActive = usePrevious(isActive); /* 1 */
```
dierat, line 87: @jinlee93 do you remember why you left this line in after commenting it out? Is it safe to remove this line in the codebase?

jinlee93: Oop yea I actually think that entire codeblock explaining the `useEffect` can be removed as I think it's pretty self-explanatory

PR #1222, opened 2022-08-11
fix(projectcard): make numberAriaLabel optional but warn

jinlee93:
### Summary:
- `numberAriaLabel` should not be required if there is no `number`
- warns if doesn't exist if `number` is passed
### Test Plan:
- wrote unit test
- no visual regression

### DIFF
--- a/src/components/ProjectCard/ProjectCard.tsx
+++ b/src/components/ProjectCard/ProjectCard.tsx
@@ -56,9 +56,9 @@ export interface Props {
    */
   number?: number;
   /**
-   * Number aria label
+   * Number aria label. Required if number prop is passed.
    */
-  numberAriaLabel: string;
+  numberAriaLabel?: string;
   /**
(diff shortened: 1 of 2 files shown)


src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -81,6 +81,11 @@ export const ProjectCard = ({
   isDragging,
   ...other
 }: Props) => {
+  if (number && !numberAriaLabel && process.env.NODE_ENV !== 'production') {
+    console.warn(
```
dierat, line 85: Should it be an error since it's an accessibility bug to not have an aria label on this number?

dierat (approved):

src/components/ProjectCard/ProjectCard.tsx
jinlee93: Yuh will change to error, thanks!

PR #1226, opened 2022-08-12
fix(drawerheader): correctly refer to color token

jinlee93:
### Summary:
- need to refer to the correct color token as the DrawerHeader no longer lives in `upcoming-components`
- will follow up with an eds patch fix
### Test Plan:
- built file refers to correct token
image


anniehu4 (approved):
weird that it didn't break the build 🤷‍♀️

PR #1223, opened 2022-08-12
chore(checkbox): update styles to match library

jinlee93:
### Summary:
[sc-202569]
- Icons are moving towards sharp icons, this updates the checkbox svgs to be sharp
- The previous checkbox icon was a custom svg in-file element and this uses the checkbox icons from the library to follow our icon practices
- Cleans up the checkbox styles and the `--checkbox-svg-size` css custom variable we used to align the label with the checkbox. Uses a flexbox centering strategy instead and is hence also more readable (imo)
- Potentially breaking: removes a small amount of space under the checkbox that used to exist 
image
### Test Plan:
- wrote unit test and updated snaps
- visual regressions only where checkboxes used (Filters, Fieldset, Checkboxes)

PR #1230, opened 2022-08-16
chore(release): 3.1.2

jinlee93:
Releases the checkbox styles changes https://github.com/chanzuckerberg/edu-design-system/commit/c94e2caee24c675e81eab6f9cff44d17a04dfcad

Patch includes this change only because it of updates needed on LP [discussion](https://github.com/chanzuckerberg/edu-design-system/pull/1223#issuecomment-1215457633) and want to separate as much as possible from upcoming larger release


dierat (approved):
Nice! Good idea splitting it out just in case 👍

PR #1233, opened 2022-08-17
fix(pageheader): remove wrapping div for overline

jinlee93:
### Summary:
- Motivation:
  - There was a request to add an area above the pageheader title for components like breadcrumbs
  - Turned out the area exists, but is wrapped with a div with some set margins
- Execution:
  - Removes the div wrapper
### Test Plan:
- unit tests pass, pageheader snap updated
- no visual regressions except for pageheader overline story

image


anniehu4 (commented):
hmm might we want that built in bottom-margin? or do you think it's better for consumers to provide it themselves?

dierat:
leo laughing meme with the text "brumbs"

dierat (approved):
I'm personally fine with consumers adding their own margin 🤷

PR #1235, opened 2022-08-19
feat(filters): make filters components controlled

jinlee93:
### Summary:
[sc-211241]
- Motivation:
  - The filters' previous api was too rigid and did not allow flexibility of styling the checkboxes since the checkbox information was passed down through props
    - Checkbox state control was done internally
    - Checkbox styling was not exposed
  - Such an api was not a pattern that was followed by EDS
- Execution:
  - Converted `<Filters>`and `<FiltersDrawer>` to be controlled
  - Made `<Filters>` a wrapper for the `<FiltersDrawer>` , which was already a wrapper for `<FiltersCheckboxField>`s
    - This allows devs to pass `<Checkbox>` or `<FiltersCheckboxField>` components as children and style it how they want
    - Also follows the api pattern found throughout EDS components 
    - Story shows how the state can be controlled
### Test Plan:
- Used the `isChromatic()` utility function to conditionally call the `play` function for better Chromatic snaps, allowing the removal of two redundant snaps just to capture the drawer
  - Chromatic snaps these two stories properly
- Only snap changes, no changes to jest functionality test
  - Unit tests pass

Screen Shot 2022-08-18 at 10 14 48 PM
Screen Shot 2022-08-18 at 10 14 53 PM


src/components/Filters/Filters.stories.tsx
```diff
@@ -12,139 +14,210 @@ export default {
...
+          onChange={
+            () => {} /* eslint-disable-line @typescript-eslint/no-empty-function */
+          }
+        />
+      </Filters.FiltersCheckboxField>
+    ),
   },
+  decorators: [
+    (Story) => (
+      <div style={{ margin: '0.25rem', height: '100vh' }}>
```
jinlee93, line 44: `100vh` is required for chromatic to create the snaps properly, or just cuts off at the Button height for the drawer

src/components/Filters/Filters.stories.tsx
```diff
@@ -12,139 +14,210 @@ export default {
...
-      { label: 'Filters label 3', value: 'segment2label3' },
-    ],
+export const Default: StoryObj<Args> = {
+  play: async ({ canvasElement }) => {
+    // We want to test visual regression for the drawer as well as the button, but don't want the drawer open initally outside Chromatic
+    if (isChromatic()) {
+      const canvas = within(canvasElement);
+      const filtersTrigger = await canvas.findByRole('button');
+      filtersTrigger.click();
+    }
```
jinlee93, line 60: Although repeated code in the other story, the play function does not execute when placed in the `export default` block

src/components/FiltersDrawer/FiltersDrawer.tsx
```diff
@@ -154,7 +84,11 @@ export const FiltersDrawer = ({
       dismissible={true}
       drawerContainerClassName={containerClassName}
       isActive={isActive}
-      onClose={() => onClose(checkedMap)}
+      onClose={() => {
+        if (onClose) {
+          onClose();
+        }
```
dierat, line 90: I forget – can we do `onClose && onClose();`?

src/components/FiltersDrawer/FiltersDrawer.tsx
```diff
@@ -166,22 +100,35 @@ export const FiltersDrawer = ({
...
-          <Button
-            className={styles['footer__button']}
-            onClick={applyFilters}
-            variant="primary"
-          >
-            Apply
-          </Button>
-        </ButtonGroup>
-      </Drawer.Footer>
+      {(onClear || onApply) && (
```
dierat, line 105: What if `onClear` or `onApply` is passed in but not both? What does the button without a method do?

src/components/Filters/Filters.tsx
```diff
@@ -1,22 +1,39 @@
...
+import { FiltersDrawer } from '../FiltersDrawer/FiltersDrawer';
 import Icon from '../Icon';
 
 export type Props = {
   /**
-   * Checkbox fields that will be displayed in the filters drawer.
+   * Text to be placed in the button that activates the Filters Drawer
    */
-  checkboxFields: CheckboxField[];
+  buttonText?: string;
```
dierat, line 12: I had to go look at the stories to be sure what this prop did because I was thinking it might be for the primary button inside the drawer. Maybe we could make it more specific like `triggerButtonText` or even just `triggerText`?

src/components/FiltersDrawer/FiltersDrawer.tsx
```diff
@@ -4,55 +4,34 @@ import { useUID } from 'react-uid';
...
 import ButtonGroup from '../ButtonGroup';
-import Checkbox from '../Checkbox';
 import Drawer from '../Drawer';
-import FiltersCheckboxField from '../FiltersCheckboxField';
 import Heading from '../Heading';
 
 export type Props = {
   /**
-   * Checkbox fields that will be displayed in the filters drawer.
+   * Checkboxes or Checkbox fields that will be displayed in the filters drawer.
```
dierat, line 12: Will these always be checkboxes or can they be radios, toggles, etc?

src/components/Filters/Filters.tsx
```diff
@@ -1,22 +1,39 @@
...
+  /**
+   * Button status variant that indicates if filters had been selected.
+   */
+  buttonStatusVariant: 'neutral' | 'brand';
```
dierat, line 16: I found this logic kinda confusing. What if instead of asking for the button status, we instead asked whether any filters have been applied? I think that would be easier to understand and one less bit of logic the dev has to write (ie `buttonStatusVariant={hasFilters ? 'brand' : 'neutral'}`).

src/components/FiltersDrawer/FiltersDrawer.tsx
```diff
@@ -4,55 +4,34 @@ import { useUID } from 'react-uid';
...
-  /**
-   * Sets the filters drawer window open or closed.
-   */
-  isActive: boolean;
-};
-
-export type CheckboxField = {
-  /**
-   * Legend text string that names the fieldset.
+   * Callback called when Clear All button is called.
```
dierat, line 20: Nit: I would avoid using exact button copy because those things shift around over time and you're probably not going to think about updating this comment (ex: "the clear button" instead of "Clear All button").

src/components/FiltersDrawer/FiltersDrawer.tsx
jinlee93: Dang, nice catch, was trying to make the individual buttons conditional as well, will fix

src/components/FiltersDrawer/FiltersDrawer.tsx
jinlee93: Yup, will do that looks cleaner 👍

src/components/Filters/Filters.tsx
jinlee93: `triggerText` works, prop concise yet informative prop names are always great :D

src/components/FiltersDrawer/FiltersDrawer.tsx
jinlee93: Think they could be any of those you mentioned, will update description

src/components/Filters/Filters.tsx
jinlee93: Mmmmmm great suggestion will change to `hasFilters` prop

src/components/FiltersDrawer/FiltersDrawer.tsx
jinlee93: Got it, will update

src/components/Filters/Filters.stories.tsx
```diff
@@ -1,150 +1,240 @@
...
-          { label: 'Filters label 2', value: 'Filters label 2' },
-          { label: 'Filters label 3', value: 'Filters label 3' },
-        ],
-      },
-    ],
-    onClose: (checkedValues) => {
-      console.log(checkedValues);
-    },
+    triggerText: 'Filters',
+    buttonStatusVariant: 'neutral',
```
dierat, line 19: Leftover `buttonStatusVariant` prop.

jinlee93: Great catch, fixed

src/components/Filters/Filters.tsx
```diff
@@ -1,22 +1,43 @@
...
   /**
-   * Checkbox fields that will be displayed in the filters drawer.
+   * Text to be placed in the button that activates the Filters Drawer
    */
-  checkboxFields: CheckboxField[];
+  triggerText?: string;
+  /**
+   * Button status variant that indicates if filters had been selected.
+   */
+  hasFilters?: boolean;
```
dierat, line 16: Maybe this could be more specific and be something like `hasSelectedFilters` or something? I know that makes it a bit longer 🤷

src/components/Filters/Filters.tsx
```diff
@@ -1,22 +1,43 @@
...
+  /**
+   * Input components, input component fields, or relevant information that will be displayed in the filters drawer.
```
dierat, line 18: Sorry to be all nitty with these comments but I think I would assume "input" meant only text input fields as opposed to checkboxes and whatnot. How about "form controls"? Like `Form controls, form fields, or other relevant information that will be displayed in the filters drawer.`.

dierat (approved):

src/components/Filters/Filters.tsx
jinlee93: All good, changed to `hasSelectedFilters`

src/components/Filters/Filters.tsx
jinlee93: No need to be sorry :D 
Definitely a better comment, changed

PR #1245, opened 2022-08-25
chore: rename disabled and error classes conventionally

jinlee93:
### Summary:
- motivation 
  - for input field migration, error and disabled handling is better styled more conventionally to our repo
- execution
  - this renames `eds-is-error` and `eds-is-disabled` which are BM remnants to `--error` and `--disabled` or the psuedoclass`:disabled` and unnests those classes from their stylesheets
  - also un-italicizes placeholder mixin to match mocks
### Test Plan:
- new unit snaps
- no visual regressions

### DIFF
--- a/src/components/FieldNote/FieldNote.module.css
+++ b/src/components/FieldNote/FieldNote.module.css
@@ -12,29 +12,28 @@
   @mixin eds-theme-typography-body-text-xs;
   margin-top: var(--eds-size-half);
   color: var(--eds-theme-color-text-neutral-subtle);
-
-  /**
-     * Field note inside a field that has an error 
-     */
-  &.eds-is-error {
-    color: var(--eds-theme-color-text-utility-error);
-  }
 }
 
 /**
   * Fieldnote icon
   */
-.field-note__icon.field-note__icon {
+.field-note__icon {
   height: var(--eds-size-2);
   width: var(--eds-size-2);
   position: relative;
   bottom: 1px;
+}
 
-  .eds-is-error & {
-    display: inline-block;
-    position: relative;
-    margin-right: var(--eds-size-1);
-  }
+/**
+ * Error variant
+ */
+.field-note--error {
+  color: var(--eds-theme-color-text-utility-error);
+}
+.field-note--error > .field-note__icon {
+  display: inline-block;
+  position: relative;
+  margin-right: var(--eds-size-1);
 }
 
 /**
--- a/src/components/TextInput/TextInput.tsx
+++ b/src/components/TextInput/TextInput.tsx
@@ -65,7 +65,7 @@ export interface Props {
    */
   onChange?: ChangeEventHandler;
   /**
-   * Input pattern
+   * Defines a regular expression that the input's value must match in order for the value to pass constraint validation.
    */
   pattern?: string;
   /**
@@ -120,58 +120,23 @@ export interface Props {
  * ```
  */
 export const TextInput = ({
-  accept,
-  'aria-describedby': ariaDescribedBy,
   className,
   disabled,
   id,
-  inputMode,
   isError,
-  min,
-  max,
-  maxLength,
-  multiple,
-  name,
-  pattern,
-  onChange,
-  placeholder,
-  readOnly,
-  required,
-  type,
-  title,
-  value,
-  defaultValue,
   ...other
 }: Props) => {
   const componentClassName = clsx(
     styles['text-input'],
+    isError && styles['error'],
     className,
-    disabled && styles['eds-is-disabled'],
-    isError && styles['eds-is-error'],
   );
 
   return (
     <input
-      accept={accept}
-      aria-describedby={ariaDescribedBy}
       className={componentClassName}
-      defaultValue={defaultValue}
       disabled={disabled}
       id={id}
-      inputMode={inputMode}
-      max={max}
-      maxLength={maxLength}
-      min={min}
-      multiple={multiple}
-      name={name}
-      onChange={onChange}
-      pattern={pattern}
-      placeholder={placeholder}
-      readOnly={readOnly}
-      required={required}
-      title={title}
-      type={type}
-      value={value}
       {...other}
     />
   );
--- a/src/design-tokens/mixins.css
+++ b/src/design-tokens/mixins.css
@@ -116,29 +116,21 @@
   /**
    * Input error state
    */
-  &.eds-is-error {
+  &.error {
     border-color: var(--eds-theme-color-border-utility-error-strong);
   }
 
   /**
    * Disabled state
    */
-  &.eds-is-disabled {
+  &:disabled {
     background-color: var(--eds-theme-color-background-disabled);
     border-color: var(--eds-theme-color-border-disabled);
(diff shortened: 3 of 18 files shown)


src/components/FieldNote/FieldNote.module.css
```diff
@@ -12,29 +12,28 @@
   @mixin eds-theme-typography-body-text-xs;
   margin-top: var(--eds-size-half);
   color: var(--eds-theme-color-text-neutral-subtle);
-
-  /**
-     * Field note inside a field that has an error 
-     */
-  &.eds-is-error {
```
dierat, line 19: I always found this naming confusing.. it sounds like a utility class.

jinlee93: Yea not sure what the intention was with these

src/upcoming-components/RadioField/RadioField.tsx
```diff
@@ -147,12 +147,11 @@ export const RadioField = ({
 
   const componentClassName = clsx(
     styles['radio-field'],
-    className,
     variant === 'inline' && styles['radio-field--inline'],
     size === 'sm' && styles['radio-field--sm'],
     inverted && styles['radio-field--inverted'],
-    isError && styles['eds-is-error'],
-    disabled && styles['eds-is-disabled'],
+    disabled && styles['radio-field--disabed'],
```
dierat, line 153:
```suggestion
    disabled && styles['radio-field--disabled'],
```

src/design-tokens/mixins.css
```diff
@@ -116,29 +116,21 @@
...
 
-  &::-webkit-input-placeholder {
-    @mixin placeholderStyles;
-  }
-
-  &::-moz-placeholder {
-    @mixin placeholderStyles;
-  }
-
-  &:-ms-input-placeholder {
```
dierat, line 141: Oh yeah I saw these the other day when I was doing the textarea thingy.. Could you test this in Firefox just to make sure `::placeholder` is enough?

src/design-tokens/mixins.css
```diff
@@ -116,29 +116,21 @@
...
+  &::placeholder {
```
dierat, line 133: Could we use a single colon, for consistency?
```suggestion
  &:placeholder {
```

src/design-tokens/mixins.css
jinlee93: Yup works in firefox, and also tested in safari, and confirmed by [caniuse](https://caniuse.com/css-placeholder)

image

src/design-tokens/mixins.css
jinlee93: Think it needs 2 colons [mdn](https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder)

src/upcoming-components/Textarea/Textarea.tsx
```diff
@@ -96,12 +96,7 @@ export const Textarea = ({
   defaultValue,
   ...other
 }: Props) => {
-  const componentClassName = clsx(
-    styles['textarea'],
-    className,
-    disabled && styles['eds-is-disabled'],
-    isError && styles['eds-is-error'],
```
dierat, line 103: Why don't we need error and disabled styling on this anymore?

dierat: Oh nvm you're handling that in the mixin now. Then my question is whether we need to do something with the `isError` and `disabled` props being deconstructed just above here.

src/upcoming-components/CheckboxField/CheckboxField.tsx
```diff
@@ -97,8 +97,6 @@ export const CheckboxField = ({
     variant === 'inline' && styles['checkbox-field--inline'],
     size === 'sm' && styles['checkbox-field--sm'],
     inverted && styles['checkbox-field--inverted'],
-    isError && styles['eds-is-error'],
-    disabled && styles['eds-is-disabled'],
```
dierat, line 101: Also wondering here if we need to do something with `disabled` prop because it doesn't seem to be used anywhere else.

src/upcoming-components/Textarea/Textarea.tsx
jinlee93: We probably do need error and disabled styling on this, will re-append the `error` style but with the syntax here (i.e. `isError && styles['error']`. Disabled will be handled via the pseudoclass `:disabled` so doesn't need a class attached. 

The error/disabled styling in the mixin is not part of this pr, was happening already, just renaming it so it's not `eds-is-error` and `eds-is-disabled`

src/design-tokens/mixins.css
dierat: Awesome, thanks for testing!

src/design-tokens/mixins.css
dierat: Awesome thanks!

src/components/TextInput/TextInput.tsx
```diff
@@ -65,7 +65,7 @@ export interface Props {
    */
   onChange?: ChangeEventHandler;
   /**
-   * Input pattern
+   * Defines a regular expression that the input's value must match in order for the value to pass constraint validation.
```
dierat, line 68: Nice!

src/upcoming-components/CheckboxField/CheckboxField.tsx
jinlee93: Yea this case the class styles have no styling attached so I removed these lines. I can go ahead and removed the disabled prop from this component, or just let it be since it may be required if we decide to move such a component to `components` folder

src/components/TextInput/TextInput.tsx
jinlee93: Copy pasta from mdn lmao

src/upcoming-components/RadioField/RadioField.tsx
jinlee93: Oof, thanks for catching

src/upcoming-components/Textarea/Textarea.tsx
jinlee93: Yup that worked:

Screen Shot 2022-08-26 at 9 53 22 AM
Screen Shot 2022-08-26 at 9 53 26 AM

src/upcoming-components/CheckboxField/CheckboxField.tsx
dierat: Maybe leave a todo comment on the disabled prop that the styling for that prop needs to be added?

jinlee93: That makes sense, will do

src/upcoming-components/Textarea/Textarea.tsx
dierat: Nice!

src/upcoming-components/CheckboxField/CheckboxField.tsx
dierat: Awesome, thanks!

dierat (approved):

PR #1246, opened 2022-08-26
feat(textinput, textfield): migrate components from EDSCandidates

jinlee93:
### Summary:
[sc-211843]
- motivation
  - there is a need for text input components in upcoming pilots
  - impending need to be used in search component
- execution
  - currently raised against another branch that will help cleanup and make this pr more readable
  - removes `inverted` label prop and `inverted` features of `<TextField>` and other `upcoming-components` since there are no designs for `inverted` variants and there is no plan to support immediately, should consider removing `inverted` support in other components
  - modifies Label and Fieldnote components to continue to be used as subcomponents of `<TextField>`
    - as mentioned in checkbox pr, Label, and now Fieldnote may possibly all be consolidated into `Text`, [should explore ticket](https://app.shortcut.com/czi-edu/story/212992) 
### Test Plan:
- axe tests pass
- no visual regression outside impacted
- unit tests pass

Screen Shot 2022-08-26 at 12 42 32 AM
Screen Shot 2022-08-26 at 12 42 34 AM
Screen Shot 2022-08-26 at 12 42 42 AM
Screen Shot 2022-08-26 at 12 42 44 AM
Screen Shot 2022-08-26 at 12 42 46 AM
Screen Shot 2022-08-26 at 12 42 47 AM
Screen Shot 2022-08-26 at 12 42 51 AM
Screen Shot 2022-08-26 at 12 42 57 AM
Screen Shot 2022-08-26 at 12 42 59 AM
Screen Shot 2022-08-26 at 12 43 01 AM
Screen Shot 2022-08-26 at 12 43 08 AM
Screen Shot 2022-08-26 at 12 43 09 AM


dierat:
>removes inverted label prop and inverted features of <TextField> and other upcoming-components since there are no designs for inverted variants and there is no plan to support immediately, should consider removing inverted support in other components

Do you think we should go through and clear all that stuff out?

jinlee93:
> > removes inverted label prop and inverted features of  and other upcoming-components since there are no designs for inverted variants and there is no plan to support immediately, should consider removing inverted support in other components
> 
> Do you think we should go through and clear all that stuff out?

Yea, made [ticket](https://app.shortcut.com/czi-edu/story/213036)

src/components/TextInput/TextInput.tsx
```diff
@@ -110,14 +110,16 @@ export interface Props {
...
+};
 
 /**
  * BETA: This component is still a work in progress and is subject to change.
  *
  * ```ts
  * import {TextInput} from "@chanzuckerberg/eds";
  * ```
+ *
+ * Text input component for one line of text. For multiple lines, consider the Textarea upcoming-component.
```
dierat, line 122: I'd prefer it if we didn't specify if some other component is in the `upcoming-components` directory because, when we move that component over, we're probably not going to remember to come back and update this comment.

jinlee93: Ah that makes sense, would changing it to just `consider the Textarea component` be more future proof?

src/components/FieldNote/FieldNote.tsx
```diff
@@ -36,34 +40,33 @@ export interface Props {
...
  * import {FieldNote} from "@chanzuckerberg/eds";
  * ```
+ *
+ * Fieldnote component wraps text to describe other components.
  */
 export const FieldNote = ({
   className,
+  disabled,
   id,
-  iconName,
```
dierat, line 43: Should we also remove this prop? From the code, it looks like this is now being passed down into the wrapper `div` via the `other` variable..

src/components/Label/Label.tsx
```diff
@@ -42,7 +38,7 @@ export interface Props {
   /**
    * The label text string
    */
-  text?: string | boolean;
```
dierat, line 45: The text couldn't been a `bool`? How odd..

jinlee93: Oh nice catch, thought I changed this to `string`, will change to `string`

src/components/TextField/TextField.tsx
```diff
@@ -151,35 +130,35 @@ export interface Props {
...
+  if (process.env.NODE_ENV !== 'production' && !label && !other['aria-label']) {
+    throw new Error('You must provide a visible label or aria-label');
+  }
+
+  const shouldRenderOverline = label || required || optionalIndicator;
+  const indicator = required
+    ? 'Required'
+    : optionalIndicator
+    ? 'Optional'
+    : null;
```
dierat, line 155: Small thing but it looks like if you move the `shouldRenderOverline` down under `indicator` you could check for `label || indicator` since `indicator` checks for `required` and `optionalIndicator`.

```
  const indicator = required
    ? 'Required'
    : optionalIndicator
    ? 'Optional'
    : null;
  const shouldRenderOverline = label || indicator;
```
I'm also wondering if you should make `shouldRenderOverline` a boolean like
```
  const shouldRenderOverline = !!(label || indicator);
```
because it looks like right now it's a string?

dierat (approved):
LGTM!

src/components/Label/Label.tsx
dierat: No you did; it's good now. I'm just confused about why they had it like that before 🤔

src/components/TextInput/TextInput.tsx
dierat: Yeah that's what I'm thinking; just drop the `upcoming-` part 👍

src/components/FieldNote/FieldNote.tsx
jinlee93: Yup will remove

src/components/TextField/TextField.tsx
jinlee93: Ohhh great suggestions, will adopt

PR #1243, opened 2022-08-25
feat(radio): migrate radio from EDScandidates and match mocks

jinlee93:
### Summary:
[sc-211844]
mocks: https://www.figma.com/file/Ff7hJoSvXX9hkAmIklnsT2/EDS-Paper-v1.0-%7C-UI-Components?node-id=2453%3A12292
- motivation
  - for 1.1 release, and for better support radio needs to be brought from EDSCandidates
  - realized common code for label, so separated that to another component `InputComponent`
    - TODO: a `Label` component already exists, but brought over from BM and many of its functionality is not useful to us, should consolidate to one component
    - made `text` into `children` to follow other component conventions
  - the `RadioField` component I feel like has little use and seems there is more favor towards the Headless Radio Group regarding that usage, so I chose not to bring it over
    - radio state management can be controlled with an array of booleans similar to checkbox anyways
- execution
  - internally, very similar to checkbox
    - there is potential here to make more common components, but not sure if they will be of use since no other input components may use them
  - uses the svg icon for radio over border to keep similar to checkbox
  - renames size variants sm -> md and md -> lg for checkbox (affects label only) to match mocks and keep consistent with radio
### Test Plan:
- unit tests pass
- no visual regression

Screen Shot 2022-08-25 at 9 08 32 AM
Screen Shot 2022-08-25 at 9 08 34 AM
Screen Shot 2022-08-25 at 9 08 36 AM
Screen Shot 2022-08-25 at 9 08 38 AM
Screen Shot 2022-08-25 at 9 08 40 AM

PR #1247, opened 2022-08-26
chore: remove mixin import if mixins aren't being used

dierat:
### Summary:
In https://github.com/chanzuckerberg/edu-design-system/pull/1243 @jinlee93 removed an unused mixin import in the `Checkbox` CSS module file, which made me realize that actually there are a bunch of these imports that aren't being used.

In this PR, I'm removing all the unnecessary mixin imports in CSS module files and also removing it from the plop generator template.

### Test Plan:
There should be no chromatic changes.


src/components/Card/Card.module.css
```diff
@@ -20,7 +18,6 @@
   background-color: var(--eds-theme-color-background-neutral-default);
 }
 
-
```
dierat, line 23: This and one other spacing change in the PR were automatically made by prettier.

jinlee93 (approved):
Thanks for the cleanup!

PR #1250, opened 2022-08-29
chore: use focus-visible instead of focus

dierat:
### Summary:
We got a comment that link focus styles seem "sticky" because after you click on them as a mouse user, the link stays focused until you click on something else. This didn't really bother people until we added the link focus style, which is a solid purple background. But if we're going to change that, we should just change all interactive elements to work the same way.

So in this PR, I'm changing `&:focus` styles to instead use `&:focus-visible`. We already have a few instances of this in the codebase, but now it will be universal. I also added a line to the documentation to this effect.

### Test Plan:
I went through all of the interactive components and recipes (but not pages) and verified that interactive elements shows a focus styles on keyboard interaction and no focus styles when clicked with mouse/trackpad.

(I didn't test pages because they're made up of components I'd already tested. I probably didn't need to test recipes either but I wanted to try to be thorough.)

### DIFF
--- a/docs/CODE_GUIDELINES.md
+++ b/docs/CODE_GUIDELINES.md
@@ -113,7 +113,7 @@ EDS uses [PostCSS Nested](https://github.com/postcss/postcss-nested) to provide 
   background: var(--eds-theme-color-background-brand-primary-strong);
 
   &:hover,
-  &:focus {
+  &:focus-visible {
     background: var(--eds-theme-color-background-brand-primary-strong-hover);
   }
 
@@ -269,6 +269,30 @@ Use:
 import { EdsThemeColorUtilitySuccessForeground } from 'src/tokens-dist/ts/colors';
 
 <Icon color={EdsThemeColorUtilitySuccessForeground} />;
+```
+
+- Use `:focus-visible` instead of `:focus`
+
+This allows focus indicators to appear for keyboard users while remaining hidden from mouse-only users.
+
+Instead of:
+
+```scss
+.button--primary {
+  &:focus {
+    @mixin focus;
+  }
+}
+```
+
+Use:
+
+```scss
+.button--primary {
+  &:focus-visible {
+    @mixin focus;
+  }
+}
 ```
 
 ## Utility classes <a name="utility-classes"></a>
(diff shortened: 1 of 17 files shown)


dierat:

dierat:
https://app.shortcut.com/czi-edu/story/211854/link-polish

jinlee93:
Sweet, and I also read that `:focus-visible` should be preferred over `:focus`, but it was some [random blog](https://pawelgrzybek.com/the-difference-between-css-focus-and-focus-visible-pseudo-class/#css-focus-vs-focus-visible) lol

jinlee93 (approved):
LGTM, thanks!

docs/CODE_GUIDELINES.md
```diff
@@ -271,6 +271,30 @@ import { EdsThemeColorUtilitySuccessForeground } from 'src/tokens-dist/ts/colors
 <Icon color={EdsThemeColorUtilitySuccessForeground} />;
 ```
 
+- Use `:focus-visible` instead of `:focus`
```
ahuth, line 274: According to MDN, `focus-visible` isn't supported by our currently supported browsers: https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#browser_compatibility

So unfortunately, I don't think it makes sense to use this, yet.

We could use https://github.com/WICG/focus-visible in LP (like we do in Along). Not sure of the best way for EDS to tie into that, though.

dierat: Woo, nice catch! So instead I should turn the `focus-visible` selectors into `focus`, right?

ahuth: I'm not sure. Do you mean going back to `:focus`, add https://github.com/WICG/focus-visible in LP, and it should just work? I think so 🤔

dierat: > I'm not sure. Do you mean going back to `:focus`, add https://github.com/WICG/focus-visible in LP, and it should just work? I think so 🤔

Yes but also turn the usage of `focus-visible` in EDS to `focus` –  the Big Medium engineers were using `focus-visible`, so we have a mixture of focus implementation methods right now.

ahuth: Ah, gotcha. Yeah, changing any focus-visible-es back to focus makes sense 👍

dierat: [PR for using :focus instead of :focus-visible](https://github.com/chanzuckerberg/edu-design-system/pull/1268)

PR #1251, opened 2022-08-29
chore(deps): remove abandoned cz-conventional-changelog dep

jinlee93:
### Summary:
- removes [cz-convetional-changelog](https://github.com/commitizen/cz-conventional-changelog) which seems abandoned
  - we don't use `npx cz` or `git cz` to make commits
  - stops dependabot alerts regarding [minimist](https://github.com/chanzuckerberg/edu-design-system/security/dependabot/17)
- doesn't seem to affect standard-version:
Screen Shot 2022-08-29 at 10 53 45 AM

### Test Plan:
- standard-version works
- no test fails in anything

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,7 +130,6 @@
     "chromatic": "^6.5.3",
     "codecov": "^3.8.3",
     "copyfiles": "^2.4.1",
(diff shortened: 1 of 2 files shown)


dierat (approved):
Nice find!

PR #1252, opened 2022-08-29
chore(deps): patch bumps for sb deps and minor bump to 6.8 for chromatic

jinlee93:
### Summary:
- keep storybook and storybook deps updated
- chromatic has jumps 2 minor versions, but looking at [changelog](https://github.com/chromaui/chromatic-cli/blob/v6.8.0/CHANGELOG.md), we should not be affected by the deprecation of `--preserve-missing` flag
### Test Plan:
- chromatic works locally
- chromatic visual regression testing and hosting works
- storybook works locally
- storybook build works and confirmed via chromatic storybook hosting


dierat (approved):
Nice! Thanks! ❤️

PR #1254, opened 2022-08-30
feat(searchbar): add SearchBar component

jinlee93:
### Summary:
[sc-211845]
- motivation
  - use cases of a SearchBar component is unique enough from the TextField component where the biggest difference is that it is likely not used in forms or submission of data, encompassing:
    - does not need such features such as required flagging, error handling, etc (yet)
    - functionality can be possibly be expanded via predictive text, etc (not in scope)
  - also used in upcoming pilots
- execution
  - created subcomponents SearchField and SearchButton wrapped by SearchBar
  - as a result, events can be controlled by owning component
### Test Plan:
- unit snaps written
- no visual regression affecting other components

Screen Shot 2022-08-29 at 6 14 30 PM
Screen Shot 2022-08-29 at 6 14 32 PM
Screen Shot 2022-08-29 at 6 14 36 PM
Screen Shot 2022-08-29 at 6 14 41 PM
Screen Shot 2022-08-29 at 6 14 45 PM


src/components/SearchBar/SearchBar.tsx
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * TODO: update this comment with a description of the component.
+ */
+export const SearchBar = ({ children, className }: Props) => {
+  const componentClassName = clsx(styles['search-bar'], className);
+
+  return <div className={componentClassName}>{children}</div>;
+};
+
+SearchBar.InputField = SearchField;
```
jinlee93, line 33: Not sure what to name this re-export

src/components/SearchButton/SearchButton.tsx
```diff
@@ -0,0 +1,38 @@
+import clsx from 'clsx';
+import React, { MouseEventHandler } from 'react';
+import styles from './SearchButton.module.css';
+import Button from '../Button';
+
+export type SearchButtonProps = {
```
jinlee93, line 6: Is there value in just using `Button` props? Or would that be not advised since we don't other props such as `variant` and `status` to be used here?

src/components/SearchBar/SearchBar.module.css
```diff
@@ -0,0 +1,11 @@
+/*------------------------------------*\
+    # SEARCH FIELD
+\*------------------------------------*/
+
+/**
+ * Search bar wrapper for the SearchField and SearchButton
+ */
+.search-bar {
+  display: flex;
+  gap: var(--eds-size-1);
```
dierat, line 10: How does this look on mobile? Do you think we should consider wrapping the button to the next line on narrower windows?

src/components/SearchButton/SearchButton.module.css
```diff
@@ -0,0 +1,11 @@
+/*------------------------------------*\
+    # SEARCH BUTTON
+\*------------------------------------*/
+
+/**
+ * Button to trigger search in a Search Bar.
+ */
+.search-button {
+  /* Matches height with input field */
+  height: calc(var(--eds-size-5) + 2 * var(--eds-theme-form-border-width));
```
dierat, line 10: Could we use `box-sizing: content-box` instead of `+ 2 * var(--eds-theme-form-border-width))`?

src/components/SearchButton/SearchButton.tsx
dierat: With so few props with such generic types, I feel like this is easier.

src/components/SearchBar/SearchBar.module.css
jinlee93: Oh good question, I'll ask design how they want it to look

src/components/SearchButton/SearchButton.module.css
jinlee93: Didn't work but `height: auto` did!

dierat (approved):
LGTM!

src/components/SearchBar/SearchBar.module.css
jinlee93: How it looks on mobile: image

And confirmed with design that this is how they want it

dierat: Awesome, thanks!

PR #1255, opened 2022-08-30
fix(fieldnote): match error icon color to mocks

jinlee93:
### Summary:
- icon was not properly receiving color token
- fixes css selection to receive color token
### Test Plan:
- unit tests pass
- visual regression on error icon color only

image


dierat (approved):
Nice!

PR #1256, opened 2022-08-30
Add back tailwind utilities for stories & recipes

anniehu4:
### Summary:
[ch213288]

Tailwind utilities are easier to use than the BM-defined ones, and especially for reference pages & recipes, are easier for app developers to copy & paste in.

Note that this matches our behavior before BM -- it's a separate question whether & how to use utilities in component styles themselves.

### Test Plan:
This PR has 5 commits:
1: Adds back all the necessary config stuff to get Tailwind working again
2-4: Converts styles in each of the types of files we want to support Tailwind utilities in to make sure it's working.
5: Turn off preflight reset styles to ensure no visual changes

I'd also expect to see no changes in chromatic

### DIFF
--- a/.storybook/pages/FeedbackOverview/FeedbackOverview.tsx
+++ b/.storybook/pages/FeedbackOverview/FeedbackOverview.tsx
@@ -31,7 +31,6 @@ import {
 } from '../../../src/tokens-dist/ts/colors';
 import NumberIconList from '../../recipes/NumberIconList';
 import { PageShell } from '../../recipes/PageShell/PageShell';
-import '../../../src/components/Utilities/Spacing.css';
 
 export interface Props {
   /**
@@ -42,15 +41,15 @@ export interface Props {
 
 export const FeedbackOverview = ({ activeIndex = 0 }: Props) => (
   <PageShell className="body--alternate">
-    <Breadcrumbs className="u-margin-bottom-md">
+    <Breadcrumbs className="mb-4">
       <BreadcrumbsItem href="#" text="My Courses" />
       <BreadcrumbsItem href="#" text="Modern World 2" />
     </Breadcrumbs>
     <PageHeader
-      className="u-margin-bottom-xl u-flex-direction-row"
+      className="!mb-8 !flex-row"
       right={
         <Button status="neutral" variant="icon">
-          <div className="u-mobile-hidden">More options</div>
+          <div className="sr-only lg:not-sr-only">More options</div>
           <Icon name="dots-vertical" purpose="decorative" />
         </Button>
       }
--- a/package.json
+++ b/package.json
@@ -17,19 +17,20 @@
     "url": "https://github.com/chanzuckerberg/edu-design-system/issues"
   },
   "files": [
-    "/lib"
+    "/lib",
+    "tailwind.config.js"
   ],
   "scripts": {
     "build": "yarn build:clean && yarn build:tokens && yarn build:js && yarn build:styles && yarn copy-to-lib",
     "build:clean": "rm -rf lib/",
-    "build:tokens": "rm -rf src/tokens-dist/ && node ./style-dictionary.config.js",
+    "build:tokens": "rm -rf src/tokens-dist/ && node ./style-dictionary.config.js && yarn copy-tokens-to-lib",
     "build:icons": "svg-sprite-generate -d src/icons/ -o src/icons/spritemap/spritemap.svg && yarn generate-icon-types",
     "build:js": "tsc --project tsconfig.build.json",
     "build:storybook": "build-storybook -o storybook-static -s src/design-tokens/tier-1-definitions/fonts",
     "build:styles": "postcss \"src/components/**/*.css\" --dir lib/ --base src/ --verbose",
     "chromatic": "chromatic",
     "commit": "./node_modules/.bin/cz",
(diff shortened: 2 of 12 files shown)


.storybook/pages/FeedbackOverview/FeedbackOverview.tsx
```diff
@@ -42,15 +41,15 @@ export interface Props {
...
 export const FeedbackOverview = ({ activeIndex = 0 }: Props) => (
   <PageShell className="body--alternate">
-    <Breadcrumbs className="u-margin-bottom-md">
+    <Breadcrumbs className="mb-4">
       <BreadcrumbsItem href="#" text="My Courses" />
       <BreadcrumbsItem href="#" text="Modern World 2" />
     </Breadcrumbs>
     <PageHeader
-      className="u-margin-bottom-xl u-flex-direction-row"
+      className="!mb-8 !flex-row"
```
anniehu4, line 49: these need to be more specific than `PageHeader` default styles

package.json
```diff
@@ -17,10 +17,11 @@
     "url": "https://github.com/chanzuckerberg/edu-design-system/issues"
   },
   "files": [
-    "/lib"
+    "/lib",
+    "tailwind.config.js"
   ],
   "scripts": {
-    "build": "yarn build:clean && yarn build:tokens && yarn build:js && yarn build:styles && yarn copy-to-lib",
+    "build": "yarn build:clean && yarn build:tokens && yarn copy-to-lib && yarn build:js && yarn build:styles",
```
anniehu4, line 24: to extend this `tailwind.config.js` file in Traject, it needs to reference `lib/tokens` from the built output. That folder gets copied over during `yarn copy-to-lib`, so we now need to move the copy step before `yarn build:styles` happens. Otherwise `yarn build:styles` complains that the folder doesn't exist

postcss.config.js
```diff
@@ -6,7 +6,8 @@ module.exports = {
       addModulesDirectories: ['node_modules'],
     }),
     require('postcss-mixins'),
-    require('postcss-nested'),
     require('postcss-simple-vars')({ variables: breakpoints }),
+    require('tailwindcss/nesting'),
```
anniehu4, line 10: tailwind's rec for nesting: https://tailwindcss.com/docs/using-with-preprocessors#nesting. it uses postcss-nested under the hood

tailwind.config.js
```diff
@@ -0,0 +1,50 @@
+const variableTokens = require('./lib/tokens/json/css-variables-nested.json');
+const staticTokens = require('./lib/tokens/json/variables-nested.json');
+
+module.exports = {
```
anniehu4, line 4: copied from LP (https://github.com/FB-PLP/traject/blob/master/tailwind.config.js), minus the `extend` block (not sure those are valid for EDS anymore). after our next publish we probably want update traject's config to read from this file again

package.json
```diff
@@ -17,19 +17,20 @@
...
   "files": [
-    "/lib"
+    "/lib",
+    "tailwind.config.js"
   ],
   "scripts": {
     "build": "yarn build:clean && yarn build:tokens && yarn build:js && yarn build:styles && yarn copy-to-lib",
     "build:clean": "rm -rf lib/",
-    "build:tokens": "rm -rf src/tokens-dist/ && node ./style-dictionary.config.js",
+    "build:tokens": "rm -rf src/tokens-dist/ && node ./style-dictionary.config.js && yarn copy-tokens-to-lib",
```
anniehu4, line 26: to extend the tailwind.config.js file in Traject, it needs to reference `lib/tokens` from the built output. this makes sure the folder gets copied over as part of token building

postcss.config.js
```diff
@@ -1,12 +1,15 @@
 const breakpoints = require('./src/design-tokens/tier-1-definitions/breakpoints');
 
 module.exports = {
-  plugins: [
-    require('postcss-import')({
+  plugins: {
```
anniehu4, line 4: updating to a more modern syntax

tailwind.config.js
```diff
@@ -0,0 +1,61 @@
...
+  corePlugins: {
+    /**
+     * Disable Tailwind's preflight (CSS reset) styles.
+     * We're only using Tailwind for stories/recipes, and some of those
+     * (esp. documentation pages) depend on browser default styles.
+     *
+     * Remove this setting if we are turning on Tailwind for component styles,
+     * so we will better match consuming apps.
+     */
+    preflight: false,
```
anniehu4, line 24: Unfortunately, it seems some of our stories depend on browser default styles 🤔  See https://www.chromatic.com/build?appId=61313967cde49b003ae2a860&number=2504. It's not the biggest deal to have preflight turned off because
1. It preserves how the stories look rn
2. Technically, having preflight on would make our story environment closer to traject. But traject has its own bootstrap nonsense that affects how things look there anyways

It'd be good to eventually turn it on / work through the needed story changes, but not high pri

jinlee93: Created [backlog ticket](https://app.shortcut.com/czi-edu/story/213428) for preflight/browser default styles

jinlee93 (approved):
Makes sense to me. Moving forward I already know there's tickets to remove utility classes and add page specific styling. Should there also be a low-pri effort to remove css module sheets if possible since our [styling guides](https://czi.atlassian.net/wiki/spaces/ETE/pages/2322268183/Styling) suggest that we prefer styling over css modules?

Also curious how much bigger the EDS bundle becomes and wondering if it affected storybook dev build times

PR #1257, opened 2022-08-30
feat(tooltip): add childDisabled prop

dierat:
### Summary:
Our `Tooltip` component currently doesn't work on disabled buttons (which is a relatively common pattern). The reason is, `Tippy` only worked on elements that have a `tabIndex`, which disabled elements don't have. The solution is to wrap the children in a `span` with a `tabIndex` of 0.

I was originally going to add documentation and a story to this effect, but then I noticed the `react-bootstrap` `OverlayTrigger` component has a `childDisabled` prop. If we use a prop like that, we can wrap the children in a `span` with a `tabIndex` so our users don't have to. (I also noticed some weirdness around needing a different `display` styling for top and bottom alignment, so we can bake that in too.

### Test Plan:
I verified the tooltip looked okay with all 4 alignments in storybook.

### DIFF
--- a/src/components/Tooltip/Tooltip.stories.tsx
+++ b/src/components/Tooltip/Tooltip.stories.tsx
@@ -184,7 +184,12 @@ export const InteractiveDisabled: StoryObj<Args> = {
         styles['trigger--spacing-left'],
       )}
     >
-      <Tooltip align="top" childDisabled={true} text={defaultArgs.text}>
+      <Tooltip
+        align="top"
+        childDisabled={true}
+        duration={0}
+        text={defaultArgs.text}
+      >
         <Button disabled variant="primary">
           Tooltip trigger
         </Button>
--- a/src/components/Tooltip/Tooltip.stories.tsx
+++ b/src/components/Tooltip/Tooltip.stories.tsx
@@ -177,7 +177,10 @@ export const Interactive: StoryObj<Args> = {
 };
 
 export const InteractiveDisabled: StoryObj<Args> = {
-  render: () => (
+  args: {
+    duration: undefined,
+  },
+  render: (args) => (
     <div
       className={clsx(
         styles['trigger--spacing-top'],
@@ -187,7 +190,7 @@ export const InteractiveDisabled: StoryObj<Args> = {
       <Tooltip
         align="top"
         childDisabled={true}
-        duration={0}
+        duration={args.duration}
         text={defaultArgs.text}
       >
         <Button disabled variant="primary">
--- a/src/components/Tooltip/Tooltip.stories.tsx
+++ b/src/components/Tooltip/Tooltip.stories.tsx
@@ -134,6 +134,19 @@ export const LongButtonText: StoryObj<Args> = {
   },
 };
 
+export const Disabled: StoryObj<Args> = {
+  args: {
+    children: (
+      // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+      <span className={clsx(styles['trigger--spacing'])} tabIndex={0}>
+        <Button disabled variant="primary">
+          Tooltip trigger
+        </Button>
+      </span>
+    ),
+  },
+};
+
 export const Interactive: StoryObj<Args> = {
   args: {
     // reset prop values defined in defaultArgs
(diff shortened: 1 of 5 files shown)


src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -134,6 +134,28 @@ export const LongButtonText: StoryObj<Args> = {
   },
 };
 
+export const Disabled: StoryObj<Args> = {
+  render: () => (
+    <div
+      className={clsx(
+        styles['trigger--spacing-top'],
+        styles['trigger--spacing-left'],
+      )}
```
dierat, line 143: One noteworthy thing here is any spacing/placement styling has to be placed on a wrapper around the `Tooltip`, not the `children`, because of the new wrapper. That could be pretty confusing for people.

dierat:
https://app.shortcut.com/czi-edu/story/208461/tooltip-doesn-t-work-on-disabled-buttons

jinlee93 (approved):
This makes sense, thanks for adding the disabled variant

jinlee93:
Pressing escape on the disabled tooltip causes the message to disappear, which is fine if expected behavior.

Is it possible to write a unit test for the disabled tooltip behavior?

dierat:
> Pressing escape on the disabled tooltip causes the message to disappear, which is fine if expected behavior.

Isn't that the expected behavior for all popups?

>Is it possible to write a unit test for the disabled tooltip behavior?

I added a unit test, but for some reason the escape functionality actually isn't working in there (even though it does seem to be when testing manually) 🤔

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -148,17 +169,34 @@ export const Interactive: StoryObj<Args> = {
...
+        <p>Hover over the button to make the tooltip appear.</p>
         <Story />
       </div>
     ),
   ],
-  play: async ({ canvasElement }) => {
-    const canvas = within(canvasElement);
-    const trigger = await canvas.findByRole('button');
-    await userEvent.hover(trigger);
-  },
```
dierat, line 163: I removed the play method because it didn't seem to be used anywhere, unless it was so the tooltip appeared on page load?

src/components/Tooltip/Tooltip.test.tsx
```diff
@@ -29,6 +29,18 @@ describe('<Tooltip />', () => {
...
+  it('should close tooltip via escape key for disabled buttons', async () => {
+    // disable animation for test
+    render(<InteractiveDisabled duration={0} />);
+    const trigger = await screen.findByTestId('disabled-child-tooltip-wrapper');
+    expect(screen.queryByTestId('tooltip-content')).not.toBeInTheDocument();
+    await userEvent.hover(trigger);
+    expect(screen.getByTestId('tooltip-content')).toBeInTheDocument();
+    // TODO: make worky
+    // await userEvent.keyboard('{esc}');
+    // expect(screen.queryByTestId('tooltip-content')).not.toBeInTheDocument();
```
dierat, line 41: This is the test that isn't working 🤔 Could it be something about using a keyboard event with a span? That's the only thing I can think of..

src/components/Tooltip/Tooltip.stories.tsx
jinlee93: The intention was to capture the message being open in an interactive story in Chromatic but seeing as the stories are flaky, think we can remove the play method as mentioned. 

Can you change the button copy to `Tooltip trigger` or something relevant?

src/components/Tooltip/Tooltip.test.tsx
jinlee93: You have to pass `duration` as an arg to the story i.e.:

```
render: (args) => (
...
        <Tooltip
          duration={args.duration}
...
```

dierat: Oh nice! Thanks! ❤️

src/components/Tooltip/Tooltip.stories.tsx
dierat: You got it 👍

src/components/Tooltip/Tooltip.stories.tsx
dierat: I added some info about this to the prop comment.

jinlee93 (approved):

PR #1258, opened 2022-08-31
feat(datasummarycard): adds DataSummaryCard recipe

jinlee93:
### Summary:
[sc-211849]
- motivation
  - Data Summary Card is used in the school and district tools pilot
  - Data Summary Card displays simple data of students to provide status at a quick glance
    - Communicates 'off-track' status if necessary
- execution
  - Wraps the `Card` component
  - Relevant mocks: [Recipe](https://www.figma.com/file/CQu6m6sxvFO0zYSrWq6jsH/Data-Summary-Card---Creation-Exploration?node-id=5961%3A12067), [usage in poc](https://www.figma.com/file/PJ6pnQy2J68vLiYJwTCy2y/SDT-Course-%3E-Students-Proof-of-Concept-Pilot?node-id=158%3A20107)
### Test Plan:
- unit tests pass
- no visual regressions
Screen Shot 2022-08-30 at 8 30 19 PM
Screen Shot 2022-08-30 at 8 30 21 PM
Screen Shot 2022-08-30 at 8 30 23 PM
Screen Shot 2022-08-30 at 8 30 25 PM

### DIFF
--- a/.storybook/recipes/DataSummaryCard/DataSummaryCard.module.css
+++ b/.storybook/recipes/DataSummaryCard/DataSummaryCard.module.css
@@ -1,0 +1,106 @@
+@import '../../../src/design-tokens/mixins.css';
+
+/*------------------------------------*\
+    # DATA SUMMARY CARD
+\*------------------------------------*/
+
+/**
+ * Data Summary Card centers and wraps header, data, and footnote.
+ */
+.data-summary-card {
+  width: 12.75rem;
+  height: 9.25rem;
+  position: relative;
+  overflow: hidden;
+
+  border: 0;
+  box-shadow: var(--eds-box-shadow-sm);
+
+  align-items: center;
+}
+.data-summary-card:hover {
+  border: var(--eds-border-width-sm) solid
+    var(--eds-theme-color-border-neutral-default);
+}
+
+/**
+ * The header wraps the title.
+ */
+.data-summary-card__header {
+  margin-top: var(--eds-size-2);
+  margin-bottom: var(--eds-size-half);
+}
+
+/**
+ * Prevents jitter of the card components when hovering due to border spawning.
+ * Transparent border can't be used due to off-track variant indicator needing full width.
+ */
+.data-summary-card:hover > .data-summary-card__header {
+  margin-top: calc(var(--eds-size-2) - var(--eds-border-width-sm));
+}
+
+/**
+ * The title communicates card organization.
+ */
+.data-summary-card__title {
+  @mixin eds-theme-typography-overline-sm;
+}
+
+/**
+ * The body wraps data and description text.
+ */
+.data-summary-card__body {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+}
+
+/**
+ * The data is the main content for the card.
+ */
+.data-summary-card__data {
+  @mixin eds-theme-typography-headline-lg;
+
+  display: flex;
+  align-items: center;
+  gap: var(--eds-size-half);
+
+  margin-bottom: var(--eds-size-half);
+}
+
+/**
+ * The data unit provides a unit of measurement for the data to make it relevant.
+ */
+.data-summary-card__data-unit {
+  @mixin eds-theme-typography-headline-sm;
+}
+
+/**
+ * The Description provides more elucidation for the data.
+ */
+.data-summary-card__description {
+  @mixin eds-theme-typography-caption-text-sm;
+  font-weight: var(--eds-font-weight-light);
+}
+
+/**
+ * Off track variant, bottom orange border indicates off track status.
+ */
+.data-summary-card__indicator--off-track {
+  width: 100%;
+  height: 8px;
+
+  position: absolute;
+  bottom: 0;
+  left: 0;
+
+  background-color: var(--eds-theme-color-icon-utility-warning);
+}
+
+/**
+ * Prevents jitter of the card components when hovering due to border spawning.
+ * Transparent border can't be used due to off-track variant indicator needing full width.
+ */
+.data-summary-card:hover > .data-summary-card__indicator--off-track {
+  height: 7px;
+}
--- a/.storybook/recipes/DataSummaryCard/DataSummaryCard.tsx
+++ b/.storybook/recipes/DataSummaryCard/DataSummaryCard.tsx
@@ -12,7 +12,7 @@ export interface Props {
   /**
    * Main data text of the card.
    */
-  data?: string;
+  dataAmount: string;
   /**
    * Text that provides a unit of measurement for the data.
    */
@@ -20,10 +20,9 @@ export interface Props {
   /**
    * Text to provide more context for the data.
    */
-
   description?: string;
   /**
-   * Title text of the data to represented.
+   * Title text of the data represented.
    */
   title: string;
   /**
@@ -37,7 +36,7 @@ export interface Props {
  */
 export const DataSummaryCard = ({
   className,
-  data,
+  dataAmount,
   dataUnit,
   description,
   title,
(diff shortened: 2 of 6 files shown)


.storybook/recipes/DataSummaryCard/DataSummaryCard.module.css
```diff
@@ -0,0 +1,106 @@
...
+ * The header wraps the title.
+ */
+.data-summary-card__header {
+  margin-top: var(--eds-size-2);
+  margin-bottom: var(--eds-size-half);
+}
+
+/**
+ * Prevents jitter of the card components when hovering due to border spawning.
+ * Transparent border can't be used due to off-track variant indicator needing full width.
```
jinlee93, line 36: See comment on the other `:hover` block

.storybook/recipes/DataSummaryCard/DataSummaryCard.module.css
```diff
@@ -0,0 +1,106 @@
...
+  background-color: var(--eds-theme-color-icon-utility-warning);
+}
+
+/**
+ * Prevents jitter of the card components when hovering due to border spawning.
+ * Transparent border can't be used due to off-track variant indicator needing full width.
+ */
+.data-summary-card:hover > .data-summary-card__indicator--off-track {
+  height: 7px;
+}
```
jinlee93, line 106: Please advise on a lighter way to prevent jitter if possible.
Tried doing `border-color: transparent` and just `border-bottom: 8px var(--eds-theme-color-orange)` but there's noticeable side borders 
image

.storybook/recipes/DataSummaryCard/DataSummaryCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+   */
+  className?: string;
+  /**
+   * Main data text of the card.
+   */
+  data?: string;
+  /**
+   * Text that provides a unit of measurement for the data.
+   */
+  dataUnit?: string;
```
jinlee93, line 19: Should I just make this `children: ReactNode`? Or just the data prop into children: `children?: ReactNode; dataUnit?: string`?

My justification for having this kind of api is that the two props `data` and `dataUnit` need to be styled differently

src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -11,7 +11,7 @@ const defaultArgs = {
 };
 
 export default {
-  title: 'Molecules/Forms/Checkbox',
+  title: 'Atoms/Forms/Checkbox',
```
jinlee93, line 14: Moving this to correct storybook folder, not related to DataSummaryCard

plop-templates/Recipe/Recipe.tsx.hbs
```diff
@@ -16,8 +16,7 @@ export const {{pascalCase name}} = ({
   className,
   ...other
 }: Props) => {
-  const componentClassName = clsx(styles['{{dashCase name}}'], className, {
-  });
+  const componentClassName = clsx(styles['{{dashCase name}}'], className);
```
jinlee93, line 19: Reordering import according to lint and removing empty block in `clsx()`, not related to DataSummaryCard

.storybook/recipes/DataSummaryCard/DataSummaryCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+  return (
+    <Card className={componentClassName} {...other}>
+      <Card.Header className={styles['data-summary-card__header']}>
+        <Heading className={styles['data-summary-card__title']} size="h3">
+          {title}
+        </Heading>
+      </Card.Header>
+      <Card.Body className={styles['data-summary-card__body']}>
+        <Text
+          as="p"
```
dierat, line 57: Isn't "p" the default value for `as`?

.storybook/recipes/DataSummaryCard/DataSummaryCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+   */
+  data?: string;
+  /**
+   * Text that provides a unit of measurement for the data.
+   */
+  dataUnit?: string;
+  /**
+   * Text to provide more context for the data.
+   */
+
```
dierat, line 23:
```suggestion
```

.storybook/recipes/DataSummaryCard/DataSummaryCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+  description?: string;
+  /**
+   * Title text of the data to represented.
```
dierat, line 26:
```suggestion
   * Title text of the data represented.
```

.storybook/recipes/DataSummaryCard/DataSummaryCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+          >
+            {description}
+          </Text>
+        )}
+      </Card.Body>
+      {variant === 'off-track' && (
+        <div
+          aria-label="off track"
+          className={styles['data-summary-card__indicator--off-track']}
+          role="img"
```
dierat, line 86: Nice 👍

.storybook/recipes/DataSummaryCard/DataSummaryCard.tsx
dierat: I guess my question is whether we want this card to be versatile enough to accommodate other kinds of content. This feels like a very specific kind of card and that it makes sense for it to have a very specific api like this. I could also see the value of having a card that just has the band across the bottom and you can put whatever you want inside; I have a feeling we're going to see that pattern show up again somewhere else now that it's in the product. But for this card specifically, I think this is fine.

.storybook/recipes/DataSummaryCard/DataSummaryCard.module.css
dierat: I think this is fine. I know it feels kind of hacky, but it doesn't look that bad to me. Thanks for paying attention to these finer details ❤️

.storybook/recipes/DataSummaryCard/DataSummaryCard.tsx
dierat: I think my only concern is that these props might be a little hard to map to their usage. Maybe `data` could be `dataAmount` so it's clearer that it's the left half of the string thingy. I'm also wondering if `data` should be a `number` instead of a string and whether `dataUnit` should mention that it should start with a forward slash since we're not adding that.

.storybook/recipes/DataSummaryCard/DataSummaryCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const componentClassName = clsx(styles['data-summary-card'], className);
+  return (
+    <Card className={componentClassName} {...other}>
+      <Card.Header className={styles['data-summary-card__header']}>
+        <Heading className={styles['data-summary-card__title']} size="h3">
+          {title}
+        </Heading>
+      </Card.Header>
+      <Card.Body className={styles['data-summary-card__body']}>
+        <Text
```
dierat, line 56: I see that both `data` and `dataUnit` are optional – what does this look like if neither are passed in? Should `data` at least be required?

.storybook/recipes/DataSummaryCard/DataSummaryCard.tsx
jinlee93: Hmmmm I think `dataAmount` makes sense but not sure if it should be a number since maybe there's a use case for displaying letter grades or something. There exists a use case for `dataUnit` to be a `%`, I'll change one of the stories to represent that.

.storybook/recipes/DataSummaryCard/DataSummaryCard.tsx
jinlee93: Yes it is, will remove line

.storybook/recipes/DataSummaryCard/DataSummaryCard.tsx
jinlee93: `data`/`dataAmount` should be required yup, will fix

dierat (approved):

PR #1259, opened 2022-09-01
chore: make upcoming components meaningful and move comments inline 

schakraborty22:
### Summary:
This is the PR for the upcoming components comments changes made for story id:
211222 - Make sure all component comments are meaningful
https://app.shortcut.com/czi-edu/story/211222/make-sure-all-component-comments-are-meaningful
211537 - Move CSS and JS comments next to what they're referencing
https://app.shortcut.com/czi-edu/story/211537/move-css-and-js-comments-next-to-what-they-re-referencing

### DIFF
--- a/src/upcoming-components/Accordion/Accordion.tsx
+++ b/src/upcoming-components/Accordion/Accordion.tsx
@@ -4,19 +4,9 @@ import styles from './Accordion.module.css';
 import AccordionPanel from '../AccordionPanel';
 
 export interface Props {
-  /**
-   * Child node(s) that can be nested inside component
-   */
-  children?: ReactNode;
-  /**
-   * CSS class names that can be appended to the component.
-   */
-  className?: string;
-  /**
-   * Inverted variant
-   * 1) Used for dark backgrounds
-   */
-  inverted?: boolean;
+  children?: ReactNode /* Child node(s) that can be nested inside component */;
+  className?: string /* CSS class names that can be appended to the component */;
+  inverted?: boolean /* 1) Used for dark backgrounds */;
 }
 
 /**
@@ -26,7 +16,7 @@ export interface Props {
  * import {Accordion} from "@chanzuckerberg/eds";
  * ```
  *
- * TODO: update this comment with a description of the component.
+ * Component is an accordian panel element with additional styling passing in
  */
 export const Accordion = ({
   children,
--- a/src/upcoming-components/AccordionPanel/AccordionPanel.tsx
+++ b/src/upcoming-components/AccordionPanel/AccordionPanel.tsx
@@ -12,37 +12,13 @@ import Icon from '../../components/Icon';
 import { EdsThemeColorIconNeutralStrong } from '../../tokens-dist/ts/colors';
 
 export interface Props {
-  /**
-   * Thinks the aria-control attribute in the accordion panel header to the id in the accordion panel body
-   */
-  ariaControls?: string;
-  /**
-   * Child node(s) that can be nested inside component
-   */
-  children?: ReactNode;
-  /**
-   * Accordion panel header slot for elements like heading
-   */
-  header?: ReactNode;
-  /**
-   * HTML id for aria-labelledby, which associates panel contents with the accordion title
-   */
-  'aria-labelledby'?: string;
-  /**
-   * CSS class names that can be appended to the component.
-   */
-  className?: string;
-  /**
-   * is Active
-   * 1) Panel is open when set to true. Close when set to false
-   */
-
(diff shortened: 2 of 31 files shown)


src/upcoming-components/Accordion/Accordion.tsx
```diff
@@ -4,19 +4,9 @@ import styles from './Accordion.module.css';
...
-  children?: ReactNode;
-  /**
-   * CSS class names that can be appended to the component.
-   */
-  className?: string;
-  /**
-   * Inverted variant
-   * 1) Used for dark backgrounds
-   */
-  inverted?: boolean;
```
jinlee93, line 19: I think leaving prop comments like this is fine, keeps it more readable and allows intellisense and jsdoc to pick up it up

src/upcoming-components/FileUploadField/FileUploadField.tsx
```diff
@@ -237,18 +160,10 @@ export const FileUploadField = ({
...
 
-    /*
-     * 1. Copy existing files from state to a new variable
-     * 2. Append new files to this new array
-     * 3. If there is a maxFiles limit and if the more files have been selected than the max,
-     *    remove those over the limit from the array.
-     * 4. Set new state with existing and newly selected files
-     */
-
-    /* 1 */
```
jinlee93, line 248: This is the kind of comment block the ticket was asking to place next to what they're referencing 👍

src/upcoming-components/Accordion/Accordion.tsx
anniehu4: +1 I prefer this style

jinlee93:
Like the comment in line 64 here should be at line 69 https://github.com/chanzuckerberg/edu-design-system/blob/27b8189d6c0c7c9bf09947278ebabc38d52b5646/src/upcoming-components/CheckboxFieldItem/CheckboxFieldItem.tsx#L69

jinlee93:
And a css file example of the comment needing to be next to the code https://github.com/chanzuckerberg/edu-design-system/blob/27b8189d6c0c7c9bf09947278ebabc38d52b5646/src/upcoming-components/Counter/Counter.module.css#L38

src/upcoming-components/Accordion/Accordion.tsx
```diff
@@ -26,7 +16,7 @@ export interface Props {
  * import {Accordion} from "@chanzuckerberg/eds";
  * ```
  *
- * TODO: update this comment with a description of the component.
+ * Component is an accordian panel element with additional styling passing in
```
dierat, line 19: For the component comments, could you try to describe what the component does? I don't think we need the "with additional styling passed in" or mentioning the props, just a basic description of the purpose of the thing. In this case, an accordion is a common component type, so I googled "accordion component" and grabbed the description from the first result https://mui.com/material-ui/react-accordion/

```suggestion
 * Component that allows the user to show and hide sections of related content on a page.
```

src/upcoming-components/AccordionPanel/AccordionPanel.tsx
```diff
@@ -52,7 +28,8 @@ export interface Props {
  * import {AccordionPanel} from "@chanzuckerberg/eds";
  * ```
  *
- * TODO: update this comment with a description of the component.
+ * Accordian panel component to toggle between accordian content and active styling for the selected accordian,
+ * with adddtional styling passing in.
```
dierat, line 32: How about something like..
```suggestion
 * Panel inside Accordion component for one section of content that can be shown and hidden by the user.
```
?

src/upcoming-components/Band/Band.tsx
```diff
@@ -25,7 +15,7 @@ export interface Props {
  * import {Band} from "@chanzuckerberg/eds";
  * ```
  *
- * TODO: update this comment with a description of the component.
+ * Component is called Band with the band syling and based on variant props ,addtional styling passed in
```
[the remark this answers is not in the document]
dierat: Actually, looking at this component for this PR makes me wonder what the purpose of this and some of the other components in `upcoming-components` and whether we should just remove them. So I would skip over the following components for upgrades because I'd like to advocate for removing them:
- `Band`
- `StackedBlockBody`
- `StackedBlockFooter`
- `StackedBlockHeader`
- `Tags`
- `TagsItem`
- `Tile`
- `TileBody`
- `TileHeader`

There are some others that I'm also pretty sure we won't use, but I'm most confident that we don't need the ones I listed above.

schakraborty22:
Made separate PR for the components folder and remove not used upcoming components.

PR #1260, opened 2022-09-02
fix(dropdownmenu): make up and down key to navigate work

jinlee93:
### Summary:
- Motivation
  - [slack context](https://czi-edu.slack.com/archives/C01BVP94BRR/p1661984635731609)
  - In consuming app for `<ProjectCard>` the `buttonDropdownItems` were not navigable with arrow keys
- Exploration
  - Refs were not being attached properly to the actual menu items and instead being attached to the helper Component
  - This is due to an attempt to flatten React Fragments until `<DropdownMenuItems>` become first child level, however, even if the helper Component wraps a React Fragment, checking for `child.type === React.Fragment` returns false in such case and results in the `<DropdownMenuItems>` not being top level 
- Execution
  - Remove flattening/cloning attempt of children in `<DropdownMenu>`
  - Pass refs as context to `<DropdownMenuItems>`
    - Allows for n number of wrapping parents
  - Adds another helper function for focusing
- Considerations
  - Move to HeadlessUI Menu may solve this and other unknowns but would need bigger lift in styling and matching component api for other consuming components, quite possibly breaking
### Test Plan:
- add story with wrapping helper component
- test story with unit tests for arrow keys
- no visual regression

### DIFF
--- a/src/components/ButtonDropdown/ButtonDropdown.stories.tsx
+++ b/src/components/ButtonDropdown/ButtonDropdown.stories.tsx
@@ -88,7 +88,10 @@ const TestComponent = ({ color }: { color: string }) => {
     </>
   );
 };
-
+/**
+ * This story is required to test DropdownMenu functionality when wrapped in a helper component.
+ * Tests refs being properly passed to DropdownMenuItem components.
+ */
 export const ComponentWrapped: StoryObj<Args> = {
   args: {
     children: <TestComponent color="blanchedalmond" />,
--- a/src/components/ButtonDropdown/ButtonDropdown.stories.tsx
+++ b/src/components/ButtonDropdown/ButtonDropdown.stories.tsx
@@ -34,7 +34,9 @@ export default {
         </DropdownMenuItem>
       </>
     ),
-    dropdownMenuTrigger: <Button>Open Dropdown</Button>,
+    dropdownMenuTrigger: (
+      <Button data-testid="trigger-button">Open Dropdown</Button>
+    ),
   },
 } as Meta<Args>;
 
@@ -57,5 +59,38 @@ export const PositionTopRight: StoryObj<Args> = {
 export const PositionBottomRight: StoryObj<Args> = {
   args: {
     position: 'bottom-right',
+  },
+};
+
+const TestComponent = ({ color }: { color: string }) => {
+  return (
+    <>
+      <DropdownMenuItem>
+        <Icon name="schedule" purpose="decorative" size="1.25rem" />
+        Item 1
+      </DropdownMenuItem>
+      <DropdownMenuItem>
+        <div style={{ backgroundColor: color }}>
+          <Icon name="schedule" purpose="decorative" size="1.25rem" />
+          Item 2
+        </div>
+      </DropdownMenuItem>
+      <DropdownMenuItem>
+        <Icon name="schedule" purpose="decorative" size="1.25rem" />
+        Item 3
+      </DropdownMenuItem>
+      <DropdownMenuItem>
+        <div style={{ backgroundColor: color }}>
+          <Icon name="schedule" purpose="decorative" size="1.25rem" />
+          Item 4
+        </div>
+      </DropdownMenuItem>
+    </>
+  );
+};
+
+export const ComponentWrapped: StoryObj<Args> = {
+  args: {
+    children: <TestComponent color="blanchedalmond" />,
   },
 };
--- a/src/components/ButtonDropdown/ButtonDropdown.test.tsx
+++ b/src/components/ButtonDropdown/ButtonDropdown.test.tsx
@@ -16,6 +16,16 @@ describe('<ButtonDropdown />', () => {
     const triggerButton = screen.getByTestId('trigger-button');
     userEvent.click(triggerButton);
     expect(screen.getByRole('button', { name: 'Item 1' })).toHaveFocus();
+    expect(triggerButton).toHaveAttribute('aria-expanded', 'true');
+  });
+
+  it('should close with esc key', () => {
+    render(<ComponentWrapped />);
+    const triggerButton = screen.getByTestId('trigger-button');
+    userEvent.click(triggerButton);
+    userEvent.keyboard('{esc}');
+    expect(triggerButton).toHaveFocus();
+    expect(triggerButton).toHaveAttribute('aria-expanded', 'false');
   });
 
   it('should focus next item or cycle to beginning on down and right arrow keys', () => {
--- a/src/components/ButtonDropdown/ButtonDropdown.test.tsx
+++ b/src/components/ButtonDropdown/ButtonDropdown.test.tsx
@@ -1,6 +1,42 @@
 import { generateSnapshots } from '@chanzuckerberg/story-utils';
+
+import { composeStories } from '@storybook/testing-react';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import React from 'react';
 import * as stories from './ButtonDropdown.stories';
+
+const { ComponentWrapped } = composeStories(stories);
 
 describe('<ButtonDropdown />', () => {
   generateSnapshots(stories);
+
+  it('should focus first item on open', () => {
+    render(<ComponentWrapped />);
+    const triggerButton = screen.getByTestId('trigger-button');
+    userEvent.click(triggerButton);
+    expect(screen.getByRole('button', { name: 'Item 1' })).toHaveFocus();
+  });
+
+  it('should focus next item or cycle to beginning on down and right arrow keys', () => {
+    render(<ComponentWrapped />);
+    const triggerButton = screen.getByTestId('trigger-button');
+    userEvent.click(triggerButton);
+    userEvent.keyboard('{arrowdown}');
+    expect(screen.getByRole('button', { name: 'Item 2' })).toHaveFocus();
+    userEvent.keyboard('{arrowdown}');
+    userEvent.keyboard('{arrowright}');
+    userEvent.keyboard('{arrowright}');
+    expect(screen.getByRole('button', { name: 'Item 1' })).toHaveFocus();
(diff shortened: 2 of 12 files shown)


src/components/ButtonDropdown/ButtonDropdown.stories.tsx
```diff
@@ -59,3 +61,36 @@ export const PositionBottomRight: StoryObj<Args> = {
...
+      </DropdownMenuItem>
+    </>
+  );
+};
+
+export const ComponentWrapped: StoryObj<Args> = {
+  args: {
+    children: <TestComponent color="blanchedalmond" />,
+  },
+};
```
jinlee93, line 96: This story emulates having wrapping helper components where arrow key navigation was previously not working. Confirmed not working by cherry-picking onto `next`, and confirmed working here.

src/components/DropdownMenuItem/DropdownMenuItem.tsx
```diff
@@ -57,57 +58,55 @@ export interface Props {
  *
  * Dropdown menu item within `DropdownMenu`
  */
-export const DropdownMenuItem = React.forwardRef<HTMLLIElement, Props>(
```
jinlee93, line 60: No longer forwards ref, instead grabs ref from context

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -45,107 +58,67 @@ export const DropdownMenu: React.FC<Props> = ({
...
+  };
 
   useEffect(() => {
-    setTimeout(() => {
-      if (isActive && childRefs) {
-        childRefs.current[0]
-          ?.querySelector<HTMLButtonElement | HTMLAnchorElement>(':first-child')
-          ?.focus();
-      }
-    }, 1);
```
jinlee93, line 60: Successfully focuses first element without this setTimeout, also reflected in unit test

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -45,107 +58,67 @@ export const DropdownMenu: React.FC<Props> = ({
...
+    }
   }, [isActive]);
 
-  /**
-   * On KeyDown
-   * 1) Find active item. If there isn't one, do nothing on Keydown
-   * 2) Set active item, next item, and previous item.
-   * 3) If right or down arrow key keyed, focus on next item.
-   * 4) If left or up arrow key keyed, focus on the previous item.
-   */
```
jinlee93, line 69: Removed comments since the code seemed more self explanatory, lemme know if I should re-add

jinlee93:
If there is a better way to send refs all the way down without using context, please advise

src/components/ButtonDropdown/ButtonDropdown.stories.tsx
```diff
@@ -59,3 +61,36 @@ export const PositionBottomRight: StoryObj<Args> = {
...
+          Item 4
+        </div>
+      </DropdownMenuItem>
+    </>
+  );
+};
+
+export const ComponentWrapped: StoryObj<Args> = {
+  args: {
+    children: <TestComponent color="blanchedalmond" />,
```
dierat, line 94: Damn this color sounds tasty

src/components/ButtonDropdown/ButtonDropdown.stories.tsx
```diff
@@ -59,3 +61,36 @@ export const PositionBottomRight: StoryObj<Args> = {
...
+    <>
+      <DropdownMenuItem>
+        <Icon name="schedule" purpose="decorative" size="1.25rem" />
+        Item 1
+      </DropdownMenuItem>
+      <DropdownMenuItem>
+        <div style={{ backgroundColor: color }}>
+          <Icon name="schedule" purpose="decorative" size="1.25rem" />
+          Item 2
+        </div>
```
dierat, line 76: Is the thing that differentiates this story from the ones above that it has these `div`s wrapping the `DropdownMenuItem` contents and the different colored backgrounds? I don't see any other differences.

src/components/ButtonDropdown/ButtonDropdown.test.tsx
```diff
@@ -1,6 +1,42 @@
...
+
+  it('should focus previous item or cycle to end on up and left arrow keys', () => {
+    render(<ComponentWrapped />);
+    const triggerButton = screen.getByTestId('trigger-button');
+    userEvent.click(triggerButton);
+    userEvent.keyboard('{arrowup}');
+    expect(screen.getByRole('button', { name: 'Item 4' })).toHaveFocus();
+    userEvent.keyboard('{arrowleft}');
+    expect(screen.getByRole('button', { name: 'Item 3' })).toHaveFocus();
+  });
```
dierat, line 41: Damn nice tests!

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -45,107 +58,67 @@ export const DropdownMenu: React.FC<Props> = ({
...
   const onKeyDown = (e: KeyboardEvent<HTMLUListElement>) => {
     if ([ESCAPE_KEYCODE].includes(e.key)) {
```
dierat, line 83: Just curious – do you understand why here they're using `includes` instead of `==`? Maybe they copy/pasted this from somewhere else that was checking against multiple keycodes..

dierat: Also could we flatten these if statements into one? Ie
```
if (e.key === ESCAPE_KEYCODE && handleOnEscDown) {
  handleOnEscDown(e);
}
```

src/components/DropdownMenuItem/DropdownMenuItem.tsx
```diff
@@ -57,57 +58,55 @@ export interface Props {
...
+  href,
+  children,
+  variant,
+  onClick,
+  status,
+  target,
+  ...other
+}: DropdownMenuItemProps) => {
+  const context = useContext(DropdownMenuContext);
+  const ref = (el: HTMLLIElement) => {
```
dierat, line 73: Could you use the full world (ie "element" instead of the truncated version? It makes it a little easier to read.

src/components/ButtonDropdown/ButtonDropdown.stories.tsx
jinlee93: Yuh the point of this story was just to have a helper component wrapping the fragment. This was the cause of the bug previously and this story tests for it.

src/components/DropdownMenu/DropdownMenu.tsx
jinlee93: Good point, I'll make the conditionals more readable

src/components/DropdownMenuItem/DropdownMenuItem.tsx
jinlee93: Yup, done

src/components/ButtonDropdown/ButtonDropdown.stories.tsx
dierat: Could you leave some kind of comment about that? I could see a future dev not realizing the significance of this test and removing it for seeming redundant.

jinlee93: Oooh yea good call, will do

src/components/DropdownMenuItem/DropdownMenuItem.tsx
```diff
@@ -57,57 +58,55 @@ export interface Props {
...
+    if (context && el && !context.refs.current.set.has(el)) {
+      context.refs.current.set.add(el);
+      context.refs.current.list.push(el);
+    }
```
dierat, line 77: Is the `set` just to make sure you don't push a ref to the `list` more than once?

jinlee93: Yea, some re-renders can cause `list` to expand with the same elements, so the set helps to prevent that

src/components/DropdownMenu/DropdownMenu.tsx
dierat: I could handle some inline comments like
```
// Close menu on escape key trigger
// Wrap around to top of menu since we're at the bottom
// Wrap around to bottom of menu since we're at the top
```
or something like that.

src/components/ButtonDropdown/ButtonDropdown.tsx
```diff
@@ -173,6 +171,7 @@ export const ButtonDropdown = ({
           ref: buttonRef,
           onClick: togglePanel,
           'aria-expanded': isActiveVar,
+          'aria-haspopup': 'menu',
```
dierat, line 174: nice

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -45,107 +60,78 @@ export const DropdownMenu: React.FC<Props> = ({
...
+      focusItem(focusIndex);
+      // prevents page from scrolling
+      e.preventDefault();
+    }
+    if (e.key === END_KEYCODE) {
+      focusIndex = refs.current.list.length - 1;
+      focusItem(focusIndex);
+      // prevents page from scrolling
+      e.preventDefault();
+    }
```
dierat, line 120: Oo nice

jinlee93: Yea was reading https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role#keyboard_interactions and realized we were missing this functionality (think Headless menu also handles this but that's out of scope for now)

src/components/DropdownMenuItem/DropdownMenuItem.tsx
dierat: Could you add a comment to that effect?

dierat (approved):
Nice cleanup!

PR #1263, opened 2022-09-02
chore(score): rename neutral variant to table

jinlee93:
### Summary:
[sc-213671]
- during design crit, discussed using the transparent border score variant advised for use in the Table component only
- decided on renaming variant to `table`
### Test Plan:
- unit tests pass
- no visual regression except change of `neutral` to `table` variant


dierat (approved):
Nice! To me "neutral" sounds like it's gray anyway.

anniehu4 (approved):

PR #1265, opened 2022-09-06
chore(tooltip)!: rename childDisabled to childNotInteractive

dierat:
### Summary:
We're seeing `Tooltip` usage in `traject` where the child is not an interactive element (like an icon that's not a button). We just added the `childDisabled` prop so the `Tooltip` supports disabled buttons and such as `children`. This prop also works for other non-interactive children, but the prop name is a little confusing to use in those situations. To make this usage clearer, I'm renaming the `childDisabled` prop to `childNotInteractive`.

### Test Plan:
I added a story that has text in a react fragment to verify the prop works for that situation.

tooltip story with plain text as the child

### DIFF
--- a/src/components/Tooltip/Tooltip.module.css
+++ b/src/components/Tooltip/Tooltip.module.css
@@ -50,9 +50,9 @@
 }
 
 /**
- * 1) When childDisabled is true and alignment is top or bottom
+ * 1) When childNotInteractive is true and alignment is top or bottom
  */
-.tooltip__child-disabled-wrapper--vertical {
+.tooltip__child-not-interactive-wrapper--vertical {
   /* Needed for the top and bottom alignment to work properly */
   display: inline-flex;
 }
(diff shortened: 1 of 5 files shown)


dierat:
https://app.shortcut.com/czi-edu/story/208461/tooltip-doesn-t-work-on-disabled-buttons

ahuth (approved):
LGTM 🤘 

Side question: what's the bang in the commit message in `chore(tooltip)!` do?

Does that make this a breaking change for the release notes / version or something?

dierat:
> LGTM 🤘
> 
> Side question: what's the bang in the commit message in `chore(tooltip)!` do?
> 
> Does that make this a breaking change for the release notes / version or something?

Yeah! It should add a "breaking" message in the changelog and bump the major version automatically the next time we cut a release.

anniehu4:
> Yeah! It should add a "breaking" message in the changelog and bump the major version automatically the next time we cut a release.

ooh nice shortcut, TIL

dierat:
> > Yeah! It should add a "breaking" message in the changelog and bump the major version automatically the next time we cut a release.
> 
> ooh nice shortcut, TIL

same lol

PR #1266, opened 2022-09-06
docs: remove beta tags and comments from mature components

dierat:
### Summary:
This PR removes a bunch of beta tags from mature existing components. In https://github.com/chanzuckerberg/edu-design-system/pull/1267 I'm updating the documentation to include information about our new beta process.

I left the beta tags on components that have been made in the last ~6 weeks. Going forward, it should be easier to just look at the releases since we're only releasing a few components in each release.

The components I didn't remove the beta tags from are:
- plop file component templates
- Everything in `upcoming-components`
- `Drawer` (and `DrawerExample`, `DrawerBody`, `DrawerFooter`, and `DrawerHeader`)
- `FieldNote`
- `Filters` (and `FiltersCheckboxField` and `FiltersDrawer`)
- `InputField` (and `InputLabel`)
- `Radio`
- `Score`
- `SearchBar` (and `SearchButton` and `SearchField`)
- `StackedBlock`
- `Table` (and `TableBody`, `TableCell`, `TableFooter`, `TableHeader`, `TableHeaderCell`, `TableObject`, `TableObjectBody`, `TableObjectFooter`, `TableObjectHeader`, and `TableRow`)
- `Textfield`

### Summary shared between the two PRs
Almost all of our components have a beta tag (in storybook and also in the component docstring) even though some of them were completed months ago. This morning we discussed what should be the threshold for component graduation out of beta status, and it was decided that:
- when a component is in beta when it's first released/announced
- upon the following release, the component is no longer in beta
- releases will probably happen about once/month

### Test Plan:
Spot-checked some of the stories to verify the beta comment and badge are gone.

PR #1267, opened 2022-09-06
docs(components): add information about beta status

dierat:
### Summary:
This PR updates the component documentation to describe our new beta process. In https://github.com/chanzuckerberg/edu-design-system/pull/1266 I'm removing a bunch of beta tags from mature existing components.

### Summary shared between the two PRs:
Almost all of our components have a beta tag (in storybook and also in the component docstring) even though some of them were completed months ago. This morning we discussed what should be the threshold for component graduation out of beta status, and it was decided that:
- when a component is in beta when it's first released/announced
- upon the following release, the component is no longer in beta
- releases will probably happen about once/month

### Test Plan:
Verify the new documentation looks okay in the visual studio preview.
component documentation with the new section visible


jinlee93 (approved):
Makes sense to me

PR #1268, opened 2022-09-07
chore: use focus instead of focus-visible

dierat:
### Summary:
When discussion https://github.com/chanzuckerberg/edu-design-system/pull/1250 we realized that the browser versions we support don't all support the `focus-visible` CSS feature, which we're currently using in a few places in the codebase.

This PR just replaces `:focus-visible` with `:focus`.

### Test Plan:
Spot check that these now show focus on both mouse and keyboard interaction.

### DIFF
--- a/src/components/TimelineNav/TimelineNav.module.css
+++ b/src/components/TimelineNav/TimelineNav.module.css
@@ -205,7 +205,7 @@
     color: var(--eds-theme-color-text-neutral-default);
   }
 
-  &:focus-visible {
+  &:focus {
     @mixin focus;
   }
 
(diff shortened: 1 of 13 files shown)


jinlee93 (approved):
Manually tested these to see if focus worked, thanks for making these changes

src/components/TimelineNav/TimelineNav.module.css
```diff
@@ -205,7 +205,7 @@
     color: var(--eds-theme-color-text-neutral-default);
   }
 
-  &:focus-visible {
+  &:focus {
```
jinlee93, line 208: Did notice from testing this tho is that the read/unread dot indicator does not have a focus indicator (not a result of this pr, is existing in NotificationList in [published storybook](https://chanzuckerberg.github.io/edu-design-system/?path=/story/molecules-lists-notificationlist--default)), filed [ticket](https://app.shortcut.com/czi-edu/story/214162)

dierat: Damn nice catch 👍

PR #1269, opened 2022-09-07
fix(notification-list): add focus state to button

dierat:
### Summary:
@jinlee93 flagged that the button in notification list items does not currently have a visual state change on focus. (This button serves as the "read" indicator, and the user can trigger it to flip the "read" state.)

This PR just adds the regular focus outline on focus.

before and after comparison showing the notification dot button now has an outline on focus

### Test Plan:
- navigated to http://localhost:9009/?path=/story/molecules-lists-notificationlist--default
- verified I could tab through the notification text and buttons
- verified there's now a focus outline on the buttons when focused
- verified I can flip the "read" status on the buttons with the keyboard

### DIFF
--- a/src/components/NotificationList/NotificationList.module.css
+++ b/src/components/NotificationList/NotificationList.module.css
@@ -81,6 +81,10 @@
   background-color: var(--eds-theme-color-text-link-strong);
   cursor: pointer;
 
+  &:focus {
+    @mixin focus;
+  }
+
   /*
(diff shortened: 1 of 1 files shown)


dierat:
https://app.shortcut.com/czi-edu/story/214162/notificationlist-mark-as-read-unread-button-needs-focus-indicator

jinlee93 (approved):
Nice, thanks!

PR #1274, opened 2022-09-09
chore(clickable-style): remove link visited state styling

dierat:
### Summary:
Right now, the `link` variant of `ClickableStyle`, `Button`, and `Link` components has a distinct visited state styling, but the visited state doesn't have a distinct hover state styling. Since our link style usually has a hover state style, the fact that it's missing here is confusing for users and looks like a bug. We looked at different options for hover state styles for the visited state for both the `brand` and `neutral` variants and ultimately decided that with two link variants, it's going to just be too many different states if you throw in visited as well and continue to confuse users. (Products that use visited states really well, like Google and Wikipedia, only have one link variant.)

So this PR removes the visited link state styling and the helpers I had added for generating randomized urls in the `Link` component stories.

### Test Plan:
Verify that:
- the default, hover, focus, and active link styles are unchanged for both variants
- there is no distinct visited state style for links for both variants
- links in the `Link` and `Text` component stories snapshots have the EDS zeroheight link for the `href` attribute


dierat:
https://app.shortcut.com/czi-edu/story/206837/nit-should-visited-links-have-a-distinct-hover-state

src/design-tokens/mixins.css
```diff
@@ -48,11 +48,6 @@
     ) !important;
     background-color: var(--eds-theme-color-background-brand-primary-strong);
   }
-
-  &:visited {
-    color: var(--eds-theme-color-text-neutral-subtle);
-    text-decoration-color: var(--eds-theme-color-text-neutral-subtle);
-  }
```
dierat, line 55: This is the crux of the PR. All the other changes are related to the randomized urls I had in the stories.

jinlee93 (approved):
Thanks for resolving the visited link issue

PR #1275, opened 2022-09-09
chore(clickable-style): lighten hover state for neutral secondary button

dierat:
### Summary:
We're going to lighten the hover state for neutral secondary (aka "tertiary") buttons. Instead of having white text on a dark background, it's going to be dark text on a light background. (It will now also be out of sync with the brand secondary button, which will still have white text on a dark background).

In terms of implementation, I just updated the neutral secondary button tokens, which are only being used in the neutral secondary button so there are no other side effects.

##### Figma
secondary brand and neutral button styles in figma, before this change

##### Storybook
https://user-images.githubusercontent.com/7761701/189383056-c1d9999a-61bd-48fa-8564-ac81d1946572.mp4

##### Figma
secondary brand and neutral button styles in figma, after this change

##### Storybook
https://user-images.githubusercontent.com/7761701/189383099-2da44e90-86cb-4c04-a2d7-ef9a68518d49.mp4

### Test Plan:
Verify the hover state for neutral secondary buttons has been updated but all other states and variants are unchanged.


dierat:
https://app.shortcut.com/czi-edu/story/213176/ui-kit-button-lighten-the-hover-and-active-state-colors

src/design-tokens/tier-3-component/buttons.json
```diff
@@ -128,7 +128,7 @@
                   "value": "{eds.theme.color.transparent.black.0}"
                 },
                 "hover": {
-                  "value": "{eds.color.neutral.600}"
+                  "value": "{eds.color.neutral.200}"
```
dierat, line 131: This is the file I changed manually. All the other changes were generated automatically.

ahuth (approved):
LGTM

Should we have seen some changes in Chromatic?

dierat:
> LGTM
> 
> Should we have seen some changes in Chromatic?

Nah because we don't have the hover styles being tested in chromatic right now. We used to have that whole grid that showed all the states on all the buttons, but that was dropped when we started the codebase over from scratch and we haven't brought it back yet. I have a ticket for it somewhere in the backlog.

PR #1278, opened 2022-09-10
chore(modal-body): pass down extra props to wrapper div

dierat:
### Summary:
The `ModalBody` component isn't passing down extra props, which means you can't put a `data-testid` on it for end-to-end tests. This PR just passes extra props down onto the wrapper `div`.

### Test Plan:
I temporarily added a `data-testid` to a `Modal.Body` in the `Modal.stories.tsx` file and verified in the browser that the attribute is being passed down.

modal story with the chrome dev tools open, inspecting the modal body, with a test id visible in the dom tree

### DIFF
--- a/src/components/ModalBody/ModalBody.tsx
+++ b/src/components/ModalBody/ModalBody.tsx
@@ -20,15 +20,19 @@ export type Props = {
   isFocusable?: boolean;
 };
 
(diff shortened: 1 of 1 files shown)


jinlee93 (approved):
Ah crap, nice catch

PR #1279, opened 2022-09-12
chore(tokens): make disabled border token neutral 300

dierat:
### Summary:
All the `disabled` tokens are neutral 300 except for the border disabled token. This makes disabled secondary (outline) buttons really light and hard to see against white backgrounds (and I'm guessing invisible on light gray backgrounds). This PR makes the border disabled token neutral 300 to be consistent with all the others.

The other component besides `Button` using this token are forms (`inputStyles` mixin and `FileUploadField`, which is in `upcoming-components`). Input fields have a neutral 300 border by default, which means a neutral 300 border for the disabled state would not differentiate the disabled state from the regular state. So I added a form border disabled token to keep the input field disabled border 100. (Alternatively, I could have added a disabled button border token and made that 300, but the form border is the odd one out. Also, I neutral 300 has a 1.75:1 contrast ratio against white, so the regular form border token should really be darker anyway. I've brought this up a couple of times but it still hasn't been fixed in figma. I'm going to make a separate task to address that.)

### Test Plan:
Verify in chromatic that:
- the disabled secondary (outline) buttons are now the same color as disabled primary (solid) buttons
- the input fields are unchanged


dierat:
https://app.shortcut.com/czi-edu/story/214669/make-disabled-button-token-darker

ahuth (approved):

dierat:
Thanks for reviewing!

PR #1281, opened 2022-09-13
feat(filterspopover): add filters popover

jinlee93:
### Summary:
[sc-212092]
- motivation
  - Drawer filters seem heavy for small amount of form inputs for filtering
  - Popover filters are being used in upcoming pilots
  - There seem to be a need for popover filters as seen in Multiselect in EDSCandidates
- execution
  - Uses <Popover> component as basis for `FiltersPopover`
  - Since HeadlessUI Popover does not natively provide a `onClose` callback, includes an effect and a ref to provide similar functionality
  - Removes `Filters` component and its `FiltersDrawer` logic into `FiltersDrawer`
  - Api is intentionally very similar to `FiltersDrawer` to make switching between the two based on screensizes easy
    - Only difference is `placement` prop for placing the popover
    - Demonstrated in `Filters` recipe 
    - Demonstrated in very similar stories for the components
  - Felt no need for `FiltersButton` subcomponent since usage would only be in the two `FiltersDrawer` and `FiltersPopover` components
    - Having such subcomponent I feel would needless increase complexity with ref forwarding and attribute passing. If custom trigger buttons are needed, scrums can just use regular `Button` and follow similar implementation
### Test Plan:
- added unit functionality tests
- added snaps
- no visual regressions affecting others

Screen Shot 2022-09-13 at 9 27 56 AM
Screen Shot 2022-09-13 at 9 28 03 AM
Screen Shot 2022-09-13 at 9 28 15 AM
Screen Shot 2022-09-13 at 9 28 28 AM


src/components/FiltersPopover/FiltersPopover.tsx
```diff
@@ -0,0 +1,162 @@
...
+  return (
+    <>
+      <Popover.Button as={React.Fragment}>
+        <Button
+          className={styles['filters-popover__button']}
+          status={buttonStatus}
+          variant={buttonVariant}
+        >
+          <Icon name="filter-list" purpose="decorative" size="1.5rem" />
+          {triggerText}
```
jinlee93, line 107: Could also have `triggerText` also replace Icon as well and have default trigger text as the icon + 'Filters'

dierat:
Btw one thing I think may come up in the Lodestone testing is the focus outline of the drawer when it first opens, which only appears on the left side of the drawer itself.

filters drawer in storybook with the drawer open, the focus 'outline' only appearing on the left side of the drawer, and text that reads 'purple focus outline only appears on left side of drawer, which may make it hard for sighted keyboard users to understand which element has focus'

dierat:
Also is the drawer supposed to have an overlay on the rest of the page?

src/components/Filters/Filters.tsx
```diff
@@ -1,115 +0,0 @@
...
-/**
- * BETA: This component is still a work in progress and is subject to change.
- *
- * ```ts
- * import {Filters} from "@chanzuckerberg/eds";
- * ```'
- *
- * A filter component with a button that triggers a drawer of checkbox filters to be selected.
- */
-export const Filters = ({
```
dierat, line 52: Are filters only going to be available in the context of a drawer or popover? Do we see no reason to expose the filter itself so it could be placed on a page where it's always visible? Just curious.

src/components/FiltersDrawer/FiltersDrawer.test.tsx
```diff
@@ -2,16 +2,9 @@ import { generateSnapshots } from '@chanzuckerberg/story-utils';
...
+import * as stories from './FiltersDrawer.stories';
 
 const { OverflowInteractive } = composeStories(stories);
-const consoleLogMock = jest.spyOn(console, 'log').mockImplementation();
-beforeEach(() => {
-  consoleLogMock.mockClear();
-});
-afterAll(() => {
-  consoleLogMock.mockRestore();
-});
```
dierat, line 14: Was this just not being used?

src/components/FiltersDrawer/FiltersDrawer.tsx
```diff
@@ -33,9 +38,9 @@ export type Props = {
    */
   onApply?: () => void;
   /**
-   * Sets the filters drawer window open or closed.
+   * Text to be placed in the button that activates the Filters Drawer
    */
-  isActive: boolean;
+  triggerText?: string;
```
dierat, line 43: I know this api was just moved over from `Filters` but, looking at this PR, I'm wondering about how much control consumers have over the look and feel of this trigger button. That seems like the kind of thing that gets shifted around in designs because designers are used to having a lot of control over buttons. In the PR description you said

>If custom trigger buttons are needed, scrums can just use regular Button and follow similar implementation

but I'm just wondering how much work that would take / what that would look like. Maybe we could have a story with that example.

src/components/FiltersPopover/FiltersPopover.tsx
dierat: I think having a standardized icon here makes sense because the action is always the same (like a next button always has a right arrow regardless of whether the text says "Next" or "Continue").

src/components/FiltersPopover/FiltersPopover.tsx
```diff
@@ -0,0 +1,162 @@
...
+   * Hooks to emulate an onClose callback for the Popover.
+   */
+  const firstRender = useRef(true);
+  useEffect(() => {
+    if (firstRender.current) {
+      firstRender.current = false;
+    } else if (!open && onClose) {
+      onClose();
+    }
+  }, [open]); // eslint-disable-line react-hooks/exhaustive-deps
```
dierat, line 88: What's eslint complaining about here?

src/components/FiltersPopover/FiltersPopover.tsx
```diff
@@ -0,0 +1,162 @@
...
+  onClear,
+  onClose,
+  open,
+  triggerText,
+  ...other
+}: FiltersPopoverRenderProps) => {
+  /**
+   * Hooks to emulate an onClose callback for the Popover.
+   */
+  const firstRender = useRef(true);
```
dierat, line 81: Could you help me explain the need for tracking the first render?

.storybook/recipes/Filters/Filters.tsx
```diff
@@ -0,0 +1,31 @@
...
+   */
+  as?: 'FiltersDrawer' | 'FiltersPopover';
+  /**
+   * Popover placement options relative to the filters trigger button.
+   */
+  placement?: FiltersPopoverProps['placement'];
+} & FiltersPopoverProps;
+
+/**
+ * Primary UI component for user interaction
```
dierat, line 20: Could you update this component comment while you're in here?

src/components/FiltersDrawer/FiltersDrawer.test.tsx
jinlee93: I think there used to be `console logs` in the code but it was no longer happening so I removed it

src/components/Filters/Filters.tsx
jinlee93: That's a good question, and I'll bring it up to design. However I'm thinking if product teams need to have filters available on the page, wouldn't it just be the `FiltersCheckboxField` component, other `<fieldsets>`, or simply other input elements that would be on the page rather than being wrapped by `FiltersPopover` or `FiltersDrawer`?

jinlee93:
> Also is the drawer supposed to have an overlay on the rest of the page?

The drawer should definitely have a veil, but I'll scope it out of this pr which already has a lot of changes

src/components/FiltersDrawer/FiltersDrawer.tsx
jinlee93: Hmmm I think this is a valid point. I'll make the `triggerText` a required `trigger: ReactNode` instead and have a `FiltersButton` subcomponent that can be used as the trigger and update the stories to reflect it. Also will add another story with another CTA kind of element to show other components can be used

src/components/FiltersPopover/FiltersPopover.tsx
jinlee93: Wants `onClose` as part of the deps, but since `onClose` manages external state, there is a callback loop that occurs. I'll add a comment about this in the code

src/components/FiltersPopover/FiltersPopover.tsx
jinlee93: Yea callback in `useEffect` always runs on first render, but the Popover renders as closed, causing `onClose` to be called on render. Tracking `firstRender` prevents this from happening

.storybook/recipes/Filters/Filters.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+      <div style={{ margin: '0.25rem', height: '100vh' }}>
+        <Story />
+      </div>
+    ),
+  ],
+} as Meta<Args>;
+
+type Args = React.ComponentProps<typeof Filters>;
+
+export const Default: StoryObj<Args> = {};
```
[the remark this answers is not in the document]
jinlee93: Good question, I left it as one story so that whoever views the story can choose the storybook control for FiltersPopover. Didn't think there was much value in Chromatic testing (flakiness and already being tested in FiltersPopover stories). Also this way, the dev can know to implement in a similar manner.

But now looking back at it, not sure how many devs use the control so I'll make the second story.

.storybook/recipes/Filters/Filters.tsx
jinlee93: YUS will do

src/components/FiltersPopover/FiltersPopover.tsx
dierat: Awesome. Could you add a comment about that too?

.storybook/recipes/Filters/Filters.stories.tsx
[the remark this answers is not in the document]
dierat: >not sure how many devs use the control

Yeah that's what I'm thinking too 👍

src/components/FiltersPopover/FiltersPopover.tsx
dierat: Is there a way to check what the previous version of a prop was in functional react components? Because I think if I were doing this in the old component format, I would use `componentDidUpdate ` and check if 1) `open` is `false` and 2) `open` previously was `true` (so I know it did close as opposed to rendering as closed to start with).

src/components/FiltersDrawer/FiltersDrawer.tsx
dierat: Awesome, thanks! Feel free to punt that to a separate PR though.

src/components/Filters/Filters.tsx
dierat: That's fair 👍

src/components/FiltersPopover/FiltersPopover.tsx
jinlee93: There [is](https://stackoverflow.com/questions/53446020/how-to-compare-oldvalues-and-newvalues-on-react-hooks-useeffect) but in the end its the same `useRef` `useEffect` combo

src/components/FiltersDrawer/FiltersDrawer.tsx
jinlee93: [Ticket](https://app.shortcut.com/czi-edu/story/215710/filtersbutton-subcomponent-and-change-trigger-to-default-to-filtersbutton) created so I don't lose track

jinlee93:
> Btw one thing I think may come up in the Lodestone testing is the focus outline of the drawer when it first opens, which only appears on the left side of the drawer itself.

Moved focus outline to inside the drawer
Screen Shot 2022-09-16 at 2 29 02 PM
Screen Shot 2022-09-16 at 2 29 24 PM

dierat:
> Moved focus outline to inside the drawer

Awesome, thanks!

src/components/Drawer/Drawer.tsx
```diff
@@ -78,13 +78,6 @@ export const Drawer = ({
   const ref = useRef<HTMLDivElement | null>(null);
   const isMountedRef = useRef(true);
 
-  /**
-   * Use effect
-   * 1) Set prevIsActive to previous isActive prop
-   * 2) If prevIsActive is defined and previous isActive prop is not equal
-   * to current isActive prop, toggle state
-   */
```
dierat, line 86: Maybe leave this block just to reduce merge conflicts with the other PR addressing comments.

src/components/FiltersPopover/FiltersPopover.tsx
```diff
@@ -0,0 +1,164 @@
...
+  useEffect(() => {
+    if (firstRender.current) {
+      firstRender.current = false;
+    } else if (!open && onClose) {
+      onClose();
+    }
+  }, [open]); // eslint-disable-line react-hooks/exhaustive-deps
+  // onClose is not included as a dependency since it usually affects external state and can cause a callback loop
```
dierat, line 90: Nit: maybe move this comment one line up so it's above the line it's referring to.

.storybook/recipes/Filters/Filters.tsx
dierat: Bump on this comment.

dierat (approved):

PR #955, opened 2022-04-06
Ifrost/global header tweaks

ifrost1:
### Summary:
- Adjusted `UtilityNav` to be more composable. Popover shouldn't be directly built into the `UtilityNav` for future flexibilty for adding more items (etc. notifications and an account link).

Addressed the conversation we had with Courtney and Sean:
- Hover states for primary nav
- Break into a top nav bar on larger screens so List Detail doesn't squish
- Visually hide the name from the Avatar in `UtilityNavItem`. I added the name `Notifications` for that since it technically triggers notifications and a name doesn't serve a purpose there.

### Test Plan:
 Ran `yarn test`. It ran with a deprecated `variant="dark` message that doesn't have anything to do with my work

`yarn lint` passed

`yarn start` spins up.

PR #1292, opened 2022-09-19
chore: use focus-visible instead of focus everywhere

dierat:
### Summary:
In https://github.com/chanzuckerberg/edu-design-system/pull/1290 I updated the buttons and links to only show the focus state when the user interacts with them using a keyboard (as opposed to with a mouse/trackpad). The implementation method was done using `focus-visible` along with a fallback block for browsers that don't support it yet. (I added a [ticket to remove the fallback blocks](https://app.shortcut.com/czi-edu/story/215843/remove-focus-fallback-css) once we're only supporting browser versions that support `focus-visible`.)

In this PR, I'm expanding to use this method on almost all `:focus` states. Exceptions include:
- `inputStyles` mixin and `SearchField` component
  - I talked to design and we determined that we want mouse users to see the focus state so they know that typing on their keyboard will add text in the input field.
- `BaselineCard` and `SkipLink`
  - In both of these cases, the link only appears on keyboard interaction already, so there's nothing to hide for mouse users.
- inverted states
  - Because we're going to remove that stuff anyway; there's no point updating it.
- `FileUploadField`
  - Because it's also using `focus-within`, and I don't want there to be a discrepancy where sometimes the focus state appears for mouse users and sometimes it doesn't.

### Test Plan:
In any of the supported browser versions (e.g. the latest version of Chrome), verify that:
- the focus state does not appear on mouse click
- the focus state does appear when tabbing to it using a keyboard

In any of the *unsupported* browser versions (e.g. Chrome 84), verify that:
- the focus state appears on both mouse click and when tabbing to it using a keyboard

### DIFF
--- a/.storybook/recipes/GlobalHeader/GlobalHeader.module.css
+++ b/.storybook/recipes/GlobalHeader/GlobalHeader.module.css
@@ -49,15 +49,29 @@
   color: var(--eds-theme-color-text-neutral-default-inverse);
 
   &:hover,
-  &:focus {
+  &:focus-visible {
     background: var(
       --eds-theme-color-button-primary-brand-background-hover
     ); /* TODO */
     border-color: transparent; /* TODO */
   }
 
-  &:focus {
+  &:focus-visible {
     @mixin focusInverted;
+  }
+
+  @supports not selector(:focus-visible) {
+    &:hover,
+    &:focus {
+      background: var(
+        --eds-theme-color-button-primary-brand-background-hover
+      ); /* TODO */
+      border-color: transparent; /* TODO */
+    }
+
+    &:focus {
+      @mixin focusInverted;
+    }
   }
 
   @media all and (min-width: $eds-bp-xl) {
--- a/.storybook/recipes/GlobalHeader/GlobalHeader.module.css
+++ b/.storybook/recipes/GlobalHeader/GlobalHeader.module.css
@@ -50,10 +50,8 @@
 
   &:hover,
   &:focus-visible {
-    background: var(
-      --eds-theme-color-button-primary-brand-background-hover
-    ); /* TODO */
-    border-color: transparent; /* TODO */
+    background: var(--eds-theme-color-button-primary-brand-background-hover);
+    border-color: transparent;
   }
 
   &:focus-visible {
@@ -63,10 +61,8 @@
   @supports not selector(:focus-visible) {
     &:hover,
     &:focus {
-      background: var(
-        --eds-theme-color-button-primary-brand-background-hover
-      ); /* TODO */
-      border-color: transparent; /* TODO */
+      background: var(--eds-theme-color-button-primary-brand-background-hover);
+      border-color: transparent;
     }
 
     &:focus {
--- a/docs/CODE_GUIDELINES.md
+++ b/docs/CODE_GUIDELINES.md
@@ -106,6 +106,26 @@ EDS uses [PostCSS Nested](https://github.com/postcss/postcss-nested) to provide 
 }
 ```
 
+#### Focus states
+
+We use focus-visible for most focus states. This ensures that the focus state will appear when the user is interacting with an element using a keyboard, but the focus state will not be present if the user if interacting with the element using a mouse. Exceptions to this rule include situations where the focus state also communicates to users that they are in some sort of edit state, e.g. when typing in an input field.
+
+However, we currently support some browser versions that do not support the focus-visible CSS feature, so we also use a fallback block in conjunciton with focus-visible.
+
+```scss
+.button--primary {
+  &:focus-visible {
+    @mixin focus;
+  }
+
+  @supports not selector(:focus-visible) {
+    &:focus {
+      @mixin focus;
+    }
+  }
+}
+```
+
 #### States and pseudo-selectors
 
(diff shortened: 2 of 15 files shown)


.storybook/recipes/GlobalHeader/GlobalHeader.module.css
```diff
@@ -49,17 +49,31 @@
...
   }
 
+  @supports not selector(:focus-visible) {
+    &:hover,
+    &:focus {
+      background: var(
+        --eds-theme-color-button-primary-brand-background-hover
+      ); /* TODO */
+      border-color: transparent; /* TODO */
+    }
```
dierat, line 70: I'm duplicating the whole block, including the `hover` property, because otherwise the `hover` styles are ignored in browsers that don't support `focus-visible`.

dierat:

.storybook/recipes/GlobalHeader/GlobalHeader.module.css
```diff
@@ -49,17 +49,31 @@
...
     @mixin focusInverted;
   }
 
+  @supports not selector(:focus-visible) {
+    &:hover,
+    &:focus {
+      background: var(
+        --eds-theme-color-button-primary-brand-background-hover
+      ); /* TODO */
+      border-color: transparent; /* TODO */
```
jinlee93, line 69: What were these `todo`s for?

docs/CODE_GUIDELINES.md
```diff
@@ -106,17 +106,43 @@ EDS uses [PostCSS Nested](https://github.com/postcss/postcss-nested) to provide
...
+    @mixin focus;
+  }
+
+  @supports not selector(:focus-visible) {
+    &:focus {
+      @mixin focus;
+    }
+  }
+}
+```
```
jinlee93, line 127: Oooh good docs update 👍

docs/CODE_GUIDELINES.md
```diff
@@ -106,17 +106,43 @@ EDS uses [PostCSS Nested](https://github.com/postcss/postcss-nested) to provide
 }
 ```
 
+#### Focus states
+
+We use focus-visible for most focus states. This ensures that the focus state will appear when the user is interacting with an element using a keyboard, but the focus state will not be present if the user if interacting with the element using a mouse. Exceptions to this rule include situations where the focus state also communicates to users that they are in some sort of edit state, e.g. when typing in an input field.
+
+However, we currently support some browser versions that do not support the focus-visible CSS feature, so we also use a fallback block in conjunciton with focus-visible.
```
jinlee93, line 113: nit: sp

```suggestion
We use focus-visible for most focus states. This ensures that the focus state will appear when the user is interacting with an element using a keyboard, but the focus state will not be present if the user is interacting with the element using a mouse. Exceptions to this rule include situations where the focus state also communicates to users that they are in some sort of edit state, e.g. when typing in an input field.

However, we currently support some browser versions that do not support the focus-visible CSS feature, so we also use a fallback block in conjunction with focus-visible.
```

.storybook/recipes/GlobalHeader/GlobalHeader.module.css
dierat: lol I have no idea. I mostly left comments and stuff alone since there's another ticket for that, but I can also clean them up while I'm in here. 🤷

docs/CODE_GUIDELINES.md
dierat: Oh, nice catch, thanks!

src/components/RadioInput/RadioInput.module.css
```diff
@@ -45,11 +45,18 @@
...
-.radio__input:focus + .radio__icon {
+.radio__input:focus-visible + .radio__icon {
   border: var(--eds-border-width-sm) solid var(--eds-theme-color-focus-ring);
   border-radius: 9999px;
 }
 
+@supports not selector(:focus-visible) {
+  .radio__input:focus + .radio__icon {
+    border: var(--eds-border-width-sm) solid var(--eds-theme-color-focus-ring);
+    border-radius: 9999px;
```
jinlee93, line 56: nit: 
```
border-radius: var(--eds-border-radius-full);
```

jinlee93 (approved):
Looks good, just commented some nits

.storybook/recipes/GlobalHeader/GlobalHeader.module.css
ahuth: TODO: figure out what the todos were

ahuth (approved):
Great solution 👍

.storybook/recipes/GlobalHeader/GlobalHeader.module.css
anniehu4: lol ?? I have a vague memory Ian was gonna update the tokens these point to or something. agree we can delete it

anniehu4: I think it was this https://github.com/chanzuckerberg/edu-design-system/pull/955#discussion_r845273010 ... not sure why they didn't remove the TODO when they updated it next

dierat:
Thanks for reviewing this! ❤️

PR #1293, opened 2022-09-19
chore(page-shell): add focus outline

dierat:
### Summary:
I happened to notice while working on something else that the `SkipLink` doesn't have an outline when focused. It is normally hidden and appears on focus, but I think it needs the outline to make it clear that that's where the user's focus is when it appears.

### Before
page shell recipe with the skip link visible, the skip link does not have a white outline

### After
page shell recipe with the skip link visible, the skip link does have a white outline

### Test Plan:
Navigate to the page shell recipe,
tab to the skip link,
and verify there's a white outline around the link.

### DIFF
--- a/.storybook/recipes/PageShell/PageShell.module.css
+++ b/.storybook/recipes/PageShell/PageShell.module.css
@@ -25,6 +25,7 @@
 
   &:focus {
     transform: translateY(0%);
+    @mixin focusInverted;
   }
 
   @media screen and (prefers-reduced-motion) {


jinlee93 (approved):
Tbqh I didn't know this link existed, but having focus indicator makes sense to make it clear

.storybook/recipes/PageShell/PageShell.module.css
```diff
@@ -25,6 +25,7 @@
 
   &:focus {
     transform: translateY(0%);
+    @mixin focusInverted;
```
ahuth, line 28: Good catch!

Do we know why it doesn't have a focus outline to begin with? Are we suppressing native outlines or something, and need to add this mixin where we want them?

dierat: Actually, looking at it again, it does have a focus outline. The issue is that the default focus outline is the same color as the nav. So this is a case where what we really need is the dual focus outline because we don't know for sure what color background the link will appear in front of. I happened to be testing it at a window width where the mobile version of the nav is present, and it was presumably built while testing with a wider window so the nav was on the left.

The skip link in LP has an interesting solution: the link has a background color and the focus outline is offset -2px to place the outline inside the background color, so it will always be purple on white.

learning platform skip link with dev tools open displaying the styles

dierat: Also apparently Along doesn't have a skip link at all.. 🤔

dierat: Anyway I'll close this PR then because:
- it doesn't solve the issue; just transforms it
- the thing I was fixing is just a recipe anyway, not a component
- the skip link in LP is fine and won't be broken when the nav starts moving to the top of the page 
- Along doesn't have a skip link to fix

PR #1294, opened 2022-09-19
chore(filtersbutton): accept custom filters triggers and have default TriggerButton

jinlee93:
[sc-215710]
### Summary:
- as part of https://github.com/chanzuckerberg/edu-design-system/pull/1281#discussion_r972191027 filters button discussion, filters trigger needed expanded functionality to accept any custom trigger element, while having default FiltersButton
- Also makes filters recipe story screen size dependent, which will be closer to use cases

https://user-images.githubusercontent.com/86632227/191121309-4334ca37-e108-4629-84a7-d3f3850c6e00.mov

### Test Plan:
- no visual regression
- unit tests pass


.storybook/recipes/Filters/Filters.tsx
```diff
@@ -1,31 +1,43 @@
...
 } & FiltersPopoverProps;
 
 /**
- * Primary UI component for user interaction
+ * Demonstrates usage of both FiltersPopover and FiltersDrawer depending on screensize.
  */
-export const Filters = ({ as, placement, ...other }: Props) => {
+export const Filters = ({ placement, ...other }: Props) => {
+  const [isLarge, setIsLarge] = useState(false);
+  const popoverBreakpoint = parseInt(breakpoint['eds-bp-md']) * 16; /* 1 */
```
dierat, line 21: Do we need the `/* 1 */` on this line?
```suggestion
  const popoverBreakpoint = parseInt(breakpoint['eds-bp-md']) * 16;
```

.storybook/recipes/Filters/Filters.tsx
```diff
@@ -1,31 +1,43 @@
...
+    }
+  };
   return (
     <div>
-      {as === 'FiltersDrawer' && <FiltersDrawer {...other} />}
-      {as === 'FiltersPopover' && (
-        <FiltersPopover placement={placement} {...other} />
-      )}
+      {!isLarge && <FiltersDrawer {...other} />}
+      {isLarge && <FiltersPopover placement={placement} {...other} />}
```
dierat, line 40: Do you think it's possible to have this kind of functionality built into a wrapper component or will devs need to add this themselves every time?

.storybook/recipes/Filters/Filters.tsx
```diff
@@ -1,31 +1,43 @@
...
+    };
+  });
+
+  const updateScreenSize = () => {
+    if (window.innerWidth >= popoverBreakpoint) {
+      setIsLarge(true);
+    } else {
+      setIsLarge(false);
+    }
+  };
```
dierat, line 36: I'm wondering if we should add debouncing and only update state if it needs to be updated to demonstrate best practices in case someone decides to copy/paste this into the downstream product.

src/components/FiltersDrawer/FiltersDrawer.tsx
```diff
@@ -37,6 +43,11 @@ export type FiltersDrawerProps = {
    * Callback called when the apply button is called.
    */
   onApply?: () => void;
+  /**
+   * Trigger element that toggles the filters drawer.
+   * Must be able to accept an 'onClick: MouseEventHandler' prop.
+   */
```
dierat, line 49: Maybe add something in the prop comments about how consumers need to pass in either `triggerElement` or `triggerText` but not both.

dierat: By `Must be able to accept an 'onClick: MouseEventHandler' prop.` do you mean it shouldn't have that prop already? I'm asking because I'm wondering if line 128 would override any `onClick` already being passed in.

.storybook/recipes/Filters/Filters.tsx
jinlee93: We do not, thanks for checking

.storybook/recipes/Filters/Filters.tsx
jinlee93: I think it would be possible, but I want to leave it up to devs since when to swap between Drawer and Popover can depend on content size or other conditions. Basically want to punt state management as much as possible lol

.storybook/recipes/Filters/Filters.tsx
jinlee93: Good call, will add debouncing

src/components/FiltersDrawer/FiltersDrawer.tsx
jinlee93: > Maybe add something in the prop comments about how consumers need to pass in either triggerElement or triggerText but not both (and that if both are passed in, only the triggerElement will be used).

Sounds good, will add comment

> By Must be able to accept an 'onClick: MouseEventHandler' prop. do you mean it shouldn't have that prop already? I'm asking because I'm wondering if line 128 would override any onClick already being passed in.

In line 128, `() => setIsActive(true)` is being passed to the `triggerElement` as `onClick`, so `triggerElement` needs to be able to accept the `onClick` function.  The comment can be changed to `Must be able to accept an 'onClick: () => void' prop.`

.storybook/recipes/Filters/Filters.tsx
jinlee93: Added debouncing and conditional, still working:


https://user-images.githubusercontent.com/86632227/191861400-28119327-ae79-4e17-9916-5cd0b723316b.mov

dierat (approved):
Thanks for making this change!

src/components/FiltersDrawer/FiltersDrawer.tsx
jinlee93: Damn good question. I made sure to call the original `onClick`, as shown: `console.log('yeehaw')` is original `<button>` `onClick`:

https://user-images.githubusercontent.com/86632227/192010584-5c436fa4-16dd-4086-bc06-8780843bf7bb.mov

dierat: Awesome, thanks!

src/components/FiltersDrawer/FiltersDrawer.tsx
```diff
@@ -101,8 +127,29 @@ export const FiltersDrawer = ({
     }
   }, [filtersBody]);
 
-  const buttonVariant = hasSelectedFilters ? 'primary' : 'secondary';
-  const buttonStatus = hasSelectedFilters ? 'brand' : 'neutral';
+  const addFiltersToTriggerElement = (trigger: ReactNode) => {
+    if (!React.isValidElement(trigger)) return;
```
jinlee93, line 131: I put this line in for accessing `props`/`cloneElement` if this helper function is copypasta'd elsewherem but not exactly necessary here since `trigger` is already being typechecked as ReactNode and truthy checked in the ternary in line 143. 

Let me know if I should remove

dierat: Seems fine to me 🤷

PR #1295, opened 2022-09-19
fix(baseline-card): stop clicks outside visible card area

dierat:
### Summary:
I noticed while working on something else that in the baseline card that is clickable, if you click in the empty space to the right of the card, it still activates the link and navigates you to another page. This PR moves the `onClick` handler to an inner wrapper that is constrained to the card content.

### Video of the bug
https://user-images.githubusercontent.com/7761701/191122393-43acc8f7-be5a-4d37-8aaf-2c1166fec591.mp4

### Test Plan:
Navigate to the `Clickable` story for the `BaselineCard` in storybook,
click in the empty space to the right of the card,
and verify that you're not navigated to another page.

### DIFF
--- a/.storybook/recipes/BaselineCard/BaselineCard.module.css
+++ b/.storybook/recipes/BaselineCard/BaselineCard.module.css
@@ -22,6 +22,17 @@
 }
 
 /**
+ * Inner wrapper that is also clickable
+ *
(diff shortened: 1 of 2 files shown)


jinlee93 (approved):
Weird lol, thanks

PR #1299, opened 2022-09-20
chore(deps): update deps and jest, remove some devdeps

jinlee93:
### Summary:
[sc-215876]
- updates more dependencies, but have elevated risk than the previous pr for deps updates:
  - updates deps in `dependencies` except for `@headlessui/react` which causes typescript and unit test errors with v1.7
  - updates jest related deps in `devdependencies`
    - follows https://jestjs.io/docs/28.x/upgrading-to-jest28 and https://jestjs.io/docs/upgrading-to-jest29
    - biggest relevant change for us is `jsdom` is no longer exported with `jest`, needs `jest-environment-jsdom`
  - removes `autoprefixer`,  `downshift`, `make-dir-cli`, `web-vitals`, and `@storybook/addon-storyshots` which I believe are not being used or are peerdependencies
    - think they were brought over as part of BM merge https://github.com/chanzuckerberg/edu-design-system/commit/c8a3bbc28fb688c94f99b0ff73a64bc4cda0b8d8#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519
    - `downshift` probably used to be used for dropdown which was refactored for more native/React/HeadlessUI
    - `@storybook/addon-storyshots` never used in favor of `generateSnapshots` from `story-utils`
    - advise if `autoprefixer`, `make-dir-cli`, or `web-vitals` are still being used
### Test Plan:
- no hiccups in github actions
- unit tests pass
- no visual regression


package.json
```diff
@@ -128,7 +128,6 @@
     "@types/svg4everybody": "^2.1.2",
     "@typescript-eslint/eslint-plugin": "^5.15.0",
     "@typescript-eslint/parser": "^5.15.0",
-    "autoprefixer": "^10.4.2",
```
jinlee93, line 131: Not sure if this is needed anywhere but looked like it wasn't being used and was not a peerdependency of anything

ahuth (approved):
LGTM

PR #1301, opened 2022-09-21
chore(filters-popover): use flex-end instead of end

dierat:
### Summary:
In `traject` we're seeing the following warning:
```
WARNING in ./node_modules/@chanzuckerberg/eds/lib/components/FiltersPopover/FiltersPopover.module.css (./node_modules/css-loader/dist/cjs.js??ruleSet[1].rules[7].use[1]!./node_modules/postcss-loader/dist/cjs.js!./node_modules/@chanzuckerberg/eds/lib/components/FiltersPopover/FiltersPopover.module.css)
Module Warning (from ./node_modules/postcss-loader/dist/cjs.js):
Warning

(29:3) autoprefixer: end value has mixed support, consider using flex-end instead
```
which seems to be saying we should use `justify-content: flex-end` instead of `justify-content: end` in `FiltersPopover`. This PR makes that change.

### Test Plan:
I made an alpha build from this branch (`5.1.1-alpha.0`), installed that in `traject`, ran localhost, and verified I no longer saw the warning.

I'm also seeing changes in chromatic that look like the styling wasn't working as expected before but is now (some elements are pushed to the right side of the screen whereas before they were on the left).
FiltersPopover:Overflow Interactive story change in chromatic
FiltersPopover:With On Clear story change in chromatic

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,6 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
-### [5.1.1-alpha.0](https://github.com/chanzuckerberg/edu-design-system/compare/v5.1.0...v5.1.1-alpha.0) (2022-09-21)
-
 ## [5.1.0](https://github.com/chanzuckerberg/edu-design-system/compare/v5.0.1...v5.1.0) (2022-09-20)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### [5.1.1-alpha.0](https://github.com/chanzuckerberg/edu-design-system/compare/v5.1.0...v5.1.1-alpha.0) (2022-09-21)
+
 ## [5.1.0](https://github.com/chanzuckerberg/edu-design-system/compare/v5.0.1...v5.1.0) (2022-09-20)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### [5.1.1-alpha.0](https://github.com/chanzuckerberg/edu-design-system/compare/v5.1.0...v5.1.1-alpha.0) (2022-09-21)
+
```
dierat, line 6: I can't remember if we want to commit the alpha build updates or not.

jinlee93: we have not been

dierat: I feel like it's messy and I don't want to have to commit every time I try out an alpha build, but I guess one reason to do it is then it's easier to figure out which alpha build has which change if you're looking for a specific one.

src/components/FiltersPopover/FiltersPopover.module.css
```diff
@@ -23,5 +23,5 @@
 .footer__button-group {
   margin-top: var(--eds-size-3);
   display: flex;
-  justify-content: end;
+  justify-content: flex-end;
```
jinlee93, line 26: Ah, good catch

jinlee93 (approved):
Thanks for the styling fix

CHANGELOG.md
dierat: >I don't think we have been committing the alpha build updates

Cool, I'll take it out.

PR #1302, opened 2022-09-21
chore: remove unused commitizen config from package.json

booc0mtaco:
### Summary:

- We removed the dependency for cz-conventional-commit in the last 6 months. This PR removes the lingering configuration in package.json, which references that dependency. Conventional commits are still enforced by linting rules (commitlint/config-conventional)

### Test Plan:

- ensure CI passes and `install` works as expected

### DIFF
--- a/package.json
+++ b/package.json
@@ -188,10 +188,5 @@
       "stylelint --fix"
     ]
   },
-  "packageManager": "yarn@3.1.1",
-  "config": {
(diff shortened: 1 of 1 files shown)


booc0mtaco:
will move to be based off `next` instead of `main`

jinlee93 (approved):
Looks good, thanks for the cleanup!

PR #1304, opened 2022-09-22
chore(layout): remove expandable prop options

dierat:
### Summary:
I happened to notice some "expandable" prop thingies in `Layout` and `LayoutSection` that don't seem to be used, and I don't think we have any plans to make these things expandable, so I'm removing them.

### Test Plan:
All automated tests pass and there are no visual changes detected by chromatic.


jinlee93 (approved):
Thanks for fighting against unused code

PR #1305, opened 2022-09-22
chore: remove ordered lists in comments

dierat:
### Summary:
As part of the work to clean up the old code patterns we don't really want anymore, I'm waging a just (imo) war against numbered lists in code comments. This PR removes numbered lists in descriptions where the number isn't pointing to a specific line of code; it's just listing out details about the block of code below (usually a CSS block).

I also tried to remove some comments I thought were redundant but I didn't want to spend too much time nit-picking these comments because there are a lot of them.

### Test Plan:
All  automated tests pass and the comments look a little better...

### DIFF
--- a/.storybook/components/IconGrid/IconGrid.css
+++ b/.storybook/components/IconGrid/IconGrid.css
@@ -3,9 +3,9 @@
 \*------------------------------------*/
 
 /**
- * 1) The icon grid is not for production and is meant solely to display
- *    a grid of icons within the pattern library. THerefore, it does not
- *    follow coding conventions
+ * The icon grid is not for production and is meant solely to display
+ * a grid of icons within the pattern library. THerefore, it does not
+ * follow coding conventions
  */
 
 .icon-grid {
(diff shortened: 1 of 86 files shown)


.storybook/components/IconGrid/IconGrid.css
```diff
@@ -3,9 +3,9 @@
 \*------------------------------------*/
 
 /**
- * 1) The icon grid is not for production and is meant solely to display
```
ahuth, line 6: I was wondering why some of these ordered lists don't even reference anything in the code 🤔

ahuth (approved):
LGTM 🤘

.storybook/components/IconGrid/IconGrid.css
dierat: Yeah it's kinda confusing because I sometimes read through the whole CSS block to see if the number is in there somewhere 🤷

dierat:
Especially because I know this is a super long one 😓

PR #1306, opened 2022-09-22
chore: remove empty CSS blocks

dierat:
### Summary:
I noticed a couple of empty CSS blocks that aren't doing anything.

### Test Plan:
All automated tests pass and there are no more references to `block-no-empty` in the codebase.

### DIFF
--- a/src/upcoming-components/Hero/Hero.module.css
+++ b/src/upcoming-components/Hero/Hero.module.css
@@ -79,11 +79,3 @@
   @mixin eds-theme-typography-heading-1;
   margin-bottom: var(--eds-size-2);
 }
-
-/**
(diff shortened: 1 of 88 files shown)


src/upcoming-components/Hero/Hero.module.css
```diff
@@ -79,11 +79,3 @@
   @mixin eds-theme-typography-heading-1;
   margin-bottom: var(--eds-size-2);
 }
-
-/**
- * Hero description
- * 1) Slot for additional hero items
- */
-/* stylelint-disable-next-line block-no-empty */
```
ahuth, line 87: I think the lint rule suppression is the best part 😆

Issue #794, opened 2022-01-20
Tag should not set `max-width`


anniehu4:
## Problem summary
`Tag.module.css` is setting `max-w-xs` (aka max-width: 20rem), which is way too wide for the Tag.

## Expected behavior
Consumer should be able to set their own max-width without having to override through a CSS module

## Screenshots
![image](https://user-images.githubusercontent.com/15840841/150239886-80c2c7b0-487f-4586-a84a-7fae950ff886.png)

Issue #686, opened 2021-10-26
Tag should accept data-testid


anniehu4:
## Problem summary
@meghailango was having trouble targeting a specific Tag in feature testing and I realized we don't allow passing through `data-testid`. This might be an issue on other components as well

## Reproduce
render
```
<Tag data-testid="test">
```

## Expected behavior
`data-testid="test"` is rendered in the DOM

PR #1307, opened 2022-09-22
fix(tag): allow data attributes on tag and remove max-width

jinlee93:
### Summary:
Addresses #686, #794
These issues are a little old but still valid
- Passes `other` props for broad data attributes for Tag
- Removes `max-width` for tag
### Test Plan:
- added `data-testid` to a story
  - no errors and seen in storybook
Screen Shot 2022-09-22 at 9 20 29 AM
  - snap updated including `data-testid`
- no other unit test failures
- no visual regressions

### DIFF
--- a/src/components/Tag/Tag.tsx
+++ b/src/components/Tag/Tag.tsx
@@ -52,6 +52,7 @@ export const Tag = ({
   icon,
   text,
   hasOutline = false,
+  ...other
 }: Props) => {
   const componentClassName = clsx(
     styles['tag'],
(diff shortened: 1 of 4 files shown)


ahuth (approved):
LGTM

src/components/Tag/Tag.tsx
```diff
@@ -52,6 +52,7 @@ export const Tag = ({
   icon,
   text,
   hasOutline = false,
+  ...other
```
ahuth, line 55: What type does `other` have? Is it `any` or `unknown`?

If so, I wonder if it makes sense to be explicit about max-width and data-testid.

jinlee93: In `other` Typescript seems to be unhappy with custom props (i.e. `yee` in this screenshot`) but is ok with props starting with `data-`
image

`max-width` will be set either by consumer via classes

PR #1310, opened 2022-09-23
chore: remove remaining numbered lists in comments

dierat:
### Summary:
This one is huge but I swear it's the last one. This PR removes the rest of the numbered lists in comments, mostly moving them inline but I also removed a couple that seemed redundant (and removed any empty TODO comments I saw as well).

### Test Plan:
Automated tests pass; no chromatic changes.

### DIFF
--- a/.storybook/recipes/BaselineCard/BaselineCard.tsx
+++ b/.storybook/recipes/BaselineCard/BaselineCard.tsx
@@ -49,15 +49,15 @@ export interface Props {
 
 /*
  * Clickable card: link properties
- * 1) Href and text are required attributes
- * 2) Text will be used by screen readers; it will only be visible when the
- * link is focused by a keyboard
- * 3) Target attribute is optional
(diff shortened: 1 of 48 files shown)


.storybook/recipes/BaselineCard/BaselineCard.tsx
```diff
@@ -82,35 +82,31 @@ export const BaselineCard = ({
   const linkRef = useRef() as MutableRefObject<HTMLAnchorElement>;
 
   /**
-   * useEffect hook
```
dierat, line 85: I removed some comments where they just reworded code that's pretty self-explanatory. (I wasn't consistent about this though because it's easier to just copy/paste than to analyze the comment and determine whether it should be removed or reworded, etc.)

ahuth (approved):

anniehu4:
it's so beautiful :')

PR #1311, opened 2022-09-23
feat(Tabs): remove and clean up props for Tabs (sub-)components

booc0mtaco:
### Summary:

[sc-216176] remove unused props from parent component, and use the exported type for the Tab sub-component in place of `any`  in the parent to ensure props are applied accurately.

### Test Plan:

- ensure components build and existing tests pass
- update snapshots to make sure props generate appropriately


src/components/Tab/Tab.tsx
```diff
@@ -39,7 +39,6 @@ export const Tab = ({
   className,
   id,
   'aria-labelledby': ariaLabelledBy,
-  title,
```
booc0mtaco, line 42: Removing this allows `title` to fall thru to the `...other`, which means we now can support title attributes on tabs (before, this named prop was not being used below, preventing the title from actually showing).

updated the snapshot to cover this from regressing.

src/components/Tab/index.ts
```diff
@@ -1 +1,2 @@
 export { Tab as default } from './Tab';
+export { Props as TabProps } from './Tab';
```
booc0mtaco, line 2: exporting in index, so we can use it as a type in the parent component

src/components/Tabs/Tabs.tsx
```diff
@@ -167,13 +121,13 @@ export const Tabs = ({
...
    */
   useEffect(() => {
-    // TODO: improve `any` type
     setId(
-      tabs().map((tab: any) => (tab.props.id ? tab.props.id : getUID(tab))),
+      tabs().map((tab: React.ReactElement<TabProps>) =>
+        tab.props.id ? tab.props.id : getUID(tab),
+      ),
     );
-    // TODO: improve `any` type
```
dierat, line 174: Thanks for improving these!

src/components/Tab/Tab.tsx
dierat: The only effect this has is the little popover text when a mouse user hovers over the tabbed panel element, right?

booc0mtaco: @dierat correct. this change is mostly to support what some of our snapshots were testing (we had a snapshot specifying title but not rendering it, which was a bug). I don't think we wanna rely on `title` in general, tho... it's apparently a bit iffy on the a11y side of things

booc0mtaco:
Full disclosure, read up on this to see what we might wanna suggest re: `title` attribute usage as I was curious about its usage, and these folks seem pretty against it https://www.tpgi.com/using-the-html-title-attribute/ Food for thought 🍔

src/components/Tab/Tab.tsx
dierat: In this case, the `aria-labelledby` attribute is overriding `title` for screen readers anyway, so I'm not concerned about it reducing the experience. What I'm wondering is whether we need it at all or if it should be nixed with the others since it's not adding much.

src/components/Tabs/Tabs.tsx
```diff
@@ -167,13 +121,13 @@ export const Tabs = ({
    * Autogenerate ids on tabs if not defined.
    */
   useEffect(() => {
-    // TODO: improve `any` type
     setId(
-      tabs().map((tab: any) => (tab.props.id ? tab.props.id : getUID(tab))),
+      tabs().map((tab: React.ReactElement<TabProps>) =>
```
ahuth, line 125: Can we update the types of `tabs()` so that TS knows what these are without us having to specify it?

I'm guessing we don't have good types for `react-children-by-type`, so there's 2 options I see:
1. Remove `react-children-by-type` and do whatever logic we need ourselves
2. Write better custom types for `react-children-by-type`
3. Pass a type param to `useCallback`

3 is the easiest. Would have a similar result to specifying here (since either way we're telling TS what to expect). But I think it's better since generally we shouldn't need to specify the types of callback args.

dierat:
> Full disclosure, read up on this to see what we might wanna suggest re: `title` attribute usage as I was curious about its usage, and these folks seem pretty against it https://www.tpgi.com/using-the-html-title-attribute/ Food for thought 🍔

> Takeaway
If you want to hide content from mobile and tablet users as well as assistive tech users and keyboard only users, use the title attribute.

the "archaic rap" meme of a fancy gentleman in a classical painting with text that reads "that retort was indeed most inflict my good man"

src/components/Tabs/Tabs.tsx
```diff
@@ -97,19 +61,9 @@ export interface Props {
...
-  hideLegend,
-  id,
   onChange,
-  overflow,
-  radioLegend,
-  radioLegendPosition,
-  required,
-  size,
-  tabsOnClick,
-  variant,
```
ahuth, line 112: Yikes, all these props aren't used?

dierat: This is one of those components that came as a template from Big Medium and was updated to work for our needs. I'm guessing these were all very useful in one or more of their previous projects.

src/components/Tabs/Tabs.tsx
ahuth: Can also do this later.

ahuth (approved):
LGTM

src/components/Tabs/Tabs.tsx
ahuth: Yea, I pretty much assume any oddities are BM's fault.

src/components/Tab/Tab.tsx
booc0mtaco: yah, from what I can glean, the snapshot only incidentally is specifying `title` so we  could also specify some other attribute supported on the tag just to demonstrate `...other` is doing what it should. 

There is no longer code directly specifying `title` so this already nixes that part

dierat: >we could also specify some other attribute supported on the tag just to demonstrate ...other is doing what it should

I don't think we need a test for the `...other` functionality. You can add something (like a `data-testid` attribute) if you want, but I'm also fine with just removing the `title` and moving on.

src/components/Tab/Tab.tsx
```diff
@@ -22,7 +22,7 @@ export interface Props {
    */
   isActiveIndex?: number;
   /**
-   * Number passed down from Tabs to show the active index state of Tabs
+   * Text used to label the tab in the tab list
```
dierat, line 25: Oo thanks for updating this 👍

dierat (approved):
Awesome work, thanks!

PR #1312, opened 2022-09-23
docs: update language around 'next' base branch

booc0mtaco:
### Summary:

From onboarding: I believe the new process is rolled out, so updating the text here to reflect how to use `next` when making new changes.

### Test Plan:

- n/a (documentation change)

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -2,24 +2,24 @@
 
 ## Git Flow
 
-TODO: CZI to update to match specific git working model.
-
 This project uses [Git-flow](https://nvie.com/posts/a-successful-git-branching-model/) to manage the development workflow. Here's a breakdown of what this means for the design system:
 
-- The `main` always represents the latest production-ready version of the design system. Every commit to `main` indicates a new release.
-- `next` represents the unpublished next Major release of the design system. **use this branch as the base of your changes**.
+- The `main` always represents the latest production-ready version of the design system. Each commit to `main` indicates a new release (either from changes on `release` or from `hotfix`).
+- `next` represents the unpublished changes for the next release of the design system. These can be standalone commits, or merged in from `feature`.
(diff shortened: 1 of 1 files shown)


docs/CONTRIBUTING.md
```diff
@@ -7,7 +7,7 @@ TODO: CZI to update to match specific git working model.
 This project uses [Git-flow](https://nvie.com/posts/a-successful-git-branching-model/) to manage the development workflow. Here's a breakdown of what this means for the design system:
 
 - The `main` always represents the latest production-ready version of the design system. Every commit to `main` indicates a new release.
-- `next` represents the unpublished next Major release of the design system. All feature work that belongs to the new architecture  TODO: revisit the existance of `next` once the new architecture is in place and launched
+- `next` represents the unpublished next Major release of the design system. **use this branch as the base of your changes**.
```
dierat, line 10: 2 things:
- you can use develop off `main` directly if it's a bugfix that you want to put out quickly without messing with publishing everything currently on `next`
- this was already in there but I'm wondering if "Major" release makes sense here because the next release off `next` could be minor or patch (or maybe I'm misunderstanding what this line is trying to say 🤔 )

booc0mtaco: @dierat yah you are right, GitFlow specifies how hot fixes can be based off `main` to expedite things (which can get merged back into `next`). I think "Major" in this sentence means closer to "substantial"? Even then, we could have one commit on `next` that fixes a typo so ...

Honestly i think inverting these bullets would be clearer, e.g., "for urgent hot fixes, etc., base work off of `main` branch" and so on. Right now, they only allude to which branch to use.

dierat: I would also be fine with tweaking the copy to include the hotfix exemption. Maybe something like

>- The `main` always represents the latest production-ready version of the design system. Every commit to `main` indicates a new release. The only commits made to this branch should be regular releases and hotfixes.
>- `next` represents the unpublished next Major release of the design system. **Use this branch as the base of your changes** for regular changes (hotfixes can be branched off `main`.

dierat (approved):
This is great, thanks!

PR #1314, opened 2022-09-24
Add types for react-children-by-type

ahuth:
[sc-196807]

### Summary:

Add types for `react-children-by-types`, to improve the types, of some of our types.

Inspired by the cleanups in #1311.

### Test Plan:

- Type checking
- Unit tests

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -83,7 +83,6 @@ export const Tabs = ({
     return allByType(children, Tab);
   }, [children]);
 
-  // When rendered, we use the anchor tag below to attach these refs
   const tabRefs = tabs().map(() => React.createRef<HTMLAnchorElement>());
 
   // we can't use the hook in an iterator like this, so generate the base and increment if needed
@@ -246,7 +245,7 @@ export const Tabs = ({
     scrollableRight && styles['tabs--scrollable-right'],
   );
 
-  const childrenWithProps = React.Children.map(tabs(), (child, i: number) => {
+  const childrenWithProps = React.Children.map(tabs(), (child, i) => {
     // Checking isValidElement is the safe way and avoids a typescript
     // error too.
     if (React.isValidElement(child)) {
@@ -266,7 +265,7 @@ export const Tabs = ({
         ref={headerRef}
       >
         <ul className={styles['tabs__list']} role="tablist">
-          {tabs().map((tab, i: number) => {
+          {tabs().map((tab, i) => {
             const isActive = activeIndexState === i;
             return (
               <li
--- a/src/components/TimelineNav/TimelineNav.tsx
+++ b/src/components/TimelineNav/TimelineNav.tsx
@@ -126,7 +126,9 @@ export const TimelineNav = ({
     return allByType(children, TimelineNavPanel);
   }, [children]);
 
-  const timelineNavItemRefs = timelineNavItems().map(() => React.createRef());
+  const timelineNavItemRefs = timelineNavItems().map(() =>
+    React.createRef<HTMLAnchorElement>(),
+  );
 
   // we can't use the hook in an iterator like this, so generate the base and increment if needed
   const [idVar, setId] = useState<string[]>([]);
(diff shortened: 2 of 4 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -161,21 +162,17 @@ export const Tabs = ({
...
       setActiveIndexState(activeIndex);
-      tabRefs[activeIndex].current.focus();
+      tabRefs[activeIndex].current?.focus();
     }
   }, [prevActiveIndex, activeIndex, tabRefs]);
 
   /**
    * Autogenerate ids on tabs if not defined.
    */
   useEffect(() => {
```
ahuth, line 172: Pretty sure none of these `useEffect`s in these files are working as expected. They're probably re-rendering unnecessarily (which may not really be that big of a deal). I'll look at that.

src/components/Tabs/Tabs.tsx
```diff
@@ -292,23 +288,19 @@ export const Tabs = ({
...
-        // @ts-expect-error TODO: fix "No overload matches this call" error
-        return React.cloneElement<Props>(child, {
-          id: idVar[i],
-          ['aria-labelledby']: ariaLabelledByVar[i],
-        });
-      }
-      return child;
-    },
-  );
+  const childrenWithProps = React.Children.map(tabs(), (child, i) => {
```
ahuth, line 291: Formatting changed, but the actual change here is removing the explicit types from `child` and `i`. TypeScript now understands what those are without us specifying them.

src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -126,7 +126,9 @@ export const TimelineNav = ({
     return allByType(children, TimelineNavPanel);
   }, [children]);
 
-  const timelineNavItemRefs = timelineNavItems().map(() => React.createRef());
+  const timelineNavItemRefs = timelineNavItems().map(() =>
+    React.createRef<HTMLAnchorElement>(),
```
ahuth, line 130: I'm not convinced this is good use of `React.createRef`, since it's creating a new ref on every render (which sort of defeats the purpose of a ref). Will investigate.

src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -160,7 +162,7 @@ export const TimelineNav = ({
       prevActiveIndex !== activeIndex
     ) {
       setActiveIndexState(activeIndex);
-      timelineNavItemRefs[activeIndex].current.focus();
+      timelineNavItemRefs[activeIndex].current?.focus();
```
ahuth, line 165: Now that we have better types, the ref isn't `any` anymore, and we actually have to handle the `null` case.

src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -342,18 +340,23 @@ export const TimelineNav = ({
...
+       * activeIndexState holds the index of the last nav item selected in the
+       * timelinNavItemRefs array.
+       */
+      element.focus();
+      /**
+       * Because the 'Back' button's onFocus listener removes the selected nav
+       * item from the tab order, we need to manually insert it back into the tab
+       * order by setting its tabIndex back to "0".
+       */
+      element.tabIndex = 0;
```
ahuth, line 357: Note the change from `'0'` to `0`. Apparently we're supposed to use numbers, not strings, but before the ref pointed to an `any`.

src/custom.d.ts
```diff
@@ -10,6 +10,17 @@ declare module '*.module.css' {
...
+
+  export function allByType<P>(
+    children: ReactNode,
+    type: JSXElementConstructor<P>,
+  ): ReactElement<P>[];
+
+  export function oneByType<P>(
+    children: ReactNode,
+    type: JSXElementConstructor<P>,
+  ): ReactElement<P>;
```
ahuth, line 25: We should contribute these types back to DefinitelyTyped.

booc0mtaco:
@ahuth some of these tweaks between PRs may overlap looks like. Just went back in and started fixing the remaining `any`s and did some similar changes for the refs. Both of these have some good stuff, so happy to consolidate next week

ahuth:
Thanks, @booc0mtaco. I mostly wanted to get some thoughts onto a branch. I'll wait to merge until #1311 is in, and will deal with any conflicts then.

anniehu4:
https://app.shortcut.com/czi-edu/story/196807/add-better-typing-for-react-children-by-type

dierat (approved):
Nice!

jeremiah-clothier (approved):

ahuth:
Thanks!

PR #1315, opened 2022-09-26
Cleanup

ahuth:
### Summary:

Follow up to #1311 and #1314.

Cleaning up some stuff:
- Removed unnecessary type casting when using `useRef`
- Removed all suppressions of the react-hooks/exhaustive-deps rule
- Simplified `useMergedRefs` (too achieve the above bullet)
- Added some tests

mr clean

### Test Plan:

- Tests
- Storybook

### DIFF
--- a/.storybook/recipes/BaselineCard/BaselineCard.tsx
+++ b/.storybook/recipes/BaselineCard/BaselineCard.tsx
@@ -1,5 +1,5 @@
 import clsx from 'clsx';
-import React, { MutableRefObject, ReactNode, useEffect, useRef } from 'react';
+import React, { type ReactNode, useEffect, useRef } from 'react';
 import styles from './BaselineCard.module.css';
 
 import { Card, CardBody, CardFooter, CardHeader, Score } from '../../../src';
@@ -78,8 +78,8 @@ export const BaselineCard = ({
    * Initialize refs to use on the card container and the link
    *  (for clickable card)
    */
-  const cardRef = useRef() as MutableRefObject<HTMLDivElement>;
-  const linkRef = useRef() as MutableRefObject<HTMLAnchorElement>;
+  const cardRef = useRef<HTMLDivElement>(null);
+  const linkRef = useRef<HTMLAnchorElement>(null);
 
   /**
    * Hook up the link functionality on component mount if a link is present
@@ -103,7 +103,7 @@ export const BaselineCard = ({
     function handleClick() {
       const isTextSelected = window.getSelection()?.toString();
       if (!isTextSelected) {
-        linkRef.current.click();
+        linkRef.current?.click();
       }
     }
 
--- a/src/components/Breadcrumbs/Breadcrumbs.test.tsx
+++ b/src/components/Breadcrumbs/Breadcrumbs.test.tsx
@@ -1,21 +1,43 @@
 import { generateSnapshots } from '@chanzuckerberg/story-utils';
-import { render } from '@testing-library/react';
+import { composeStories } from '@storybook/testing-react';
+import { render, screen, waitFor } from '@testing-library/react';
 import React from 'react';
-import { Breadcrumbs } from './Breadcrumbs';
-import * as BreadcrumbsStoryFile from './Breadcrumbs.stories';
+import * as stories from './Breadcrumbs.stories';
+
+const { LongList } = composeStories(stories);
 
 describe('<Breadcrumbs />', () => {
-  generateSnapshots(BreadcrumbsStoryFile);
-  it('throws an error if there are invalid children', () => {
-    expect(() => {
-      render(
-        <Breadcrumbs>
-          <Breadcrumbs.Item href="#" text="Breadcrumb 1" />
-          <button>Breadcrumb 2</button>
-        </Breadcrumbs>,
-      );
-    }).toThrow(
-      /Only <Breadcrumbs.Item>, <BreadcrumbsItem>, or React.Fragment of aforementioned components allowed/,
-    );
+  generateSnapshots(stories);
+
+  describe('truncation', () => {
+    it('truncates when its content overflows', async () => {
+      render(<LongList />);
+      const list = screen.getByRole('list');
+
+      Object.defineProperty(list, 'clientWidth', { value: 100 });
+      Object.defineProperty(list, 'scrollWidth', { value: 200 });
+      expect(list.clientWidth).toBeLessThan(list.scrollWidth);
+
+      window.dispatchEvent(new Event('resize'));
+
+      await waitFor(() => {
+        expect(screen.getAllByRole('listitem').length).toEqual(4);
+      });
+    });
+
+    it('does not truncate when its content does not overflow', async () => {
+      render(<LongList />);
+      const list = screen.getByRole('list');
+
+      Object.defineProperty(list, 'clientWidth', { value: 200 });
+      Object.defineProperty(list, 'scrollWidth', { value: 200 });
+      expect(list.clientWidth).toBeGreaterThanOrEqual(list.scrollWidth);
+
+      window.dispatchEvent(new Event('resize'));
+
+      await waitFor(() => {
+        expect(screen.getAllByRole('listitem').length).toEqual(12);
+      });
+    });
   });
 });
--- a/src/components/Breadcrumbs/Breadcrumbs.tsx
+++ b/src/components/Breadcrumbs/Breadcrumbs.tsx
@@ -1,6 +1,6 @@
 import clsx from 'clsx';
 import debounce from 'lodash.debounce';
-import React, { ReactNode } from 'react';
+import React, { type ReactNode } from 'react';
 import { useUID } from 'react-uid';
 import styles from './Breadcrumbs.module.css';
 import { flattenReactChildren } from '../../util/flattenReactChildren';
@@ -47,36 +47,32 @@ export const Breadcrumbs = ({
   const ref = React.useRef<HTMLUListElement>(null);
 
   /**
-   * Checks if breadcrumbs would overflow
-   */
-  const updateShouldTruncate = () => {
-    setShouldTruncate(false);
-    if (
-      ref &&
-      ref.current &&
-      ref.current.clientWidth < ref.current.scrollWidth
-    ) {
-      setShouldTruncate(true);
-    }
-  };
-  const debouncedUpdateShouldTruncate = debounce(updateShouldTruncate, 200);
-
-  /**
    * Needs useLayoutEffect over useEffect since it needs to be run before paint.
    * TODO: with React 18, might be able to use useEffect https://github.com/reactjs/reactjs.org/blob/d14cbdca2445cd676526c4c52e1e106342ff7bb3/content/docs/hooks-reference.md?plain=1#L155
    */
   React.useLayoutEffect(() => {
+    const updateShouldTruncate = () => {
+      const willOverflow = ref.current
+        ? ref.current.clientWidth < ref.current.scrollWidth
+        : false;
+
+      setShouldTruncate(willOverflow);
+    };
+
+    const debouncedUpdateShouldTruncate = debounce(updateShouldTruncate, 200);
+
     updateShouldTruncate();
     window.addEventListener('resize', debouncedUpdateShouldTruncate);
     return () => {
       window.removeEventListener('resize', debouncedUpdateShouldTruncate);
     };
-  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+  }, []);
 
   /**
    * Flattens breadcrumb items that may be wrapped in React.Fragment into an array so they can be manipulated easily.
    */
   const breadcrumbsItems = flattenBreadcrumbsItems(children);
+
   /**
    * Finds the second to last breadcrumb item for mobile Breadcrumbs back icon.
    */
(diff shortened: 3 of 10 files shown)


src/components/Breadcrumbs/Breadcrumbs.test.tsx
```diff
@@ -1,21 +1,43 @@
...
 import React from 'react';
-import { Breadcrumbs } from './Breadcrumbs';
-import * as BreadcrumbsStoryFile from './Breadcrumbs.stories';
+import * as stories from './Breadcrumbs.stories';
+
+const { LongList } = composeStories(stories);
 
 describe('<Breadcrumbs />', () => {
-  generateSnapshots(BreadcrumbsStoryFile);
-  it('throws an error if there are invalid children', () => {
```
ahuth, line 9: Deleted this test, because it causes any other tests to fail (not sure why).

Figured we're better off having tests testing production behavior, instead of a test testing dev behavior.

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -46,37 +46,33 @@ export const Breadcrumbs = ({
...
    * Needs useLayoutEffect over useEffect since it needs to be run before paint.
    * TODO: with React 18, might be able to use useEffect https://github.com/reactjs/reactjs.org/blob/d14cbdca2445cd676526c4c52e1e106342ff7bb3/content/docs/hooks-reference.md?plain=1#L155
    */
   React.useLayoutEffect(() => {
+    const updateShouldTruncate = () => {
+      const willOverflow = ref.current
+        ? ref.current.clientWidth < ref.current.scrollWidth
+        : false;
+
+      setShouldTruncate(willOverflow);
```
ahuth, line 59: Note the change of logic from

```ts
setShouldTruncate(false);
if (someConditions) {
  setShouldTruncate(true);
}
```

to

```ts
setShouldTruncate(someConditions);
```

Result should be the same, since `setState` updates get batched. Everything seems the same in storybook and in the added unit tests, too.

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -46,37 +46,33 @@ export const Breadcrumbs = ({
...
-    }
-  };
-  const debouncedUpdateShouldTruncate = debounce(updateShouldTruncate, 200);
-
   /**
    * Needs useLayoutEffect over useEffect since it needs to be run before paint.
    * TODO: with React 18, might be able to use useEffect https://github.com/reactjs/reactjs.org/blob/d14cbdca2445cd676526c4c52e1e106342ff7bb3/content/docs/hooks-reference.md?plain=1#L155
    */
   React.useLayoutEffect(() => {
+    const updateShouldTruncate = () => {
```
ahuth, line 54: This function is only used in this effect, so moving it in removes the needs to worry about it as a dep.

src/components/DragDrop/DragDrop.tsx
```diff
@@ -78,60 +77,63 @@ export const DragDrop = ({
   /**
    * Target drag drop and drag drop inner DOM elements
    */
-  const dragDropInnerRef = useRef() as MutableRefObject<HTMLDivElement>;
-  const dragDropRef = useRef() as MutableRefObject<HTMLElement>;
+  const dragDropInnerRef = useRef<HTMLDivElement>(null);
+  const dragDropRef = useRef<HTMLElement | null>(null);
```
ahuth, line 81: Specifying the `| null` type makes the ref "mutable" from TS's perspective.

Not sure why the types are written like that, but that's how it is 🤷

src/components/DragDrop/DragDrop.tsx
```diff
@@ -78,60 +77,63 @@ export const DragDrop = ({
...
-    let childrenWidth = 0;
-
-    overflowListItems.forEach(function (item) {
-      const itemMarginLeft = parseInt(
-        getComputedStyle(item).marginLeft.slice(0, -2),
-      );
-      childrenWidth += item.clientWidth + itemMarginLeft;
-    });
-    if (overflowListItems) {
+    if (dragDropInnerRef.current) {
```
ahuth, line 86: Now that we're not type casting (the `as` has been removed), we have to handle the `null` case. Handling it here by adding a top-level `if`.

src/components/DragDrop/DragDrop.tsx
```diff
@@ -78,60 +77,63 @@ export const DragDrop = ({
...
-        setIsStart(true);
-      } else {
+      overflowListItems.forEach(function (item) {
+        const itemMarginLeft = parseInt(
+          getComputedStyle(item).marginLeft.slice(0, -2),
+        );
+        childrenWidth += item.clientWidth + itemMarginLeft;
+      });
+
+      if (dragDropRef.current) {
```
ahuth, line 100: May be hard to tell in the diff, but I've removed the

```ts
if (overflowListItems) {
```

check. That array will always be truthy, so it's not necessary.

Replaced with this

```ts
if (dragDropRef.current) {
```

which is needed because we have to do the `null` check now.

src/components/Drawer/Drawer.tsx
```diff
@@ -75,8 +76,8 @@ export const Drawer = ({
    */
   const [activeFocus, setActiveFocus] = useState(isActive);
   const windowRef = useRef<HTMLElement | null>(null);
-  const ref = useRef<HTMLDivElement | null>(null);
   const isMountedRef = useRef(true);
+  const onCloseRef = useRef(onClose);
```
ahuth, line 80: Capturing a ref to the `onClose` prop, so we can use it in callbacks and effects without worrying about whether the dep has changed or not.

src/components/Drawer/Drawer.tsx
```diff
@@ -75,8 +76,8 @@ export const Drawer = ({
    */
   const [activeFocus, setActiveFocus] = useState(isActive);
   const windowRef = useRef<HTMLElement | null>(null);
-  const ref = useRef<HTMLDivElement | null>(null);
```
ahuth, line 78: Not used

src/components/OverflowList/OverflowList.tsx
```diff
@@ -42,62 +41,68 @@ export const OverflowList = ({ className, children, ...other }: Props) => {
...
-    let childrenWidth = 0;
-
-    overflowListItems.forEach(function (item) {
-      const itemMarginLeft = parseInt(
-        getComputedStyle(item).marginLeft.slice(0, -2),
-      );
-      childrenWidth += item.clientWidth + itemMarginLeft;
-    });
-    if (overflowListItems) {
+    if (overflowListInnerRef.current) {
```
ahuth, line 50: This code is definitely copy and pasted.

src/components/Tabs/Tabs.tsx
```diff
@@ -186,37 +187,39 @@ export const Tabs = ({
   /**
    * Handles if scroll fade indicators should be displayed.
    */
-  const handleTabsScroll = debounce(
```
ahuth, line 189: Will debounce once in the effect, instead of on any renders.

src/hooks/useMergedRefs.ts
```diff
@@ -1,57 +1,33 @@
...
- * React.Ref uses the readonly type `React.RefObject` instead of
- * `React.MutableRefObject`, We pretty much always assume ref objects are
- * mutable (at least when we create them), so this type is a workaround so some
- * of the weird mechanics of using refs with TS.
- */
-export type AssignableRef<ValueType> =
-  | {
-      bivarianceHack(instance: ValueType | null): void;
-    }['bivarianceHack']
-  | React.MutableRefObject<ValueType | null>;
```
ahuth, line 17: Really just wanted to get rid of the lint rule suppression in here, but decided to simplify while I was at it. Should make it much more clear what's going on here.

src/hooks/useMergedRefs.ts
```diff
@@ -1,57 +1,33 @@
...
   if (typeof ref === 'function') {
     ref(value);
   } else {
-    try {
-      ref.current = value;
-    } catch (error) {
-      throw new Error(`Cannot assign value "${value}" to ref "${ref}"`);
-    }
+    // @ts-expect-error RefObjects have a readonly type, but we're going to assign to it anyway.
+    ref.current = value;
```
ahuth, line 12: Could probably muck around with the types to get a MutableRefObject here, but I figured this was the simplest/cleanest/easiest.

src/hooks/useMergedRefs.ts
```diff
@@ -1,57 +1,33 @@
...
+ * @example
  *
- * @param refs Refs to fork
+ * const combinedRef = useMergedRefs(refA, refB);
+ * return <span ref={combinedRef}>hi</span>;
  */
-export function useMergedRefs<RefValueType = any>(
-  ...refs: (AssignableRef<RefValueType> | null | undefined)[]
-) {
-  return useCallback((node: any) => {
```
ahuth, line 51: Don't think we need the `useCallback` here, because
- Typically we don't use it when passing a callback ref - e.g. `<span ref={(node) => console.log(node)} />`
- As far as I can tell, `ref` isn't handled like other props, and we don't need to protect against re-renders

src/hooks/useMergedRefs.ts
```diff
@@ -1,57 +1,33 @@
...
-) {
+function assignRef<Val>(ref: Ref<Val>, value: Val) {
   if (ref == null) return;
   if (typeof ref === 'function') {
     ref(value);
   } else {
-    try {
-      ref.current = value;
-    } catch (error) {
-      throw new Error(`Cannot assign value "${value}" to ref "${ref}"`);
```
ahuth, line 36: Don't need the try-catch, since we're handling every possible cause of `Ref<T>`. That's the beauty of static typing.

src/hooks/useMergedRefs.ts
```diff
@@ -1,57 +1,33 @@
...
+    // @ts-expect-error RefObjects have a readonly type, but we're going to assign to it anyway.
+    ref.current = value;
   }
 }
 
 /**
- * Passes or assigns a value to multiple refs (typically a DOM node). Useful for
- * dealing with components that need an explicit ref for DOM calculations but
- * also forwards refs assigned by an app.
+ * Assign one value to multiple refs.
```
ahuth, line 17: Kind of tempted to delete this entire file, since it's
- Used in one place (UtilityNavItemThing)
- Overly generic (the one place its used is only dealing with RefObjects, not RefObjects+RefCallbacks)

src/components/OverflowList/OverflowList.tsx
dierat: Could we invert this if check to avoid indenting further? Like
```
      if (!overflowListInnerRef.current) {
        return;
      }

      /**
       * Target the actual overflow list inside overflow list
       */
      const overflowListItems = Array.from(
        overflowListInnerRef.current.children,
      );
      [...]
```

ahuth: Good idea

src/components/DragDrop/DragDrop.tsx
```diff
@@ -96,7 +99,8 @@ export const DragDrop = ({
       );
       childrenWidth += item.clientWidth + itemMarginLeft;
     });
-    if (overflowListItems) {
```
ahuth, line 99: `overflowListItems` is always truthy (since it's an array), and we don't need this conditional. We do, however, now need to do a null check on `dragDropRef.current`.

src/components/Breadcrumbs/Breadcrumbs.test.tsx
```diff
@@ -1,21 +1,43 @@
...
-          <Breadcrumbs.Item href="#" text="Breadcrumb 1" />
-          <button>Breadcrumb 2</button>
-        </Breadcrumbs>,
-      );
-    }).toThrow(
-      /Only <Breadcrumbs.Item>, <BreadcrumbsItem>, or React.Fragment of aforementioned components allowed/,
-    );
+  generateSnapshots(stories);
+
+  describe('truncation', () => {
```
dierat, line 12: Thanks for the great tests ❤️

dierat (commented):
Makes sense to me 👍 Thanks for the cleanup! ❤️

.storybook/recipes/BaselineCard/BaselineCard.tsx
```diff
@@ -78,8 +78,8 @@ export const BaselineCard = ({
    * Initialize refs to use on the card container and the link
    *  (for clickable card)
    */
-  const cardRef = useRef() as MutableRefObject<HTMLDivElement>;
-  const linkRef = useRef() as MutableRefObject<HTMLAnchorElement>;
```
jeremiah-clothier, line 82: Checking my understanding, `MutableRefObject` is for when you want to manipulate the value of `myref.current` yourself. Otherwise, if the code is just assigning `ref={myref}` then you're fine without it?

jeremiah-clothier (approved):

.storybook/recipes/BaselineCard/BaselineCard.tsx
ahuth: Yes, that's right.

Although normally I wouldn't expect to manually give a ref its type (since we create them through `useRef` or `createRef`.

But I guess for function params or something we'd have to think about it.

ahuth:
Thanks!

PR #1316, opened 2022-09-26
docs: readme updates, remove flowgen

anniehu4:
### Summary:

@gulbrand-czi pointed out that the README needs some small updates:
- variant names are kebab-case, not camelCase
- we're no longer supporting flowgen (this PR goes ahead and removes the flowgen scripts as well -- use TS, dammit!)

### Test Plan:
n/a


ahuth (approved):
LGTM 🤘

PR #1320, opened 2022-09-28
Tabs cleanup #3

ahuth:
### Summary:

More tabs cleanup:
- Remove unused prop
- Simplify getting tab children, refs, and ids
- Don't clone elements that won't be shown
- Fix some logic
- Use good keys

### Test Plan:

- Storybook
- Tests

### DIFF
--- a/src/components/Tabs/Tabs.test.tsx
+++ b/src/components/Tabs/Tabs.test.tsx
@@ -9,6 +9,7 @@ const { Default } = composeStories(stories);
 
 describe('<Tabs />', () => {
   generateSnapshots(stories);
+
   it('should focus and select with keyboard controls', () => {
     render(<Default />);
     const firstTab = screen.getByRole('tab', { name: 'Tab Title 1' });
@@ -31,5 +32,13 @@ describe('<Tabs />', () => {
       'aria-selected',
       'true',
     );
+  });
+
+  it('changes the active tab when activeIndex changes', () => {
+    const { rerender } = render(<Default activeIndex={0} />);
+    expect(screen.getByRole('heading', { name: 'Tab 1' })).toBeInTheDocument();
+
+    rerender(<Default activeIndex={1} />);
+    expect(screen.getByRole('heading', { name: 'Tab 2' })).toBeInTheDocument();
   });
 });
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -2,10 +2,11 @@ import clsx from 'clsx';
 import debounce from 'lodash.debounce';
 import React, {
   type ReactNode,
+  useCallback,
+  useEffect,
+  useMemo,
   useRef,
   useState,
-  useEffect,
-  useCallback,
   type KeyboardEvent,
 } from 'react';
 import { allByType } from 'react-children-by-type';
@@ -60,21 +61,32 @@ export const Tabs = ({
   onChange,
   ...other
 }: Props) => {
+  const getUID = useUIDSeed();
   const ref = useRef<number | undefined>();
   const headerRef = useRef<HTMLDivElement>(null);
   const [activeIndexState, setActiveIndexState] = useState(activeIndex);
   const [scrollableLeft, setScrollableLeft] = useState<boolean>(false);
   const [scrollableRight, setScrollableRight] = useState<boolean>(false);
 
-  /** Get all children that are Tab components. Ignore any others. */
-  const getTabs = useCallback(() => {
+  /** Children that are actually tabs. Any others are ignored. */
+  const tabs = useMemo(() => {
     return allByType(children, Tab);
   }, [children]);
 
-  const tabRefs = getTabs().map(() => React.createRef<HTMLAnchorElement>());
-  const [tabPanelIds, setTabPanelIds] = useState<string[]>([]);
-  const [tabIds, setTabIds] = useState<string[]>([]);
-  const getUID = useUIDSeed();
+  const tabRefs = useMemo(
+    () => tabs.map(() => React.createRef<HTMLAnchorElement>()),
+    [tabs],
+  );
+
+  const tabPanelIds = useMemo(
+    () => tabs.map((tab, i) => tab.props.id || getUID('panel' + i)),
+    [tabs, getUID],
+  );
+
+  const tabIds = useMemo(
+    () => tabs.map((tab) => tab.props['aria-labelledby'] || getUID(tab)),
+    [tabs, getUID],
+  );
 
   /**
    * Get previous prop
@@ -104,21 +116,6 @@ export const Tabs = ({
       tabRefs[activeIndex].current?.focus();
     }
   }, [prevActiveIndex, activeIndex, tabRefs]);
-
-  // Set the tab and tab panel ids. Autogenerate them if necessary.
-  useEffect(() => {
-    setTabPanelIds(
-      getTabs().map((tab) => (tab.props.id ? tab.props.id : getUID(tab))),
-    );
-
-    setTabIds(
-      getTabs().map((tab) =>
-        tab.props['aria-labelledby']
-          ? tab.props['aria-labelledby']
-          : getUID(tab),
-      ),
-    );
-  }, [getTabs, getUID]);
 
   /**
    * Handles if scroll fade indicators should be displayed.
@@ -212,7 +209,7 @@ export const Tabs = ({
     scrollableRight && styles['tabs--scrollable-right'],
   );
 
-  const childrenWithProps = React.Children.map(getTabs(), (child, i) => {
+  const childrenWithProps = React.Children.map(tabs, (child, i) => {
     // Checking isValidElement is the safe way and avoids a typescript
     // error too.
     if (React.isValidElement(child)) {
@@ -232,7 +229,7 @@ export const Tabs = ({
         ref={headerRef}
       >
         <ul className={styles['tabs__list']} role="tablist">
-          {getTabs().map((tab, i) => {
+          {tabs.map((tab, i) => {
             const isActive = activeIndexState === i;
             return (
               <li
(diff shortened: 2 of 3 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -44,11 +45,6 @@ export interface Props {
    * Passed down to initially set the activeIndex state
    */
   activeIndex?: number;
-  /**
-   * The array of items to be passed into the component. The array must take on the specified shape
-   * TODO: improve `any` type
-   */
-  items?: Array<any>;
```
ahuth, line 51: Not used. Looks like copy-pasta.

src/components/Tabs/Tabs.tsx
```diff
@@ -65,66 +61,37 @@ export const Tabs = ({
...
+  const activeTabPanelId = useUID();
   const headerRef = useRef<HTMLDivElement>(null);
   const [activeIndexState, setActiveIndexState] = useState(activeIndex);
   const [scrollableLeft, setScrollableLeft] = useState<boolean>(false);
   const [scrollableRight, setScrollableRight] = useState<boolean>(false);
 
-  /** Get all children that are Tab components. Ignore any others. */
-  const getTabs = useCallback(() => {
+  /** Children that are actually tabs. Any others are ignored. */
+  const tabs = useMemo(() => {
```
ahuth, line 72: Hat tip to @jeremiah-clothier for the idea.

By using `useMemo`, we only iterate `children` looking for `Tab`s once per render, instead of like 7 times per render.

src/components/Tabs/Tabs.tsx
```diff
@@ -65,66 +61,37 @@ export const Tabs = ({
...
   }, [children]);
 
-  const tabRefs = getTabs().map(() => React.createRef<HTMLAnchorElement>());
-  const [tabPanelIds, setTabPanelIds] = useState<string[]>([]);
-  const [tabIds, setTabIds] = useState<string[]>([]);
-  const getUID = useUIDSeed();
+  const tabRefs = useMemo(
+    () => tabs.map(() => React.createRef<HTMLAnchorElement>()),
+    [tabs],
+  );
```
ahuth, line 79: We only need to re-generate refs when tabs themselves are regenerated, instead of on every render.

src/components/Tabs/Tabs.tsx
```diff
@@ -65,66 +61,37 @@ export const Tabs = ({
...
-  function usePrevious(activeIndex: number) {
-    useEffect(() => {
-      ref.current = activeIndex;
-    });
-    return ref.current;
-  }
+  const tabIds = useMemo(
+    () => tabs.map((tab) => tab.props['aria-labelledby'] || getUID(tab)),
+    [tabs, getUID],
+  );
```
ahuth, line 84: We don't need to store the generate ids in state.

In fact, doing so was kind of weird, if you think about it.

We never want to re-render based on the ids changing... That's actually backwards, we re-generate ids based on other things changing.

src/components/Tabs/Tabs.tsx
```diff
@@ -65,66 +61,37 @@ export const Tabs = ({
...
-   * If prevActiveIndex is defined and previous value prop is not equal
-   * to current value prop, toggle state.
-   */
+  // Set the active tab if the `activeIndex` prop changes.
   const prevActiveIndex = usePrevious(activeIndex);
   useEffect(() => {
-    if (
-      (prevActiveIndex !== undefined || null) &&
-      prevActiveIndex !== activeIndex
-    ) {
```
ahuth, line 107: [Mentioned in Slack](https://czi-edu.slack.com/archives/C01BVP94BRR/p1664287719631189). This logic is goofy. Almost certainly not what the author intended.

src/components/Tabs/Tabs.tsx
```diff
@@ -217,16 +184,9 @@ export const Tabs = ({
...
-  const childrenWithProps = React.Children.map(getTabs(), (child, i) => {
-    // Checking isValidElement is the safe way and avoids a typescript
-    // error too.
-    if (React.isValidElement(child)) {
-      return React.cloneElement<Props>(child, {
-        id: tabPanelIds[i],
-        'aria-labelledby': tabIds[i],
-      });
-    }
-    return child;
```
ahuth, line 229: No need to clone every child, since we're only rendering one of them (the active tab panel).

src/components/Tabs/Tabs.tsx
```diff
@@ -237,24 +197,23 @@ export const Tabs = ({
...
-          {getTabs().map((tab, i) => {
+          {tabs.map((tab, i) => {
             const isActive = activeIndexState === i;
             return (
               <li
                 className={clsx(
                   styles['tabs__item'],
                   isActive && styles['eds-is-active'],
                 )}
-                key={'tabs-item-' + i}
```
ahuth, line 248: Best practice is to use a unique identifier for `key`, not the loop index.

https://reactjs.org/docs/lists-and-keys.html

src/components/Tabs/Tabs.tsx
```diff
@@ -237,24 +197,23 @@ export const Tabs = ({
...
+                key={tabIds[i]}
                 role="presentation"
               >
                 <a
-                  aria-controls={tabPanelIds[i]}
```
ahuth, line 252: Note that every anchor tag now `aria-controls` the same id. The active tab panel's id.

I think it was weird how this previously pointed to ids that weren't rendered (since only one panel is actually rendered).

src/components/Tabs/Tabs.tsx
```diff
@@ -237,24 +197,23 @@ export const Tabs = ({
...
+                  aria-controls={activeTabPanelId}
                   aria-selected={isActive}
                   className={styles['tabs__link']}
-                  href={`#${tabPanelIds[i]}`}
+                  href={`#${activeTabPanelId}`}
                   id={tabIds[i]}
-                  key={'tab-' + i}
```
ahuth, line 257: Pointless key.

src/components/Tabs/Tabs.tsx
```diff
@@ -65,66 +61,37 @@ export const Tabs = ({
...
   useEffect(() => {
-    if (
-      (prevActiveIndex !== undefined || null) &&
-      prevActiveIndex !== activeIndex
-    ) {
+    if (prevActiveIndex != null && prevActiveIndex !== activeIndex) {
       setActiveIndexState(activeIndex);
       tabRefs[activeIndex].current?.focus();
     }
   }, [prevActiveIndex, activeIndex, tabRefs]);
```
ahuth, line 93: I'm tempted to delete `usePrevious` and just do something like

```ts
useEffect(() => {
  if (activeIndex != null) {
    setActiveIndexState(activeIndex);
    tabRefs[activeIndex].current?.focus();
  }
}, [activeIndex, tabRefs]);
```

which would do the correct thing when the `activeIndex` prop changes.

However, I think that would get messed up if tabs are dynamically added or removed (by activating the original `activeIndex` tab, instead of the currently active one).

src/components/Tabs/Tabs.tsx
```diff
@@ -65,66 +61,37 @@ export const Tabs = ({
   onChange,
   ...other
 }: Props) => {
-  const ref = useRef<number | undefined>();
+  const getUID = useUIDSeed();
+  const activeTabPanelId = useUID();
```
ahuth, line 65: Only maintaining a single id for the active tab panel, instead of ids for every tab panel.

Since we only actually render one tab panel at a time (the active one), I think this makes sense. Better off not referencing ids that aren't actually in the DOM.

src/components/Tabs/Tabs.tsx
```diff
@@ -237,24 +197,23 @@ export const Tabs = ({
...
+                key={tabIds[i]}
                 role="presentation"
               >
                 <a
-                  aria-controls={tabPanelIds[i]}
+                  aria-controls={activeTabPanelId}
                   aria-selected={isActive}
                   className={styles['tabs__link']}
-                  href={`#${tabPanelIds[i]}`}
+                  href={`#${activeTabPanelId}`}
```
ahuth, line 215: All tabs now have their `href` pointing to the lone active tab panel id.

Maybe that's weird, maybe it's not. Either way, a future PR is going to change these to button elements, since they're not actually links.

src/components/Tabs/Tabs.tsx
ahuth: I think this is pretty similar to the `useRef([])` solution [discussed at our team meeting](https://docs.google.com/document/d/1xnmVlJALAWiSNZdgr3aWWiki0GY2rEaDu_UDEKktocc/edit#), except better/easier/simpler.

src/components/Tabs/Tabs.test.tsx
```diff
@@ -32,4 +33,12 @@ describe('<Tabs />', () => {
...
     );
   });
+
+  it('changes the active tab when activeIndex changes', () => {
+    const { rerender } = render(<Default activeIndex={0} />);
+    expect(screen.getByRole('heading', { name: 'Tab 1' })).toBeInTheDocument();
+
+    rerender(<Default activeIndex={1} />);
+    expect(screen.getByRole('heading', { name: 'Tab 2' })).toBeInTheDocument();
+  });
```
dierat, line 43: Nice

src/components/Tabs/Tabs.tsx
dierat: In examples that I've seen of accessible tab panels, the `aria-controls` id on a tab button points to the panel that would appear if that button is pressed and the tab becomes active. https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-manual.html

dierat (approved):
Makes sense to me 👍

src/components/Tabs/Tabs.tsx
ahuth: It is kind of weird though 🤔

ahuth:
Thanks!

PR #1323, opened 2022-09-29
More tabs cleanup

ahuth:
### Summary:

Yet more Tabs cleanup:
- Fix tab ids (more info inline)
- Change where aria-labelledby is applied
- ~~Remove scroll styling~~ (nvm, I didn't realize this was intentional. Thought it was part of BM's template)
- Minor type cleanup

### Test Plan:

- Tests
- Storybook

### DIFF
--- a/src/components/Tabs/Tabs.module.css
+++ b/src/components/Tabs/Tabs.module.css
@@ -21,6 +21,33 @@
 }
 
 /**
+ * Fade scrollable indicators to display if there is scrollable area on either side.
+ *
+ * The color "white" is arbitrary and any non transparent color can be used here.
+ */
+.tabs--scrollable-left {
+  -webkit-mask-image: -webkit-linear-gradient(
+    left,
+    transparent,
+    white var(--eds-size-8)
+  );
+}
+
+.tabs--scrollable-right {
+  -webkit-mask-image: -webkit-linear-gradient(right, transparent);
+}
+
+.tabs--scrollable-left.tabs--scrollable-right {
+  -webkit-mask-image: -webkit-linear-gradient(
+    left,
+    transparent,
+    white var(--eds-size-8),
+    white calc(100% - var(--eds-size-8)),
+    transparent 100%
+  );
+}
+
+/**
  * Tabs list
  *
  * Actual unordered list of tabs.
--- a/src/components/Tabs/Tabs.module.css
+++ b/src/components/Tabs/Tabs.module.css
@@ -21,33 +21,6 @@
 }
 
 /**
- * Fade scrollable indicators to display if there is scrollable area on either side.
- *
- * The color "white" is arbitrary and any non transparent color can be used here.
- */
-.tabs--scrollable-left {
-  -webkit-mask-image: -webkit-linear-gradient(
-    left,
-    transparent,
-    white var(--eds-size-8)
-  );
-}
-
-.tabs--scrollable-right {
-  -webkit-mask-image: -webkit-linear-gradient(right, transparent);
-}
-
-.tabs--scrollable-left.tabs--scrollable-right {
-  -webkit-mask-image: -webkit-linear-gradient(
-    left,
-    transparent,
-    white var(--eds-size-8),
-    white calc(100% - var(--eds-size-8)),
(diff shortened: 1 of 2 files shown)


src/components/Tabs/Tabs.module.css
```diff
@@ -20,33 +20,6 @@
...
-
-.tabs--scrollable-left.tabs--scrollable-right {
-  -webkit-mask-image: -webkit-linear-gradient(
-    left,
-    transparent,
-    white var(--eds-size-8),
-    white calc(100% - var(--eds-size-8)),
-    transparent 100%
-  );
-}
```
ahuth, line 48: Removed this scroll styling, because

1. It uses non-standard CSS
2. Seems like it has low value

Here's a pic of the styling (that's been deleted):

Screen Shot 2022-09-29 at 1 40 47 PM

It's the sorta-transparency on the left.

Removing this cleans up a lot of code from Tabs.

src/components/Tabs/Tabs.tsx
```diff
@@ -79,7 +75,7 @@ export const Tabs = ({
   );
 
   const tabIds = useMemo(
-    () => tabs.map((tab) => tab.props['aria-labelledby'] || getUID(tab)),
+    () => tabs.map((tab) => tab.props.id || getUID(tab)),
```
ahuth, line 78: Finally realized that it makes no sense to use `aria-labelledby` here as the id for a tab...

Think 🤔 about it like this. If we have

```jsx
<h2 id="heading">Stuff!</h2>

<Tabs>
  <Tab aria-labelledby="heading" title="Thing 1" />
</Tabs>
```

we'll end up with

```html
<h2 id="heading">Stuff!</h2>

<ul>
  <li id="heading">Thing 1</li>
</ul>
```

The ids are duplicated, and the logic here is clearly wrong.

src/components/Tabs/Tabs.tsx
```diff
@@ -176,27 +122,19 @@ export const Tabs = ({
...
-      <div
-        className={headerClassName}
-        onScroll={(e) => handleTabsScroll(e.target as HTMLDivElement)}
-        ref={headerRef}
-      >
-        <ul className={styles['tabs__list']} role="tablist">
+    <div className={clsx(styles['tabs'], className)} {...other}>
+      <div className={styles['tabs__header']}>
+        <ul
+          aria-labelledby={ariaLabelledBy}
```
ahuth, line 134: Moved the `aria-labelledby` to the element with the tablist role, because that's how https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-manual.html does it. It also makes sense that its the tablist that would be labelled, not some random div.

I will note, however, that the WAI-ARIA tabs example is just that - an example. It's not a guideline or requirement. But I like following it when possible.

src/components/Tabs/Tabs.module.css
dierat: >2. Seems like it's low value

The purpose of this transparency fade is to communicate to the user that the tabs extend beyond the edge of the page. I think that's unexpected and many users would be confused by that. The fade was design's solution to communicating that non-standard UX to the user. I don't think we should remove it unless we have an alternative solution.

ahuth: Heh, I sort of assumed this came from BM's template. But if we designed it this way, I'll add it back.

I do still question the non-standard CSS (there's probably other ways of doing it). There's also the browser's scrollbar, which indicates that there are more things (and can be scrolled).

But anyway, will add it back. Bummer it takes JS to do 🤔

ahuth: Oh yeah, that's true. The scrollbars only show up when you're already scrolling.

I think I just don't like this solution because it seems like a lot of trouble to show a bit of transparency (will people even notice it?). But that doesn't mean it's wrong.

dierat: >will people even notice it?

The designer originally also had a right/left caret button on top of the transparent area, but honestly that just looked like a hassle to build so we decided to simplify a little and just do the transparency.

dierat (approved):

PR #1325, opened 2022-10-06
chore(tooltip): expose appendTo prop and add docstring

jinlee93:
### Summary:
Exposes the `appendTo` prop from Tippy with docstring
### Test Plan:
- No visual regressions
- Unit tests pass

### DIFF
--- a/src/components/Tooltip/Tooltip.tsx
+++ b/src/components/Tooltip/Tooltip.tsx
@@ -13,6 +13,12 @@ type TooltipProps = {
    */
   align?: 'top' | 'right' | 'bottom' | 'left';
   /**
+   * The element or ref to append the tooltip to.
+   * Defaults to the body element.
+   * 'parent' is suggested if used in a modal.
+   */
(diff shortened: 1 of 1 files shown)


src/components/Tooltip/Tooltip.tsx
```diff
@@ -12,6 +12,12 @@ type TooltipProps = {
    * but our CSS currently only supports the 4 main sides.
    */
   align?: 'top' | 'right' | 'bottom' | 'left';
+  /**
+   * The element or ref to append the tooltip to.
+   * Defaults to the body element.
+   * 'parent' is suggested if used in a modal.
+   */
+  appendTo?: 'parent' | Element | ((ref: Element) => Element);
```
ahuth, line 20: Does Tippy only take a ref callback, and not a ref object?

dierat: This is copied from the tippy repo. https://github.com/atomiks/tippyjs/blob/8f7b2df156ae39c4df23c481b77ab59fe7f08c44/src/types.ts#L67

PR #1326, opened 2022-10-06
chore(dropdownmenu): close with tab, revert to :focus, close on click

jinlee93:
### Summary:
[sc-216210]
- [slack context](https://czi-edu.slack.com/archives/CTFV79JH4/p1665004863841489)
Motivation: 
- After opening button dropdown via click, the focus lands on the first item, but previously due to `:focus-visible` this is not indicated
- This means that there needs to be separate hover and focus indicators for the dropdown menu items
- In safari, apparently clicking a `<button>` [doesn't focus it](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#clicking_and_focus) for some odd reason and hence keyboard functionality is lost after clicking button

Execution:
- go back to `:focus` from `:focus-visible` and separate from `:hover`
- close dropdown menu on click and on tab to make safari not focusing not a problem (sorry double negative)
  - Items are no longer focusable via tab
  - This follows same behavior as default HeadlessUI menu and Bootstrap menu
  - Items are still cyclable via arrow keys and home / end keys
  - Since items are anchor or button elements anyways, closing is fine 
### Test Plan:
- manual safari testing:

https://user-images.githubusercontent.com/86632227/194253918-ed6bcd61-20e9-4a90-b6db-edeb2ba5927f.mov

- same expected behavior on chrome, and unit tests exist for keyboard navigation
- no visual regression

### DIFF
--- a/src/components/ButtonDropdown/ButtonDropdown.tsx
+++ b/src/components/ButtonDropdown/ButtonDropdown.tsx
@@ -2,7 +2,7 @@ import clsx from 'clsx';
 import React, { ReactNode, SyntheticEvent } from 'react';
 import styles from './ButtonDropdown.module.css';
 import { DropdownMenu } from '../..';
-import { ESCAPE_KEYCODE } from '../../util/keycodes';
+import { ESCAPE_KEYCODE, TAB_KEYCODE } from '../../util/keycodes';
 import type { ClickableStyleProps } from '../ClickableStyle';
 
 interface FocusEvent<T = Element> extends SyntheticEvent<T> {
@@ -141,7 +141,7 @@ export const ButtonDropdown = ({
    * If the escape key is struck, close the panel.
    */
   function handleKeyDown(e: React.KeyboardEvent) {
-    if (e.key === ESCAPE_KEYCODE) {
+    if (e.key === ESCAPE_KEYCODE || e.key === TAB_KEYCODE) {
       closePanel();
     }
   }
@@ -172,14 +172,6 @@ export const ButtonDropdown = ({
     },
   );
 
-  const handleBlur = (event: FocusEvent<HTMLDivElement>) => {
-    if (!event.currentTarget.contains(event.relatedTarget as HTMLElement)) {
-      if (isActiveVar) {
-        closePanel();
-      }
-    }
-  };
-
   const componentClassName = clsx(
     styles['button-dropdown'],
     isActiveVar && styles['eds-is-active'],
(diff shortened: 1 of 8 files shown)


src/components/ButtonDropdown/ButtonDropdown.tsx
```diff
@@ -189,18 +176,13 @@ export const ButtonDropdown = ({
...
-      onBlur={handleBlur}
-      ref={ref}
-      {...other}
-    >
+    <div className={componentClassName} ref={ref} {...other}>
       {dropdownMenuTriggerWithProps}
       <DropdownMenu
         className={styles['button-dropdown__dropdown-menu']}
-        handleOnEscDown={(e: React.KeyboardEvent) => handleKeyDown(e)}
+        handleOnClick={closePanel}
```
jinlee93, line 183: Would have loved to have just used the existing event listeners in lines 106 but doing so prevents click events from being captured by the grandchildren `<DropdownMenuItem>` due to the native event handling occurring and dismounting the children before the React onClick could be called

jinlee93:
Migration to HeadlessUI Menu could have helped prevent this, but that comes with its own can of things to figure out

dierat:
>same expected behavior on chrome, and unit tests exist for keyboard navigation

Should we add a test for closing on tab key since that's new behavior?

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -134,6 +149,7 @@ export const DropdownMenu: React.FC<Props> = ({
       <div className={componentClassName} {...other}>
         <ul
           className={styles['dropdown-menu__list']}
+          onClick={handleOnClick}
```
dierat, line 152: Can you explain this part to me? I'm curious if it could possibly be overriding the click behavior on the items. I went to the course planner in storybook in the chromatic build for this PR, and the items in the dropdown list for moving the cards around aren't doing anything for me.

jinlee93: Since `buttonDropdown` controls the active/open state for `DropdownMenu` this calls the `handleOnClick` callback which closes sets it closed in `buttonDropdown`.

I don't think any of the buttons in the dropdown in the course planner edit page in EDS had attached functionality to begin with as they don't work in https://chanzuckerberg.github.io/edu-design-system/?path=/story/pages-course-planner-edit--default and lack of `onClick` being passed in https://github.com/chanzuckerberg/edu-design-system/blob/main/.storybook/pages/CoursePlannerEdit/CoursePlannerEdit.tsx#L288-L332

dierat: Awesome, thanks!

PR #1329, opened 2022-10-06
chore: remove upcoming-components and references from system

booc0mtaco:
### Summary:

This creates a PR which de-lists the candidate components and references in other files, to streamline the presentation of the repository contents. This also updates documentation to refer to this change, in case any of the removed components serve as a good basis for implementation.

### Test Plan:

- n/a (unused components were not under test ATM)


src/upcoming-components/Accordion/Accordion.tsx
```diff
@@ -1,35 +0,0 @@
...
- * ```
- *
- * TODO: update this comment with a description of the component.
- */
-export const Accordion = ({ children, className, ...other }: Props) => {
-  const componentClassName = clsx(styles['accordion'], className);
-  return (
-    <dl className={componentClassName} {...other}>
-      {children}
-    </dl>
```
jinlee93, line 31: I think this is a good reason to remove the accordion component.

This suggests that the accordion uses will be descriptive in nature when that may not be the case in use cases and can be applying the wrong semantic element. 

Even the [w3 accordion](https://www.w3.org/WAI/ARIA/apg/example-index/accordion/accordion) example shows form elements in the panel

booc0mtaco: Good callout. This is also a nice way to document concerns about any of these as we go. Anyone referring to this PR will see the discussion and can make informed choices.

src/upcoming-components/CheckboxField/CheckboxField.tsx
```diff
@@ -1,116 +0,0 @@
...
-/**
- * BETA: This component is still a work in progress and is subject to change.
- *
- * ```ts
- * import {CheckboxField} from "@chanzuckerberg/eds";
- * ```
- *
- * TODO: update this comment with a description of the component.
- */
-export const CheckboxField = ({
```
jinlee93, line 71: Fieldset component fulfills this purpose

src/upcoming-components/Textarea/Textarea.module.css
```diff
@@ -1,13 +0,0 @@
...
-
-/*------------------------------------*\
-    #TEXTAREA
-\*------------------------------------*/
-
-/**
- * Inherits base textarea styles 
- */
-.textarea {
-  @mixin inputStyles;
```
jinlee93, line 11: This is probably valuable part of the textarea component, where the `inputStyles` mixin is reusable and keeps in line with theming of the inputfield.

src/upcoming-components/TextareaField/TextareaField.tsx
```diff
@@ -1,215 +0,0 @@
...
-   * Value passed down from higher levels for initial state
-   */
-  value?: string;
-  /**
-   * Default value passed down from higher levels for initial state
-   */
-  defaultValue?: string;
-}
-
-export const TextareaField = ({
```
jinlee93, line 123: Think when it's time to build the `Textarea` component, the `TextareaField` may come naturally and could consider using our `Fieldset` component there. Don't think there's much value in these props that is not already come across

docs/COMPONENTS.md
```diff
@@ -37,6 +37,8 @@ Each EDS component is documented in Storybook, which surfaces each component's A
 
 ### Creating a component
 
+**NOTE**: Refer to this [Pull Request](https://github.com/chanzuckerberg/edu-design-system/pull/1329) for examples of potential upcoming components. If you see one you like, you can use it as a baseline for implementation.
```
jinlee93, line 40: Could also create a tag or cut a branch before this pr lands, so that it's more easily searchable in the github searchbox as well 
image

jinlee93 (approved):
thanks for the cleanup!

docs/COMPONENTS.md
booc0mtaco: Good idea. I can push a tag, and refer to that here

dierat (approved):
Thanks for making the repo cleaner and less confusing!

timzchang (approved):
Always down for streamlining the codebase :)

PR #1330, opened 2022-10-07
chore(dropdownmenu, textfield, inputfield): clean up types and onclick, forward ref 

jinlee93:
### Summary:
Due to `projectCard` using `buttonDropdown`/`dropdownMenu` and not `popover` some issues arise as it tries to use secondary functionality when dropdown menus should purely be primary functionality (aka anchor and button elements only)
image
This panel with the `TextField` and `ButtonGroup` is rendered when a menu item is clicked, so it initially made sense to use a dropdown menu.

[While projectCard is being worked on to use popover instead](https://app.shortcut.com/czi-edu/story/224122/use-popover-and-not-dropdown-in-project-card), we need to support this secondary functionality in a dropdown menu.

Hence changes in this pr: 
- no longer closes dropdownmenu on close
- forwards ref to the input element in `TextField` and `InputField`
- clean up work with types

### Test Plan:
- alpha build created with the changes from this pr, and is being used in https://github.com/FB-PLP/traject/pull/12436
- no visual regressions in both eds or in LP using alpha build
- in both eds and LP using alpha build:
  - project card dropdown should cycle via arrow, home, and end keys when in menu mode
  - project card dropdown should close with escape or tab when in menu mode
- in LP using alpha build [course plan edit story](https://614b969675171d003ac6f98a-tlejigwjjt.chromatic.com/?path=/story/teachers-courses-courseplan-courseplanwizard--course-plan-edit), arrive at the previous screenshot dropdown by clicking the three dots on a card and click "Adjust instructional blocks":
image
  - in secondary func dropdown:
    - tab should cycle through tabbable elements
    - if tab leaves tabbable elements (shift+tab on inputfield or tab on last button), dropdown should close
    - escape closes dropdown

### DIFF
--- a/src/components/ButtonDropdown/ButtonDropdown.tsx
+++ b/src/components/ButtonDropdown/ButtonDropdown.tsx
@@ -180,7 +180,6 @@ export const ButtonDropdown = ({
       {dropdownMenuTriggerWithProps}
       <DropdownMenu
         className={styles['button-dropdown__dropdown-menu']}
-        handleOnClick={closePanel}
         handleOnEscDown={(e) => handleKeyDown(e)}
         handleOnTabDown={(e) => handleKeyDown(e)}
         isActive={isActiveVar}
--- a/src/components/ButtonDropdown/ButtonDropdown.tsx
+++ b/src/components/ButtonDropdown/ButtonDropdown.tsx
@@ -2,15 +2,13 @@ import clsx from 'clsx';
 import React, { ReactNode } from 'react';
 import styles from './ButtonDropdown.module.css';
 import { DropdownMenu } from '../..';
-import { ESCAPE_KEYCODE, TAB_KEYCODE } from '../../util/keycodes';
 import type { ClickableStyleProps } from '../ClickableStyle';
 
 export interface Props {
-  buttonAriaLabel?: string;
   /**
-   * Makes button full width
+   * Aria label to be attacehd to the dropdown trigger button.
    */
-  fullWidth?: boolean;
+  buttonAriaLabel?: string;
   /**
    * Adds status to the button (e.g. error, success)
    */
@@ -24,9 +22,27 @@ export interface Props {
    */
   buttonVariant?: ClickableStyleProps<'button'>['variant'];
   /**
+   * Prop used to pass in `DropdownMenuItem` child components
+   */
+  children?: ReactNode;
+  /**
+   * CSS class names that can be appended to the component.
+   */
+  className?: string;
+  /**
    * Disables the field and prevents editing the contents
    */
   disabled?: boolean;
+  /**
+   * Prop used to pass in the dropdown menu trigger (dropdownTrigger={<Button />}). This
+   * allows for maximum flexbility with extending the button and passing in props from the
+   * outside
+   */
+  dropdownMenuTrigger?: ReactNode;
+  /**
+   * Makes button full width
+   */
+  fullWidth?: boolean;
   /**
    * Determines type of clickable
    * - default renders a dropdown menu to the bottom left of the button
@@ -42,20 +58,6 @@ export interface Props {
    * - **reset** The clickable is a reset clickable (resets the form-data to its initial values)
    */
   type?: 'button' | 'reset' | 'submit';
-  /**
-   * Prop used to pass in the dropdown menu trigger (dropdownTrigger={<Button />}). This
-   * allows for maximum flexbility with extending the button and passing in props from the
-   * outside
-   */
-  dropdownMenuTrigger?: ReactNode;
-  /**
-   * Prop used to pass in `DropdownMenuItem` child components
-   */
-  children?: ReactNode;
-  /**
-   * CSS class names that can be appended to the component.
-   */
-  className?: string;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/components/ButtonDropdown/ButtonDropdown.tsx
```diff
@@ -2,15 +2,13 @@ import clsx from 'clsx';
 import React, { ReactNode } from 'react';
 import styles from './ButtonDropdown.module.css';
 import { DropdownMenu } from '../..';
-import { ESCAPE_KEYCODE, TAB_KEYCODE } from '../../util/keycodes';
```
jinlee93, line 5: Moved all event key functionality to `<DropdownMenu>`

src/components/ButtonDropdown/ButtonDropdown.tsx
```diff
@@ -42,20 +58,6 @@ export interface Props {
...
-   */
-  dropdownMenuTrigger?: ReactNode;
-  /**
-   * Prop used to pass in `DropdownMenuItem` child components
-   */
-  children?: ReactNode;
-  /**
-   * CSS class names that can be appended to the component.
-   */
-  className?: string;
```
jinlee93, line 58: Just alphabetizing the props

src/components/ButtonDropdown/ButtonDropdown.tsx
```diff
@@ -180,9 +171,7 @@ export const ButtonDropdown = ({
       {dropdownMenuTriggerWithProps}
       <DropdownMenu
         className={styles['button-dropdown__dropdown-menu']}
-        handleOnClick={closePanel}
-        handleOnEscDown={(e) => handleKeyDown(e)}
-        handleOnTabDown={(e) => handleKeyDown(e)}
+        closeDropdownMenu={closePanel}
```
jinlee93, line 174: No longer closes on click, (should be brought back once project card uses popover).

src/components/ButtonDropdown/ButtonDropdown.tsx
```diff
@@ -180,9 +171,7 @@ export const ButtonDropdown = ({
       {dropdownMenuTriggerWithProps}
       <DropdownMenu
         className={styles['button-dropdown__dropdown-menu']}
-        handleOnClick={closePanel}
-        handleOnEscDown={(e) => handleKeyDown(e)}
-        handleOnTabDown={(e) => handleKeyDown(e)}
```
jinlee93, line 185: Since not handling key presses here, just have to send close dropdown menu callback

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -143,6 +135,13 @@ export const DropdownMenu: React.FC<Props> = ({
     }
   };
 
+  const handleOnClick: MouseEventHandler = (e) => {
+    focusIndex = refs.current.list.findIndex((ref) =>
+      ref.contains(e.target as HTMLElement),
+    );
```
jinlee93, line 141: Finds which item had been clicked, and sets `focusIndex` accordingly so up and down arrow keys select the next or previous item

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -143,6 +135,13 @@ export const DropdownMenu: React.FC<Props> = ({
...
+    focusItem(focusIndex);
```
jinlee93, line 142: Required since [Safari doesn't focus `<button>` elements on click](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#clicking_and_focus)

src/components/InputField/InputField.tsx
```diff
@@ -1,5 +1,5 @@
 import clsx from 'clsx';
-import React, { ChangeEventHandler } from 'react';
+import React, { ChangeEventHandler, forwardRef } from 'react';
```
jinlee93, line 2: This file change is just forwarding the ref

src/components/TextField/TextField.tsx
```diff
@@ -1,13 +1,13 @@
...
+import React, { ChangeEventHandler, forwardRef, ReactNode } from 'react';
 import { useUID } from 'react-uid';
 import styles from './TextField.module.css';
 import FieldNote from '../FieldNote';
 import InputField from '../InputField';
 import Label from '../Label';
 import Text from '../Text';
 
-export interface Props {
+export type Props = React.InputHTMLAttributes<HTMLInputElement> & {
```
jinlee93, line 10: 1/2 changes besides forwarding the ref in this file is extending the HTMLInputProps to attach to the input element. Was already being done via `{...other}` but this would show what's available

src/components/TextField/TextField.tsx
```diff
@@ -56,6 +56,10 @@ export interface Props {
    * HTML label text
    */
   label?: string;
+  /**
+   * Maximum value allowed for the input, if type is 'number'. When the input value matches this maximum, the plus button becomes disabled.
+   */
+  max?: number;
```
jinlee93, line 62: #2/2 change besides forwarding the ref in this file is matching the `max` type in `InputField` to prevent TS error

src/components/DropdownMenu/DropdownMenu.tsx
dierat: Could you move this and the comment above into the code?

dierat (approved):
Thanks for fixing this; it looks really complicated lol

jinlee93:
> Thanks for fixing this; it looks really complicated lol

@dierat I feel like this is a result of fixing Safari specific behavior regarding traditional `onBlur` and focus. AND once the project card is switched over to `Popover` not sure where this component would be used besides brumbs. At which point could get to looking at using Menu from Headless for this kind of purpose maybe it will help clean up some code and be more accessible

PR #1333, opened 2022-10-12
chore(table): remove TableObject

jinlee93:
[sc-217415]
### Summary:
- removes TableObject component
  - only provided a `overflow` variant with some masking to visually indicate horizontal overflow but I'm deciding to leave that behavior up to parent
    - wasn't being used anywhere
  - confusing why there is a Table and a TableObject component
- sort state is externally controlled now
- moved sorting story to Table.stories and made it interactive
- possibly incoming: show memoization of sorted table to make less expensive
### Test Plan:
- TableObject stories are gone
- No visual regression in Feedback Overview page
  - this used to use TableObject but not the `overflow` variant
- No unrelated visual regression either


src/components/Table/Table.stories.tsx
```diff
@@ -152,6 +153,12 @@ export const Stacked: StoryObj<Args> = {
       </>
     ),
   },
+  parameters: {
+    viewport: {
+      defaultViewport: 'mobile2',
+    },
+    chromatic: { viewports: [414] },
+  },
```
jinlee93, line 161: the point of this story is to show the mobile cards in smaller breakpoints

src/components/Table/Table.stories.tsx
```diff
@@ -243,3 +250,85 @@ export const AlignTableCellContentRight: StoryObj<Args> = {
...
+  const values = [
+    { col1: 'Value 1', col2: 'Value A' },
+    { col1: 'Value 3', col2: 'Value B' },
+    { col1: 'Value 2', col2: 'Value C' },
+    { col1: 'Value 4', col2: 'Value D' },
+  ];
+  const [sortDirection, setSortDirection] =
+    useState<SortDirectionsType>('default');
+  const [tableValues, setTableValues] = useState({ ...values });
+  const onSortClick = () => {
```
dierat, line 264: Just curious—what was the reasoning for pulling the sort logic out of the component?

src/components/TableObjectBody/TableObjectBody.tsx
```diff
@@ -1,148 +0,0 @@
...
-  /**
-   * Target table body and table body inner DOM elements.
-   */
-  const tableObjectBodyRef = useRef<HTMLDivElement>(null);
-  const tableObjectBodyInnerRef = useRef<HTMLDivElement>(null);
-
-  /**
-   * Set right and left gradients on tables
-   */
-  const setShadows = () => {
```
dierat, line 55: Are we thinking that tables won't use this behavior or they'll implement it themselves every time? If it's the latter, should we move this logic into a story so devs can see how it's done?

src/components/Table/Table.stories.tsx
jinlee93: The `Table` component never had sort logic to begin with, unless you're asking to add sorting logic to the `Table` component, in which case there would need to be context being thrown around and there is less value to be gained if the parent component can simply sort the data to begin with

src/components/TableObjectBody/TableObjectBody.tsx
jinlee93: This adds masking like in `Tabs` to indicate overflow.

However I don't think we should own overflow behavior and the parent can add as wished either via `overflow: scroll` to always show scroll bar or some masking.

Is there a way to extract the masking from Tabs and export it for use? Kind of like our mixins but also for available outside EDS

src/components/Table/Table.stories.tsx
dierat: I guess I'm thinking of the `toggleSort` logic that was pulled out of `TableHeaderCell`, but it's strange how that logic updated the icon without hooking into anything else.

jinlee93: Yea that was used for which icon to display only and had no actual sorting functionality hooking into rows or cells

src/components/TableObjectBody/TableObjectBody.tsx
dierat: > Is there a way to extract the masking from Tabs and export it for use? Kind of like our mixins but also for available outside EDS

I haven't thought about it, but it would be nice to make that reusable if it's a pattern we want to keep using. I remember seeing it on a table design that was brought to office hours once, where they had one column fixed in view and the rest of the table scrolled, and I think they were using some sort of fade to indicate a section of the row was scrolled out of view.

dierat (approved):

src/components/TableObjectBody/TableObjectBody.tsx
jinlee93: If you don't mind, I'm going to punt adding the overflow story and the masking into another ticket

dierat: Sounds good 👍

jinlee93: As an update to this, I had a discussion with @ahuth and agree that we should keep the table component dumber and not display overflow behavior, as:
- discourage overpopulating tables
- discourage use in Modals
- would require a div wrapper with use case specific functionality
- should leave it up to consumers to handle their own use case (and hence discouraging cases where tables need overflow and is not resolved by whole page scrolling)

PR #1334, opened 2022-10-12
Prepare for shared lint configs

ahuth:
[sc-224372]

### Summary:

Prepare for using our shared lint configs by:

- Bringing in prettier-config-js
- Using the default import ordering

Also for some reason 😬:

- Build the tokens on CI before linting... otherwise linting fails (since tailwind.config.js imports them). No idea why this doesn't happen on main. Investigating.

This unfortunately touches almost every file. Biggest changes are:

- No spacing within brackets - `{ foo }` -> `{foo}`
- Tailwind class ordering
- Different import ordering

### Test Plan:

- none


.eslintrc
```diff
@@ -52,8 +52,7 @@
       {
         "alphabetize": {
           "order": "asc"
-        },
-        "groups": ["builtin", "external", "sibling", "parent", "index"]
+        }
```
ahuth, line 55: We're now getting the default groups of `["builtin", "external", "parent", "sibling", "index"]`.

Basically the same, except parent comes before sibling.

I'd like to actually do away with "groups" and just alphabetically sort the import paths. But at this point it makes sense to go with what we landed on for Traject and the shared configs.

.github/workflows/test.yml
```diff
@@ -84,6 +84,11 @@ jobs:
       - name: Install Dependencies ⬆️
         run: yarn install --immutable
 
+      # Build the design tokens. Our Tailwind config imports these, and linting will fail if not
+      # present.
+      - name: Build design tokens (needed for linting to succeed)
+        run: yarn build:tokens
```
ahuth, line 90: Investigating what the heck is going on here.

Linting failed on CI because:
- tailwind.config.js imports tokens
- If the tokens aren't present, linting fails

Adding this here ensures the tokens are present.

No idea why this wasn't an issue until now, though... 😬🤔

.prettierignore
```diff
@@ -1,6 +1,7 @@
 storybook-static
 node_modules
 lib
+src/tokens-dist
```
ahuth, line 4: Added src/tokens-dist because otherwise prettier tries to fix some stuff up, and the linting steps fail on CI.

Figured it makes sense to just go with the formatting style-dictionary gives us for these.

ahuth:
Man, what is going on with this PR? CI is green now, but now Chromatic shows some changes...

ahuth:
I'm going to try this whole enterprise from scratch.

PR #1336, opened 2022-10-12
chore(tabs): add size for tabs right mask

jinlee93:
### Summary:
- noticed right mask for tabs was not being styled correctly, adds size to show mask
image

### Test Plan:
- right mask exists in deployed storybook

### DIFF
--- a/src/components/Tabs/Tabs.module.css
+++ b/src/components/Tabs/Tabs.module.css
@@ -34,7 +34,11 @@
 }
 
 .tabs--scrollable-right {
-  -webkit-mask-image: -webkit-linear-gradient(right, transparent);
+  -webkit-mask-image: -webkit-linear-gradient(
+    right,
+    transparent,
+    white var(--eds-size-8)
+  );
 }
 
 .tabs--scrollable-left.tabs--scrollable-right {
--- a/src/components/Tabs/Tabs.stories.tsx
+++ b/src/components/Tabs/Tabs.stories.tsx
@@ -1,4 +1,5 @@
 import { StoryObj, Meta } from '@storybook/react';
+import { within } from '@storybook/testing-library';
 import React from 'react';
 
 import { Tabs } from './Tabs';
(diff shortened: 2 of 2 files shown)


dierat:
Do we have a chromatic test covering this case?

src/components/Tabs/Tabs.module.css
```diff
@@ -34,7 +34,11 @@
 }
 
 .tabs--scrollable-right {
-  -webkit-mask-image: -webkit-linear-gradient(right, transparent);
+  -webkit-mask-image: -webkit-linear-gradient(
+    right,
+    transparent,
+    white var(--eds-size-8)
```
dierat, line 40: Could you help me explain what this changes exactly? I can't tell the difference in Chrome.

jinlee93:
> Do we have a chromatic test that catches the scrollable styling?

We do not, I can add one

src/components/Tabs/Tabs.module.css
jinlee93: Re introduces the right masking when there is no more left content but there is more right content (also in chrome):

Currently:
Screen Shot 2022-10-12 at 12 52 49 PM

This change:
Screen Shot 2022-10-12 at 12 52 47 PM

dierat: Awesome, thanks!

dierat (approved):

jinlee93:
@dierat Added some stories for testing the masking, think there is value in keeping these?

dierat:
> @dierat Added some stories for testing the masking, think there is value in keeping these as they don't work when going from the default story to the scroll stories since the play function runs after the viewport change?

But it works in chromatic? I'm fine with keeping them if they work in chromatic 👍

jinlee93:
> But it works in chromatic? I'm fine with keeping them if they work in chromatic 👍

Sounds good to me

PR #1337, opened 2022-10-12
fix(icons): remove mask from play icon

booc0mtaco:
### Summary:

[sc-201336] This icon included a bounding rectangle in figma, which was causing a strange appearance in Firefox, and was inconsistent with the other icons. Import a new icon that does not have this mask, but preserves the same dimensions. Tested in Firefox and other Webkit-based browsers.

### Test Plan:

- test icon layout in chrome / firefox : http://localhost:9000/?path=/story/atoms-icons-icongrid--default
- ensure icon is not clipped


dierat (approved):
pics or it didn't happen

ahuth (approved):
LGTM 🤘 

Thanks for taking care of Firefox (which I and a few of our users use).

anniehu4:
We could consider turning on Firefox tests for certain stories in [Chromatic](https://www.chromatic.com/manage?appId=61313967cde49b003ae2a860) 🤔

booc0mtaco:
@dierat ok, fine lol

Screen Shot 2022-10-12 at 12 28 21 PM

PR #1338, opened 2022-10-12
chore(build): run prettier on tokens dist after building them

jinlee93:
### Summary:
- runs prettier on built tokens after building them
- resolves the double quote being spat out from style dictionary causing lint errors
### Test Plan:
- no lint errors after running `yarn build:tokens`
- tokens build unaffected

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "scripts": {
     "build": "yarn build:clean && yarn build:tokens && yarn build:js && yarn build:styles && yarn copy-to-lib",
     "build:clean": "rm -rf lib/",
-    "build:tokens": "rm -rf src/tokens-dist/ && node ./style-dictionary.config.js && yarn copy-tokens-to-lib",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
   "scripts": {
     "build": "yarn build:clean && yarn build:tokens && yarn build:js && yarn build:styles && yarn copy-to-lib",
     "build:clean": "rm -rf lib/",
-    "build:tokens": "rm -rf src/tokens-dist/ && node ./style-dictionary.config.js && yarn copy-tokens-to-lib",
+    "build:tokens": "rm -rf src/tokens-dist/ && node ./style-dictionary.config.js && yarn copy-tokens-to-lib && yarn prettier-tokens-dist",
```
ahuth, line 26: Nit (really doesn't matter) -

Could run prettier before copying to lib, so that lib and src match.

But it's just formatting differences. Obviously nothing in the logic/values.

ahuth (approved):
Nice 🤘

package.json
jinlee93: Makes sense, will move to before copying lib

dierat (approved):

PR #1340, opened 2022-10-12
Use JS shared lint config

ahuth:
### Summary:

[sc-224372]

Use JS ~~and Prettier~~ shared lint ~~configs~~ config.

~~One major formatting change resulting from this is changing to no spaces between brackets~~

(removed the shared Prettier config, because it breaks the build. More info in another comment)

### Test Plan:

CI


dierat:
Oops your tests are failing with `src/tokens-dist/ts/colors.ts: Error: Cannot find module './lib/tokens/json/css-variables-nested.json'`

.eslintrc
```diff
@@ -5,27 +5,20 @@
...
     "testing-library",
     "@chanzuckerberg/stories"
   ],
   "extends": [
-    "eslint:recommended",
+    "@chanzuckerberg/eslint-config-edu-js",
     "@chanzuckerberg/eslint-config-edu-ts",
-    "plugin:jsx-a11y/recommended",
     "plugin:prettier/recommended",
-    "plugin:react/recommended",
```
ahuth, line 17: Not related to this line.

I originally had the shared prettier config in this PR, but it breaks the build 😑 

- It brings in `prettier-plugin-tailwindcss`
- which seems to cause Prettier to lint `tailwind.config.js` during the styles build
- which breaks because `tailwind.config.js` imports files that don't exist before the build completes
- in fact, at that point, the build can never complete, because it breaks because it's not complete :inception:

Removed from this PR, and we can look more into it later.

dierat (approved):

ahuth:
Thanks

PR #1341, opened 2022-10-13
Auto accept chromatic

ahuth:
### Summary:

Auto accept chromatic changes on the `next` and `main` branches.

I noticed there were changes on `next`. Definitely indicates some flakiness, but I think it's best to assume that what's on `main`/`next` is correct (or at least that's what we do in LP and Along... but what do YOU think?).

### Test Plan:

🤞

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,6 +64,7 @@ jobs:
             --storybook-build-dir storybook-static \
             --exit-zero-on-changes \
             --exit-once-uploaded \
+            --auto-accept-changes '{main,next}' \
             --skip 'dependabot/**'
       - name: Run Accessibility Tests 💛
         run: yarn run storybook:axeOnly


.github/workflows/test.yml
```diff
@@ -64,6 +64,7 @@ jobs:
             --storybook-build-dir storybook-static \
             --exit-zero-on-changes \
             --exit-once-uploaded \
+            --auto-accept-changes '{main,next}' \
```
ahuth, line 67: Not totally super sure this is the right syntax for matching 2 branches 😬 

https://www.chromatic.com/docs/cli#chromatic-options sez we can use globbing (via picomatch), and the interwebz tells me this is how to glob for a list of items.

Figured we can merge this and I'll monitor builds to see what happens.

dierat (approved):
Makes sense to me. @jinlee93 any reservations?

booc0mtaco (approved):
I think this makes sense. The changes will still appear on individual PRs, but once merged any changes that appear in `next`/`main` will not bug us anymore?

ahuth:
> The changes will still appear on individual PRs, but once merged any changes that appear in next/main will not bug us anymore?

Sort of. If there's a flaky story, whatever happens on main/next will be the baseline. But if a PR does something different (due to flakiness or intended changes or whatever) they'll show up on the PR.

ahuth:
So we should def still address the flakiness.

This change mostly ensures:
- CI runs for next/main always finish
- CI runs for next/main are considered the "truth" (AKA baseline, I guess)

PR #1342, opened 2022-10-13
chore(table): remove unnecessary table sorted values state

jinlee93:
### Summary:
- removes the `tableValues` state from the Table Sorted Example story
- also maps the table values into TableRow to mimic real world use
### Test Plan:
- no visual regression
- unit tests pass


ahuth (approved):
LGTM 🤘

PR #1343, opened 2022-10-13
feat(ClickableStyle): use type check on status and variant

booc0mtaco:
### Summary:

[sc-216457] Check whether the status/variant combination is valid using typescript discriminated unions and some other type changes. Users of leaf components will get checks on the types automatically (violations will show with suggested values to use in code). Currently, users of `ClickableStyle` won't receive that benefit as extending the types to support the Union didn't work as expected.

#### For Discussion

- Should we preserve existing fallback JS tests and checks? 
- What use cases are there (ex: React-router `<Link/>`) for using this component directly?

### Test Plan:

- Fix existing TypeScript violations in FiltersButton component
- make sure existing fallback tests continue to pass


src/components/Button/Button.tsx
```diff
@@ -36,7 +36,7 @@ export type ButtonProps = ButtonHTMLElementProps & {
    * - **reset** The clickable is a reset clickable (resets the form-data to its initial values)
    */
   type?: 'button' | 'reset' | 'submit';
-};
+} & VariantStatus;
```
booc0mtaco, line 39: Pull in the `VariantStatus` which pairs the different variants to the available statuses

src/components/FiltersButton/FiltersButton.tsx
```diff
@@ -33,16 +33,22 @@ export const FiltersButton = forwardRef<HTMLButtonElement, FiltersButtonProps>(
...
+        };
 
     return (
       <Button
         className={componentClassName}
         ref={ref}
-        status={buttonStatus}
-        variant={buttonVariant}
         {...other}
+        {...variantStatus}
```
booc0mtaco, line 51: apply after `{...other}` to make sure any values in rest are replaced with those calculated in this component directly.

jeremiah-clothier (commented):
General question: as types get more specific / complicated like this. Is there value to testing they actually work. For example, maintain some code/test file and use `//@ts-expect-error` to ensure there is a valid TS error when props are provided incorrectly?

booc0mtaco:
> General question: as types get more specific / complicated like this. Is there value to testing they actually work. For example, maintain some code/test file and use `//@ts-expect-error` to ensure there is a valid TS error when props are provided incorrectly?

I think that can be a slippery slope. IMO it may be a code smell if the types become so complex that: one feels the need to test them, or they behave unpredictably. For this case, it might make sense, given the base component has some manual checks, but ideally these types would all live in ClickableStyle and we would not have to test things directly. my 2¢

src/components/ClickableStyle/ClickableStyle.tsx
```diff
@@ -111,17 +138,19 @@ const throwInvalidPropComboError = (variant: Variant, status: Status) => {
...
       variant = 'secondary',
       ...other
-    }: ClickableStyleProps<IComponent>,
-    ref: React.ForwardedRef<HTMLElement>,
-  ) => {
+    } = props;
+
     if (
       !getPropCombinationIsValid(variant, status) &&
       process.env.NODE_ENV !== 'production'
```
dierat, line 156: You can remove all the invalid prop console log stuff in here and the whole `ClickableStyle` test file now.

dierat:
This is great! The type checks are going to be so much more effective than console errors.

src/components/ClickableStyle/ClickableStyle.tsx
booc0mtaco: Agreed. Any things building directly upon `ClickableStyle` should use the exported union to lock the props to the valid combinations. Removed!

src/components/Button/Button.tsx
```diff
@@ -17,10 +17,10 @@ export type ButtonProps = ButtonHTMLElementProps & {
    * Toggles clickable that fills the full width of its container
    */
   fullWidth?: boolean;
-  status?: ClickableStyleProps<'button'>['status'];
+
```
dierat, line 20: Small thing—did you mean to leave empty lines here where the old props were?

dierat (approved):

src/components/Button/Button.tsx
booc0mtaco: Ah good catch 🎣 . I did not. Once we figured out this approach i scrambled to implement it. I suppose prettier won't clean up empty lines on its own in some cases

jeremiah-clothier (approved):

jeremiah-clothier:
> I think that can be a slippery slope. IMO it may be a code smell if the types become so complex that: one feels the need to test them, or they behave unpredictably. For this case, it might make sense, given the base component has some manual checks, but ideally these types would all live in ClickableStyle and we would not have to test things directly. my 2¢

Okay, I wasn't strongly opinionated on the matter, and I am fine without tests. I know I've seen tests for types before, and was just curious about your thoughts.

booc0mtaco:
@jeremiah-clothier Yah, i go both ways on this and have changed my mind in the past. In my mind it's similar to the debate around where form validation should happen. Good in the UI, but *mandatory* on the backend (and sharing validation logic is nice when possible).

It will be good to make sure we do have tests around things that are dynamic at runtime, and could go wrong b/c the server sent invalid data, etc. for example so pls keep me honest! :)

PR #1344, opened 2022-10-14
chore(buttondropdown): apply class name to dropdown menu

jinlee93:
### Summary:
- applies the className prop in `ButtonDropdown` to `DropdownMenu` and not the wrapper
- the trigger button and the items can already be styled individually, just not the menu through `ButtonDropdown`
### Test Plan:
- no visual regression
- unit tests pass

### DIFF
--- a/src/components/ButtonDropdown/ButtonDropdown.tsx
+++ b/src/components/ButtonDropdown/ButtonDropdown.tsx
@@ -27,7 +27,7 @@ export interface Props {
    */
   children?: ReactNode;
   /**
-   * CSS class names that can be appended to the component.
+   * CSS class names that can be appended to the component to further style the dropdown menu.
    */
   className?: string;
   /**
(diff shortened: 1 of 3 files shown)


src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -136,7 +136,6 @@ export const ProjectCard = ({
       {buttonDropdownItems && (
         <CardFooter className={styles['project-card__footer']}>
           <ButtonDropdown
-            className={styles['project-card__button-dropdown']}
```
jinlee93, line 139: Unused styling, so removed, and the snap changes are a result of this

src/components/ButtonDropdown/ButtonDropdown.tsx
```diff
@@ -168,11 +168,17 @@ export const ButtonDropdown = ({
     position === 'bottom-right' && styles['button-dropdown--bottom-right'],
     className,
```
dierat, line 169: So it's being applied to both elements now?

jinlee93: Good catch 🤦  will remove here

dierat (approved):

PR #1351, opened 2022-10-17
chore(tab): export props as type

dierat:
### Summary:
I noticed a console warning in storybook about `Tab` not having an export called `Props`:
```
./src/components/Tab/index.ts 2:0-42"export 'Props' (reexported as 'TabProps') was not found in './Tab'
```

console warning in storybook

This PR removes the warning by exporting `Props` as a `type`.

### Test Plan:
Boot up storybook,
open the browser console,
and verify the warning's gone.

### DIFF
--- a/src/components/Tab/index.ts
+++ b/src/components/Tab/index.ts
@@ -1,2 +1,2 @@
 export { Tab as default } from './Tab';
-export { Props as TabProps } from './Tab';
+export type { Props as TabProps } from './Tab';


src/components/Tab/index.ts
```diff
@@ -1,2 +1,2 @@
 export { Tab as default } from './Tab';
-export { Props as TabProps } from './Tab';
+export type { Props as TabProps } from './Tab';
```
ahuth, line 2: Dang, seems like `@typescript-eslint/consistent-type-imports` could've/should've found this 😞

ahuth (approved):

dierat:
Thanks for reviewing!

PR #1352, opened 2022-10-19
fix(icons): remove play-arrow hard-coded fill color

dierat:
### Summary:
In https://github.com/FB-PLP/traject/pull/12532 we updated the `play-arrow` icon to no longer have a mask. This involved reexporting the icon from figma, which brought along a hard-coded fill color. That hard-coded fill color blocked the icon color from being changed. This PR removes the `fill` attribute to be consistent with the other icons.

### Test Plan:
Navigate to the [icon story in storybook](http://localhost:9000/?path=/docs/atoms-icons-icon--default),
switch to the "docs" tab (if it's not there already),
in the props list, change the icon to "play-arrow" and the color to something random (like grape 800),
and verify the icon color changes to the selected color.

### DIFF
--- a/src/icons/play-arrow.svg
+++ b/src/icons/play-arrow.svg
@@ -1,3 +1,3 @@
 <svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
-<path d="M8 19V5L19 12L8 19Z" fill="#5D6369"/>
+<path d="M8 19V5L19 12L8 19Z" />
 </svg>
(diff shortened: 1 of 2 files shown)


booc0mtaco (commented):
Let's check with design on this first, as something isn't quite right with this icon and the source needs to be updated too.

booc0mtaco (approved):
This answers a question i had now... how come the spritemap seems to be empty... it's because none of the icons have a fill color so they are transparent when previewed on github :)

ahuth (approved):
LGTM

PR #1355, opened 2022-10-21
chore(course-planner-edit): update empty state illustration

dierat:
### Summary:
The empty state for the drag and drop feature in the course planner edit page has an outdated illustration, which keeps confusing design. It doesn't actually effect the end product because this is a recipe. I verified the product is already using the correct image.

### Test Plan:
Navigate to the course planner edit story in storybook,
and verify the empty state illustration has the hand pointing to the **right** (because the user moves the card from the left column into the empty space on the right).

#### Before
course planner edit page in storybook with the old illustration where the hands points to the left

#### After
course planner edit page in storybook with the new illustration where the hands points to the right


dierat:
https://app.shortcut.com/czi-edu/story/215520/update-the-drag-and-drop-illustration

ahuth (approved):
LGTM

dierat:
Thanks for reviewing!

PR #1362, opened 2022-10-31
chore(build): remove source map

jinlee93:
### Summary:
[sc-225828]
- Thought about trying to change the source to the `lib` folder but doesn't make too much sense since the source file itself is not part of the publish
- Including the source files in the publish is unnecessary and makes it heavier
- Source maps don't seem too useful as devs hopefully don't need to debug EDS components from consuming apps and and its currently not working state can either block or spam warnings
- As a result remove source maps
### Test Plan:
- Build is unaffected besides source maps

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -13,7 +13,6 @@
     "outDir": "lib",
     "resolveJsonModule": true,
     "skipLibCheck": true,
-    "sourceMap": true,
     "strict": true,
     "suppressImplicitAnyIndexErrors": true,
     "target": "es6"


booc0mtaco (approved):
Seems fair. It would be trivial to revert this if something truly depends on it, but in its current form it is not helping anyone sounds like.

jeremiah-clothier (approved):

ahuth (approved):
LGTM 🤘 

I think shipping source maps for EDS has limited utility, and doesn't seem to work anyway (without the source files being present 🤔).

PR #1363, opened 2022-11-01
docs(recipes): add recipes overview documentation

dierat:
### Summary:
First pass at adding some documentation on specifically how we use recipes on the EDS team. Feel free to nitpick the text I wrote up for it.

### Test Plan:
Tested it in localhost.

recipes documentation page in storybook running locally

getting started documentation page in storybook running locally, there's an arrow pointing to the new link to the new recipes documentation page


.storybook/components/Recipes/RecipesDocumentation.tsx
```diff
@@ -0,0 +1,25 @@
+import type { ReactNode } from 'react';
```
dierat, line 1: This whole file is just copy/pasted from the other docs files. The only change is the component name on line 16. We could split this out into a reusable component and just leave the `.stories.tsx` files for all 3.
- [Documentation](https://github.com/chanzuckerberg/edu-design-system/blob/08078155c77160222708c6b8f0ef7161006365b4/.storybook/components/Docs/Documentation.tsx)
- [GettingStartedDocumentation](https://github.com/chanzuckerberg/edu-design-system/blob/08078155c77160222708c6b8f0ef7161006365b4/.storybook/components/GettingStarted/GettingStartedDocumentation.tsx)

.storybook/components/Recipes/Recipes.stories.tsx
```diff
@@ -0,0 +1,23 @@
+import type { StoryObj } from '@storybook/react';
+import React from 'react';
+import { RecipesDocumentation } from './RecipesDocumentation';
+import RecipesDocs from '../../../docs/RECIPES.md';
+
+export default {
+  title: '',
```
dierat, line 7: Without a title, it automatically gets dumped in the "Recipes" folder by default.

dierat:
https://app.shortcut.com/czi-edu/story/227000/add-recipes-documentation-in-storybook

jinlee93 (approved):
This looks good to me and have no objections to how recipes are defined here

booc0mtaco (approved):
This makes sense to me. The recipe docs file is basically a FAQ (which we can add to if people come to us with other common questions). Nice one.

I'm still not 100% on board with this recipe concept :) but they do serve some good uses and are very pragmatic

dierat:
> I'm still not 100% on board with this recipe concept :) but they do serve some good uses and are very pragmatic

If you feel strongly against recipes, I think you should consider writing up a proposal for how we should be operating differently, maybe as an RFC? I don't know how tied the rest of the team is to this pattern, but I'm open to considering other paths.

PR #1368, opened 2022-11-03
chore(avatar, avatar-image): move to recipes

dierat:
### Summary:
The `Avatar` component is very specialized and can only represent a version of the Learning Platform avatar. It's not being used in the Learning Platform and never will be because the actual component is very complex and has many possible versions. As such, the `Avatar` is not being exported in EDS and is not used anywhere else; we just use it in stories to more accurately represent the Learning Platform UI. Therefore, I think the `Avatar` should actually be a recipe, not a component. I talked to design and they agreed we don't actually need this to be a full component.

This PR updates the `Avatar` (and its helper `AvatarImage` into the "recipes" area to make it clear that it is not actually intended to be used outside of EDS examples.

This is not marked as a breaking change because it wasn't being exported from EDS yet anyway.

EDS storybook with the Avatar and AvatarImage components in the recipes section

### Test Plan:
Boot up storybook and verify that:
- `Avatar` and `AvatarImage` are now in the "recipes" area
- there are no changes to the `GlobalHeader` recipe; the avatar still renders as expected
- the `Section` story now uses a help icon instead of the avatar for testing the case where there's something in front of the title


dierat:
https://app.shortcut.com/czi-edu/story/225829/move-avatar-and-avatarimage-into-recipes

src/components/Section/Section.stories.tsx
```diff
@@ -17,6 +16,10 @@ export default {
 
 type Args = React.ComponentProps<typeof Section>;
 
+const HelpIcon = (
+  <Icon name="help" purpose="informative" size="1.375rem" title="help" />
+);
+
```
dierat, line 22: I changed both the `titleBefore` and `titleAfter` stories to use an icon without a button so it's easier to see how much spacing is added by the component between the text and whatever's being inserted before/after it.

### Before
section component in storybook before this change

### After
section component in storybook after this change

jinlee93 (approved):
Makes sense to me, also not seeing use of it in LP so does make sense to move it to recipes, thanks!

dierat:
Oh also, I made this a `chore` and not a breaking change because it wasn't being exported from the package yet anyway.

booc0mtaco (approved):
Yah, this `Avatar` component isn't really doing anything, and the implementation in LP for the actual image has code to make the parts vary per user (different colors, etc.) but other teams/projects would likely take a different approach for their brand.

I think this makes sense to move, but I could also see us eventually deleting them tbh (`Avatar` is basically a DIV with a static subcomponent..)

dierat:
> I think this makes sense to move, but I could also see us eventually deleting them tbh (`Avatar` is basically a DIV with a static subcomponent..)

@booc0mtaco good point; this is basically just a helper component for the `GlobalHeader` now. We could move it into that component so it doesn't clutter up the recipe list.

dierat:
I'm going to merge this and open a separate PR where I collapse the `Avatar` into the `GlobalHeader` so we can have a separate discussion about where exactly this should live.

PR #1371, opened 2022-11-04
docs(recipes): move to Documentation section

dierat:
### Summary:
Follow-up to https://github.com/chanzuckerberg/edu-design-system/pull/1363 in which I added recipes documentation to storybook.

I didn't notice that when adding that documentation in the "Recipes" section (by not adding a title to the story file but calling the story itself "Recipes"), a side effect was the "Recipes" section was moved up in the sidebar, between "Atoms" and "Molecules". It's normally between "Organisms" and "Pages".

I actually think the organization of the sidebar is pretty important, so this PR moves the recipes documentation to the "Documentation" section. I'm open to other ideas where it might be more discoverable. One option, if we do want to keep it in the "Recipes" section, is to add a title like "Recipes/_RecipesDocumentation" so it's still at the top of the contents in the folder. 

### Test Plan:
Verify that, in storybook:
- the "Recipes" documentation is now in the "Documentation" section
- the "Recipes" section is now between "Organisms" and "Pages"
- the link in "Getting started" to the "Recipes" documentation still works


dierat:
Thanks for reviewing!

PR #1372, opened 2022-11-07
docs: import examples in component comments to one line with backticks

dierat:
### Summary:
This is part of a series of updates I'm making to the "component comments" (docstrings on components that appear on the docs tab in storybook) to be more consistent and render properly.

There have been some issue with the `import` example that appears at the top of the component comments. For some reason, it sometimes appears as `[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object]` with white text and a light gray outer glow. Right now there are 2 components in particular where this is happening:
- `Text`
  - to repro, navigate to https://chanzuckerberg.github.io/edu-design-system/?path=/docs/atoms-text-text--body and click one of the "Show code" buttons below
- `Dropdown`
  - to repro, navigate to https://chanzuckerberg.github.io/edu-design-system/?path=/docs/molecules-forms-dropdown--default and refresh the page

As usual, I have no idea what's really going on, but it seems to be somehow related to the fact that these code examples use a code block (with a line of 3 backticks, the code on a new line, and another line with 3 backticks) but is only one line. If I duplicate the import line, it's fine and I can no longer repro the issue. This only seems to be a problem on pages that have other code blocks in the component comment.

Since it's just one line, my suggestion is that we change the import statements to just use 2 backticks on one line instead of a code block.

I'm suggesting we do this:
`import {Text} from "@chanzuckerberg/eds";`
instead of this:
```
import {Text} from "@chanzuckerberg/eds";
```
as long as it's only one line of code. In this PR, I'm making this change throughout the codebase to be consistent and avoiding running into this again when someone adds a code block with some example usage somewhere else.

### Test Plan:
- navigate to http://localhost:9000/?path=/docs/atoms-text-text--body
- refresh the page
- click one of the "Show code" buttons below
- verify the import statement at the top of the page is still readable text

### DIFF
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -94,9 +94,7 @@ const variantToIconAssetsMap: {
  * @deprecated The Banner component is deprecated and will be removed in an upcoming release.
  * Please visit Zeroheight to find the right notification component for your needs: https://eds.czi.design/
  *
- * ```ts
- * import {Banner} from "@chanzuckerberg/eds";
- * ```
+ * `import {Banner} from "@chanzuckerberg/eds";`
  *
  * A banner used to provide and highlight information to a user or ask for a decision or action.
  *
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -26,12 +26,9 @@ export type CheckboxProps = Omit<CheckboxInputProps, 'id'> & {
 };
 
 /**
- * ```ts
- * import {Checkbox} from "@chanzuckerberg/eds";
- * ```
- * ```ts
- * import {CheckboxInput, CheckboxLabel} from '@chanzuckerberg/eds';
- * ```
+ * `import {Checkbox} from "@chanzuckerberg/eds";`
+ *
+ * `import {CheckboxInput, CheckboxLabel} from '@chanzuckerberg/eds';`
  *
  * Checkbox control indicating if something is selected or unselected.
  *
--- a/src/components/Popover/Popover.tsx
+++ b/src/components/Popover/Popover.tsx
@@ -73,11 +73,9 @@ export const defaultPopoverModifiers = [
 /**
  * BETA: This component is still a work in progress and is subject to change.
  *
- * A styled popover built on 'headless UI' popover. Consider using the Dropdown component, instead.
+ * `import {Popover} from "@chanzuckerberg/eds";`
  *
- * ```ts
- * import {Popover} from "@chanzuckerberg/eds";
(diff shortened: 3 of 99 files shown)


src/components/Banner/Banner.tsx
```diff
@@ -94,9 +94,7 @@ const variantToIconAssetsMap: {
  * @deprecated The Banner component is deprecated and will be removed in an upcoming release.
```
dierat, line 94: This component comment is currently not appearing at all; will address in a separate PR.

src/components/Checkbox/Checkbox.tsx
```diff
@@ -26,12 +26,9 @@ export type CheckboxProps = Omit<CheckboxInputProps, 'id'> & {
...
 
 /**
- * ```ts
- * import {Checkbox} from "@chanzuckerberg/eds";
- * ```
- * ```ts
- * import {CheckboxInput, CheckboxLabel} from '@chanzuckerberg/eds';
- * ```
+ * `import {Checkbox} from "@chanzuckerberg/eds";`
+ *
```
dierat, line 30: Empty line added to force them to appear on separate lines.

src/components/Popover/Popover.tsx
```diff
@@ -73,11 +73,9 @@ export const defaultPopoverModifiers = [
 /**
  * BETA: This component is still a work in progress and is subject to change.
  *
- * A styled popover built on 'headless UI' popover. Consider using the Dropdown component, instead.
```
dierat, line 76: I moved this line down because the import should be at the top of the comment (unless there's a BETA line, in which case that one's at the top).

jinlee93:
Interesting that this is also only with the first line of the docstring and not lines after like:
```
 * ```tsx
 * <HorizontalStepper activeIndex={0} steps={['Step 1', 'Step 2', 'Step 3']} />
 * ```
```
and that shows up fine

src/components/Checkbox/Checkbox.tsx
jinlee93: I like separate 👍 Distinguishes that you can do either of the options rather than both

jinlee93 (approved):
Thanks for fixing the docs, `[object Object]` doesn't really mean much I guess

jinlee93:
Oh could you also update `Component.tsx.hbs` to 
```
 * ` import {{inBraces name}} from "@chanzuckerberg/eds";`
```

booc0mtaco (approved):
Lots of tweaks. Thanks! looks good to me

dierat:
Thanks for the reviews!

PR #1373, opened 2022-11-07
chore(card, table): add data-bootstrap-override data attributes for LP override

jinlee93:
### Summary:
- adds `data-bootstrap-override` attributes to target to override global styling in LP
### Test Plan:
- alpha publish [7.0.0-alpha.3](https://www.npmjs.com/package/@chanzuckerberg/eds/v/7.0.0-alpha.3)doesn't break EDS Card and Table in lp
  - [pr to update to 7.0 once released](https://github.com/FB-PLP/traject/tree/jlee/EDS7.0)
  - [viewCoursePlanner story which uses both tables and cards](https://614b969675171d003ac6f98a-iearkjpqyq.chromatic.com/?path=/story/teachers-courses-courseplan-viewcourseplan--view-course-planner-story)
- no visual regression of cards and tables and their used pages/recipes

### DIFF
--- a/.storybook/pages/FeedbackOverview/FeedbackOverview.tsx
+++ b/.storybook/pages/FeedbackOverview/FeedbackOverview.tsx
@@ -239,8 +239,8 @@ export const FeedbackOverview = ({ activeIndex = 0 }: Props) => {
                         <Table.HeaderCell>Final Product</Table.HeaderCell>
                       </Table.Row>
                       <Table.Row variant="header">
-                        <Table.Cell className="py-2"></Table.Cell>
-                        <Table.Cell className="py-2">
+                        <Table.Cell className="!py-2"></Table.Cell>
+                        <Table.Cell className="!py-2">
                           <NumberIconList>
                             <NumberIcon
                               aria-label="Item 1"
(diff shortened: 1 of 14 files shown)


.storybook/pages/FeedbackOverview/FeedbackOverview.tsx
```diff
@@ -239,8 +239,8 @@ export const FeedbackOverview = ({ activeIndex = 0 }: Props) => {
                         <Table.HeaderCell>Final Product</Table.HeaderCell>
                       </Table.Row>
                       <Table.Row variant="header">
-                        <Table.Cell className="py-2"></Table.Cell>
-                        <Table.Cell className="py-2">
+                        <Table.Cell className="!py-2"></Table.Cell>
```
dierat, line 242: Can you elaborate on why we need the `!important` modifiers now?

jinlee93: It doesn't override the padding styles in TableCell.module.css:
```css
.table-cell {
  @mixin eds-theme-typography-body-text-sm;

  padding: var(--eds-size-2) var(--eds-size-1);
}
```
but now I'm thinking I might just add data attributes like `data-bootstrap-override="table-cell"` and override in LP instead of removing the `:where` pseudoclasses instead, so we wouldn't need the `!important` modifiers

PR #1374, opened 2022-11-07
docs: all components and recipes have meaningful comments

dierat:
### Summary:
We already went through and updated most of the components to have meaningful comments, but there are still a couple of lingering ones that have the old default undescriptive comment or have no description of the component/recipe.

I'm also removing comments on documentation pages because those are worthless.

### Test Plan:
Go through the components that have been updated and verify their updated component comments appear in the docs tab in storybook.

### DIFF
--- a/plop-templates/Recipe/Recipe.tsx.hbs
+++ b/plop-templates/Recipe/Recipe.tsx.hbs
@@ -10,7 +10,7 @@ export interface Props {
 }
 
 /**
- * Primary UI component for user interaction
+ * TODO: update this comment with a description of the recipe.
  */
 export const {{pascalCase name}} = ({
   className,
--- a/src/components/Heading/Heading.tsx
+++ b/src/components/Heading/Heading.tsx
@@ -59,6 +59,10 @@ const getComputedAs = (size: HeadingSize) => {
 
 /**
  * `import {Heading} from "@chanzuckerberg/eds";`
+ *
+ * A component for styling heading text (`<h1>`-`<h6>`).
+ *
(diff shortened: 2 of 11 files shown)


plop-templates/Recipe/Recipe.tsx.hbs
```diff
@@ -10,7 +10,7 @@ export interface Props {
 }
 
 /**
- * Primary UI component for user interaction
+ * TODO: update this comment with a description of the recipe.
```
dierat, line 13: This todo comment now matches the one in the component plop template file.

src/components/Heading/Heading.tsx
```diff
@@ -59,6 +59,10 @@ const getComputedAs = (size: HeadingSize) => {
 
 /**
  * `import {Heading} from "@chanzuckerberg/eds";`
+ *
+ * A component for styling heading text (`<h1>`-`<h6>`).
+ *
+ * Be careful to pass the correct heading element via the `as` prop to avoid skipping heading levels because [heading levels increasing by only one level at a time is important for screen reader users.](https://www.w3.org/WAI/tutorials/page-structure/headings/)
  */
```
booc0mtaco, line 67: nit: I wonder why this one has a space between the comment block and the export... would this break the intellisense in VSCode?

dierat: Do you mean the popover with the preview of the comment content? If so, that seems fine. I guess they've outsmarted empty lines between the components and their comments. (I vaguely remember an issue where the documentation wasn't appearing in storybook because of a space like this.)

dierat: That being said, I can still remove the space.

booc0mtaco: ha, yeah i remember it getting confused if the block wasn't right up against the code. Good to know it doesn't hurt nowadays

booc0mtaco (approved):

PR #1375, opened 2022-11-08
chore: replace the @ to default in the compiled css-variables-nested.json

schakraborty22:
### Summary:
The task for this ticket is to replace all the '@' in the css-variables-nested.json, for which needed to enhance the style-dictionary.config.js to find the deeply nested @ in the icon object

### Test Plan:
1) yarn build
2) yarn test

### DIFF
--- a/src/components/BreadcrumbsItem/BreadcrumbsItem.module.css
+++ b/src/components/BreadcrumbsItem/BreadcrumbsItem.module.css
@@ -102,7 +102,7 @@
  * Breadcrumbs Icon - a separator between breadcrumb links.
  */
 .breadcrumbs__separator {
-  color: var(--eds-theme-color-icon-neutral-subtle);
+  @apply text-icon-neutral-subtle-default;
   margin-left: var(--eds-size-1);
   margin-right: var(--eds-size-1);
   cursor: default;
--- a/src/components/DropdownButton/DropdownButton.module.css
+++ b/src/components/DropdownButton/DropdownButton.module.css
@@ -25,8 +25,7 @@
   align-items: center;
 
   cursor: pointer;
-
-  background-color: var(--eds-theme-color-background-neutral-default);
+  @apply bg-background-neutral-default-default;
 }
 
 /**
--- a/src/tokens-dist/json/css-variables-nested.json
+++ b/src/tokens-dist/json/css-variables-nested.json
@@ -185,24 +185,24 @@
         "background": {
           "neutral": {
             "default": {
-              "@": "var(--eds-theme-color-background-neutral-default)",
-              "hover": "var(--eds-theme-color-background-neutral-default-hover)"
+              "hover": "var(--eds-theme-color-background-neutral-default-hover)",
+              "default": "var(--eds-theme-color-background-neutral-default)"
             },
             "subtle": {
-              "@": "var(--eds-theme-color-background-neutral-subtle)",
-              "hover": "var(--eds-theme-color-background-neutral-subtle-hover)"
+              "hover": "var(--eds-theme-color-background-neutral-subtle-hover)",
+              "default": "var(--eds-theme-color-background-neutral-subtle)"
             },
             "medium": {
-              "@": "var(--eds-theme-color-background-neutral-medium)",
-              "hover": "var(--eds-theme-color-background-neutral-medium-hover)"
+              "hover": "var(--eds-theme-color-background-neutral-medium-hover)",
+              "default": "var(--eds-theme-color-background-neutral-medium)"
             }
           },
           "brand": {
             "primary": {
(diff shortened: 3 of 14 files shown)


src/tokens-dist/json/css-variables-nested.json
```diff
@@ -185,24 +185,24 @@
         "background": {
           "neutral": {
             "default": {
-              "@": "var(--eds-theme-color-background-neutral-default)",
-              "hover": "var(--eds-theme-color-background-neutral-default-hover)"
+              "hover": "var(--eds-theme-color-background-neutral-default-hover)",
+              "default": "var(--eds-theme-color-background-neutral-default)"
```
jinlee93, line 189: Can you show how this would be used with Tailwind utility classes?

dierat:
https://app.shortcut.com/czi-edu/story/224552/replace-the-to-default-in-the-compiled-css-variables-nested-json

src/tokens-dist/json/css-variables-nested.json
schakraborty22: let me check..thanks

schakraborty22: @jinlee93  adding Tailwind utility classes is working after replacing "@" to "default" in css-variables-nested.json ,please check and let me know thanks

src/components/DropdownButton/DropdownButton.module.css
```diff
@@ -25,8 +25,7 @@
   align-items: center;
 
   cursor: pointer;
-
-  background-color: var(--eds-theme-color-background-neutral-default);
+  @apply bg-background-neutral-default-default;
```
jinlee93, line 28: Hi @schakraborty22 sorry for getting back to this late but we want to find a way to compile the TW styles to use `bg-background-neutral-default` and not `bg-background-neutral-default-default`. This is occurring due to `css-variables-nested.json` being 
```css
"background": {
  "neutral": {
    "default": {
      "hover": "var(--eds-theme-color-background-neutral-default-hover)",
      "default": "var(--eds-theme-color-background-neutral-default)"
    }
  }
}
```
Previously 
```css
```css
"background": {
  "neutral": {
    "default": {
      "hover": "var(--eds-theme-color-background-neutral-default-hover)",
      "@": "var(--eds-theme-color-background-neutral-default)"
    }
  }
}
```
would become `bg-background-neutral-default-@" which we're also trying to avoid.

src/components/BreadcrumbsItem/BreadcrumbsItem.module.css
```diff
@@ -102,7 +102,7 @@
  * Breadcrumbs Icon - a separator between breadcrumb links.
  */
 .breadcrumbs__separator {
-  color: var(--eds-theme-color-icon-neutral-subtle);
+  @apply text-icon-neutral-subtle-default;
```
jinlee93, line 105: This will only work with Storybook and consumers with tailwind enabled with configs extending our config, so I think these kinds of change should be punted and combined with enabling TW with general EDS component use

booc0mtaco:
https://app.shortcut.com/czi-edu/story/224552/replace-the-to-default-in-the-compiled-css-variables-nested-json

booc0mtaco:
Could someone provide a little summary of why we want to do this (versus doing nothing, or something different)? Trying to imagine the context and not sure if I get it 😖. Were the `@` used initially, and we don't want to do that anymore?

jinlee93:
> Could someone provide a little summary of why we want to do this (versus doing nothing, or something different)? Trying to imagine the context and not sure if I get it 😖. Were the `@` used initially, and we don't want to do that anymore?

Yea of course!
- The `colors-background.json` and other similar files are where we define our tier 2 / 3 tokens, looks like: 
image
- In this screenshot, `{ eds: {theme: {color: {background: {neutral: {default: {@: {value: "{eds.color.neutral.white}"` compiles to `EdsThemeColorBackgroundNeutralDefault` in `src/tokens-dist/ts/colors.ts` and css custom property `--eds-theme-color-background-neutral-default`. 
- However the formatter used to generate `css-variables-nested.json` (primarily used for TW) compiles to `{ eds: {theme: {color: {background: {neutral: {default: {@: {value: "var(--eds-theme-color-background-neutral-default)"` which when used with `@apply` or TW utility classes, becomes `-color-background-neutral-default-@`, but I would like that to be `-color-background-neutral-default` to be similar with what's in the mocks, and the tokens that are generated.
- This pr changes `-@` to `-default`, which is not preferred since that would become `color-background-neutral-default-default`
- https://github.com/chanzuckerberg/edu-design-system/pull/1387 resolves this

Will be copy/pasting this block into #1387 and the linked Jira tickets

PR #1376, opened 2022-11-08
Add accordion component

jinlee93:
### Summary:
- WIP
- Adds accordion component
- Should follow [w3 accordion pattern](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/)
- [Mocks](https://www.figma.com/file/kCX4vzCtunnJSKQ1cwvksH/Accordions%3A-Creation?node-id=7866%3A29625)

RFC:
- Considering making the accordion row a separate component i.e. `<AccordionRow>`, and the `<Accordion>` the wrapping stack with the React context for `variant: compact` and `hasOutline` (or something similar)
  - i.e.: 
```tsx
<Accordion>
  <Accordion.Row>
    <Accordion.Button>
      Title 1
    </Accordion.Button>
    <Accordion.Panel>
      Content 1
    </Accordion.Button>
  </Accordion.Row>
  <Accordion.Row>
    <Accordion.Button>
      Title 2
    </Accordion.Button>
    <Accordion.Panel>
      Content 2
    </Accordion.Button>
  </Accordion.Row>
</Accordion>
```
  - This solves for an outlined variant with slightly different paddings and no bottom border on the last accordion row
  - This could be needlessly complicated and adds boilerplating when slapping some TW classes can also solve this (but this would be a lot of added classnames from consumer)
  - However the aforementioned w3 accordion pattern describes the accordion as the entire thing like this code snippet here and not the individual sections

Upcoming (done):
- [x] Rotating indicator icon
- [x] Chromatic stories with `defaultOpen: true` for visual regression testing
- [x] Unit tests to open / close with mouse and kb
- [x] CSS comments and removing TODO comments for real ones

### Test Plan:
- stories written
- snaps written
- unit tests written

https://user-images.githubusercontent.com/86632227/200771010-14d7ea5a-d203-4ccc-af42-d0482e0ee016.mov

https://user-images.githubusercontent.com/86632227/200771025-b4fa4cd8-d829-48b5-8565-230737d65494.mov


src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,84 @@
...
+}: Props) => (
+  <AccordionContext.Provider value={{ variant }}>
+    <Disclosure {...other}>
+      {({ open }) => (
+        <div
+          className={clsx(
+            styles['accordion'],
+            className,
+            open && classNameOpen,
+            !open && classNameClosed,
```
jinlee93, line 72: Pattern used in LP

src/components/AccordionButton/AccordionButton.module.css
```diff
@@ -0,0 +1,49 @@
+/*------------------------------------*\
+    # ACCORDION BUTTON
+\*------------------------------------*/
+
+/**
+ * Accordion Button, wraps the heading and open indicator icon.
+ */
+.accordion-button.accordion-button {
```
jinlee93, line 8: This and the rest of the the `AccordionButton` styles had difficulty overriding some of the styling in [ClickableStyle](https://github.com/chanzuckerberg/edu-design-system/blob/next/src/components/ClickableStyle/ClickableStyle.module.css#L10) hence the double selectors. @dierat do you have any suggestions for avoiding the double selectors?

src/components/Accordion/Accordion.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+          Suscipit dui, nunc sit dui tellus massa laoreet tellus.
+        </Accordion.Panel>
+      </Accordion>
+    </div>
+  ),
+};
+
+/**
+ * The following stories are mostly for visual regression testing to capture the open state.
+ */
```
dierat, line 93: Should this comment be after the `StackedCompact` story and before `DefaultOpen` instead?

dierat:
>Considering making the accordion row a separate component i.e. <AccordionRow>, and the <Accordion> the wrapping stack with the React context for variant: compact and hasOutline (or something similar)

My understanding of an `Accordion` is that only one section can be open at a time, and that's something else we can control if all the sections are wrapper in one big component vs having a list of accordions with no logical wrapping. Did you happen to touch on this with design, whether multiple sections can be open at the same time or not?

src/components/AccordionButton/AccordionButton.module.css
```diff
@@ -0,0 +1,49 @@
...
+  transform: rotate(180deg);
+}
+
+/**
+ * Animates the icon rotation when opening and closing.
+ */
+.accordion-button > .accordion-button__icon {
+  transition-property: transform;
+  transition-duration: var(--eds-anim-move-medium);
+  transition-timing-function: var(--eds-anim-ease);
```
dierat, line 44: Just curious, is there a particular reason you split this across 3 lines? Is it for readability?

```suggestion
  transition: transform var(--eds-anim-move-medium) var(--eds-anim-ease);
```

src/components/AccordionButton/AccordionButton.tsx
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * Callback called when accordion is closed.
+   */
+  onClose?: () => void;
+};
+
+/**
+ * BETA: This component is still a work in progress and is subject to change.
+ *
+ * The interactive button subcomponent that houses the heading text and open indicator icon for the EDS Accordion component.
```
dierat, line 28: Nit: "interactive button" seems a little redundant.

src/components/Accordion/Accordion.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+    render(<Default />);
+    expect(screen.queryByTestId('accordion-panel')).not.toBeInTheDocument();
+    const accordionButton = screen.getByTestId('accordion-button');
+    userEvent.click(accordionButton);
+    expect(screen.getByTestId('accordion-panel')).toBeInTheDocument();
+    userEvent.click(accordionButton);
+    expect(screen.queryByTestId('accordion-panel')).not.toBeInTheDocument();
+  });
+
+  it('should open and close Accordion panel with space and enter keys on the Accordion button', () => {
```
dierat, line 24: Thanks for the great tests as always ❤️

dierat:
Have you looked at the hover and focus states with design? Just wondering since it's not in the mocks.

src/components/AccordionButton/AccordionButton.module.css
dierat: Mm yeah that's tough. I guess if we want we could try using `:where`, or we oculd combine the class names so something like `.clickable-style--primary.clickable-style--brand` would become `.clickable-style--primary--brand`. For adding the classes to the component, we could change
```
      [...]
      status === 'brand' && styles['clickable-style--brand'],
``` 
to 
```
      styles['clickable-style--primary'],
      [...]
      styles[`clickable-style--${variant}--${status}`],
``` 
What do you think? Worth it? Not just for the `Accordion` but also for consumers in LP, if we want people to be able to override the `Button` and `Link` styles like this.

src/components/AccordionButton/AccordionButton.module.css
dierat: Also, we can't this stuff go in `.accordion-button__icon {`? Is it a specificity issue?

src/components/Accordion/Accordion.stories.tsx
jinlee93: Yes 🤦 thanks for catching

src/components/AccordionButton/AccordionButton.tsx
jinlee93: True, will change to just `button`

src/components/AccordionButton/AccordionButton.module.css
jinlee93: Nah I like the condensed, thanks for the suggestion.

And yea it's more specificity issues https://github.com/chanzuckerberg/edu-design-system/blob/e519e4c6cfc0f849354142c2967a261c23a78b4c/src/components/ClickableStyle/ClickableStyle.module.css#L23-L25

jinlee93:
> Have you looked at the hover and focus states with design? Just wondering since it's not in the mocks.

Oh good Q, will bring up with Sean

dierat (approved):
LGTM.

src/components/AccordionButton/AccordionButton.module.css
jinlee93: The specificity issue here was with `.clickable-style` itself and not the more specific variant styling.

I want to say it wouldn't be super necessary if we want to intentionally constrain `Buttons` `Links` and other `ClickableStyle` to the options available, and with `AccordionButton` I can use a native `button` instead and apply similar styling that I would get from using the component `<Button>`.

For other cases, do you think there would be other similar case as this AccordionButton? Would we lose potential consumers if the styling was not overridable?

src/components/Accordion/Accordion.stories.tsx
```diff
@@ -0,0 +1,134 @@
...
+  component: Accordion,
+  parameters: {
+    badges: [BADGE.BETA],
+  },
+  args: {
+    children: (
+      <>
+        <Accordion.Button data-testid="accordion-button">
+          <Accordion.Title as="h2">Massa quam egestas massa.</Accordion.Title>
+        </Accordion.Button>
```
booc0mtaco, line 18: Putting this here in the example instead of in the code below for clarity

Is there a case to support where the `Accordion.Button` will have something other than an `Accordion.Title` component (that is not a `"h2"`)? We could streamline usage by providing one component in the implementation (e.g., `Accordion.Header` or similar), and hiding the fact that headless has this button/title separation.

In this example, `Accordion.Header` would have the `as` prop if needed, which would get applied to the Disclosure  title sub-component. Thoughts?

Another consideration: I assume all the `Accordion.Title` instances should have the same, so perhaps configuring the component to use in the current `Accordion.Title` could be handled on `Accordion`

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,91 @@
...
+   */
+  className?: string;
+  /**
+   * Additional classnames passed in for styling active only when open.
+   */
+  classNameOpen?: string;
+  /**
+   * Additional classnames passed in for styling active only when not open.
+   */
+  classNameClosed?: string;
```
booc0mtaco, line 25: Do other components follow this pattern for applying things based on state? I would lean toward us applying some state class, which consumers would use in local css modules  to `@apply` any styling tweaks. For example, panels have some class `.accordion__open` or `__closed`, where the combo of that an an id (or class added via `className`) lets people set what they want

jinlee93: I like this pattern you're suggesting. Made a change to: 
```tsx
<Disclosure defaultOpen={defaultOpen}>
  {({ open }) => (
    <div
      className={clsx(
       styles['accordion'],
       open && 'eds-accordion--open',
       !open && 'eds-accordion--closed',
       className,
     )}
      {...other}
    >
     {children}
   </div>
 )}
</Disclosure>
```
Is this along the lines of what you were thinking of?

src/components/Accordion/Accordion.module.css
```diff
@@ -0,0 +1,12 @@
+/*------------------------------------*\
+    # ACCORDION
+\*------------------------------------*/
+
+/**
+ * Accordion used to hide or reveal content downwards when interacted.
+ */
+.accordion {
+  width: 100%;
```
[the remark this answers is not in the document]
jinlee93: It should be full-width since it's a block, will remove

src/components/Accordion/Accordion.stories.tsx
jinlee93: > In this example, Accordion.Header would have the as prop if needed, which would get applied to the Disclosure title sub-component. Thoughts?
This makes sense but if a consumer wants the style the Title, would they nest the text in another styled element?

>Another consideration: I assume all the Accordion.Title instances should have the same, so perhaps configuring the component to use in the current Accordion.Title could be handled on Accordion
Are you suggesting receiving the title as a prop instead of using this subcomponent structure?

src/components/AccordionButton/AccordionButton.module.css
dierat: >The specificity issue here was with .clickable-style itself and not the more specific variant styling.

So would we need to use `:where` to reduce that specificity to allow you to override it more easily?

>with AccordionButton I can use a native button instead and apply similar styling that I would get from using the component <Button>. (lmk your thoughts on this)

I was actually wondering this and I think it depends on how design feels about the hover state. If we want the hover state to be consistent with the other icon buttons, I think using the `Button` EDS component makes sense because that will keep them in sync. If Sean wants something else, it might be easier to use a `button` element.

src/components/Accordion/Accordion.tsx
booc0mtaco: Yah exactly! one small check is to see if we use open/close or expand/collapse anywhere. No need to fix if we vary on this, but a small thing we would want to make consistent as we go.

jinlee93: There's no precedent for this and the state based classNames were just brought over from EDSCandidates (as there are some consuming components that need this). 
I am down to keep this pattern going forward if this kind of need arises in other components.

src/components/Accordion/Accordion.stories.tsx
booc0mtaco: > This makes sense but if a consumer wants the style the Title, would they nest the text in another styled element?

I was fuzzy about how much we would want to allow titles to be styled. if the header can only be text, then we could simplify the component interface a bit:

```
<Accordion.Header as="h2" data-testid="accordion-button">Test Text</Accordion.Header>
``` 
=> 
```
<Accordion.Button>
     <Accordion.Title as="h2">Test Text</Accordion.Title>
</Accordion.Button>
```

> Are you suggesting receiving the title as a prop instead of using this subcomponent structure?

I accidentally a word :P "should have the same **header level**"

This might break a pattern, but my thought was to DRY the use of `as="h2"` across an accordion's individual sections by moving **that** prop to `Accordion`. e.g., `<Accordion as="h2">...</Accordion>` which would apply that to the relevant children. Hope that clears it up

src/components/AccordionButton/AccordionButton.module.css
dierat: >For other cases, do you think there would be other similar cases as this AccordionButton? Would we lose potential consumers if the styling was not overridable?

Ideally it would only be cases like this where they're making a reusable thing that is different enough from the standard button that it requires unique styling, and if it is very different from our standard buttons, that also begs the question whether we should just use a native `button` instead. So I think I'm okay with people having to increase the specificity to override it since it should be pretty rare. What do you think?

jinlee93: Yea I agree with that.
I'll update this button with the native `button` once I get feedback regarding the focus/hover states.

src/components/Accordion/Accordion.tsx
dierat: Counterpoint—I think it's kinda annoying when people need to dig into the guts of a thing to figure out how to style different states. I like that props are exposed and documented in a way that suggests they are available for customization.

src/components/Accordion/Accordion.stories.tsx
jinlee93: Ahhhh thanks for the clarification.
I agree with that the `as` can be set in the highest level `Accordion` component and passed down as necessary. 
I think the `Accordion.Title` component does need to accept styling via `className` and allow elements other than text so will keep this sort of structure.

I have a follow up pr coming with the `Accordion` being the wrapper and `Accordion.Row` being the individual rows so I will move the `as` there

src/components/Accordion/Accordion.tsx
jinlee93: Wait I think I am leaning with Diedra here. Due to our recommendation of CSS modules, the resulting classNames are slightly jumbled and would need to apply `:global` to `eds-accordion--open` and `eds-accordion--closed` to target, which is linted against and discouraged. To target a specific use of accordion, the open/closed selectors would have to be combined with another selector as well, increasing the specificity.

booc0mtaco: Oh this `:global` thing I didn't know about. Does this mean that adding class names for states don't show as written, unless we make it global? Sounds like we don't want people to know or use our class names in consuming code. 

I usually discourage props that have state and modifiers as part of their naming, e.g., className${state}. They have a bad habit of expanding and becoming inconsistent over time, so I treat it like a code smell. I didn't see any other such `classNameX` props in other components yet either, so now wondering if we need these two?

jinlee93: Adding class names show as written, but a class name such as `styles[.accordion]` will show as `Accordion-module__accordion__2yPgy`.
The `eds-accordion--closed` will add as written since it's not a css module classname, but in LP's stylesheet, would have to target like `:global(.eds-is--closed)`

dierat: >I didn't see any other such classNameX props in other components yet either, so now wondering if we need these two?

Just because it hasn't come up yet. We also don't have any class names for states for consumers to reuse in their styles yet either.

jinlee93: How do y'all feel about just nixing `classNameOpen` / `classNameClosed ` and telling the LP consumer "tough"?

dierat: I'm just concerned that we'll lose on adoption if we don't have some kind of an escape hatch. I'd rather have the class names they can use with `global` than nothing at all.

booc0mtaco: I'm in favor of making this a discussion topic, but trimming it for now. We can treat it as a feature request if it comes up. Happy to facilitate in the right meeting for such topics

anniehu4: also calling out that these were added organically in LP because we didn't have design guidance so product designers were in total free space -- now that we're working with Sean, we can afford to be stricter. we may end up needing to re-add it, but that should be a thoughtful convo with our design + product design

jinlee93: Trimmed the extra classname props and instead added a story about using the `open` render prop to possibly control classnames

src/components/AccordionButton/AccordionButton.tsx
```diff
@@ -0,0 +1,95 @@
...
+                onClose();
+              }
+            }
+          }}
+          status="neutral"
+          variant="icon"
+          {...other}
+        >
+          <Heading className={headingClassName} size={headingAs}>
+            {children}
```
jinlee93, line 79: Removed the `Accordion.Title` subcomponent for use in `Accordion.Button` directly. The styling is hard to override anyways to begin with due to double class selectors and this can be selected in LP via `> h2` or wrap their title in a styled element if they really disagree with the EDS styling

src/components/AccordionButton/AccordionButton.module.css
jinlee93: Sean is fine with the styling here except for the border-radius, which I updated here and will update accordingly in the other Accordion pr

PR #1377, opened 2022-11-08
docs: misc cleanup for component comments

dierat:
### Summary:
This PR contains a grab bag of various component comment cleanups, such as:
- removing any references to "Primary UI component for user interaction", which was the old default component comment and is still hanging around in a couple of places where the comment was updated to include actual useful information
- adding backticks around elements being used as text (mostly in `Text.tsx`)
- code blocks use `tsx` instead of `ts`
- various grammar touch-ups (like adding periods at the end of sentences)
- general wordsmithing to reduce redundancy

### Test Plan:
Viewed the updated components in storybook to verify their documentation has updated.

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -171,7 +171,9 @@ export const ModalContent = (props: ModalContentProps) => {
  * NOTE: You must have at least one focusable element in the modal contents, for keyboard
  * accessibility reasons. (The close button counts as a focusable element.)
  *
- * @example
+ * Example usage:
+ *
+ * ```tsx
  * <Modal>
  *   <Modal.Header>
  *     <Modal.Title>{modalTitle}</Modal.Title>
@@ -182,6 +184,7 @@ export const ModalContent = (props: ModalContentProps) => {
  *     {modalFooterContent}
  *   </Modal.Footer>
  * </Modal>
+ * ```
  */
 export const Modal = (props: ModalProps) => {
   const {
--- a/src/components/Text/Text.tsx
+++ b/src/components/Text/Text.tsx
@@ -57,18 +57,21 @@ export type Props = {
  * `import {Text} from "@chanzuckerberg/eds";`
  *
  * There are two perceived use cases for the text component.
- * One is to decorate <p> and <span> with thematic variants.
- * Defaults to <p> and should pass as="span" to set as <span>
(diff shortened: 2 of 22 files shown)


src/components/Text/Text.tsx
```diff
@@ -57,18 +57,21 @@ export type Props = {
  * `import {Text} from "@chanzuckerberg/eds";`
  *
  * There are two perceived use cases for the text component.
- * One is to decorate <p> and <span> with thematic variants.
- * Defaults to <p> and should pass as="span" to set as <span>
+ * One is to decorate `<p>` and `<span>` with thematic variants.
+ * Defaults to `<p>` and should pass `as="span"` to set as `<span>`.
```
dierat, line 61: Before:
text component in storybook before this change

After:
text component in storybook after this change

src/components/Modal/Modal.tsx
```diff
@@ -171,7 +171,9 @@ export const ModalContent = (props: ModalContentProps) => {
  * NOTE: You must have at least one focusable element in the modal contents, for keyboard
  * accessibility reasons. (The close button counts as a focusable element.)
  *
- * @example
```
dierat, line 174: Because of this line, nothing below this point was showing up.

dierat:
https://app.shortcut.com/czi-edu/story/227525/clean-up-component-comments-in-storybook

src/components/Modal/Modal.tsx
jinlee93: Weird. What kind of docs should we revise for avoiding these kinds of jsdoc tags?

jinlee93 (approved):
Thanks for the cleanup!

src/components/Modal/Modal.tsx
dierat: Sorry, I'm not following. Do you mean "advise"? I was planning on just putting `@deprecated` at the end of the comments and avoiding using the others until this gets fixed.

jinlee93: Sorry to clarify, it seems that Storybook isn't happy with certain jsdoc tags like `@example`. Do we have a place to put that information so we don't add them in the future?

dierat: Oh I see. I can can add it to the code guidelines. Good call 👍

PR #1378, opened 2022-11-09
chore(Tooltip): replace tier 1 token usage in styles

dierat:
### Summary:
In order for theming to work properly, components should only be using tier 2 or 3 tokens (except in cases of deprecated components or variants which we don't want people to theme anyway). There are a couple of instances where tier 1 tokens are being used, notably in the `Tooltip` component. I'll address the others in a separate PR.

### Test Plan:
Verify there are no color changes to the `Tooltip` component when open.


jinlee93 (approved):
Thanks!

PR #1380, opened 2022-11-09
chore(icons): remove second fill color variable from status icons

dierat:
### Summary:
While looking for tier 1 color tokens used in components, I noticed the 4 "status" icons, which have a fill color baked in, each have a variable for overriding the fill color that is not being used. The [original intent](https://github.com/chanzuckerberg/edu-design-system/pull/1104) was to give consumers a way to override the fill color. But there is another way to do that, which is to override the other variable being used to determine the color.

For example, the success icon fill code looks like this:
`fill="var(--eds-color-success-icon, var(--eds-theme-color-icon-utility-success))"`

The `--eds-color-success-icon` is the variable that is only used here, and not defined anywhere, so consumers can customize the color of this icon. Hopefully people don't actually have to override the fill color here, but if they absolutely need to, they could also override `--eds-theme-color-icon-utility-success`.

This PR removes the unused color variables from the 4 status icons.

Thoughts?

### Test Plan:
Verify there are no visual changes in these icons.


jinlee93:
I'm messing around with this a little bit and it seems that the color is very hard to override. Not doable via the `color` prop nor inline styling via the inspector 
image
The only way might be overriding `--eds-theme-color-icon-utility-success` as mentioned.

Is there any value in surfacing how to override that token or do we not expect overrides to not be needed at all for these icons? Maybe just wait until a request?

dierat:
> Is there any value in surfacing how to override that token or do we not expect overrides to not be needed at all for these icons? Maybe just wait until a request?

@jinlee93 I don't think they'll be needed but they're not hurting anyone, just confusing. We could add documentation instead of removing them.

jinlee93:
> > Is there any value in surfacing how to override that token or do we not expect overrides to not be needed at all for these icons? Maybe just wait until a request?
> 
> @jinlee93 I don't think they'll be needed but they're not hurting anyone, just confusing. We could add documentation instead of removing them.

I'm down for just removing tbqh, thanks for answering

jinlee93 (approved):
Thanks for the cleanup, as always :D

dierat:
> > > Is there any value in surfacing how to override that token or do we not expect overrides to not be needed at all for these icons? Maybe just wait until a request?
> > 
> > 
> > @jinlee93 I don't think they'll be needed but they're not hurting anyone, just confusing. We could add documentation instead of removing them.
> 
> I'm down for just removing tbqh, thanks for answering

Okay cool, let's remove them and wait to see if we get a request for an easier way to override them 👍

PR #1381, opened 2022-11-10
Makes Accordion the stack of items and AccordionItem the button/panel

jinlee93:
### Summary:
- [mocks](https://www.figma.com/file/kCX4vzCtunnJSKQ1cwvksH/Accordions%3A-Creation?node-id=7888%3A29473)
- Motivation:
  - Accordion component should refer to the stack of items, rather than the individual item
  - There could be a common context for all items
  - Outline variant of the Accordion needs to be supported
- Implementation:
  - Previous `<Accordion>` component made into the stack and has the AccordionContext
    - Context now includes `headingAs` to assign the `as` for all AccordionTitle subcomponents
      - If `as` is provided to AccordionTitle, overrides the parent `headingAs`
  - `hasOutline` is also part of context to adjust spacing in subcomponents as needed
### Test Plan:
- more stories added for visual regression testing of `outline` and `compact outline`

Outline variants (other variants not screencapped since they were already displayed in the other accordion pr):

https://user-images.githubusercontent.com/86632227/201228203-475f8b48-eba4-4c35-ac70-6dd2eda60be0.mov

https://user-images.githubusercontent.com/86632227/201228206-8124626c-c26c-4bf4-b084-770b955e56b4.mov


src/components/Accordion/Accordion.tsx
```diff
@@ -16,76 +17,81 @@ type Props = {
...
 };
 
-export const AccordionContext = createContext<{ variant: Props['variant'] }>({
-  variant: undefined,
+export const AccordionContext = createContext<{
+  headingAs: HeadingElement;
+  hasOutline?: Props['hasOutline'];
+  variant?: Props['variant'];
+}>({
+  headingAs: 'h2',
```
jinlee93, line 38: @booc0mtaco dried the `as` prop for `AccordionTitle` as discussed

src/components/AccordionItem/AccordionItem.module.css
```diff
@@ -0,0 +1,19 @@
+/*------------------------------------*\
+    # ACCORDION ITEM
```
jinlee93, line 2: Name suggestions? Accordion Item v. Accordion Row v. Accordion Section?

dierat: I would prefer "section" or "row" over "item". For some reason "section" sounds right to me, but I'm also fine with "row".

src/components/AccordionItem/AccordionItem.module.css
```diff
@@ -0,0 +1,19 @@
...
+\*------------------------------------*/
+
+/**
+ * Accordion Item houses one Accordion Button subcomponent and respect Accordion Panel subcomponent.
```
dierat, line 6: I'm kinda confused by the wording here.
```suggestion
 * Accordion Item houses one Accordion Button subcomponent and its relevant Accordion Panel subcomponent.
```

src/components/Accordion/Accordion.tsx
```diff
@@ -16,76 +17,81 @@ type Props = {
...
-  classNameClosed?: string;
-  /**
-   * Whether panel is expanded by default.
-   */
-  defaultOpen?: boolean;
+  headingAs: HeadingElement;
   /**
    * Compact variant shrinks the Accordion size.
    */
   variant?: 'compact';
```
dierat, line 30: Btw if there's only one variant, does it make more sense to rename the prop to be specific to that variant? Ie `isCompact?: boolean`  vs `variant?: 'compact'`. Just putting the question out there; I'm not asking you to change it.

booc0mtaco: @dierat This is a good callout! We actually have two variants with one hidden. `compact` and non-`compact` which is the default if `variant` is missing. [Our guidelines](https://github.com/chanzuckerberg/edu-design-system/blob/main/docs/CODE_GUIDELINES.md#variants) don't make it super clear what to do in the case of default variants... seems to imply "when variant is missing, apply default style"?

In this case we may end up with a bigger variant at some point in the future, so I think having this prop allows for that without as much refactoring work. We might also consider applying a default/regular class when no specific variant is applied, which we might be doing elsewhere.

dierat: >We actually have two variants with one hidden.

Sorry, that's what I meant—one default and one alternate variant, which could easily be a boolean.

> In this case we may end up with a bigger variant at some point in the future

Then maybe we should call the "compact" size "sm" and the default "md" so it's easier to tack on a "lg" later on?

src/components/AccordionItem/AccordionItem.module.css
jinlee93: meant to say "its respective" 🤦‍♂️ but down for "its relevant"

dierat: Either's fine 👍

src/components/AccordionItem/AccordionItem.module.css
jinlee93: Will go with row since that term is also used in the mocks

src/components/Accordion/Accordion.tsx
jinlee93: I'm down to call the prop `size` and the options `sm` and `md`, aligns with the mocks better too

PR #1383, opened 2022-11-10
fix(readme): update workflow badge directory

jinlee93:
updates the workflow badge directory so it shows up when landing on our Github page

image


dierat (approved):
Thanks!

PR #1387, opened 2022-11-17
chore(tailwind): remove -@ in token bulding

jinlee93:
### Summary:
[EDS-736]
[EDS-259]
Relates to #1375, except instead of replacing `-@` with `-default`, assigns the appropriate `child-grandchild` key to the respective value.

**<i>Please let me know if I could word the helper utility function comment any better to make it clearer.</i>**

- The `colors-background.json` and other similar files are where we define our tier 2 / 3 tokens, looks like: 
image
- In this screenshot, `{ eds: {theme: {color: {background: {neutral: {default: {@: {value: "{eds.color.neutral.white}"` compiles to `EdsThemeColorBackgroundNeutralDefault` in `src/tokens-dist/ts/colors.ts` and css custom property `--eds-theme-color-background-neutral-default`. 
- However the formatter used to generate `css-variables-nested.json` (primarily used for TW) compiles to `{ eds: {theme: {color: {background: {neutral: {default: {@: {value: "var(--eds-theme-color-background-neutral-default)"` which when used with `@apply` or TW utility classes, becomes `-color-background-neutral-default-@`, but I would like that to be `-color-background-neutral-default` to be similar with what's in the mocks, and the tokens that are generated.

### Test Plan:
newly built tokens work with @apply or tailwind utility classnames in storybook, e.g. `bg-background-neutral-brand-primary-strong` is `--eds-theme-color-background-brand-primary-strong`

[EDS-736]: https://czi-tech.atlassian.net/browse/EDS-736?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EDS-259]: https://czi-tech.atlassian.net/browse/EDS-259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/style-dictionary.config.js
+++ b/style-dictionary.config.js
@@ -89,14 +89,38 @@ function minifyCSSVarDictionary(obj) {
   return toRet;
 }
 
+/**
+ * Tokens with key '@' mean the value of the parent is the value of '@', e.g.:
+ * {background: {neutral: {@: 'value' }}} is compiled to `background-neutral-default-@: 'value'`
+ * but we want this to look like `background-neutral: 'value'`.
+ *
+ * This helper function makes this happen by
+ * 1) Scanning great grandchildren for the key '@'
+ * 2) If such key exists, child and grandchild names are combined to make the new child key and value of the great grandchild '@' key is assigned to the new child key
+ * 2.5) The great grandchild '@' key/value pair is deleted for housekeeping.
+ * 3) If such key does not exist, but grand child is an object, recurses with the child to repeat this process.
+ */
+function formatEdsTokens(obj) {
+  for (const name in obj) {
+    if (typeof obj[name] === 'object') {
+      for (const nestedName in obj[name]) {
+        if (obj[name][nestedName]['@']) {
+          obj[name + '-' + nestedName] = obj[name][nestedName]['@'];
+          delete obj[name][nestedName]['@'];
+        } else if (typeof obj[name][nestedName] === 'object') {
+          formatEdsTokens(obj[name]);
+        }
+      }
+    }
+  }
+}
+
 EDSStyleDictionary.registerFormat({
   name: 'json/nested-css-variables',
   formatter: function (dictionary) {
-    return JSON.stringify(
-      minifyCSSVarDictionary(dictionary.properties),
-      null,
-      2,
-    );
+    const minifiedCssDictionary = minifyCSSVarDictionary(dictionary.properties);
+    formatEdsTokens(minifiedCssDictionary);
+    return JSON.stringify(minifiedCssDictionary, null, 2);
   },
 });
 
--- a/style-dictionary.config.js
+++ b/style-dictionary.config.js
@@ -90,9 +90,29 @@ function minifyCSSVarDictionary(obj) {
 }
 
 /**
- * Tokens with key '@' mean the value of the parent is the value of '@', e.g.:
- * {background: {neutral: {@: 'value' }}} is compiled to `background-neutral-default-@: 'value'`
+ * Tokens with key '@' are the base value of the parent, e.g.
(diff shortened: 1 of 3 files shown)


style-dictionary.config.js
```diff
@@ -89,14 +89,38 @@ function minifyCSSVarDictionary(obj) {
...
+ * Tokens with key '@' mean the value of the parent is the value of '@', e.g.:
+ * {background: {neutral: {@: 'value' }}} is compiled to `background-neutral-default-@: 'value'`
+ * but we want this to look like `background-neutral: 'value'`.
+ *
+ * This helper function makes this happen by
+ * 1) Scanning great grandchildren for the key '@'
+ * 2) If such key exists, child and grandchild names are combined to make the new child key and value of the great grandchild '@' key is assigned to the new child key
+ * 2.5) The great grandchild '@' key/value pair is deleted for housekeeping.
+ * 3) If such key does not exist, but grand child is an object, recurses with the child to repeat this process.
+ */
```
jinlee93, line 102: Please help clarify the comment, the function, or both

dierat:
>Relates to https://github.com/chanzuckerberg/edu-design-system/pull/1375, except instead of replacing -@ with -default, assigns the appropriate child-grandchild key to the respective value.

Does this PR replace https://github.com/chanzuckerberg/edu-design-system/pull/1375 and render it redundant?

style-dictionary.config.js
```diff
@@ -89,14 +89,38 @@ function minifyCSSVarDictionary(obj) {
   return toRet;
 }
 
+/**
+ * Tokens with key '@' mean the value of the parent is the value of '@', e.g.:
+ * {background: {neutral: {@: 'value' }}} is compiled to `background-neutral-default-@: 'value'`
+ * but we want this to look like `background-neutral: 'value'`.
```
dierat, line 95: This first block is the only part of this PR I think could be better. After seeing the change happening in `src/tokens-dist/json/css-variables-nested.json`, the `formatEdsTokens` code was pretty easy to understand. Nested loops are always a little confusing, but I think this one's fine. So my suggestion is to put some of that json from the changes in `src/tokens-dist/json/css-variables-nested.json` into the comment since I found that very helpful. This makes it a pretty long comment though, so I understand if you want to leave the example out or make it more compact. And my version could probably also be a lot easier to understand, but it's my best idea at the moment.

```suggestion
 * Tokens with key '@' mean the value of the parent is the value of '@', e.g.:
 * {background: {neutral: {@: 'value' }}} is compiled to `background-neutral-default-@: 'value'`
 * but we want this to look like `background-neutral: 'value'`.
```
```suggestion
 * Tokens with key '@' are the base value of the parent, e.g.
 * {background: {neutral: {@: 'value' }}} is compiled to `background-neutral-default-@: 'value'`,
 * but we want this to look like `background-neutral: 'value'`.
 * 
 * This function moves the '@' token out to be a sibling of the parent with the '@' part of
 * the name removed.
 *
 * Example:
 * "neutral": {
 *   "default": {
 *     "@": "var(--eds-theme-color-border-neutral-default)",
 *     "hover": "var(--eds-theme-color-border-neutral-default-hover)"
 *   },
 * },
 *
 * will be changed to
 * 
 * "neutral": {
 *   "default": {
 *     "hover": "var(--eds-theme-color-border-neutral-default-hover)"
 *   },
 * },
 * "neutral-default": "var(--eds-theme-color-border-neutral-default)",
```

dierat (approved):

style-dictionary.config.js
jinlee93: I'm down for adding the example for clarification, as there is value in showing what was happening and what we want out of the helper function

PR #1385, opened 2022-11-14
chore: add wireframe theme demo

dierat:
### Summary:
Ticket: https://czi-tech.atlassian.net/browse/EDS-710

This PR adds a wireframe theme demo, which has 2 main purposes:
- demonstrate how theming in EDS currently works and what it looks like
- set up the wireframe variables so they can be reused in prototypes for other projects

I only updated the variables that I needed to get this working, and I have a ticket to go through the rest of the variables and update them or remove them from the demo files.

There are 2 pages, taken from the Along product and devolved to a wireframe state.

#### Logged out homepage
wireframe theme demo logged out homepage

#### Watch page, top of the page
wireframe theme demo watch page, top of page

#### Watch page, bottom of the page
wireframe theme demo watch page, bottom of page

#### Theming documentation page
theming documentation page in storybook

### Test Plan:
Test the wireframe theme demo in storybook. There is no functionality beyond being able to navigate between the two pages; the other interactive components aren't wired up to any logic because it's only for visual demonstration purposes.

PR #1388, opened 2022-11-17
chore: add along themed demo

dierat:
### Summary:
Ticket: https://czi-tech.atlassian.net/browse/EDS-708

#### Overview
Follow-up to https://github.com/chanzuckerberg/edu-design-system/pull/1385 in which I added a demo of 2 pages using a wireframe theme that uses a different set of colors and fonts than the regular EDS styles. This PR adds a second demo that uses the same 2 pages but with different colors, fonts, and now it has images as well. These 2 demos together showcase what the code would look like for an alpha prototype and a mature product that has a full brand associated with it.

#### Implementation
It would be cool if we could just take the components from the wireframe demo and just give it different css variables, but because we have images in there and some styles can't be overridden with only css variables yet, I think we need to duplicate the code. Maybe we could split out sections into components that can be shared, but I'm not convinced it's worth the effort since I don't expect us to do any work in these after the theming project is complete.

#### Outstanding work
I left comments that start with `THEMING-NOTE:` in places where the system does not easily allow styling of a thing and I had to use kind of hacky methods to override the styling. We can talk through what the follow-up actions should be for them.

#### Logged out homepage
along demo logged out homepage in storybook

#### Watch page
along demo watch page in storybook, scrolled to the top of the page

### Test Plan:
Test the along demo in storybook, verifying that the logged out story shows the logged out homepage and the watch page story shows the watch page.


.storybook/pages/AlongDemo/AlongDemo.tsx
```diff
@@ -0,0 +1,479 @@
...
+
+const LoggedOutPage = ({ onLogin }: { onLogin: () => void }) => (
+  <div className={styles['logged-out-page']}>
+    <div className="flex flex-col items-center mb-16 p-8 grow">
+      <header className="flex gap-4">
+        <img
+          alt="placeholder for decorative illustration"
+          className={styles['logged-out-page__header-img']}
+          src={PlaceholderImage}
+        />
```
dierat, line 79: This is still using placeholder images because I actually don't know how to import a non-svg image in EDS..

.storybook/pages/AlongDemo/AlongDemo.tsx
```diff
@@ -0,0 +1,481 @@
...
+
+const LoggedOutPage = ({ onLogin }: { onLogin: () => void }) => (
+  <div className={styles['logged-out-page']}>
+    <div className="flex flex-col items-center mb-16 p-8 grow">
+      <header className="flex gap-4">
+        <img
+          alt="placeholder for decorative illustration"
+          className={styles['logged-out-page__header-img']}
+          // TODO: use along-user-illustration-1.png
+          src={PlaceholderImage}
```
dierat, line 79: This is still using placeholder images because I actually don't know how to import a non-svg image in EDS..

ahuth (approved):

.storybook/pages/AlongDemo/AlongDemo.tsx
ahuth: Since Storybook is using Webpack, I'd expect to be able to import it just like the svg - e.g. `import thing from './thing.png'`

That doesn't work?

ahuth: Ooooooohhhhhh makes sense.

I think we can resolve that like we did for .svg, .md, .css, etc -> https://github.com/chanzuckerberg/edu-design-system/blob/8fb00a105cb865d415077d7f0dae8a9fa0dc4607/src/custom.d.ts#L1-L4

dierat: Ahh that was it, thanks! 💖

dierat:
Thanks for reviewing!

PR #1389, opened 2022-11-18
feat(menu): add HeadlessUI-based menu component

booc0mtaco:
### Summary:

[EDS-739]: Implementation: Popover-like components

- Add in new atomic components for reuse (popover container, popover list item)
- Use Headless UI's dropdown to implement select and menu respectively
- Mark replaced components as deprecated where relevant
- Light refactoring to use popover container in existing popover component

### Test Plan:

- [x] Storybook
- [x] Unit tests (behavior)

[EDS-739]: https://czi-tech.atlassian.net/browse/EDS-739?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Dropdown/Dropdown.module.css
```diff
@@ -44,7 +44,7 @@
 .dropdown__options {
   @mixin eds-theme-typography-body-text-md;
   /* Allow Dropdown font to shrink in contexts where body font size is smaller. */
-  font-size: inherit;
```
booc0mtaco, line 47: file contains whitespace fixes picked up by prettier 🤷

booc0mtaco:
Not handled in this PR at the moment 

- using popperJS for advanced positioning options (to have parity with `Popover`)
- Implementing a new import of the current `Dropdown` as `Select`

There's also an a11y violation with the value for disabled text on `PopoverLineItem`, so will follow up with design on this one.

src/components/ButtonDropdown/ButtonDropdown.tsx
```diff
@@ -5,6 +5,7 @@ import styles from './ButtonDropdown.module.css';
 import { DropdownMenu } from '../..';
 import type { ClickableStyleProps } from '../ClickableStyle';
 
+// TODO: This component is deprecated and will be replaced by the Menu component
```
dierat, line 8: Can you put this deprecated info in the component docstring instead? You can look at the `Banner` component for an example of the current pattern. It has a message at the beginning and a jsdoc tag at the end.
- https://github.com/chanzuckerberg/edu-design-system/blob/main/src/components/Banner/Banner.tsx#L94
- https://github.com/chanzuckerberg/edu-design-system/blob/main/src/components/Banner/Banner.tsx#L113

src/components/ButtonDropdown/ButtonDropdown.stories.tsx
```diff
@@ -10,6 +11,7 @@ export default {
   title: 'Organisms/Interactive/ButtonDropdown',
   parameters: {
     layout: 'centered',
+    badges: [BADGE.DEPRECATED],
```
dierat, line 14: Thank you!

src/components/PopoverListItem/__snapshots__/PopoverListItem.test.ts.snap
```diff
@@ -0,0 +1,73 @@
...
+
+exports[`<PopoverListItem /> Disabled story renders snapshot 1`] = `
+<div
+  class="popover-list-item popover-list-item--disabled"
+>
+  <div
+    class="popover-list-item__no-icon"
+  />
+  Disabled
+</div>
```
dierat, line 47: This is the specific element that's triggering the a11y failure, and I think it's happening because this is a `div` no `aria-disabled` attribute. (It may also need a `role`; I don't know exactly how the test works.)

Disabled interactive elements are exempt from color contrast requirements in WCAG, so it makes sense that this story is failing the test since it's not marked up as disabled.

Source: https://webaim.org/articles/contrast/
>WCAG 2.0 defines four types of "incidental" text that are not required to meet the contrast requirements.
>
>- Inactive: An inactive element, like a disabled Submit button ( Submit ), is identified visually by its lower-contrast state.
>- Pure decoration: Decorative text that is not meant to be read. An example of this might be a picture of a bookshelf on a library homepage. The titles of the books are not meant to be read by the user.
>- Not visible to anyone: Text that is meant to be hidden, like an invisible [skip link](https://webaim.org/techniques/skipnav/) would not need to meet any contrast requirements until it becomes visible.
>- Part of a picture that contains significant other visual content: Text that is not an important part of the information in the image, like a name tag on the shirt of a person in a photo of a party, does not need to meet any contrast requirements.

src/components/Popover/Popover.test.tsx
```diff
@@ -17,6 +17,7 @@ describe('<Popover />', () => {
   });
 
   it('should open Popover with trigger button', () => {
+    // this test is redundant b/c of the snapshot above
```
dierat, line 20: That's fair. Should we remove it? Will this be done in a different PR? Is this a TODO comment?

src/components/Popover/Popover.stories.tsx
```diff
@@ -49,6 +49,7 @@ export const Default: StoryObj<Args> = {
 
 export const Arrow: StoryObj<Args> = {
   args: {
+    // maybe gets shallow merged with the default args (?)
```
dierat, line 52: I'm having trouble understanding the purpose of this comment. Was this a note to yourself?

src/components/Menu/__snapshots__/Menu.test.tsx.snap
```diff
@@ -0,0 +1,139 @@
...
+        class="popover-list-item"
+      >
+        <div
+          class="popover-list-item__no-icon"
+        />
+        Trigger Action
+      </div>
+    </a>
+    <div
+      aria-disabled="true"
```
dierat, line 113: This element did not trigger the a11y test because it has the expected markup.

src/components/Menu/Menu.module.css
```diff
@@ -0,0 +1,30 @@
...
+}
+
+.menu__popover {
+  max-width: var(--eds-l-max-width);
+  margin-top: var(--eds-size-1);
+  position: absolute;
+}
+
+.menu__button {
+  background-color: var(--light-color-button-background);
```
dierat, line 19: Where's this CSS variable from?

dierat:
Is this the figma file for this component? https://www.figma.com/file/Ff7hJoSvXX9hkAmIklnsT2/branch/4xGGJyfPCeoAvYX5Fjssdf/EDS-Paper-v1.1-%7C-UI-Components?node-id=7831%3A24711&t=F8tnXSnksp62mEAr-1

src/components/Menu/Menu.module.css
```diff
@@ -0,0 +1,30 @@
...
+}
+
+/* // is getting overridden by a style in clickablestyle */
+.menu__button--icon {
+  color: var(--eds-theme-color-button-secondary-neutral-text);
```
dierat, line 24: Let's use the token design is using.
```suggestion
  color: var(--eds-theme-color-icon-neutral-default);
```

src/components/Menu/Menu.module.css
dierat, line 19:
```suggestion
  background-color: var(--eds-theme-color-form-background);
```

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,119 @@
...
+};
+
+/**
+ * Trigger for the popover menu (dropdown)
+ */
+const MenuButton = ({ children, className, ...other }: MenuButtonProps) => {
+  const buttonClassNames = clsx(styles['menu__button'], className);
+  return (
+    <HeadlessMenu.Button
+      as={ClickableStyle}
```
dierat, line 51: Have you talked to design about the hover state we're getting from `ClickableStyle` and whether that's what they want?

dierat:
I dunno how annoying this is to change but, in playing with the menu in storybook, I see that the keyboard focus state for the items in the menu is the light background color change, but I don't think that's enough to indicate focus for keyboard users. The menu has the blue focus outline, but it's going to be hard for users with poorer vision to see which specific item in the list is focused.

In this example, the "MDN Menu" item is currently focused via the keyboard.
menu component in storybook, open with the MDN Menu list item focused

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,119 @@
...
+        className={styles['menu__button--icon']}
+        name="expand-more"
+        purpose="decorative"
+        size="1.5rem"
+      />
+    </HeadlessMenu.Button>
+  );
+};
+
+/**
```
dierat, line 67: @jinlee93 are we putting subcomponents in the same file/directory now or still splitting them up?

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,119 @@
...
+ * Individual menu items, styled with the popover list item component
+ */
+const MenuItem = ({
+  children,
+  href,
+  icon,
+  onClick,
+  ...other
+}: MenuItemProps) => {
+  const destinationUrl = !onClick ? href : undefined;
```
dierat, line 92: Can you explain this line to me? Why not just pass in `href`?

src/components/PopoverListItem/PopoverListItem.module.css
```diff
@@ -0,0 +1,32 @@
...
+  cursor: pointer;
+}
+
+.popover-list-item--active {
+  background-color: var(--eds-theme-color-background-neutral-default-hover);
+}
+
+.popover-list-item--disabled {
+  color: var(--eds-theme-color-text-disabled);
+  pointer-events: none;
```
dierat, line 22: Is this necessary? Looks like you're not passing in any interactive props into the item if it's disabled, so what pointer events are you removing? `cursor: not-allowed;` might be good though; that's what we use on disabled buttons.

src/components/PopoverListItem/PopoverListItem.module.css
```diff
@@ -0,0 +1,32 @@
...
+}
+
+.popover-list-item__icon {
+  padding-right: var(--eds-size-1-and-half);
+}
+
+/* right padding applies space for the icon and the padding for that icon container */
+.popover-list-item__no-icon {
+  padding-right: calc(var(--eds-size-3) + var(--eds-size-1-and-half));
```
dierat, line 31: Are you adding them together because there's no `4-and-a-half`? I would just pass it in in `rem`s.. We don't use the `size` tokens everywhere. (I'm still not convinced size tokens like these are useful.)

src/components/PopoverListItem/PopoverListItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+   */
+  icon?: IconName;
+}
+
+/**
+ * BETA: This component is still a work in progress and is subject to change.
+ *
+ * `import {PopoverListItem} from "@chanzuckerberg/eds";`
+ *
+ * This abstracts the structure of a item in a popover, when the popover contains a
```
dierat, line 34:
```suggestion
 * This abstracts the structure of an item in a popover, when the popover contains a
```

src/components/PopoverListItem/PopoverListItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+export const PopoverListItem = React.forwardRef<HTMLDivElement, Props>(
+  ({ active, className, disabled, children, icon, ...other }, ref) => {
+    const componentClassName = clsx(
+      styles['popover-list-item'],
+      active && styles['popover-list-item--active'],
+      disabled && styles['popover-list-item--disabled'],
+      className,
+    );
+
+    // use `as` to use a/button when appropriate
```
dierat, line 51: What's this comment referring to? Is it a TODO comment?

src/components/Popover/Popover.test.tsx
booc0mtaco: yah left a note for myself that this should be removed and left in PR to confirm. Ticket to clean up is https://czi-tech.atlassian.net/browse/EDS-753 and I will trim the comment now :)

src/components/Popover/Popover.stories.tsx
booc0mtaco: Yah, it's mostly a note that can be removed. @jinlee93 and I were pairing on this and he wasn't sure how the object gets used when consumed will the `args` merge or does the default one get overwritten?

src/components/Menu/Menu.module.css
booc0mtaco: that's a GOOD question. I don't know why but some JS files with a bunch of variables popped up for me and I chose one from there. 😕 

The correct one (matches the variable in the figma) is `--eds-theme-color-form-background`

booc0mtaco:
> Is this the figma file for this component? https://www.figma.com/file/Ff7hJoSvXX9hkAmIklnsT2/branch/4xGGJyfPCeoAvYX5Fjssdf/EDS-Paper-v1.1-%7C-UI-Components?node-id=7831%3A24711&t=F8tnXSnksp62mEAr-1

That's the one. Sean is out this week but I think we will have a few more concrete questions for him once he's back.

src/components/Menu/Menu.module.css
```diff
@@ -0,0 +1,30 @@
...
+  margin-top: var(--eds-size-1);
+  position: absolute;
+}
+
+.menu__button {
+  background-color: var(--light-color-button-background);
+}
+
+/* // is getting overridden by a style in clickablestyle */
+.menu__button--icon {
```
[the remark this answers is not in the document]
booc0mtaco: That's annoying. does it only apply to the things which are using clickable style (since it has `>` rules)? anything else forcing this pattern?

With that in mind, and for this case, it actually appears to need ***three*** of these since the descendant selector 
 and `svg` breaks the tie on the two classes being used on `ClickableStyle` :( [0,3,0] vs [0,2,1] I'll do for now, but worth discussing more (separately)

src/components/Menu/Menu.tsx
booc0mtaco: Nah, that is a good call. If you saw the design file, there are some cases and behaviors missing. I am going to add this to the list. Another missing thing is the disabled states (which I left a comment about earlier)

src/components/Menu/Menu.tsx
jinlee93: I actually wanted to ask you about this.
I am down to start putting subcomponents in the same file/directory as a lot of these subcomponents we really wouldn't be reusing.
However, the file does get really long, stuff could be harder to find, and css class names could get even more complex
@booc0mtaco @dierat

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,119 @@
...
+  /**
+   * Allow custom classes to be applied to the menu container.
+   */
+  className?: string;
+};
+
+export type MenuItemProps = ExtractProps<typeof HeadlessMenu.Item> & {
+  href?: string;
+  icon?: IconName;
+  onClick?: () => any;
```
jinlee93, line 25:
```suggestion
  onClick?: MouseEventHandler<HTMLAnchorElement>;
```
would this work?

src/components/Dropdown/Dropdown.module.css
```diff
@@ -56,7 +56,7 @@
   border-radius: var(--eds-border-radius-md);
   box-shadow: var(--eds-box-shadow-md);
   /* Ensures the options display over the underlying body elements. */
-  z-index: var(--eds-z-index-100); 
+  z-index: var(--eds-z-index-100);
```
jinlee93, line 59: Just an FYI bootstrap has its own set of z-indexes that may clash i.e. [bootstrap 3 in LP](https://github.com/twbs/bootstrap/blob/v3.4.1/less/variables.less#L274-L280) .

Haven't come across that many difficult issues regarding yet

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,119 @@
...
+ * A set of menu items in the Menu
+ *
+ * @param props Props used on the set of menu items
+ * @see https://headlessui.com/react/menu#menu-items
+ */
+const MenuItems = (props: MenuItemsProps) => (
+  <HeadlessMenu.Items
+    as={PopoverContainer}
+    className={clsx(styles['menu__popover'])}
+    data-testid="menu-content"
```
jinlee93, line 77: Can we place this `data-testid` attribute on line 27 of `Menu.stories.tsx`?
This way the testid attribute is not publicly exposed (although would that matter?)

src/components/Popover/Popover.stories.tsx
jinlee93: I believe the `args` are shallow merged, tested via 
image
Which results in `className: 'haw'` and `placement:'top-start'`

src/components/Menu/Menu.tsx
booc0mtaco: My 2¢ on this one is that "bound" sub-components should live in the same folder together. Similar to how headless bundles the parts together. A rule I used in the past was "components get their own folder IFF they can be composed independent of other components".

From what I understand, BM encouraged breaking things down into separate folders, but I think that might be why some of the naming for these popover/dropdowns got wonky...? I chopped some of the reusable bits (`PopoverContainer`, `PopoverListItem`) into separate things and used the common-denominator component as a basis for the name

src/components/PopoverContainer/PopoverContainer.tsx
```diff
@@ -0,0 +1,39 @@
...
+  className?: string;
+  children: React.ReactNode;
+}
+
+// Default settings for any popover container using PopperJS
+export const defaultPopoverModifiers = [
+  {
+    name: 'offset',
+    options: {
+      offset: [0, 10], // spaces the popover from the trigger element
```
jinlee93, line 18: I think the mocks have an offset of 8px

src/components/Menu/Menu.tsx
booc0mtaco: Sure, I added this as a behavior preference . if an onClick exists, it will remove the value set for the href so that it doesn't try and do both onClick handling and URL navigation

src/components/PopoverListItem/PopoverListItem.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+  parameters: {
+    badges: [BADGE.BETA],
+  },
+} as Meta<Args>;
+
+type Args = React.ComponentProps<typeof PopoverListItem>;
+
+export const Default: StoryObj<Args> = {
+  args: {
+    children: <>Default list item</>,
```
jinlee93, line 19: nit:
```suggestion
    children: 'Default list item'
```

src/components/PopoverListItem/PopoverListItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+import Icon from '../Icon';
+import type { IconName } from '../Icon';
+
+export interface Props {
+  /**
+   * Whether the list item is treated as highlighted in its container
+   */
+  active?: boolean;
+  children: ReactNode;
```
jinlee93, line 14:
```suggestion
  /**
   * Child node(s) that can be nested inside component
   */
  children: ReactNode;
```

src/components/PopoverListItem/PopoverListItem.module.css
booc0mtaco: good catch. an earlier version of this included an anchor which did have behaviors, and this prevented bubbling up to that to trigger actions.

The cursor is undefined from design side, but I can add that for consistency, and remove this `pointer-events` property.

src/components/PopoverListItem/PopoverListItem.module.css
booc0mtaco: I use the two values to show where the quantity comes from. Just above (L26) we use 1.5 for padding, and here the icon size happens to be 3. Lots of components are using various sizes so didn't think they were not part of the existing patterns.

This is fairly complex too. The sizes have a relationship to one another, based on some base size and multipliers ... having tokens for each one means defining different sizes is more work than it has to be IMO. Something to discuss (moving to defining a base grid value and using `calc()` or similar to get other quantities)

src/components/PopoverListItem/PopoverListItem.tsx
booc0mtaco: I can't remember. it was a reminder tho, so chucking it

src/components/Menu/Menu.tsx
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * BETA: This component is still a work in progress and is subject to change.
+ *
+ * `import {Menu} from "@chanzuckerberg/eds";`
+ *
+ * Creates a list of actions which appear next to a trigger component. This should
+ * be used when there is a discrete number of actions such as user navigations (e.g.,
+ * a profile menu with links to settings) or a set of actions
+ */
+export const Menu = (props: MenuProps) => {
```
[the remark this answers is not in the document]
booc0mtaco: In our xfn meeting we landed on using Menu/Select for the two components to avoid ambiguity. All three of these things sorta drop down, and we already have other (deprecated) components using "Dropdown" in the name, so trying to avoid that in the component itself.

The figma currently has conflicting naming too, so will take a note for that to get tidied up

src/components/Dropdown/Dropdown.module.css
booc0mtaco: I think this file just had whitespace changes in it, so the lines changed should just be that.

src/components/Menu/Menu.tsx
booc0mtaco: I have worked in libs where such testIDs got stripped out on publish, i think? I can't remember. Do we have something similar here?

I will move that tho, so we don't have to worry about it

src/components/PopoverContainer/PopoverContainer.tsx
booc0mtaco: This data is a little weird right now. I am not using these values in this PR, but moved this from another file, so all popover containers would have their settings in one place. This does reveal a discrepancy between popovers tho... once they all implement popperJS, they should all use **this** config (this will also prevent circular dependencies) and we would change the values to 8 if that's correct for everyone.

src/components/PopoverContent/PopoverContent.tsx
```diff
@@ -88,7 +84,7 @@ export const PopoverContent = ({
       as="article"
       className={componentClassName}
     >
-      <div className={bodyComponentClassName}>{children}</div>
+      <PopoverContainer className={bodyClassName}>{children}</PopoverContainer>
       {showArrow && (
```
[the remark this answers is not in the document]
booc0mtaco: haha, I will add it to the list. I think for menu / select we don't want arrow according to [figma](https://www.figma.com/file/Ff7hJoSvXX9hkAmIklnsT2/branch/4xGGJyfPCeoAvYX5Fjssdf/EDS-Paper-v1.1-%7C-UI-Components?node-id=7831%3A24711&t=WIALim2FJJoByEXT-0) but it is undefined at the moment which popover-like things should support arrows

src/components/PopoverListItem/PopoverListItem.stories.tsx
booc0mtaco: ah i didn't know it would allow strings like that; will update

booc0mtaco:
> I dunno how annoying this is to change but, in playing with the menu in storybook, I see that the keyboard focus state for the items in the menu is the light background color change, but I don't think that's enough to indicate focus for keyboard users. The menu has the blue focus outline, but it's going to be hard for users with poorer vision to see which specific item in the list is focused.
> 
> In this example, the "MDN Menu" item is currently focused via the keyboard. menu component in storybook, open with the MDN Menu list item focused

@dierat Gonna add this to the list, as I did follow up with sean if we should have focus ring and it sounded like that wasn't needed. it

src/components/PopoverContainer/PopoverContainer.tsx
jinlee93: Makes sense, thanks

src/components/Menu/Menu.tsx
jinlee93: Good question, not sure if test ids are being stripped out but would be good to have

src/components/Menu/Menu.tsx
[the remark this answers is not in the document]
dierat: Thanks!

src/components/Menu/Menu.tsx
dierat: >A rule I used in the past was "components get their own folder IFF they can be composed independent of other components".

This sounds right to me. I think it's confusing to have a component in its own folder if it can only be used in conjunction with another specific component—they should be bundled together. I also find it weird that there are directories with no `*.stories.ts/x` file (because subcomponents aren't exposed in storybook).

>However, the file does get really long, stuff could be harder to find, and css class names could get even more complex

I can also see splitting them up into separate files to make it cleaner, but they can still live in the same directory.

src/components/Menu/Menu.tsx
dierat: Cool! Could you leave a code comment to that effect here?

src/components/PopoverListItem/PopoverListItem.module.css
dierat: >I use the two values to show where the quantity comes from. Just above (L26) we use 1.5 for padding, and here the icon size happens to be 3.

Ah, cool. Could you move the comment from line 29 down one step so it's above this styling?

src/components/Menu/Menu.module.css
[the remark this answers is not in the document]
dierat: >does it only apply to the things which are using clickable style (since it has > rules)?

That's the one I've noticed coming up recently. Maybe we should look into lower the specificity in `ClickableStyle` to make these overrides easier. Though I don't necessarily want it to be easy to override them because I would prefer it if buttons were consistent wherever possible.. 🤔

src/components/Menu/Menu.tsx
dierat: But if we decide to move away from the one-component-per-directory rule, we should update the [code guidelines](https://chanzuckerberg.github.io/edu-design-system/iframe.html?viewMode=story&id=documentation-guidelines--code-guidelines&args=#component-rules).

booc0mtaco: @dierat @jinlee93 yah, I don't have too strong an opinion about how we do the internals of the component, as some cases may benefit from separate subcomponent files more than others. I will turn the guidelines update into a ticket, and we can have a brief discussion at one of our upcoming meetings (or async via comments on that one). How does that sound?

src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -18,6 +18,7 @@ import {
   TAB_KEYCODE,
 } from '../../util/keycodes';
 
+// This component is deprecated and will be replaced by the Menu component
```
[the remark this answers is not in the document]
booc0mtaco: sounds good. also another supporting fact for grouping things together since marking subcomponents as deprecated gets complicated if they are all separated out.

src/components/Menu/Menu.tsx
dierat: Sounds good, thanks!

booc0mtaco:
* Added https://czi-tech.atlassian.net/browse/EDS-754 to handle adding the `Select` export from EDS
* Added https://czi-tech.atlassian.net/browse/EDS-755 to handle doing popperJS (re-)factoring

dierat:
> > I dunno how annoying this is to change but, in playing with the menu in storybook, I see that the keyboard focus state for the items in the menu is the light background color change, but I don't think that's enough to indicate focus for keyboard users. The menu has the blue focus outline, but it's going to be hard for users with poorer vision to see which specific item in the list is focused.
> > In this example, the "MDN Menu" item is currently focused via the keyboard. menu component in storybook, open with the MDN Menu list item focused
> 
> @dierat Gonna add this to the list, as I did follow up with sean if we should have focus ring and it sounded like that wasn't needed. it

We need it. This is just going to come back as an accessibility issue via Lodestone when someone uses it in the product if we don't add a high contrast focus ring to the individual list items for keyboard users.

jinlee93 (approved):
Thanks for all the Dropdown work and clarification!

booc0mtaco:
Added final ticket to track some questions that came up here https://czi-tech.atlassian.net/browse/EDS-759

PR #1390, opened 2022-11-18
fix(banner, page-level-banner): use aside instead of article

dierat:
### Summary:
Ticket: https://czi-tech.atlassian.net/browse/EDS-110

We're currently using the [article](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article) element tag for the `Banner` and `PageLevelBanner` components, but, after some discussion, we decided that [aside](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside) is a better semantic choice here.

### Test Plan:
Snapshots for `Banner` and `PageLevelBanner` show the components are now using `aside` instead of `article`.

### DIFF
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -145,7 +145,7 @@ export const Banner = ({
   const variantComputed = variant === 'neutral' ? 'neutral-strong' : variant;
 
   return (
-    <article className={componentClassName}>
+    <aside className={componentClassName}>
       <Icon
         className={styles['banner__icon']}
         name={variantToIconAssetsMap[variant].name}
@@ -186,7 +186,7 @@ export const Banner = ({
           />
         </Button>
       )}
-    </article>
+    </aside>
   );
 };
 Banner.displayName = 'Banner';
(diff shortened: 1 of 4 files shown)


jinlee93 (approved):
Makes sense to me.

On a side note, when are we planning on dropping support for the `<Banner>` component since it's been listed as deprecated?

dierat:
> when are we planning on dropping support for the `<Banner>` component since it's been listed as deprecated?

@jinlee93 do you mean not updating it like I am in this PR? I think if we didn't have the `PageLevelBanner`, I wouldn't have bothered updating it, but since it's easy to update both of them at the same time, I figured I might as well. Then again, it's used like 43 times in LP so even if it's deprecated, maybe it would still be worth these kinds of accessibility improvements even if we don't want it used in new places.

jinlee93:
> > when are we planning on dropping support for the `<Banner>` component since it's been listed as deprecated?
> 
> @jinlee93 do you mean not updating it like I am in this PR? I think if we didn't have the `PageLevelBanner`, I wouldn't have bothered updating it, but since it's easy to update both of them at the same time, I figured I might as well. Then again, it's used like 43 times in LP so even if it's deprecated, maybe it would still be worth these kinds of accessibility improvements even if we don't want it used in new places.

Makes sense to me, thanks for the explanation

PR #1391, opened 2022-11-18
chore: use css variables instead of js variables for color in icons

dierat:
### Summary:
Ticket: https://czi-tech.atlassian.net/browse/EDS-705

In order to support theming, we only want to use CSS variables, not JavaScript variables, for styling components. This will allow other downstream products to override the variables we use for styling. The `Icon` component is typically colored by passing in a JS variable via the `color` prop. This is fine in other products, and it's fine in storybook files, but we don't want to use this convention in components themselves.

This PR updates components with colored `Icon`s to use css variables for the color, and I updated the documentation that says to do the opposite.

There is also one icon color change—the `DrawerHeader` was overriding the close button icon color to be lighter. I checked the [figma design for the Filter component](https://www.figma.com/file/Ff7hJoSvXX9hkAmIklnsT2/EDS-Paper-v1.1-%7C-UI-Components?node-id=7452%3A24882&t=509XHeTp0Hscikjj-1), and it's actually using the default icon button color. I would prefer we not override default button styling unless we have a really good reason, so I'm removing the color to match figma.

### Test Plan:
- no visual changes to the `PrimaryNav` and `TimelineNav` components
- the `DrawerHeader` close button icon is a little darker and now matches the figma mock
- the code guidelines have been updated

### DIFF
--- a/docs/CODE_GUIDELINES.md
+++ b/docs/CODE_GUIDELINES.md
@@ -276,7 +276,17 @@ Use:
 }
 ```
 
-- Avoid using classes on one component to alter the styling of a different component. For example, instead of:
+- Use CSS variables instead of JavaScript variables for styling in components. This enables theming the components because our theming system relies on overrides CSS variables.
+
+Instead of:
+
+```tsx
+import { EdsThemeColorUtilitySuccessForeground } from 'src/tokens-dist/ts/colors';
+
+<Icon color={EdsThemeColorUtilitySuccessForeground} />;
+```
+
+Use:
 
 ```css
 .banner__icon {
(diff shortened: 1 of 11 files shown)


jinlee93:
What will happen to the `color` prop for `<Icon>`?

dierat:
> What will happen to the `color` prop for `<Icon>`?

@jinlee93 I was thinking we'd leave it because this is only an issue for components exported from EDS to be used and themed in other products. Those other products can continue passing JS variables into `Icon`.

jinlee93 (approved):
> @jinlee93 I was thinking we'd leave it because this is only an issue for components exported from EDS to be used and themed in other products. Those other products can continue passing JS variables into Icon.

Makes sense to me, thanks for the updates!

dierat:
Thanks for reviewing!

PR #1394, opened 2022-11-19
chore(modal-header): use tier 3 token for brand header background color

dierat:
### Summary:
Ticket: https://czi-tech.atlassian.net/browse/EDS-704

We only want tier 2 and tier 3 tokens to be used in components so that downstream products using the components can easily override the token values using css variables. We have just 1 left I believe, which is in the `ModalHeader` brand background color. We synced with design today and decided a new tier 3 token would be added, and it would be called "color modal brand header background". This PR adds the new token and uses it in the `ModalHeader` `brand` variant.

### Test Plan:
No visual changes in the brand modal.

### DIFF
--- a/src/design-tokens/tier-3-component/modals.json
+++ b/src/design-tokens/tier-3-component/modals.json
@@ -1,0 +1,19 @@
+{
+  "eds": {
+    "theme": {
+      "color": {
+        "modal": {
+          "brand": {
+            "header": {
+              "background": {
+                "@": {
+                  "value": "{eds.color.brand.grape.500}"
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+}
--- a/src/tokens-dist/json/css-variables-nested.json
+++ b/src/tokens-dist/json/css-variables-nested.json
@@ -645,6 +645,14 @@
           },
           "label": "var(--eds-theme-color-form-label)"
         },
+        "modal": {
+          "brand": {
+            "header": {
+              "background": {}
+            },
+            "header-background": "var(--eds-theme-color-modal-brand-header-background)"
+          }
+        },
         "text-highlight": {
           "foreground": "var(--eds-theme-color-text-highlight-foreground)",
           "background": "var(--eds-theme-color-text-highlight-background)"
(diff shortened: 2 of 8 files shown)


src/design-tokens/tier-3-component/modals.json
```diff
@@ -0,0 +1,19 @@
...
+                  "value": "{eds.color.brand.grape.500}"
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+}
```
booc0mtaco, line 19: it's beautiful :)

booc0mtaco (approved):
it's possible there should also be a token for the text color/size for the branded modal header too?

src/tokens-dist/json/css-variables-nested.json
```diff
@@ -645,6 +645,14 @@
           },
           "label": "var(--eds-theme-color-form-label)"
         },
+        "modal": {
+          "brand": {
+            "header": {
+              "background": {}
```
jinlee93, line 651: Oof created ticket to fix this https://czi-tech.atlassian.net/browse/EDS-743?atlOrigin=eyJpIjoiNDE5MGExMjMyYjNhNGE3NGEwMjNmYjcwNDQ3MjVhN2IiLCJwIjoiaiJ9

dierat:
> it's possible there should also be a token for the text color/size for the branded modal header too?

@booc0mtaco how about I make a ticket to discuss with design so we keep in sync?

src/tokens-dist/json/css-variables-nested.json
dierat: Thanks!

booc0mtaco:
> > it's possible there should also be a token for the text color/size for the branded modal header too?
> 
> @booc0mtaco how about I make a ticket to discuss with design so we keep in sync?

Sounds good 👍

src/tokens-dist/json/css-variables-nested.json
jinlee93: @dierat I think we can continue with merging since I don't think not having this token in TW would be that blocking

dierat: Awesome, thanks!

dierat:
Thanks for reviewing!

PR #1395, opened 2022-11-21
chore(wireframe-demo): update remaining color variables

dierat:
### Summary:
Ticket: https://czi-tech.atlassian.net/browse/EDS-734

In https://github.com/chanzuckerberg/edu-design-system/pull/1385 I added a wireframe theme demo, which overrides a bunch of eds tokens with tokens specifically made for wireframe prototypes. I only updated the EDS tokens I needed in order to make the demo work and therefore didn't touch tokens that were not present in the demo (and I also skipped over some tokens that were already gray because it didn't make a difference).

But the wireframe theme will, we think, be used in prototypes that may use more components than the ones I did, so, just to be safe, I'm overriding all the color tokens with wireframe tokens. Now this set of tokens can be copy/pasted into a different app using EDS to apply the wireframe theme and it should just work. I also removed all the tokens I don't expect a wireframe to need to override, like border widths, z-indices, and legacy colors (which aren't used in EDS). I also sorted the tokens by token name. The order had become kind of messy because I was just pulling out the tokens when I encountered them and dumped them in rough groups.

There are a lot of tokens here, and not all of the components are represented in the [wireframe kit in figma](https://www.figma.com/file/eirEfrTnkc2jIqQzK5AVpN/Wireframe-Kit-1.0?node-id=0%3A1&t=FhuSzkEcO0UjSs6k-1), so, to make things a little faster, I converted them based on their tier 1 token level. If a themed token used eds tier 1 token level 100 or 200 (e.g. `--eds-color-other-ruby-100`, `--eds-color-brand-grape-200`, etc), I set it to use `--wireframe-gray-a`. The full conversation looked like this:
- 100 or 200 -> wireframe gray a
- 300 or 400 -> wireframe gray b
- 500 or 600 -> wireframe gray c
- 700 -> wireframe gray d
- 800 -> wireframe gray e

I based the conversation table on the patterns I had already seen components in the wireframe kit. (There is one exception here which is buttons—I made them consistent across the different "status" / hue options even though in EDS they use different levels because some hues appear darker or lighter than others.)

### Test Plan:
There should be no visual changes in the wireframe demo in storybook (except for the form and sidebar border, which is now darker and closer to the form border color in EDS). I used chromatic to verify this.

I did not test components that are not already in the demo (except buttons because I wanted to verify that they're consistent).

### DIFF
--- a/.storybook/pages/WireframeDemo/GlobalStyles.module.css
+++ b/.storybook/pages/WireframeDemo/GlobalStyles.module.css
@@ -11,6 +11,9 @@
  */
 .root {
   /* Level 1 wireframe theme tokens */
+  --wireframe-anim-duration: 0;
+  --wireframe-font-family-primary: 'Balsamiq Sans', sans-serif;
+  --wireframe-font-family-secondary: 'Redacted Script', sans-serif;
   --wireframe-white: #ffffff;
   --wireframe-black: #000000;
   --wireframe-gray-a: #f4f6f8;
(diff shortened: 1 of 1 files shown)


.storybook/pages/WireframeDemo/GlobalStyles.module.css
```diff
@@ -11,6 +11,9 @@
  */
 .root {
   /* Level 1 wireframe theme tokens */
+  --wireframe-anim-duration: 0;
+  --wireframe-font-family-primary: 'Balsamiq Sans', sans-serif;
+  --wireframe-font-family-secondary: 'Redacted Script', sans-serif;
```
dierat, line 16: I added these so all the overrides are using a wireframe token or `transparent` (which I guess I could add a token for if we wanted but eh).

jinlee93 (approved):
Looks good to me! And I agree that the sort does make it easier to look at.

> Now this set of tokens can be copy/pasted into a different app using EDS to apply the wireframe theme and it should just work.
Nice, love this idea

PR #1396, opened 2022-11-21
chore(icons): use tokens for baked-in fill colors

dierat:
### Summary:
Ticket: https://czi-tech.atlassian.net/browse/EDS-745

We have a few icons that have colors baked in. This is very rare because consumers should usually be able to override colors, but these are special cases where we've decided hard-coded fill colors makes sense.

But some of them are not using tokens, either using "white" or a hex code. They should at least be using css variable color tokens so downstream products can override those colors based on the product's theme. In this PR, I found all the hard-coded icon colors that are not using css variable tokens and updated them.

After this change, there should only be `fill` attributes with tokens or the word `none` in the spritemap.

### Test Plan:
There should be no visible changes in the icon grid story, which I verified using chromatic.


jinlee93 (approved):
Agree with the use of tokens here, thanks!

dierat:
Thanks for reviewing! ❤️

PR #1397, opened 2022-11-21
chore(along-demo): tidy up css variables

dierat:
### Summary:
Part of https://czi-tech.atlassian.net/browse/EDS-741

Couple of small updates for the Along theming demo CSS variables:
- fix copy/paste error where it says "WIREFRAME" at the top
- update comment above the EDS overrides to match the one in the [wireframe demo](https://github.com/chanzuckerberg/edu-design-system/pull/1395/files)
- remove the comment separating tier 2 and tier 3 token overrides
- sort the variables

### Test Plan:
No visual changes picked up by chromatic.

### DIFF
--- a/.storybook/pages/AlongDemo/GlobalStyles.module.css
+++ b/.storybook/pages/AlongDemo/GlobalStyles.module.css
@@ -1,7 +1,7 @@
 @import '../../../src/design-tokens/mixins.css';
 
 /*------------------------------------*\
-    # GLOBAL STYLES FOR WIREFRAME DEMO
+    # GLOBAL STYLES FOR ALONG DEMO
 \*------------------------------------*/
 
 /**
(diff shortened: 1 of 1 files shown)


.storybook/pages/AlongDemo/GlobalStyles.module.css
```diff
@@ -62,75 +62,71 @@
...
-  --eds-theme-color-text-link-brand: var(--along-color-space-500);
-  --eds-theme-color-text-link-neutral: var(--along-color-space-500);
-  --eds-theme-color-background-brand-primary-default: var(--along-color-white);
-  --eds-theme-color-border-brand-primary-strong: var(--along-color-blurple-500);
-  --eds-theme-color-form-border: var(--along-color-sneaker-300);
-  --eds-theme-color-icon-brand-primary: var(--along-color-blurple-500);
-  --eds-theme-color-background-neutral-medium: var(--along-color-sneaker-100);
   --eds-theme-color-border-neutral-default: var(--along-color-space-200);
-
-  /* Level 3 tokens */
```
dierat, line 91: The removal of these 2 lines is the only change in the EDS variable override section aside from sorting.

jinlee93 (approved):
Nice. Is a codemod/utility/plop possible to generate styles like this?

dierat:
> Nice. Is a codemod/utility/plop possible to generate styles like this?

@jinlee93 mm interesting idea. We've talked about putting them in the "create react app" template so if people are spinning up new prototypes with the wireframe theme, it would already be there for them to get started with.

dierat:
> > Nice. Is a codemod/utility/plop possible to generate styles like this?
> 
> @jinlee93 mm interesting idea. We've talked about putting them in the "create react app" template so if people are spinning up new prototypes with the wireframe theme, it would already be there for them to get started with.

Clarification: we're talking about putting the *wireframe* theme variables in the create react app template. At the moment I'm thinking that if/when they get to the brand stage, they would update the tier 1 tokens to use their own system and then update the eds variable overrides to point to those new tier 1 tokens. I wonder if there's something we can do to make that transition easier. But, then again, that's probably pretty far off in the future. We first need an idea to be built up as a prototype and then prove via user testing that it's worth investing the time and money into giving it its own brand.

PR #1405, opened 2022-11-29
fix(Menu): sync width of trigger and popover

booc0mtaco:
### Summary:

In [our designs](https://www.figma.com/file/Ff7hJoSvXX9hkAmIklnsT2/branch/4xGGJyfPCeoAvYX5Fjssdf/EDS-Paper-v1.1-%7C-UI-Components?node-id=7831%3A24711&t=tafQ4kj7AkliGAew-0), we have the popover content for menus match the width of the trigger, but the initial implementation didn't follow this pattern. use a non-fragment react node to allow width to be determined by DOM structure

### Test Plan:

- [x] update snapshots
- [x] view in storybook

### DIFF
--- a/src/components/Menu/Menu.module.css
+++ b/src/components/Menu/Menu.module.css
@@ -10,9 +10,10 @@
 }
 
 .menu__popover {
-  max-width: var(--eds-l-max-width);
   margin-top: var(--eds-size-1);
+  max-width: var(--eds-l-max-width);
   position: absolute;
+  width: 100%;
 }
 
 .menu__button {
--- a/src/components/Menu/Menu.module.css
+++ b/src/components/Menu/Menu.module.css
@@ -11,17 +11,18 @@
 
 .menu__popover {
   margin-top: var(--eds-size-1);
+  min-width: max-content;
   max-width: var(--eds-l-max-width);
   position: absolute;
   width: 100%;
 }
 
-.menu__button {
+.menu__button.menu__button {
   background-color: var(--eds-theme-color-form-background);
 }
 
 /* Needed to create higher specificity than ClickableStyle. TODO: improve */
-.menu__button--icon.menu__button--icon.menu__button--icon {
+.menu__button--icon.menu__button--icon.menu__button--icon.menu__button--icon {
   color: var(--eds-theme-color-icon-neutral-default);
 }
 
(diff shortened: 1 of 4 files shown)


booc0mtaco:
Screen Shot 2022-11-29 at 16 45 02

booc0mtaco:
One thing left is to set a minimum width in cases where the trigger not full width:

- in design, the width of container when the trigger is not full width is 375px
- otherwise, widths of the trigger and popover are the same

src/components/Menu/Menu.module.css
```diff
@@ -10,17 +10,19 @@
...
 }
 
-.menu__button {
+.menu__button.menu__button {
   background-color: var(--eds-theme-color-form-background);
 }
 
 /* Needed to create higher specificity than ClickableStyle. TODO: improve */
-.menu__button--icon.menu__button--icon.menu__button--icon {
+.menu__button--icon.menu__button--icon.menu__button--icon.menu__button--icon {
```
jinlee93, line 25: oof

src/components/Menu/Menu.tsx
```diff
@@ -39,7 +39,8 @@ export type MenuItemsProps = ExtractProps<typeof HeadlessMenu.Items>;
  * a profile menu with links to settings) or a set of actions
  */
 export const Menu = (props: MenuProps) => {
-  return <HeadlessMenu {...props} />;
+  const menuClassNames = clsx(props.className, styles['menu']);
+  return <HeadlessMenu as="div" className={menuClassNames} {...props} />;
```
jinlee93, line 43: Since you haven't extracted `className` out of `props`, would it be applied again with `{...props}`?

src/components/Menu/Menu.module.css
booc0mtaco: Oof indeed. I think we need to revisit how this clickablestyle thing works, as it creates a lot of specificity challenges :S

src/components/Menu/Menu.tsx
booc0mtaco: Good call, it would. i can extract that one out to prevent doubling

jinlee93 (approved):
Looks good to me!

booc0mtaco:
Screen Shot 2022-11-30 at 13 22 35

PR #1403, opened 2022-11-29
chore(deps): update commitlint monorepo to v17 (major)

czi-frontend-infra:
This PR contains the following updates:

| Package | Type | Change |
|---|---|---|
| [@commitlint/cli](https://commitlint.js.org/) ([source](https://togithub.com/conventional-changelog/commitlint)) | devDependencies | [`^16.2.1` -> `^17.0.0`](https://renovatebot.com/diffs/npm/@commitlint%2fcli/16.2.4/17.3.0) |
| [@commitlint/config-conventional](https://commitlint.js.org/) ([source](https://togithub.com/conventional-changelog/commitlint)) | devDependencies | [`^16.2.1` -> `^17.0.0`](https://renovatebot.com/diffs/npm/@commitlint%2fconfig-conventional/16.2.4/17.3.0) |
| [@commitlint/prompt-cli](https://commitlint.js.org/) ([source](https://togithub.com/conventional-changelog/commitlint)) | devDependencies | [`^16.2.1` -> `^17.0.0`](https://renovatebot.com/diffs/npm/@commitlint%2fprompt-cli/16.2.4/17.3.0) |

---

### Release Notes

<details>
<summary>conventional-changelog/commitlint (@&#8203;commitlint/cli)</summary>

### [`v17.3.0`](https://togithub.com/conventional-changelog/commitlint/blob/HEAD/@&#8203;commitlint/cli/CHANGELOG.md#&#8203;1730-httpsgithubcomconventional-changelogcommitlintcomparev1721v1730-2022-11-21)

[Compare Source](https://togithub.com/conventional-changelog/commitlint/compare/v17.2.0...v17.3.0)

**Note:** Version bump only for package [@&#8203;commitlint/cli](https://togithub.com/commitlint/cli)

### [`v17.2.0`](https://togithub.com/conventional-changelog/commitlint/blob/HEAD/@&#8203;commitlint/cli/CHANGELOG.md#&#8203;1720-httpsgithubcomconventional-changelogcommitlintcomparev1712v1720-2022-10-31)

[Compare Source](https://togithub.com/conventional-changelog/commitlint/compare/v17.1.2...v17.2.0)

##### Features

[...]

PR #1406, opened 2022-11-29
Use renovate to auto bump deps

jinlee93:
### Summary:
[EDS-758]
- Uses Renovate to auto bump EDS dependencies
- Runs Every Monday 0600 PST
- Supposed to:
 - based off `next` branch
 - assign `edu-design-system` [team](https://docs.renovatebot.com/configuration-options/#reviewers) as reviewers
 - have one concurrent pr 
 - bump non major versions, except for < 0.1.0 and > 0.1.0 and < 1.0.0
 - and some other similar behavior as the [LP renovate](https://github.com/FB-PLP/traject/blob/master/.github/renovate.json5)
### Test Plan:
- tested somewhat working in `jlee/renovate-run` where renovate did create pr's #1403 and #1404 
  - need someone to confirm this is correct behavior since it was 2 concurrent prs and attempted to bump major versions
  - didn't assign `edu-design-system` team 
  - these might be some funky behavior due to not having a ["special token"](https://github.com/renovatebot/github-action#token) so I want to land this to see if it is working correctly with the cron job
  - is correctly based off `next`
- CI runs with renovate prs

[EDS-758]: https://czi-tech.atlassian.net/browse/EDS-758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -30,9 +30,6 @@
       matchDatasources: ['npm'],
       stabilityDays: 3,
     },
-    {
-      enabled: false,
-    },
     // Our peerDependencies for React are as low as 16.8.0, so we should support React v16
     {
       matchPackageNames: ['react', 'react-dom'],
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -10,6 +10,7 @@
   prConcurrentLimit: 1,
   repositories: ['FB-PLP/traject'],
   rangeStrategy: 'bump', // Always bump minor/patch to latest
+  reviewers: ['team:edu-design-system'],
 
   major: {
     rangeStrategy: 'replace',
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -8,7 +8,7 @@
   internalChecksFilter: 'strict', // PRs will be skipped unless a non-pending version is available (applies to "stabilityDays" check only)
   labels: ['npm-dependencies'],
   prConcurrentLimit: 1,
-  repositories: ['FB-PLP/traject'],
+  repositories: ['chanzuckerberg/edu-design-system'],
   rangeStrategy: 'bump', // Always bump minor/patch to latest
   reviewers: ['team:edu-design-system'],
 
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,0 +1,84 @@
+{
+  $schema: 'https://docs.renovatebot.com/renovate-schema.json',
+
+  baseBranches: ['next'],
+  enabledManagers: ['npm'],
+  extends: ['config:base', 'group:allNonMajor'],
+  fetchReleaseNotes: false, // Only fetch release notes for breaking changes
+  internalChecksFilter: 'strict', // PRs will be skipped unless a non-pending version is available (applies to "stabilityDays" check only)
+  labels: ['npm-dependencies'],
+  prConcurrentLimit: 1,
+  repositories: ['FB-PLP/traject'],
+  rangeStrategy: 'bump', // Always bump minor/patch to latest
+
+  major: {
+    rangeStrategy: 'replace',
+    fetchReleaseNotes: true,
+  },
+
+  ignoreDeps: [
+    // svg-spritemap-webpack-plugin v4 requires webpack 5 but EDS currently uses webpack 4 for storybook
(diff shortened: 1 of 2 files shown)


jinlee93:
@jeremiah-clothier Does it need the `enabled: true` config or will it run since the default is not `enabled: false` here? In general do you expect it to work as in the pr summary?

.github/renovate.json5
```diff
@@ -0,0 +1,85 @@
+{
+  $schema: 'https://docs.renovatebot.com/renovate-schema.json',
+
+  baseBranches: ['next'],
+  enabledManagers: ['npm'],
+  extends: ['config:base', 'group:allNonMajor'],
+  fetchReleaseNotes: false, // Only fetch release notes for breaking changes
+  internalChecksFilter: 'strict', // PRs will be skipped unless a non-pending version is available (applies to "stabilityDays" check only)
```
jinlee93, line 8: https://github.com/renovatebot/github-action#configurationfile

.github/renovate.json5
```diff
@@ -0,0 +1,85 @@
...
+    {
+      // Hold back from creating branches until 3 or more days have elapsed since the version
+      // was released. Note: this only works for packages that provide a release timestamp header
+      // and does not ensure all updates are "stable".
+      matchDatasources: ['npm'],
+      stabilityDays: 3,
+    },
+    {
+      enabled: false,
+    },
```
jeremiah-clothier, line 35: What does this do?

jeremiah-clothier (approved):

.github/renovate.json5
jinlee93: Cause a bug probably, will remove

.github/renovate.json5
```diff
@@ -0,0 +1,82 @@
...
+  baseBranches: ['next'],
+  enabledManagers: ['npm'],
+  extends: ['config:base', 'group:allNonMajor'],
+  fetchReleaseNotes: false, // Only fetch release notes for breaking changes
+  internalChecksFilter: 'strict', // PRs will be skipped unless a non-pending version is available (applies to "stabilityDays" check only)
+  labels: ['npm-dependencies'],
+  prConcurrentLimit: 1,
+  repositories: ['chanzuckerberg/edu-design-system'],
+  rangeStrategy: 'bump', // Always bump minor/patch to latest
+  reviewers: ['team:edu-design-system'],
```
jeremiah-clothier, line 13: If this syntax doesn't work to assign a group, then one option could be to make edu-design-system codeowners of the package.json file and then they'll get auto assigned.

jinlee93: Oooh thats a nice workaround, will keep in mind, thanks

.github/renovate.json5
```diff
@@ -0,0 +1,82 @@
...
+    // typescript errors that need to be manually resolved.
+    {
+      matchPackageNames: ['typescript'],
+      groupName: 'typescript',
+    },
+    {
+      matchDepTypes: ['devDependencies'],
+      groupName: 'devDependencies',
+    },
+  ],
```
jeremiah-clothier, line 77: I wonder if this is overriding the `group:allNonMajor` plugin and causing major updates to be grouped for devDependencies like in https://github.com/chanzuckerberg/edu-design-system/pull/1404

jinlee93: Oh makes sense. How would you recommend then grouping `devDeps` separate from prod deps? Or does renovate do that automatically?

jinlee93: Should I just land and see what happens on Monday?

jeremiah-clothier: I'm not sure off the top of my head.

* What's the end state you're looking for?
* Generally, nonMajor updates are safe, what's the motivation for wanting to group devDeps and prodDeps separately and override the behavior of `group:allNonMajor`?

jinlee93: That is a good question. The motivation was that through separation, if a bug rose from an update, the scope of reversion would be smaller. However I do agree that nonMajor updates are generally safe and this worry is probably trivial.

PR #1407, opened 2022-12-01
feat(Select): add select component

booc0mtaco:
### Summary:

[EDS-754]: create a new `Select` component, based closely on the
existing `Dropdown` component. Make use of subcomponents introduced
in the `Menu` component, like PopoverListItem and PopoverContainer,
and some light refactoring to simplify component structure for better
alignment with Menu.

- mark appropriate components as deprecated
- merge button trigger component into select
- add specificity to class naming on modifiers
- tidy up API by removing options flagged as unused

### Test Plan:

- [x] storybook tests
- [x] chromatic

[EDS-754]: https://czi-tech.atlassian.net/browse/EDS-754?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Select/Select.module.css
```diff
@@ -0,0 +1,115 @@
+@import '../../design-tokens/mixins.css';
+
+/*------------------------------------*\
```
booc0mtaco, line 3: Next up is renaming all the classes and such to be `select` based

src/components/Select/Select.stories.module.css
```diff
@@ -0,0 +1,35 @@
+/*------------------------------------*\
```
booc0mtaco, line 1: Next up is renaming all the classes to be `select`-based

src/components/Select/Select.module.css
```diff
@@ -0,0 +1,115 @@
...
+}
+
+/**
+ * The caret icon to decorate the select trigger button, animated to rotate.
+ */
+.select-button__icon {
+  transform: rotate(0);
+
+  transition: transform var(--eds-anim-move-medium) ease-out;
+  @media (prefers-reduced-motion) {
```
jinlee93, line 107: nice

src/components/Select/Select.stories.tsx
```diff
@@ -0,0 +1,223 @@
...
+  component: Select,
+  parameters: {
+    badges: [BADGE.BETA],
+    layout: 'centered',
+  },
+  decorators: [
+    (Story) => (
+      <div
+        style={{
+          margin: '1rem', // Provides spacing to see activity around the select field
```
jinlee93, line 22: If layout is `centered` does this still make a difference?

src/components/Select/Select.stories.tsx
```diff
@@ -0,0 +1,223 @@
...
+      variant="compact"
+    />
+  ),
+};
+
+export const UsingFunctionChildrenProp: StoryObj = {
+  render: () => <InteractiveExampleUsingFunctionChildren />,
+};
+
+// This story just opens the dropdown automatically so chromatic can test it.
```
jinlee93, line 208: 1. You can use [`isChromatic`](https://www.chromatic.com/docs/ischromatic) to add play functions to the above stories if you want to snap the open state similarly (unless ofc your intention was to snap the button)
2. If you don't want to snap, you can always [disable](https://www.chromatic.com/docs/ignoring-elements#ignore-stories) the snaps in Chromatic
3. Something about timing and animations may not snap very nicely so can be flaky in Chromatic, just something to watch out for
  - Did just see a possibly new Chromatic feature regarding pausing [CSS animations](https://www.chromatic.com/docs/animations#css-animations) if you wanted to check that out

src/components/Select/Select.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+        </Select.Options>
+      </Select>
+    );
+    const renderMethod = () => {
+      render(dropdownWithDropdownLabel);
+    };
+
+    expect(renderMethod).not.toThrow(Error);
+  });
+});
```
jinlee93, line 85: Do you think there is value in testing the keyboard navigation functionality or would we just be testing HeadlessUI functionality that we don't own?

src/components/Select/Select.tsx
```diff
@@ -0,0 +1,357 @@
...
+  /**
+   * Optional className for additional options menu styling.
+   *
+   * When not using the compact variant, if optionsClassName is provided please
+   * include the width property to define the options menu width.
+   */
+  optionsClassName?: string;
+};
+
+function childrenHaveLabelComponent(children?: ReactNode): boolean {
```
jinlee93, line 57: Would this work with React fragments?

jinlee93:
Think you also need to add
```
```
to `src/index.ts`

src/components/Select/Select.stories.tsx
booc0mtaco: Yah, it doesn't seem necessary anymore. I will remove this decorator to simplify things

src/components/Select/Select.stories.tsx
booc0mtaco: Oh we may have to do a little bit of a dive into some of this. I will add this flag as I would like one of these to have a snap with it open. 

(this particular test was brought over from the `Dropdown` implementation)

src/components/Select/Select.test.tsx
booc0mtaco: Yah, the latter. if there are any bugs with the keyboard behavior, we would essentially file a bug with upstream library. Now, if we were adding some custom keyboard handling, we def would want to test that.

src/components/Select/Select.tsx
booc0mtaco: Some of the factoring in this came from the existing tests I inherited here. I was trying to change only a little of that code, but these functions obfuscate what is being tested a little bit, in general.

I suspect there is something in react that actually does this (or a better job of it), so was inclined to leave it for a later investigation

src/components/Select/Select.tsx
```diff
@@ -0,0 +1,357 @@
...
+}
+
+/**
+ * `import {Select} from "@chanzuckerberg/eds";`
+ *
+ * EDS Select. Used to select one option from a list of options.
+ *
+ * Built on top of the Headless UI Listbox: https://headlessui.dev/react/listbox#basic-example
+ *
+ * TODO: document using class names to alter widths
```
jinlee93, line 83: Leftover todo?

jinlee93 (approved):
Looks good to me but would love to solicit @dierat's opinions

src/components/Dropdown/Dropdown.tsx
```diff
@@ -114,6 +112,10 @@ function childrenHaveLabelComponent(children?: ReactNode): boolean {
 /**
  * `import {Dropdown} from "@chanzuckerberg/eds";`
  *
+ * Note: this component has been deprecated and will be removed in a future release.
+ * Please use the Select component instead for forms, and the Menu component for other
+ * types of popover-like dropdowns.
+ *
```
dierat, line 118: Could you add `@deprecated` to the bottom of this component docstring and add the deprecated badge to `Dropdown.stories.tsx`?

src/components/Select/Select.tsx
```diff
@@ -0,0 +1,357 @@
...
+    ) {
+      return true;
+    } else if ('props' in child && child.props.children) {
+      return childrenHaveLabelComponent(child.props.children);
+    }
+    return false;
+  });
+}
+
+/**
```
dierat, line 76:
```suggestion
/**
 * BETA: This component is still a work in progress and is subject to change.
 *
```

dierat:
Looks good!

src/components/Select/Select.tsx
booc0mtaco: yah sorta? I wasn't sure how this currently works at first, but I did include a test with the result in, so i can address that

src/components/Dropdown/Dropdown.tsx
dierat: Thanks for adding the `@deprecated` tag! Can you also add the badge in `Dropdown.stories.tsx`?

src/components/Select/Select.module.css
```diff
@@ -0,0 +1,115 @@
...
+  margin-bottom: var(--eds-size-half);
+}
+
+/**
+ * Visually hides the label element if a aria-label is passed,
+ * but element is kept for screen reading accessibility.
+ */
+.u-is-vishidden {
+  @mixin visuallyHide;
+}
```
dierat, line 39: Where is this being used?

booc0mtaco: oh yah this was a carry-over from the dropdown component. I can remove this

src/components/Dropdown/Dropdown.tsx
booc0mtaco: Yep, this is recorded [here](https://github.com/chanzuckerberg/edu-design-system/blob/next/src/components/Dropdown/Dropdown.stories.tsx#L16) which came in with my other popover ticket

dierat: Awesome, thanks!

src/components/Select/Select.module.css
dierat: Thanks! I really wish we had a linter for this..

dierat (approved):
Looks great, thanks!

booc0mtaco:
Thanks all!

PR #1408, opened 2022-12-02
Add progress bar

jinlee93:
### Summary:
[EDS-735]
- motivation:
  - progress bar is needed to indicate how progress of a task
  - existing <DataBar> component has complex logic for tooltips, overflow state, miniscule values state, etc
  - as a progress indicator, needs to be [accessible](https://w3c.github.io/aria/#progressbar)
- implementation:
  - although looks very similar to data bar, children and styling is simple enough where I felt reusability was not necessary and actually burdensome
  - uses `progressbar` role and associated aria attributes
  - synced with design about empty states, segment count, visual representation accuracy (written in ticket body)
  - [ticket](https://czi-tech.atlassian.net/browse/EDS-761) made for misaligned caption token

Screen Shot 2022-12-01 at 6 36 30 PM
With unit of measurement:
Screen Shot 2022-12-01 at 6 36 34 PM
Empty state:
Screen Shot 2022-12-01 at 6 36 37 PM
Complete state:
Screen Shot 2022-12-01 at 6 36 43 PM

### Test Plan:
- no other visual regression
- interactive story available for manual testing

[EDS-735]: https://czi-tech.atlassian.net/browse/EDS-735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+export type Props = {
+  /**
+   * CSS class names that can be appended to the component.
+   */
+  className?: string;
+  /**
+   * Label associated with naming the progress bar.
+   */
+  label: string;
```
jinlee93, line 14: I'm making this required and being strict about this since mocks doesn't suggest not having a label

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,98 @@
...
+   */
+  segmentCount: number;
+  /**
+   * The value that each segment represents.
+   */
+  segmentValue: number;
+  /**
+   * An optional unit of measurement to describe the progress.
+   */
+  unit?: string;
```
jinlee93, line 31: Some suggestions on prop naming is appreciated

jinlee93:
Note that there are no variants for this component and the segments themselves are not exposed for mutability. This is intentional due to lack of variants in the mocks and color changes can be done via theming.

jinlee93:
Synced with Sean and got approval for design

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,98 @@
...
+    console.warn('Progress bar segment count should be capped at 10');
+  }
+
+  // Creates an array of length segmentCount and fills them with styled divs to represent the segments.
+  const segments = segmentCount
+    ? Array.from({ length: segmentCount }, (_, index) => (
+        <div
+          className={styles['progress-bar__segment']}
+          key={'progress-bar__segment-' + index}
+          style={{ width: `${(segmentValue / max) * 100}%` }}
```
dierat, line 60: I noticed while playing with the interactive one that there's a little jumpiness when the last segment is completed. I think it's because the width of the segments doesn't take into account the gap between them, so they take up more space when the bar doesn't have all the segments in yet. For example, when the bar is 1200px wide and has max 10 segments, the segments are 120px wide when the bar is not full and 116px wide when the bar is full.

https://user-images.githubusercontent.com/7761701/205736755-f86a056e-d4f2-4af1-809b-ffd27b47b426.mp4

Not blocking but wanted to let you know in case you wanted to try to fix it. I don't know how often users will even see the bar change from being incomplete to complete in front of them, so it might be the kind of thing we notice more when demoing the component.

jinlee93: Good callout, will tinker with some css calc to make this look better

src/components/ProgressBar/ProgressBar.tsx
jinlee93: We are using it in `Checkbox`, `Radio`, `TextField`, but more importantly `DataBar` which pattern should be more similar

src/components/ProgressBar/ProgressBar.tsx
jinlee93: Yea I considered `unitOfMeasurement` but if a % sign or decimal is being used, wouldn't be applicable

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,98 @@
...
+   */
+  unit?: string;
+};
+
+/**
+ * BETA: This component is still a work in progress and is subject to change.
+ *
+ * `import {ProgressBar} from "@chanzuckerberg/eds";`
+ *
+ * A progress bar component that indicating the completion progress of a task.
```
dierat, line 39:
```suggestion
 * A progress bar component that indicates how much progress has been made on a task.
```
The current wording kinda makes it sound like this is a kind of badge that says the task is done as opposed to communicating how close it is to being done.

src/components/ProgressBar/ProgressBar.stories.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+type Args = React.ComponentProps<typeof ProgressBar>;
+
+export const Default: StoryObj<Args> = {};
+
+export const WithUnit: StoryObj<Args> = {
+  args: {
+    unit: '%',
+  },
+};
```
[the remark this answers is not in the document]
jinlee93: Can we sync about how we can use a render prop could allow arbitrary text? Not very familiar with creating render props, so some knowledge transfer would be appreciated

src/components/ProgressBar/ProgressBar.tsx
jinlee93: Good suggestion, will change

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,98 @@
...
+  label,
+  segmentCount,
+  segmentValue,
+  max,
+  unit,
+  ...other
+}: Props) => {
+  if (process.env.NODE_ENV !== 'production' && segmentCount > 10) {
+    console.warn('Progress bar segment count should be capped at 10');
+  }
```
dierat, line 52: I'm not sure how I feel about this warning. If it were a hard requirement, I would suggest we use types (something ugly like `segmentCount = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10;`), but if it's a recommendation, we could end up with warnings clogging up the terminal when the component is working as designed. I think it's enough to have this in the docstring for the component and in design documentation in zeroheight

jinlee93: Ah agree that typing as enum numbers would be better here

dierat: Ugh yikes. This SO question has another idea but that might also be an issue here: https://stackoverflow.com/questions/39494689/is-it-possible-to-restrict-number-to-a-certain-range

I'm also fine with just putting it in the documentation. I just don't want to keep adding to the browser console mess LP already has if we can help it.

booc0mtaco: I am OK with us removing this code constraint as a warning. When designs are produced using this component, they should follow the recommendations in zeroheight and whatever we document. But if someone does use more segments, the implementation will handle it to a certain degree (if it breaks the rules, it's a conversation that the designers and engineers need to talk thru for that page)

Handling > 10 but documenting "only ten or less, please" follows the [Robustness Principle](https://en.wikipedia.org/wiki/Robustness_principle) w/o making too much noise...

jinlee93: Sounds good, will remove warning and just keep as type number

src/components/ProgressBar/ProgressBar.tsx
jinlee93: This took a little more calculation than I had previously anticipated lol:
``` tsx
// Calculates the segment width to prevent jumping once progress bar is full
const segmentsPerBar = Math.ceil(max / segmentValue);
const gapMultiplier = (segmentsPerBar - 1) / segmentsPerBar;
const segmentWidthBeforeGap = (segmentValue / max) * 100;
const segmentWidth = `calc(${segmentWidthBeforeGap}% - var(--eds-size-half) * ${gapMultiplier})`;
return (
  <div
    className={styles['progress-bar__segment']}
    key={'progress-bar__segment-' + index}
    style={{ width: segmentWidth }}
  />
);
```
Let me know if you think this either needs more comments or is overkill.
Also updated the interactive story so the segment value and max can be changed to test this.
@dierat 

https://user-images.githubusercontent.com/86632227/206040757-41103fc4-aefe-4e81-aba0-9091a15cbad3.mov

src/components/ProgressBar/ProgressBar.stories.tsx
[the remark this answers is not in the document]
jinlee93: Discussed sync, will be replaced with the `caption` prop which will override pre-generated caption

src/components/ProgressBar/ProgressBar.tsx
jinlee93: Discussed sync, will be replaced with the `caption` prop which will override pre-generated caption

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,97 @@
...
+}: Props) => {
+  // Creates an array of length segmentCount and fills them with styled divs to represent the segments.
+  const segments = segmentCount
+    ? Array.from({ length: segmentCount }, (_, index) => {
+        // Calculates the segment width to prevent jumping once progress bar is full
+        const segmentsPerBar = Math.ceil(max / segmentValue);
+        const gapMultiplier = (segmentsPerBar - 1) / segmentsPerBar;
+        const segmentWidthBeforeGap = (segmentValue / max) * 100;
+        const segmentWidth = `calc(${segmentWidthBeforeGap}% - var(--eds-size-half) * ${gapMultiplier})`;
+        return (
```
dierat, line 58: Super nit: can we get an empty line between the segment width calculation and the jsx to make it clearer that that whole block is just for calculating the width? Could also split it out into a separate function and add tests.
```suggestion

        return (
```

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,97 @@
...
+  segmentCount,
+  segmentValue,
+  max,
+  ...other
+}: Props) => {
+  // Creates an array of length segmentCount and fills them with styled divs to represent the segments.
+  const segments = segmentCount
+    ? Array.from({ length: segmentCount }, (_, index) => {
+        // Calculates the segment width to prevent jumping once progress bar is full
+        const segmentsPerBar = Math.ceil(max / segmentValue);
```
dierat, line 54: How is `segmentsPerBar` different from `segmentCount`?

src/components/ProgressBar/ProgressBar.tsx
dierat: Damn that's nice! That's cool that you got it to figure out a percentage so it doesn't need to recalculate when the progress bar changes width. I would maybe add that to your comment, that it creates a percentage, because I was confused why you were doing it this way until I saw that in the `calc` line.

dierat (approved):
Nice!

dierat:
I put out a [PR for disabling the wireframe demo in chromatic](https://github.com/chanzuckerberg/edu-design-system/pull/1417), which I see is showing up as a change here.

src/components/ProgressBar/ProgressBar.tsx
jinlee93: `segmentsPerBar` is the amount of segments in a bar before the bar is "full" so if a `segmentValue` is 2 and the max value is 5, then `segmentsPerBar` would be 3

jinlee93: Will rename to `segmentsPerFullBar` to be more intuitive

src/components/ProgressBar/ProgressBar.tsx
jinlee93: Added another explaining the percentage calc and the subtraction of the gap space

booc0mtaco:
Let's walk thru how the segments themselves are generated. I am a little surprised it was possible for them to jump around based on progress. But I know I have a prior implementation in mind, so I am imagining this code works like that lol

Not a problem here; just want to make sure I am not biasing myself into being confusion basically

src/components/ProgressBar/ProgressBar.stories.tsx
```diff
@@ -0,0 +1,136 @@
...
+        label="Progress Bar Label"
+        max={100}
+        segmentCount={1}
+        segmentValue={100}
+      />
+    </div>
+  ),
+};
+
+const InteractiveProgressBar = () => {
```
booc0mtaco, line 64: Would it be possible to use storybook controls to build the test interface? if we add `args` to render and spread that on this component, for example?

jinlee93: Some of the args are accounted for via input controls 
image
and the purpose of this story is to show how the state of the progress bar can be managed in code, as the other stories provide controls via `args` 

However, I can pass the other props, i.e. `label`, `className`, and `caption` if that provides extra value

jinlee93:
> Let's walk thru how the segments themselves are generated. I am a little surprised it was possible for them to jump around based on progress. But I know I have a prior implementation in mind, so I am imagining this code works like that lol
> 
> Not a problem here; just want to make sure I am not biasing myself into being confusion basically

In [`ProgressBar.tsx`](https://github.com/chanzuckerberg/edu-design-system/blob/1e6dd727ce43dfa71f13f257103442a30ebeff7c/src/components/ProgressBar/ProgressBar.tsx#L51-L68):
```tsx
const segments = segmentCount
  ? Array.from({ length: segmentCount }, (_, index) => {
      // Calculates the segment width to prevent jumping once progress bar is full.
      // Creates a percentage for the segment width and subtracts the space gaps would take via css calc.
      const segmentsPerFullBar = Math.ceil(max / segmentValue);
      const gapMultiplier = (segmentsPerFullBar - 1) / segmentsPerFullBar;
      const segmentWidthBeforeGap = (segmentValue / max) * 100;
      const segmentWidth = `calc(${segmentWidthBeforeGap}% - var(--eds-size-half) * ${gapMultiplier})`;

      return (
        <div
          className={styles['progress-bar__segment']}
          key={'progress-bar__segment-' + index}
          style={{ width: segmentWidth }}
        />
      );
    })
  : undefined;
```
Creates an array of `segmentCount` length, and fills with almost identical `<div>` items (except the `key`), which are the segments. Segments have a calculated width dependent on the ratio of number of segments it takes to fill a bar to the full width of the bar, represented as a percentage. Then the space the gap would take is subtracted.

jinlee93:
As discussed sync, the progress bar has been refactored to use props `maxValue`, `currentValue`, and `totalSegments` and stories have been updated to reflect.
`maxValue` represents the amount of progress that needs to be made, `currentValue` is the amount of progress that has been made, and `totalSegments` is the number of segments to divide the progress, which is optional and defaults to 1 so the common use case of simply representing progress without segmentation requires less props.

@dierat @booc0mtaco

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,97 @@
...
+import styles from './ProgressBar.module.css';
+import Text from '../Text';
+
+export type Props = {
+  /**
+   * CSS class names that can be appended to the component.
+   */
+  className?: string;
+  /**
+   * Overrides the default "totalValue / max" in cases where a different caption may be desired.
```
dierat, line 12: Should this be changed now that the props have been renamed?
```suggestion
   * Overrides the default "currentValue / maxValue" in cases where a different caption may be desired.
```

booc0mtaco (approved):

PR #1409, opened 2022-12-02
Update databar styles

jinlee93:
### Summary:
- whilst working on the progress bar noticed that the databar styles were not in sync with the mocks
- syncs styles with the mocks and cleans up some now obsolete code
  - stylistically, shrinks spacing between the label wrapper and the databar from 1rem to 0.5rem and the height of the databar from 1rem to 0.875rem
- impact is indicated by visual diffs of the recipes/pages, and in LP the only page impacted is the [`CoursePlanEdit` page](https://www.chromatic.com/test?appId=614b969675171d003ac6f98a&id=638974d54f0d45e3f3992aa2). This uses the [`estimatedInstructionalBlocks` lp component](https://www.chromatic.com/test?appId=614b969675171d003ac6f98a&id=638974d54f0d45e3f3992aa3) where visual diffs are again represented.
  - Change is safe since only one page is impacted, and only in about 10px of vertical spacing
### Test Plan:
- alpha publish v7.1.0-alpha.1 reflects these changes and are shown https://www.chromatic.com/build?appId=614b969675171d003ac6f98a&number=15188

### DIFF
--- a/src/components/DataBar/DataBar.module.css
+++ b/src/components/DataBar/DataBar.module.css
@@ -18,6 +18,7 @@
   /* Rounds the sides rather than making the bar elliptical. */
   border-radius: var(--eds-border-radius-full);
 
+  /* Prevents segments from overflowing outside borders */
   overflow: hidden;
 }
 
--- a/src/components/DataBar/DataBar.module.css
+++ b/src/components/DataBar/DataBar.module.css
@@ -28,6 +28,8 @@
 .data-bar__label {
   @mixin eds-theme-typography-title-sm;
   color: var(--eds-theme-color-text-neutral-strong);
+
+  margin-bottom: 0;
 }
 
 /**
--- a/src/components/DataBar/DataBar.module.css
+++ b/src/components/DataBar/DataBar.module.css
@@ -8,15 +8,17 @@
  * The data bar represents the relative completion of a task.
  */
 .data-bar {
-  --data-bar-height: var(--eds-size-2);
+  --data-bar-height: 0.875rem;
   height: var(--data-bar-height);
 
   background-color: var(--eds-theme-color-background-neutral-subtle);
 
   border: var(--eds-theme-color-border-neutral-default) solid
     var(--eds-theme-border-width);
-  /* Rounds the sides rather than making the bar elliptical. */  
-  border-radius: var(--eds-border-radius-full); 
+  /* Rounds the sides rather than making the bar elliptical. */
+  border-radius: var(--eds-border-radius-full);
+
+  overflow: hidden;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/components/DataBar/DataBar.module.css
```diff
@@ -8,15 +8,18 @@
  * The data bar represents the relative completion of a task.
  */
 .data-bar {
-  --data-bar-height: var(--eds-size-2);
+  --data-bar-height: 0.875rem;
```
booc0mtaco, line 11: We should validate with design the spacing between things. I've seen a few cases where the spacing used by design doesn't match these `eds-size` values. If we have tokens, we should stick with them!

src/components/DataBarSegment/DataBarSegment.tsx
```diff
@@ -10,11 +10,6 @@ export type Props = {
    * CSS class names that can be appended to the component.
    */
   className?: string;
-  /**
-   * Indicates if this segment should be rounded on the right side.
-   * Used for data bars that are 100% complete.
-   */
-  isRoundRight?: boolean;
```
booc0mtaco, line 17: ++

booc0mtaco (approved):
One thing we want to validate with design are things that don't align with 4px units. What pixel size does this rem value map to?

Otherwise looks good to me!

jinlee93:
> One thing we want to validate with design are things that don't align with 4px units. What pixel size does this rem value map to?

It maps to 14px. Design made the data bar with 12px, and had a 1px border making it 14px total

booc0mtaco:
@jinlee93 Also small-ish thing. The commits in here should be of type `fix` or `feat` instead of `chore`. Usually chore is reserved for things like library updates, config changes, stuff like that. This could also be a `refactor`

* [Spec](https://www.conventionalcommits.org/en/v1.0.0-beta.2/#specification)
* [Conventional Changelog Types](https://github.com/conventional-changelog/commitlint/tree/master/%40commitlint/config-conventional)

src/components/DataBar/DataBar.module.css
jinlee93: This is 14px and I think this makes sense since the data bar was made to be 12px, with a 1px border.
Also is the same height as the progress bar.
However will check with design to be sure

jinlee93:
> @jinlee93 Also small-ish thing. The commits in here should be of type `fix` or `feat` instead of `chore`. Usually chore is reserved for things like library updates, config changes, stuff like that. This could also be a `refactor`

Ah got it, should I change the messages and force push?

src/components/DataBar/DataBar.module.css
booc0mtaco: ah ok. so it's 12px + borders on both sides? in that case, I think we are good, and this value could be a calc of the appropriate tokens (eds-size-1-and-half? and 2x border size or similar.

We wouldn't want the height to break if either of these tokens change

jinlee93: Could make it a calc to make it more obvious yea

booc0mtaco:
> > @jinlee93 Also small-ish thing. The commits in here should be of type `fix` or `feat` instead of `chore`. Usually chore is reserved for things like library updates, config changes, stuff like that. This could also be a `refactor`
> 
> Ah got it, should I change the messages and force push? @booc0mtaco

If possible, thanks!

booc0mtaco (approved):
🚢

PR #1410, opened 2022-12-02
Node 18.12.1, Yarn 3.3.0

ahuth:
### Summary:

This PR updates:
- Node to 18.12.1, so that it's on the same version as our Remix edu stack (which makes testing with `npm link` more convenient for me)
- Yarn to 3.3.0, because why not

### Test Plan:

CI


.github/workflows/docs.yml
```diff
@@ -11,10 +11,11 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - name: Setup node 😻
+        uses: actions/setup-node@v3
         with:
-          node-version: 16.x
-          cache: "yarn"
+          node-version-file: '.node-version'
+          cache: 'yarn'
```
ahuth, line 18: Err, I think the quotes changed while copy-and-pasting. Can change it back if we prefer.

.github/workflows/docs.yml
```diff
@@ -11,10 +11,11 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - name: Setup node 😻
+        uses: actions/setup-node@v3
         with:
-          node-version: 16.x
-          cache: "yarn"
+          node-version-file: '.node-version'
```
ahuth, line 17: Can now point actions/setup-node at our `.node-version` file, instead of manually specifying.

.github/workflows/test.yml
```diff
@@ -7,10 +7,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Use Node.js
-        uses: actions/setup-node@v2
+      - name: Setup node 😻
+        uses: actions/setup-node@v3
         with:
-          node-version: 16.x
+          node-version-file: '.node-version'
```
jinlee93, line 13: Didn't know this was possible, pretty cool

.github/workflows/docs.yml
jinlee93: Didn't know this was possible, pretty cool

jinlee93 (approved):
Sweet, thanks for the updates

ahuth:
Thanks!

PR #1411, opened 2022-12-05
chore: add linter for tier 1 color css variables

dierat:
### Summary:
Ticket: https://czi-tech.atlassian.net/browse/EDS-707

For the purposes of theming, we only want to use tier 2 and tier 3 color CSS variables in component styling in EDS (with some rare exceptions). This PR adds a stylelint rule to enforce that.

I followed the implementation in LP for setting up a custom internal stylelint plugin.

#### Question:
It insists that `stylelint` and `dedent` now need to be prod dependencies instead of dev dependencies. Does anyone know why? 

### Test Plan:
- verify there are no stylelint errors
- add a tier 1 color CSS variable use somewhere in a component and verify it triggers the linter


yarn.lock
```diff
@@ -2035,6 +2036,7 @@ __metadata:
     intersection-observer: ^0.12.2
     jest: ^29.0.3
     jest-environment-jsdom: ^29.0.3
+    jest-preset-stylelint: ^6.0.0
```
dierat, line 2039: This one is required for the test to run, but, like `dedent` and `stylelint`, this should really be in dev dependencies.

.eslintrc
```diff
@@ -4,10 +4,7 @@
     "browser": true,
     "node": true
   },
-  "plugins": [
-    "testing-library",
-    "@chanzuckerberg/stories"
-  ],
+  "plugins": ["testing-library", "@chanzuckerberg/stories"],
```
dierat, line 7: This was an automated change.

dierat, line 7: This was an automated change.

scripts/stylelint/rules/no-tier-1-color-variable.js
```diff
@@ -0,0 +1,44 @@
+/* eslint-env node */
+
+const dedent = require('dedent');
+const stylelint = require('stylelint');
+
+const ruleName = 'eds/no-tier-1-color-variable';
+
+const messages = stylelint.utils.ruleMessages(ruleName, {
+  rejected: dedent`
+    Don't use tier 1 color CSS variables directly. Choose a tier 2 or tier 3 variable that has the same color value with a name relevant to the style you're adding. (Example: a neutral border might use --eds-theme-color-border-neutral-default.) If such a variable does not exist, talk to design about adding a token for this use case.
```
jinlee93, line 10: Is the addition of the `dedent` package only for this message?

jinlee93 (approved):
Makes sense to add stylelint, looks good pending one question

PR #1413, opened 2022-12-06
chore(wireframe-demo): use same reflection text as along demo

dierat:
### Summary:
This is a really nitty thing, but I want to change the reflection text on the watch page in the wireframe theme demo to use the same copy that I'm using in the along theme demo. That way, the text is the same length, and it's easier to compare the two themes.

wire theme demo watch page with shortened reflection text
wire theme demo watch page, unchanged

### Test Plan:
Verify in storybook locally that the wireframe theme demo watch page reflection text is now the same length as the reflection text in the along theme demo watch page reflection text.


jinlee93 (approved):
Nice. I somewhat know how to do math as well :D

dierat:
Thanks for reviewing! 😀

PR #1150, opened 2022-06-21
chore(storybook): update storybook deps and remove webpack 5

jinlee93:
### Summary:
- update storybook deps
- remove webpack 5
  - https://github.com/chanzuckerberg/edu-design-system/pull/1111#discussion_r891203373
  - felt more storybook crashes with webpack 5 use
### Test Plan:
- unit tests pass
- no chromatic fails

PR #1414, opened 2022-12-07
Use webpack 5 or legacy node option to be compatible with node v18

jinlee93:
### Summary:
- With the recent upgrade to node v18, I noticed my storybook dev env broke when I clean installed
- This is due to storybook relying on webpack 4 as default, and [webpack 4 not supporting some openssl hashing starting with node v17](https://github.com/webpack/webpack/issues/14532#issuecomment-947513562)
- There are some options to proceed here:
  - Use webpack 5 with storybook
    - Previously reverted in #1150 due to [lack of need, lack of need to include webpack 5](https://github.com/chanzuckerberg/edu-design-system/pull/1111/files#r891203373)
    - Represented in latest commit of this pr [`4128b7f`](https://github.com/chanzuckerberg/edu-design-system/compare/next...4128b7f)
  - Use legacy node option `NODE_OPTIONS=--openssl-legacy-provider` 
    - Represented in first commit of this pr [`4b2edee`](https://github.com/chanzuckerberg/edu-design-system/compare/next...4b2edee)
    - Leaning towards this option as it is done similarly in [`axe-storybook-testing`](https://github.com/chanzuckerberg/axe-storybook-testing/commit/cbafe3e2bd499dbe0af6696280256157285a0ced)
  - Go back to node v16.13?
- Would enjoy advice on how to proceed
- Maybe its my personal lol
### Test Plan:
- Local dev works on all 3 options listed


ahuth (approved):
LGTM 🤘 

Going with the workaround makes sense to me. But any of the options will work.

Also I suspect that Storybook 7 will resolve this since it should be on Webpack 5.

booc0mtaco (approved):
It sounds like we will have an action item to remove these legacy options in relevant repo.s once storybook 7 comes out. I like the approach of using `NODE_OPTIONS=--openssl-legacy-provider` over adding a dependency, as it's a more direct fix to the issue we are seeing. 

Could any other repositories have this issue? (traject, along, etc.)

jinlee93:
> It sounds like we will have an action item to remove these legacy options in relevant repo.s once storybook 7 comes out. I like the approach of using `NODE_OPTIONS=--openssl-legacy-provider` over adding a dependency, as it's a more direct fix to the issue we are seeing.
> 
> Could any other repositories have this issue? (traject, along, etc.)

Seems like traject and along are both on node v16 so no problems yet

PR #1421, opened 2022-12-07
Set renovate version and prevent onboarding pr

jinlee93:
### Summary:
- Saw that putting only [major version of renovate](https://github.com/chanzuckerberg/edu-design-system/actions/runs/3634187356/jobs/6132023885#step:1:35) is not recognized
- Sets renovate version to prevent
- Also prevents onboarding renovate pr according to https://github.com/renovatebot/github-action#configurationfile
### Test Plan:
- https://github.com/chanzuckerberg/edu-design-system/actions/workflows/update-dependencies.yml update deps jobs ran with push on this branch (which was reverted cuz we want it on cron)
- Saw that renovate action works since it generated the onboarding prs (which I closed since we don't want this behavior)
image
- Will have to see if job runs correctly on Monday as scheduled

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -12,6 +12,9 @@
   rangeStrategy: 'bump', // Always bump minor/patch to latest
   reviewers: ['team:edu-design-system'],
 
+  onboarding: false,
+  requireConfig: false,
+
   major: {
     rangeStrategy: 'replace',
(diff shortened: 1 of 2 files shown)


jeremiah-clothier (approved):
LGTM, thanks for going through and setting this up for EDS!

PR #1422, opened 2022-12-08
feat(LoadingIndicator): add loader indicator component

booc0mtaco:
### Summary:

Add in a loader component, which can be used when there's a non-finite amount of time until loading is completed. Allow for some preliminary props to control size, and apply color overrides by CSS (blocking the component interface).

use the [react loader spinner](https://mhnpd.github.io/react-loader-spinner/docs/components/oval)'s `Oval` implementation. 

Future use case, apply an `as` prop to allow any other loader types using the same component (for example).

### Test Plan:

- [x] storybook
- [x] unit tests (for coverage)

### DIFF
--- a/src/components/LoadingIndicator/LoadingIndicator.tsx
+++ b/src/components/LoadingIndicator/LoadingIndicator.tsx
@@ -1,0 +1,83 @@
+import clsx from 'clsx';
+import React from 'react';
+
+import { Oval } from 'react-loader-spinner';
+import styles from './LoadingIndicator.module.css';
+
+export type Props = {
+  /**
+   * CSS class names that can be appended to the component.
+   */
+  className?: string;
+
+  /**
+   * Layout size of the loader. This affects the overall size and associated
+   * stroke width.
+   */
+  size?: 'sm' | 'md' | 'lg';
+
+  /**
+   * Aria label of the oval. Default is "loading". Will be overridden if ariaLabel is passed in props
+   * (Shadowed from OvalProps)
+   *
+   * See: https://mhnpd.github.io/react-loader-spinner/docs/components/oval
+   */
+  ariaLabel?: string;
+
+  /**
+   * Whether the oval is visible. Default is true.
+   * (Shadowed from OvalProps)
+   *
+   * See: https://mhnpd.github.io/react-loader-spinner/docs/components/oval
+   */
+  visible?: boolean;
+};
+
+// Pixel sizes corresponding to EDS size units
+const loaderSize = {
+  sm: 24, // --eds-size-2-and-half
+  md: 40, // --eds-size-5
+  lg: 56, // --eds-size-7
+};
+
+// Given a loader size, the stroke widths can change
+const loaderStrokeSize = {
+  sm: 2,
+  md: 3,
+  lg: 4,
+};
+
+/**
+ * BETA: This component is still a work in progress and is subject to change.
+ *
+ * `import {LoadingIndicator} from "@chanzuckerberg/eds";`
+ *
+ * Show a loading state within a component container, or a specific page or area, where
+ * arbitrary content will appear once some request has completed.
+ */
+export const LoadingIndicator = ({
+  ariaLabel = 'loading',
+  className,
+  size = 'md',
+  visible = true,
+  ...other
+}: Props) => {
+  const componentClassName = clsx(styles['loading-indicator'], className);
+
+  // setting the colors to be transparent since we override in CSS
+  // (and have a lint rule prevent token variable use in components)
+  return (
+    <Oval
+      ariaLabel={ariaLabel}
+      color="transparent"
+      height={loaderSize[size]}
+      secondaryColor="transparent"
+      strokeWidth={loaderStrokeSize[size]}
+      strokeWidthSecondary={loaderStrokeSize[size]}
+      visible={visible}
+      width={loaderSize[size]}
+      wrapperClass={componentClassName}
+      {...other}
+    />
+  );
+};
--- a/src/index.ts
+++ b/src/index.ts
@@ -50,6 +50,7 @@ export { default as LayoutLinelengthContainer } from './components/LayoutLinelen
 export { default as LayoutSection } from './components/LayoutSection';
 export { default as Link } from './components/Link';
 export { default as LinkList } from './components/LinkList';
+export { default as LoadingIndicator } from './components/LoadingIndicator';
 export { default as Logo } from './components/Logo';
 export { default as Menu } from './components/Menu';
 export { default as Modal } from './components/Modal';
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
     "react-beautiful-dnd": "^13.1.1",
     "react-children-by-type": "^1.1.0",
     "react-focus-lock": "^2.9.2",
+    "react-loader-spinner": "^5.3.4",
     "react-popper": "^2.3.0",
     "react-portal": "^4.2.2",
     "react-uid": "^2.3.2",
--- a/src/components/LoadingIndicator/LoadingIndicator.module.css
+++ b/src/components/LoadingIndicator/LoadingIndicator.module.css
@@ -1,0 +1,21 @@
+/*------------------------------------*\
+    # LOADINGINDICATOR
+\*------------------------------------*/
+
+/**
+ * LoadingIndicator
+ */
+.loading-indicator {
+  background-color: transparent;
+}
+
+/* override for `color` */
+.loading-indicator > svg path {
+  stroke: var(--eds-theme-color-background-brand-primary-strong);
+}
+
+/* override for `secondaryColor` */
+.loading-indicator > svg circle {
+  stroke: var(--eds-theme-color-background-neutral-medium);
+  stroke-opacity: 1;
+}
--- a/src/components/LoadingIndicator/LoadingIndicator.stories.ts
+++ b/src/components/LoadingIndicator/LoadingIndicator.stories.ts
@@ -1,0 +1,48 @@
+import { BADGE } from '@geometricpanda/storybook-addon-badges';
+import type { StoryObj, Meta } from '@storybook/react';
+import type React from 'react';
+
+import { LoadingIndicator } from './LoadingIndicator';
+
+export default {
+  title: 'Atoms/LoadingIndicator',
+  component: LoadingIndicator,
+  parameters: {
+    layout: 'centered',
+    badges: [BADGE.BETA],
+  },
+} as Meta<Args>;
+
+type Args = React.ComponentProps<typeof LoadingIndicator>;
+
+export const Default: StoryObj<Args> = {};
+
+export const Small: StoryObj<Args> = {
+  args: {
+    size: 'sm',
+  },
+};
+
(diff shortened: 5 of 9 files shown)


src/components/Loader/Loader.tsx
```diff
@@ -0,0 +1,76 @@
+import clsx from 'clsx';
+import React from 'react';
+
+import { Oval } from 'react-loader-spinner';
+import styles from './Loader.module.css';
+
+// TODO: some props are shadowed from react-loader-spinner. Improve?
+export type Props = {
```
jeremiah-clothier, line 8: This seems fine because we can also document the props ourselves with comments like you've done. Also, the types for `ariaLabel` or `visible` are straightforward

jeremiah-clothier (dismissed):
I'm surprised chromatic handles the animation consistently and doesn't flake

src/index.ts
```diff
@@ -78,6 +78,7 @@ export { default as SearchBar } from './components/SearchBar';
 export { default as SearchButton } from './components/SearchButton';
 export { default as SearchField } from './components/SearchField';
 export { default as Section } from './components/Section';
+export { default as Select } from './components/Select';
```
jinlee93, line 81: Nice lol

jinlee93:
Can I ask why this library for spinners over the others e.g. `react-spinners`?

jinlee93:
The small seems very err small for lack of a better word.
image
Are there any accessibility requirements for loaders?

src/components/Loader/Loader.stories.ts
```diff
@@ -0,0 +1,49 @@
...
+  component: Loader,
+  parameters: {
+    badges: [BADGE.BETA],
+  },
+} as Meta<Args>;
+
+type Args = React.ComponentProps<typeof Loader>;
+
+export const Default: StoryObj<Args> = {
+  args: {},
```
jinlee93, line 18: nit: not needed 
```suggestion
```

src/components/Loader/Loader.tsx
```diff
@@ -0,0 +1,76 @@
...
+  md: 2,
+  lg: 3,
+};
+
+/**
+ * BETA: This component is still a work in progress and is subject to change.
+ *
+ * `import {Loader} from "@chanzuckerberg/eds";`
+ *
+ */
```
jinlee93, line 56: Think we generally include a short blurb about the component here in the docstring

src/components/Loader/Loader.tsx
```diff
@@ -0,0 +1,76 @@
...
+   * Layout size of the loader. This affects the overall size and associated
+   * stroke width.
+   */
+  size?: 'sm' | 'md' | 'lg';
+
+  /**
+   * Aria label of the oval. Default is oval-loading. Will be overridden if ariaLabel is passed in props
+   * (Shadowed from OvalProps)
+   *
+   * @see https://mhnpd.github.io/react-loader-spinner/docs/components/oval
```
jinlee93, line 24: Unfortunately these jsdoc tags are not super compatible with Storybook, so these urls are not visible in the docs tab of the story page.
image

src/index.ts
jinlee93: Btw our plop command will generate recommended files and insert this line for you i.e. `yarn plop Loader` or `yarn plop Select`

booc0mtaco: lolz, yeah. For that select component I made a copy of the Dropdown directory, so bypassed plop in that case

src/components/Loader/Loader.tsx
booc0mtaco: Yah forgot to add something as a little starter. We will have some revisions to this once Sean gets a design up for comparison.

src/components/Loader/Loader.tsx
booc0mtaco: Sad. While they won't show in Storybook (yet?), they do show up in the IDE, so an engineer can see the doc.s and have an active link while doing their work. 

Screen Shot 2022-12-08 at 11 35 50

src/components/Loader/Loader.tsx
booc0mtaco: yah, we can always revisit this in case there's a much better way to pluck them from the spinner type directly. I did try to do a type union, but there's some issue with the lib since it doesn't export the interface for `OvalProps` (plus, we don't want to export all the props anyhow... this component is a subset that may change in the future)

src/components/Loader/Loader.tsx
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * CSS class names that can be appended to the component.
+   */
+  className?: string;
+
+  /**
+   * Layout size of the loader. This affects the overall size and associated
+   * stroke width.
+   */
+  size?: 'sm' | 'md' | 'lg';
```
booc0mtaco, line 18: I am thinking we may want to have some EDS utility types for stuff like this. Codify all the well-known props so they don't vary unexpectedly. This one in particular might not be ideal tho, as not every component has all three sizes. Food for thought 🍔

booc0mtaco:
@jinlee93 

* This library had a good variety of options AND was a bit smaller when unpacked, so went with it. When comparing the two components that had a similar style, this one came with a good set of props to cover the use cases Sean and I discussed as well. If we ever decide we don't want any of the other spinner types, we should probably pull out the SVG from this lib, as that extra bloat won't ever be needed. Lastly, `react-spinners` uses regular DOM elements `<span>` and CSS inline animations, while this one uses SVGs with relevant animation tags so it's a bit more self-contained.
* I made the sizes a little small on purpose to prompt a discussion on what the stroke sizes should be. Since the design is still pending, we can hold this PR until that is confirmed OR merge it in, then modify to suit. a11y-wise, we do have the label on this particular thing (and we should likely be using an `aria-live` somewhere when this component is rendered versus dismissed, but that's outside the scope of this one)

@jeremiah-clothier 

* I was curious about that too, so made a few pushes to see if the snapshots were stable. I think they are b/c the animation is triggered inside the SVG instead of via JS or thru CSS animations. Were it CSS, it might behave different based on the test page being `ready`, and JS obvs would be flakey since it might fire ... or not.

src/components/Loader/Loader.tsx
jinlee93: Makes sense to me, thanks for addressing

jinlee93 (approved):
Looks good pending mock updates and stroke width / general approval from design / Sean

src/components/Loader/Loader.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * BETA: This component is still a work in progress and is subject to change.
+ *
+ * `import {Loader} from "@chanzuckerberg/eds";`
+ *
+ * Show a loading state within a component container, or a specific page or area, where
+ * arbitrary content will appear once some request has completed.
+ */
+export const Loader = ({
```
dierat, line 58: `Loader` seems really vague to me and could mean several things (like a component that fetches data); I'd really prefer something like `LoadingSpinner`.

dierat:
What's the screen reader experience like?

src/components/Loader/Loader.tsx
booc0mtaco: @dierat fair point. I have used `LoadingIndicator` in the past, which is less ambiguous and much less likely to conflict with other stuff.

I'm waiting for a figma before merging this so once we go thru that, I will make the naming update

booc0mtaco:
> What's the screen reader experience like?

I am awaiting a POV from design on this one. I can imagine a few possible scenarios: the indicators are `aria-hidden` and designs include some text that gets read out, or the current `aria-label` exists as the sole representation of what is happening in the UI. Maybe some other options?

Related: there is probably a better default `aria-label` to apply than what the base component outputs (e.g., change the default to "Loading, please wait." or similar)

jeremiah-clothier (approved):

src/components/LoadingIndicator/LoadingIndicator.tsx
```diff
@@ -0,0 +1,81 @@
...
+  className?: string;
+
+  /**
+   * Layout size of the loader. This affects the overall size and associated
+   * stroke width.
+   */
+  size?: 'sm' | 'md' | 'lg';
+
+  /**
+   * Aria label of the oval. Default is oval-loading. Will be overridden if ariaLabel is passed in props
```
dierat, line 20: Can we use a different default aria label? "oval loading" sounds pretty weird. Just "loading" would be better.

src/components/LoadingIndicator/LoadingIndicator.tsx
```diff
@@ -0,0 +1,81 @@
...
+   * (Shadowed from OvalProps)
+   *
+   * @see https://mhnpd.github.io/react-loader-spinner/docs/components/oval
```
dierat, line 23: Do we need the `@see` jsdoc tag here? It'd be nice to have this url show up on the storybook docs page.

src/components/LoadingIndicator/LoadingIndicator.tsx
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * Whether the oval is visible. Default is true.
+   * (Shadowed from OvalProps)
+   *
+   * @see https://mhnpd.github.io/react-loader-spinner/docs/components/oval
+   */
+  visible?: boolean;
+};
+
+// Pixel sizes corresponding to EDS size units (pixel = rem * 16)
```
dierat, line 36: I don't understand what the `(pixel = rem * 16)` part means. Are you saying a rem is 16 pixels?

src/components/LoadingIndicator/LoadingIndicator.tsx
booc0mtaco: Yah, certainly. This default comes from the library itself, but doesn't make much sense

src/components/LoadingIndicator/LoadingIndicator.tsx
booc0mtaco: Interesting, i didn't realize the `@see` was causing it to be hidden by storybook. If i remove that jsdoc directive, it will show up in storybook? I can test that.

src/components/LoadingIndicator/LoadingIndicator.tsx
booc0mtaco: Yep that's it. I can actually remove that and let the size comments tell people where to find the value

src/components/LoadingIndicator/LoadingIndicator.tsx
booc0mtaco: Screen Shot 2022-12-16 at 11 04 13

dierat: Yeah for some reason storybook just cannot handle those `@` tags and won't render it or anything after it 🤷

jinlee93:
@booc0mtaco Just saw that we had a spinner [Icon](https://chanzuckerberg.github.io/edu-design-system/?path=/story/atoms-icons-icon--medium&args=name:spinner) and is being used in the `loading` variant of the [`Button`](https://chanzuckerberg.github.io/edu-design-system/?path=/story/molecules-buttons-button--loading). However, a quick regex search thru LP suggests this variant `<Button[\S\s*]*loading` isn't being used.

PR #1425, opened 2022-12-08
fix(deploy): apply workaround for node 18

jinlee93:
### Summary:
- `deploy:docs` command also broke with node 18, applies workaround similar to running storybook
### Test Plan:
- `deploy:docs` runs without erroring

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@chanzuckerberg/eds",
-  "version": "7.0.0",
+  "version": "7.1.0",
   "description": "The React-powered design system library for Chan Zuckerberg Initiative education web applications",
   "author": "CZI <edu-frontend-infra@chanzuckerberg.com>",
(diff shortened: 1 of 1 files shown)


jeremiah-clothier (approved):
LGTM.

Is there no github check that this command works which would prevent the PR from merging?

jinlee93:
> LGTM.
> 
> Is there no github check that this command works which would prevent the PR from merging?

Not as of now... not sure how that would be implemented besides actually running the command

jinlee93:
> LGTM.
> 
> Is there no github check that this command works which would prevent the PR from merging?

I think generally this wouldn't have happened with the [node upgrade](https://github.com/webpack/webpack/issues/14532#issuecomment-947807590) with better comms about the deprecation. Will land for now to unblock main build, but created a [ticket](https://czi-tech.atlassian.net/browse/EDS-767) to track

PR #1426, opened 2022-12-08
Make plop templates closer to linted status

jinlee93:
### Summary:
- Got annoyed of red squiggles for autogenerated files
- There are more lint errors when using plop but those are intentional
### Test Plan:
- `yarn plop` doesn't have related lint errors


booc0mtaco (approved):
I meant to do this as well! A few other things to check:

* is a similar change needed in the .stories file?
* some of the default files complain about the TSX extension since what gets plopped has no JSX in it. maybe the files should start out as TS files?

jinlee93:
> * is a similar change needed in the .stories file?

I think this is dependent on the next question haha

> * some of the default files complain about the TSX extension since what gets plopped has no JSX in it. maybe the files should start out as TS files?

Good question. For test files, will change to TS file.

For story files, I think most of the stories end up being TSX files but in the dev workflow, stories are written later... As a result I didn't touch it. How do you feel @booc0mtaco

dierat (approved):
Thanks for updating!

booc0mtaco:
> For story files, I think most of the stories end up being TSX files but in the dev workflow, stories are written later... As a result I didn't touch it. How do you feel @~booc0mtaco

This makes sense, tho my only remaining gripe is that the lint rules fills this whole file with red squigglies immediately, so the experience is a little jarring esp. for newer folk or less experienced eng. That said, this. would force people do add a JSX story before proceeding which might be good :)

PR #1428, opened 2022-12-09
chore: remove utility classes except typography

dierat:
### Summary:
In the Big Medium project, utility classes were introduced. These classes were intended to be used inline in storybook pages that showed EDS components in a page context. Using these classes allowed the developers to easily add spacing and typography changes without having to add a CSS module file.

Later, we reintroduced Tailwind, to be used in storybook pages instead of utility classes. These are an improvements for us because the class names are shorter, you don't have to import them into the file, we don't have to maintain them, and this is the same pattern used in `traject`.

Usage of utility classes were already removed in those storybook pages, so now we just have to remove the classes themselves.

This PR:
- removes remaining imports of the utility class files into storybook pages that were already updated to not use the classes anymore
- remove the storybook pages displaying these utility classes
- remove the utility classes
- update documentation that refers to utility classes

There is one exception here, which is typography. The typography utility classes are being used in the storybook pages for the type ramp. I'll need to update those pages before removing the typography classes, so I'll leave that for a separate PR.

### Test Plan:
- Atoms/Utilities only has typography classes in it now
- no changes registered in chromatic for remaining pages
- no lint errors around missing imports


docs/LAYOUT.md
```diff
@@ -89,11 +87,4 @@ However, it's important to be able to control spacing between components, so EDS
 
 ### Adding or Modifying Layouts or Grids
 
-`Layout` and `Grid` components should contain the minimum number of patterns to build most product layouts and compositions. Adding additional variants should be handled with careful consideration.
-
-1. Validate the need for adding an additional layout/variant.
-2.
```
dierat, line 95: I thought this was really weird, and I don't know if the author had a plan for what #2 would say, so I collapsed this line into the paragraph above.

dierat:
Jira ticket: https://czi-tech.atlassian.net/browse/EDS-257

jinlee93 (approved):
Thanks for these changes. Feels good to be removing these finally

PR #1430, opened 2022-12-09
chore: remove typography utility classes

dierat:
### Summary:
Ticket: https://czi-tech.atlassian.net/browse/EDS-257

Follow-up to https://github.com/chanzuckerberg/edu-design-system/pull/1428, in which I removed all the utility classes except typography. This PR removes the typography utility classes.

Because the utility classes were being used to make the typography tokens pages in storybook, I
- moved the styles into the tokens directory
- converted them into `.modules.css` files
- updated the imports
- deleted the pages that were displaying the typography utility classes.
- renamed the classes to make more sense now that they're no longer utility classes

So the classes kind of still exist but now they're only intended to be used for displaying the type ramp.

### Test Plan:
- Atoms/Utilities is missing in storybook
- type ramp pages in storybook are unchanged
- no lint errors around missing imports


jinlee93 (approved):
Also makes sense to remove typography utility classes, thanks

PR #1431, opened 2022-12-10
Use rollup to build JS and styles

ahuth:
Don't merge unless we need EDS support for Remix ASAP

### Summary:

Use Rollup to build the JS and styles, to better support server-side rendering (such as with Remix apps).

Basically this lets us produce a single .css file, instead of delegating css-module processing to consuming apps.

Once Remix supports css modules natively (see https://github.com/remix-run/remix/discussions/4631) we won't need this.

### Test Plan:

- [ ] Publish an alpha version and test in LP
- [ ] Publish an alpha version and test in Remix


rollup.config.mjs
```diff
@@ -0,0 +1,24 @@
...
+    dir: 'lib',
+    format: 'es',
+    preserveModules: true,
+    preserveModulesRoot: 'src',
+  },
+  external: ['../../icons/spritemap/spritemap.svg', 'react-children-by-type'],
+  plugins: [
+    commonjs(),
+    postcss({
+      extract: true,
```
ahuth, line 17: Extract all the styles to a single .css file.

This should be a comment.

rollup.config.mjs
```diff
@@ -0,0 +1,24 @@
...
+      modules: true,
```
ahuth, line 18: Enable css modules. Should also be a comment.

rollup.config.mjs
```diff
@@ -0,0 +1,24 @@
...
+  input: 'src/index.ts',
+  output: {
+    dir: 'lib',
+    format: 'es',
+    preserveModules: true,
+    preserveModulesRoot: 'src',
+  },
+  external: ['../../icons/spritemap/spritemap.svg', 'react-children-by-type'],
+  plugins: [
+    commonjs(),
```
ahuth, line 15: Convert some CJS modules into ESM. Needed so that rollup can handle some of the node_modules deps.

I think an alternative would be to compile EDS to CJS instead of ESM (wow, those acronyms).

README.md
```diff
@@ -16,9 +16,11 @@ npm install --save @chanzuckerberg/eds
 yarn add @chanzuckerberg/eds
 ```
 
-Import the CSS tokens somewhere in your app, e.g. an `init.ts` or `app.ts` file:
+Import the CSS stylesheet and tokens somewhere in your app, e.g. an `init.ts` or `app.ts` file:
 
 ```js
+import '@chanzuckerberg/eds/lib/index.css';
```
ahuth, line 22: One downside of this is you have to import ALL the css, even if you only use one component.

I suspect that's not a big deal for our use case, but need to do some testing to see if unused styles get removed.

ahuth:
Closing. I don't think this will be necessary because Remix is implementing css module support in https://github.com/remix-run/remix/pull/4852.

I will say, though, that Rollup was quite nice to work with. (Other than its postcss plugin seems inactive).

README.md
ahuth: Coming back to this, I think it'll be the same result in Remix apps regardless of whether we're shipping css modules or bundled css.

PR #1432, opened 2022-12-10
feat(toggle): bring to eds 

jinlee93:
### Summary:
[mocks](https://www.figma.com/file/kQiv4yCKXgJXI804sGd9Kh/Toggles?node-id=8396%3A23748&t=4qc7uN7Z6V7HtXUJ-4)
[EDS-765]
- motivation: toggle is an interactive component like a checkbox to gather input regarding two states, generally a checked and an unchecked state. Pretty commonly used, we should have it in our design system.
- implementation:
  - Headless Switch wrapper
  - Circle indicator of toggle status translates between left and right
  - Background color changes between brand and neutral
  - Interaction with label also works

Still need to sync with design about toggle width, is 48 px here and 55 px in mocks and the cursor for the disabled state
- [x] synced, made 56px to match mocks

Default:
Screen Shot 2022-12-09 at 9 11 05 PM
Enabled:
Screen Shot 2022-12-09 at 9 11 07 PM
Disabled:
Screen Shot 2022-12-09 at 9 11 08 PM
No visible label:
Screen Shot 2022-12-09 at 9 11 10 PM
Custom:
Screen Shot 2022-12-09 at 9 11 12 PM

### Test Plan:
- toggle works in stories
- no unit tests written since we would be testing Headless functionality

[EDS-765]: https://czi-tech.atlassian.net/browse/EDS-765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,116 @@
...
+      className,
+    )}
+    {...other}
+  >
+    {/* This span is the circular part of the toggle */}
+    <span className={styles['toggle__button-circle']} />
+  </Switch>
+);
+
+const ToggleWrapper = Switch.Group;
```
jinlee93, line 75: In order to use the `<Switch.Label>` headlessui component, it needs to be wrapped in this `Switch.Group` component

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * @example
+ * <Toggle.Wrapper as="Fragment" className={styles.customWrapperStyles}>
+ *   <Toggle.Label> Some label </Toggle.Label>
+ *   <Toggle.Button onChange={onChange} checked={checked} className={customCssModulesClassname} />
+ * </Toggle.Wrapper>
+ *
+ */
+export const Toggle = ({ label, ...other }: ToggleProps) => {
+  // TODO-JL: Consider if (children) return <Switch.Group>{children}</Switch.Group>
```
jinlee93, line 96: Should this component be improved for dev experience where we check for presence of `children` and it automatically detects custom Toggle use? i.e.:
right now:
```tsx
<Toggle.Wrapper>
  <Toggle.Button>
  <Toggle.Label>
</Toggle.Wrapper>
```
but with detection of children we can do
```tsx
<Toggle.Wrapper>
  <Toggle.Button>
  <Toggle.Label>
</Toggle.Wrapper>
```
via something like
```tsx
if (children)
return <ToggleWrapper>{children}</Toggle>
```

but the logic there regarding types gets complicated, and using `<ToggleWrapper>` not that big of a lift for consumers anyways, and probably discourages customization.

Lmk any thoughts

src/components/Toggle/Toggle.tsx
dierat: Maybe add this as a code comment?

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,116 @@
...
+ * @example
+ * <Toggle checked={checked} label="Lorem Ipsum" onChange={onChange} />
+ *
+ * @example
+ * <Toggle.Wrapper as="Fragment" className={styles.customWrapperStyles}>
+ *   <Toggle.Label> Some label </Toggle.Label>
+ *   <Toggle.Button onChange={onChange} checked={checked} className={customCssModulesClassname} />
+ * </Toggle.Wrapper>
+ *
+ */
```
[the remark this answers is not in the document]
dierat: They seem to be stopping the examples from appearing on the docs tab in storybook. Or did you not intend for them to be visible there?
docs tab of toggle story in storybook

src/components/Toggle/Toggle.tsx
dierat: I'm fine with `<Toggle.Wrapper>` 🤷 I can see it possibly being confusing at first because I might expect `Toggle.Wrapper` to wrap `Toggle`, like `<Toggle.Wrapper><Toggle></Toggle></Toggle.Wrapper>`, but I don't think it's too bad.

src/components/Toggle/Toggle.module.css
```diff
@@ -0,0 +1,85 @@
+@import '../../design-tokens/mixins.css';
+
+.toggle__button {
+  border: 0;
```
dierat, line 4: Could we get a transparent border instead of no border? I'm concerned about whether this thing would show up in windows high contrast mode (where background colors become white but borders become 100% opacity black).

src/components/Toggle/Toggle.module.css
```diff
@@ -0,0 +1,85 @@
...
+}
+
+.toggle__wrapper {
+  display: flex;
+  align-items: center;
+
+  gap: var(--eds-size-1);
+}
+
+.toggle__button-circle {
```
dierat, line 62: I'm also wondering if we should have a transparent border on the circle for the same reason mentioned on the button.

src/components/Toggle/Toggle.module.css
booc0mtaco: Using a transparent border also helps prevents components giggling around between states (e.g., adding/removing that border can cause it to move around)

booc0mtaco:
@jinlee93 What does the focus state look like?

src/index.ts
```diff
@@ -78,6 +77,7 @@ export { default as SearchBar } from './components/SearchBar';
 export { default as SearchButton } from './components/SearchButton';
 export { default as SearchField } from './components/SearchField';
 export { default as Section } from './components/Section';
+export { default as Select } from './components/Select';
```
booc0mtaco, line 80: 😭

jinlee93:
> @jinlee93 What does the focus state look like?

Screen Shot 2022-12-12 at 9 18 10 AM
Screen Shot 2022-12-12 at 9 18 13 AM

src/components/Toggle/Toggle.module.css
jinlee93: I can add since visually wouldn't make a difference

src/components/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  return label ? (
+    <ToggleWrapper as="div" className={styles['toggle__wrapper']}>
+      <ToggleButton {...other} />
+      <ToggleLabel>{label}</ToggleLabel>
+    </ToggleWrapper>
+  ) : (
+    <ToggleButton {...other} />
+  );
```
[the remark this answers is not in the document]
jinlee93: The label also renders with `label` prop as text, and seems the `label` prop has priority over `aria-label` for the accessible name since it is implemented via `aria-labelledby`
image

I haven't thought about expected behavior but I am fine with this since 1) it is expected via `aria-labelledby` overriding other names and 2) the visible label being the accessible name makes sense

src/components/Toggle/Toggle.module.css
```diff
@@ -0,0 +1,86 @@
+@import '../../design-tokens/mixins.css';
+
+.toggle__button {
+  border: var(--eds-theme-border-width) solid transparent;
```
dierat, line 4: Could you add comments to the transparent border styling?

dierat (approved):

PR #1434, opened 2022-12-12
fix(colors): align link color with figma

jinlee93:
### Summary:
[EDS-770]
Noticed while pairing with Sean about toggle that the icon link color was not aligned between figma and code, updating to match.
### Test Plan:
There shouldn't be any components affected since Toggle is the first to use these colors

[EDS-770]: https://czi-tech.atlassian.net/browse/EDS-770?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


booc0mtaco (approved):
In the new year, we should sync with design on the token values, as I have also seen a few mismatches and there's fuzziness around which is correct (figma or code)

PR #1435, opened 2022-12-12
fix: apply consistency updates to popovers

booc0mtaco:
### Summary:

align styles, layout, and focus changes to `Menu`, `Select` and their shared sub-components. make sure the spacing works the same between the components, their text and icons, and that they match the mocks where noted. This is applying design feedback to the current implementations, but changes no behavioral details.

### Test Plan:

- [x] update snapshots
- [x] review chromatic differences
- [x] compare to figma branch

Select Component (before/after):
Screen Shot 2022-12-12 at 17 47 07

Screen Shot 2022-12-12 at 17 44 33

Menu Component (before/after): 
Screen Shot 2022-12-12 at 17 46 42

Screen Shot 2022-12-12 at 17 44 44

(essentially the "after" pictures show how this updates the two components to remove unwanted inconsistencies)

### DIFF
--- a/src/components/Menu/Menu.module.css
+++ b/src/components/Menu/Menu.module.css
@@ -1,3 +1,5 @@
+@import '../../design-tokens/mixins.css';
+
 /*------------------------------------*\
     # MENU
 \*------------------------------------*/
@@ -18,7 +20,12 @@
 }
 
 .menu__button.menu__button {
+  @mixin eds-theme-typography-body-text-md;
+
+  color: var(--eds-theme-text-neutral-subtle);
   background-color: var(--eds-theme-color-form-background);
+  border-color: var(--eds-theme-color-form-border);
+  font-weight: var(--eds-font-weight-light);
 }
 
 /* Needed to create higher specificity than ClickableStyle. TODO: improve */
--- a/src/components/PopoverContainer/PopoverContainer.module.css
+++ b/src/components/PopoverContainer/PopoverContainer.module.css
@@ -1,3 +1,5 @@
+@import '../../design-tokens/mixins.css';
+
 /*------------------------------------*\
  # POPOVER CONTAINER
 \*------------------------------------*/
@@ -10,4 +12,8 @@
   overflow: auto;
   box-shadow: var(--eds-box-shadow-md);
   background-color: var(--eds-theme-color-background-neutral-default);
+
+  &:focus-visible {
+    @mixin focus;
+  }
 }
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -21,6 +21,7 @@ export type MenuProps = ExtractProps<typeof HeadlessMenu> & {
 };
 
 export type MenuItemProps = ExtractProps<typeof HeadlessMenu.Item> & {
+  className?: string;
   href?: string;
   icon?: IconName;
   onClick?: MouseEventHandler<HTMLAnchorElement>;
@@ -60,7 +61,7 @@ const MenuButton = ({ children, className, ...other }: MenuButtonProps) => {
         className={styles['menu__button--icon']}
         name="expand-more"
         purpose="decorative"
-        size="1.5rem"
+        size="1.25rem"
       />
     </HeadlessMenu.Button>
   );
@@ -85,6 +86,7 @@ const MenuItems = (props: MenuItemsProps) => (
  */
 const MenuItem = ({
   children,
+  className,
   href,
   icon,
   onClick,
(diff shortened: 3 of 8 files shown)


src/components/PopoverContainer/PopoverContainer.module.css
```diff
@@ -10,4 +12,8 @@
   overflow: auto;
   box-shadow: var(--eds-box-shadow-md);
   background-color: var(--eds-theme-color-background-neutral-default);
+
+  &:focus-visible {
```
booc0mtaco, line 16: by applying focus rings to the shared component, any future components we use can make use of this via the `as` prop. All Headless UI components work like this, so they get proper focus ring behavior thru this now

src/components/Menu/Menu.module.css
```diff
@@ -18,7 +20,12 @@
 }
 
 .menu__button.menu__button {
+  @mixin eds-theme-typography-body-text-md;
```
booc0mtaco, line 23: some of the styles being overridden are b/c `ClickableStyle` has a few things that conflict here

timzchang: Q for my learning: what does @mixin do here? A quick google mentions sass mixins but I don't think that's what this is.

timzchang (approved):
looks good to me

src/components/Menu/Menu.module.css
booc0mtaco: I'll ask @dierat / @jinlee93 to add context as I'm actually not sure how this is made to work using CSS Modules 🤔

If it works like Sass would, it basically imports a snippet of css properties and values into the current rule. I think if you search for `@define-mixin` you can see what properties it will grab

src/components/PopoverContainer/PopoverContainer.module.css
```diff
@@ -10,4 +12,8 @@
...
+    @mixin focus;
+  }
```
jinlee93, line 18: We actually [need to have a fallback](https://github.com/chanzuckerberg/edu-design-system/blob/main/docs/CODE_GUIDELINES.md#focus-states) for `focus-visible` use since we do support some browser versions that don't support `focus-visible`

src/components/Menu/Menu.module.css
jinlee93: Yea, this specific mixin is defined in [`typography-usage.css`](https://github.com/chanzuckerberg/edu-design-system/blob/next/src/design-tokens/tier-2-usage/typography-usage.css). 

It is sass like lang provided via [postcss](https://github.com/postcss/postcss-mixins) for bundling styling together. Being said, we can probably start using the [Tailwind @layer directive](https://tailwindcss.com/docs/functions-and-directives#layer) instead so we don't have confusion like this and reduce our dependency count.

I can investigate that

src/components/PopoverContainer/PopoverContainer.module.css
booc0mtaco: I may need to revisit how this [polyfill](https://www.npmjs.com/package/focus-visible) works , and what it means for things we style in EDS and consuming apps. I'll look at other examples to see what we should do.

What's the browser we support that doesn't do this? https://caniuse.com/mdn-css_selectors_focus-visible

jinlee93: Ah completely forgot we had the polyfill for `focus-visible`, think it should work then

jinlee93: Looking at the list of [supported browsers](https://czi.atlassian.net/wiki/spaces/LP/pages/2057733851/Edu+Tech+Inclusive+Browser+Support+Policy) I think it's Edge 84-85 and Safari 14-15.3, and hopefully firefox gets autoprefixed for it

jinlee93 (approved):
Looks good to me!

src/components/PopoverContainer/PopoverContainer.module.css
dierat: The polyfill is in `traject` and `along`, not EDS, so we have been adding the fallback for `:focus-visible`. I think at the time we decided to use `:focus-visible` + the `@supports` fallback, the polyfill wasn't in `traject` yet. I would make the case that we should keep the fallback because we don't know if the app using EDS has the polyfill, but if you think we should drop the fallback because the polyfill is now in place in our 2 current apps, we should update the docs and code in EDS.

https://github.com/chanzuckerberg/edu-design-system/blob/e0288ce88d4cccbf5a1e3fc5cd79b5b8ee0ce4fd/docs/CODE_GUIDELINES.md?plain=1#L121

booc0mtaco: @dierat good note. I will make a ticket for fixing this. as those new pilot get set up, there'd be an extra step for each of them to remember to add that polyfill, set it up properly, etc. yuck!

PR #1437, opened 2022-12-14
refactor(Popover): sync popover with menu and select components

booc0mtaco:
### Summary:

- use `PopoverContainer` and `PopoverListItem` in the existing `Popover` Sub-components
- Merge the `PopoverContent` and `PopoverButton` components into `Popover` (similar to `Menu` and `Select`)
- update the documentation to use the `"fpo"` blocks 

### Test Plan:

- [x] make sure the existing tests continue to pass for `Popover`
- [x] chromatic tests pass


booc0mtaco:
**Pending**:

* remove the separate sub-components
* add the subcomponents to the story file for documentation purposes (similar to select)

src/components/Menu/Menu.stories.tsx
```diff
@@ -3,6 +3,7 @@ import type { StoryObj, Meta } from '@storybook/react';
 import React from 'react';
 
 import { Menu } from './Menu';
+import type { MenuProps } from './Menu';
```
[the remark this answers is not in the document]
booc0mtaco: good question. using the extraction was causing a collision with a named private interface in HeadlessUI code (something like `MenuPropsArgs` or so). By using the already-defined props, it avoids this reuse of the interface. 

I think it would only apply to HeadlessUI. so weird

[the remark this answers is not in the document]
booc0mtaco: good question. using the extraction was causing a collision with a named private interface in HeadlessUI code (something like `MenuPropsArgs` or so). By using the already-defined props, it avoids this reuse of the interface. 

I think it would only apply to HeadlessUI. so weird

src/components/Popover/Popover.tsx
```diff
@@ -16,19 +17,8 @@ export type PopoverProps = ExtractProps<typeof HeadlessPopover> & {
...
-    | 'bottom-end'
-    | 'right-start'
-    | 'right-end'
-    | 'left-start'
-    | 'left-end'
-    | 'top'
-    | 'right'
-    | 'bottom'
-    | 'left';
+  placement?: PopperJSOptions['placement'];
```
jinlee93, line 20: Iirc I had this written out since the popover did some funky things with placement: `auto` or `auto-start` or `auto-end` (sorry forget which)

src/components/Popover/Popover.stories.tsx
```diff
@@ -23,19 +30,14 @@ export default {
           Open Popover
         </Popover.Button>
         <Popover.Content data-testid="popover-content">
-          <div style={{ width: '250px' }}>
-            My popover content Lorem ipsum dolor sit amet, consectetur
-            adipiscing elit, sed do eiusmod
-          </div>
+          <div className="fpo m-2 p-6">Popover Content goes here</div>
```
jinlee93, line 33: Oh snap I like the `fpo` use

src/index.ts
```diff
@@ -65,8 +65,6 @@ export { default as PageHeader } from './components/PageHeader';
 export { default as PageLevelBanner } from './components/PageLevelBanner';
 export { default as Panel } from './components/Panel';
 export { default as Popover } from './components/Popover';
-export { default as PopoverButton } from './components/PopoverButton';
```
jinlee93, line 68: Nice, as subcomponents these shouldn't have been default exported in the first place

src/components/Popover/Popover.tsx
```diff
@@ -1,12 +1,13 @@
 import { Popover as HeadlessPopover } from '@headlessui/react';
 import type { Options as PopperJSOptions } from '@popperjs/core';
+import clsx from 'clsx';
 import * as React from 'react';
```
jinlee93, line 4: nit, while you're in here:
```suggestion
import * as React from 'react';
```
just making it look like the other components lol

jinlee93 (approved):
Looks good pending one props q

src/components/Popover/Popover.tsx
booc0mtaco: Lemme test this out today. I can restore if it is misbehaving

booc0mtaco: @jinlee93 So the popper JS documentation is a little undercooked for my tastes. It doesn't explain how the modifiers and placement work. The behavior works as documented but by default isn't desirable for popovers like select or menus (it will open to whichever side has the most clear space, meaning the left and right WAY more than we'd want).

I'll create a ticket to use the modifiers for Popper to only filp between top and bottom for the components we want. This would be very desirable so that the popover can open downward or upward (another reason not to call them all dropDOWNs :))

jinlee93: That makes sense to me, thanks!

jinlee93 (approved):
Looks good!

PR #1438, opened 2022-12-15
Replace spritemap to support SSR

ahuth:
[EFI-827]

### Summary:

To support server-side rendering in Remix, this PR replaces the svg spritesheet (and its generation process)

```html
<!-- icons/spritemap/spritemap.svg -->
<svg xmlns="http://www.w3.org/2000/svg">
  <symbol id="add" viewbox="0 0 100 100"><path d="123" /></symbol>
  <symbol id="close" viewbox="0 0 100 100"><path d="456" /></symbol>
</svg>
```

with a JS object

```js
// icons/spritemap.tsx
export default {
  add: {
    viewBox: "0 0 100 100",
    content: <path d="123" />,
  },
  close: {
    viewBox: "0 0 100 100",
    content: <path d="456" />,
  },
}
```

Making this change does a couple things for us:
- **_Works better with Remix's SSR_** - Remix can only really handle imports of JS files (at least until [css module support](https://github.com/remix-run/remix/pull/4852) lands). Now we'll no longer have an import to an SVG file. There will only be JS  and CSS file imports.
- **_No change to Icon's API_** - we'll still reference icons by name - e.g. `<Icon name="add" />`. There's no change to EDS's public API.

There are some downsides to this approach:
- **_Not tree-shakeable_** - (the spritemap.svg also has this issue) all icons end up Webpack bundles regardless of how many icons are actually used.
- **_Repeated svg paths_** - unlike the spritemap.svg solution, we'll now repeat the path data for every instance of an icon. The severity of this issue depends on which icons are used on a page, and how much they're repeated. Hilariously, repeating the bookmark icon definition on https://demo.summitlearning.org/v2/teacher/courses actually **_decreases_** page weight, because the icon is so simple.

**Note**: I don't think this is the optimal solution or final say in how EDS handles svgs. But it makes it work with SSR, and doesn't make things worse (and actually simplifies the build process). In the future we should talk some more about how we do want svgs to be handled (and consider breaking API changes).

### Test Plan:

- [x] CI
- [x] Local testing in Storybook
- [x] Compare sizes of contents in the package (they're almost exactly the same)
- [x] Publish an alpha version and see if anything dramatic happens with Traject bundle sizes - https://github.com/FB-PLP/traject/pull/13037

[EFI-827]: https://czi-tech.atlassian.net/browse/EFI-827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Icon/Icon.tsx
```diff
@@ -140,11 +138,12 @@ export const Icon = (props: IconProps) => {
...
     width: size,
     xmlns: 'http://www.w3.org/2000/svg',
-    viewBox,
+    viewBox: name ? icons[name].viewBox : viewBox,
   };
+
   // allow passing custom SVGs to render, otherwise
   // load from the spritemap of EDS icons
-  const computedSvg = children || <use xlinkHref={`${icons}#${name}`} />;
+  const computedSvg = name ? icons[name].content : children;
```
ahuth, line 146: Note the logic change here to match `viewBox` above.
- Before - `children` took precedence over `name` (if you provide both, you get `children`)
- Now - `name` takes precedence over `children` (if you provide  both, you get the named icon)

Can revert this if it seems risky or undesired.

package.json
```diff
@@ -164,8 +160,6 @@
     "style-dictionary": "^3.7.1",
     "stylelint": "^14.16.0",
     "stylelint-config-recommended": "^9.0.0",
-    "svg-sprite-generator": "^0.0.7",
```
jinlee93, line 167: Good to see this go

.github/renovate.json5
```diff
@@ -20,10 +20,7 @@
     fetchReleaseNotes: true,
   },
 
-  ignoreDeps: [
-    // svg-spritemap-webpack-plugin v4 requires webpack 5 but EDS currently uses webpack 4 for storybook
-    'svg-spritemap-webpack-plugin',
```
jinlee93, line 25: Nice

.storybook/main.js
```diff
@@ -42,19 +39,4 @@ module.exports = {
...
-          prefix: false,
-          generate: {
-            symbol: true,
-          },
-        },
-      },
-    );
-    config.plugins.push(SVGSpritesPlugin);
-    return config;
-  },
```
jinlee93, line 59: Oooh is the storybook startup faster now?

jinlee93 (approved):
I like it

src/icons/spritemap.tsx
```diff
@@ -0,0 +1,795 @@
+import React from 'react';
+
+export type IconName = keyof typeof icons;
+
+const icons = {
+  'account-circle': {
```
jeremiah-clothier, line 6: > Not tree-shakeable - (the spritemap.svg also has this issue) all icons end up Webpack bundles regardless of how many icons are actually used

Is this file tree-shakable if we export icons directly one-by-one. For example `export const accountCircle`, `export const addAlarm` instead of exporting one giant default export?

Someone with more insight on icons would have to verify if this is even possible. Im not sure if having the name be `account-circle` instead of `accountCircle` is important.

.github/renovate.json5
```diff
@@ -20,10 +20,7 @@
     fetchReleaseNotes: true,
   },
 
-  ignoreDeps: [
-    // svg-spritemap-webpack-plugin v4 requires webpack 5 but EDS currently uses webpack 4 for storybook
```
jeremiah-clothier, line 24: :o I didn't know we weren't using the webpack5 builder for storybook. Did we have difficulties switching over or have we just never bothered?

src/icons/spritemap.tsx
jeremiah-clothier: Actually looking more closely it seems like having one big object is important because consumers of the `<Icon />` will pass `name="account-circle"` which is how we know what icon to use.

I'll leave the comment up though

ahuth: > Is this file tree-shakable if we export icons directly one-by-one

Yes, but only if we change the `<Icon>` API (as you noted).

This PR is step 1 (which is no worse than the current state for the most part), and we should talk about whether we want follow on steps to support better tree shaking.

.github/renovate.json5
jinlee93: No difficulties switching over but EDS is currently bundled via `tsc` and didn't want to include webpack in the deps

jeremiah-clothier (approved):

.github/renovate.json5
ahuth: Should be able to use Webpack 5 without including the dep in the final package, but still may not be worth the effort to bring it in and configure.

booc0mtaco (approved):
This seems like good first step, with the one caveat that adding/updating icons requires a little more care than before. Definitely worth revisiting w.r.t. tree-shaking, as we might get new (or more?) icons some day

ahuth:
> with the one caveat that adding/updating icons requires a little more care than before

@booc0mtaco, can you expand on this a bit? I think it should be about the same as before, although slightly different process (add to spritemap.tsx instead of dropping svg into the directory).

booc0mtaco:
> > with the one caveat that adding/updating icons requires a little more care than before
> 
> @booc0mtaco, can you expand on this a bit? I think it should be about the same as before, although slightly different process (add to spritemap.tsx instead of dropping svg into the directory).

@ahuth Emphasis on "little" :) I imagine a case where some icon has a complex SVG structure (grouped paths), and someone trying to import it does something unintentional like copying only one of the paths into the TSX sprite sheet file. Our tooling and visual inspection should catch syntax missteps, but saving an svg file into a directory is even more trivial. We used to let an eng essentially ignore the structure of the SVG content.

I don't think it's a huge delta in difficulty tho

src/icons/spritemap.tsx
anniehu4: 2 crumbs for thought:
1. one of the (theoretical) benefits of the separate SVG files was designers could easily add/edit/view them from the codebase. they'll need to do this much less now, but we should still think about how to facilitate that in the new structure
2. do we need tree-shaking? I think the existing spritemap is 60kB large. Brad Frost made this post when we were deciding architecture (march 2022) and there were some suggestions for keeping the `<Icon>` dev-ex while also allowing for tree shaking: https://twitter.com/brad_frost/status/1509160842062929923

.storybook/main.js
ahuth: Not that I can tell, but I wasn't real scientific about it.

src/icons/spritemap.tsx
ahuth: Interesting twitter poll 🤔 

The developer experience of `<Icon name={...} />` is nice, especially with autocomplete.

Also 60kb (66 for the spritemap.tsx) isn't too bad. Maybe even insignificant in the grand scheme of things.

booc0mtaco: Yah i also think the DX of a `name` prop is very nice. Having done it both ways relatively recently, having separate components is especially useful when there are LOTS of possible complex icons and Illustrations, and it's known that a small subset of them might be used in any given application. That said, we also had a component like `<Icon />` that imported EVERY INDIVIDUAL ICON so ...

ahuth: Maybe the next level would be webpack + esbuild plugins that analyze which `name`s are actually used and prunes the rest 🤯

PR #1439, opened 2022-12-15
Remove prop types

ahuth:
### Summary:

This PR:
- Removes prop-types (noticed it was there while working on #1438)
- Converts the files that had prop-types from JS to TS
- Removes a couple that aren't actually used

### Test Plan:

- Type checking
- Local Storybook


.storybook/components/DesignTokens/Tier1/FontSizes.jsx
```diff
@@ -1,30 +0,0 @@
-import PropTypes from 'prop-types';
```
booc0mtaco, line 1: Were these ones being used b/c they aren't used anywhere?

booc0mtaco (approved):
nice catch 🎣

.storybook/components/DesignTokens/Tier1/FontSizes.jsx
ahuth: That's right, I can't find any usages of them anywhere.

Looking at c8a3bbc28fb688c94f99b0ff73a64bc4cda0b8d8, these were probably part of Big Medium's template.

ahuth:
Thanks

.storybook/components/DesignTokens/Tier1/FontSizes.jsx
jinlee93: I went back in time for a while and don't think either of these were being used

PR #1440, opened 2022-12-17
feat(badge): add badge component

jinlee93:
### Summary:
[EDS-769]
- motivation: need an indicator to visually notify for changes to an element
- execution: small pill shaped container is placed on the upper righthand corner of wrapped child component
  - What not to place into badges, however, is mostly reliant on documentation

Screen Shot 2022-12-16 at 4 03 09 PM
Screen Shot 2022-12-16 at 4 03 11 PM
Screen Shot 2022-12-16 at 4 03 15 PM
Screen Shot 2022-12-16 at 4 03 17 PM
Screen Shot 2022-12-16 at 4 03 22 PM
Screen Shot 2022-12-16 at 4 03 29 PM

### Test Plan:
- chromatic snapshots

[EDS-769]: https://czi-tech.atlassian.net/browse/EDS-769?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Badge/Badge.stories.tsx
+++ b/src/components/Badge/Badge.stories.tsx
@@ -8,6 +8,7 @@ export default {
   title: 'Molecules/Messaging/Badge',
   component: Badge,
   subcomponents: {
+    'Badge.Dot': Badge.Dot,
     'Badge.Icon': Badge.Icon,
     'Badge.Text': Badge.Text,
   },
@@ -19,16 +20,18 @@ export default {
 
 type Args = React.ComponentProps<typeof Badge>;
 
-export const Empty: StoryObj<Args> = {
+export const Dot: StoryObj<Args> = {
   args: {
-    empty: true,
     children: (
-      <div
-        aria-label="Ava has notification"
-        className="fpo flex items-center justify-center h-8 w-8"
-      >
-        Ava
-      </div>
+      <>
+        <div
+          aria-label="Ava has notification"
+          className="fpo flex items-center justify-center h-8 w-8"
+        >
+          Ava
+        </div>
+        <Badge.Dot />
+      </>
     ),
   },
 };
--- a/src/components/Badge/Badge.stories.tsx
+++ b/src/components/Badge/Badge.stories.tsx
@@ -3,14 +3,13 @@ import type { StoryObj, Meta } from '@storybook/react';
 
 import React from 'react';
 import { Badge } from './Badge';
-import Icon from '../Icon';
 
 export default {
-  // TODO-JL: Appropriate storybook folder?
   title: 'Molecules/Messaging/Badge',
   component: Badge,
   subcomponents: {
-    'Badge.Wrapper': Badge.Wrapper,
+    'Badge.Icon': Badge.Icon,
+    'Badge.Text': Badge.Text,
   },
   parameters: {
     layout: 'centered',
@@ -20,60 +19,79 @@ export default {
 
 type Args = React.ComponentProps<typeof Badge>;
 
-export const Default: StoryObj<Args> = {
-  render: () => (
-    <Badge.Wrapper>
+export const Empty: StoryObj<Args> = {
+  args: {
+    empty: true,
+    children: (
       <div className="fpo flex items-center justify-center h-8 w-8">Ava</div>
-      <Badge />
-    </Badge.Wrapper>
-  ),
+    ),
+  },
+};
+
+export const EmptyText: StoryObj<Args> = {
+  args: {
+    children: (
+      <>
+        <div className="fpo flex items-center justify-center h-8 w-8">Ava</div>
+        <Badge.Text>{''}</Badge.Text>
+      </>
+    ),
+  },
 };
 
 export const SmallNumber: StoryObj<Args> = {
-  render: () => (
-    <Badge.Wrapper>
-      <div className="fpo flex items-center justify-center h-8 w-8">Ava</div>
-      <Badge>1</Badge>
-    </Badge.Wrapper>
-  ),
+  args: {
+    children: (
+      <>
+        <div className="fpo flex items-center justify-center h-8 w-8">Ava</div>
+        <Badge.Text>1</Badge.Text>
+      </>
+    ),
+  },
 };
 
 export const LargeNumber: StoryObj<Args> = {
-  render: () => (
-    <Badge.Wrapper>
-      <div className="fpo flex items-center justify-center h-8 w-8">Ava</div>
-      <Badge>999</Badge>
-    </Badge.Wrapper>
-  ),
+  args: {
+    children: (
+      <>
+        <div className="fpo flex items-center justify-center h-8 w-8">Ava</div>
+        <Badge.Text>999</Badge.Text>
+      </>
+    ),
+  },
 };
 
 export const OverNineThousand: StoryObj<Args> = {
-  render: () => (
-    <Badge.Wrapper>
-      <div className="fpo flex items-center justify-center h-8 w-8">Ava</div>
-      <Badge>99+</Badge>
-    </Badge.Wrapper>
-  ),
+  args: {
+    children: (
+      <>
+        <div className="fpo flex items-center justify-center h-8 w-8">Ava</div>
+        <Badge.Text>99+</Badge.Text>
+      </>
+    ),
+  },
 };
 
 export const IconBadge: StoryObj<Args> = {
-  render: () => (
-    <Badge.Wrapper>
-      <div className="fpo flex items-center justify-center h-8 w-8">Ava</div>
-      <Badge>
-        <Icon name="alarm" purpose="decorative" size="1rem" />
-      </Badge>
-    </Badge.Wrapper>
-  ),
+  args: {
+    children: (
+      <>
+        <div className="fpo flex items-center justify-center h-8 w-8">Ava</div>
+        <Badge.Icon name="alarm" />
+      </>
+    ),
+  },
 };
 
 export const LargeBadgeableObject: StoryObj<Args> = {
-  render: () => (
-    <Badge.Wrapper>
-      <div className="fpo flex items-center justify-center w-96 h-96">
-        Badge-able Obj
-      </div>
-      <Badge>10</Badge>
-    </Badge.Wrapper>
-  ),
+  args: {
+    children: (
+      <>
+        <div className="fpo flex items-center justify-center w-96 h-96">
+          Badge-able Obj
+        </div>
+        <Badge.Text>10</Badge.Text>
+      </>
+    ),
+  },
 };
--- a/src/components/Badge/Badge.stories.tsx
+++ b/src/components/Badge/Badge.stories.tsx
@@ -23,7 +23,12 @@ export const Empty: StoryObj<Args> = {
   args: {
     empty: true,
     children: (
-      <div className="fpo flex items-center justify-center h-8 w-8">Ava</div>
+      <div
+        aria-label="Ava has notification"
+        className="fpo flex items-center justify-center h-8 w-8"
+      >
+        Ava
+      </div>
     ),
   },
 };
@@ -32,7 +37,12 @@ export const EmptyText: StoryObj<Args> = {
   args: {
     children: (
       <>
-        <div className="fpo flex items-center justify-center h-8 w-8">Ava</div>
+        <div
+          aria-label="Ava has notification"
+          className="fpo flex items-center justify-center h-8 w-8"
+        >
+          Ava
+        </div>
         <Badge.Text>{''}</Badge.Text>
       </>
     ),
@@ -43,7 +53,12 @@ export const SmallNumber: StoryObj<Args> = {
   args: {
     children: (
       <>
-        <div className="fpo flex items-center justify-center h-8 w-8">Ava</div>
+        <div
+          aria-label="Ava has 1 notification"
+          className="fpo flex items-center justify-center h-8 w-8"
+        >
+          Ava
+        </div>
         <Badge.Text>1</Badge.Text>
       </>
     ),
@@ -54,7 +69,12 @@ export const LargeNumber: StoryObj<Args> = {
   args: {
     children: (
       <>
-        <div className="fpo flex items-center justify-center h-8 w-8">Ava</div>
+        <div
+          aria-label="Ava has 999 notifications"
+          className="fpo flex items-center justify-center h-8 w-8"
+        >
+          Ava
+        </div>
         <Badge.Text>999</Badge.Text>
       </>
     ),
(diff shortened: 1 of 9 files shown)


src/components/Badge/Badge.stories.tsx
```diff
@@ -0,0 +1,61 @@
+import { BADGE } from '@geometricpanda/storybook-addon-badges';
+import type { StoryObj, Meta } from '@storybook/react';
+
+import React from 'react';
+import { Badge } from './Badge';
+import Icon from '../Icon';
+
+export default {
+  // TODO-JL: Appropriate storybook folder?
+  title: 'Molecules/Messaging/Badge',
```
jinlee93, line 10: Placed in Molecules/Messaging because it seemed most appropriate, advise if needed

src/components/Badge/Badge.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+};
+
+export const LargeNumber: StoryObj<Args> = {
+  args: {
+    badge: '999',
+  },
+};
+
+// TODO-JL: cheeky story naming
+export const LargerThanLargeNumber: StoryObj<Args> = {
```
jinlee93, line 40: Kind of a cheeky story name to show that numbers above 999 should be represented with 99+, maybe I'll rename to `NinetyNinePlus` or something?

src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,45 @@
...
+import React from 'react';
+import styles from './Badge.module.css';
+
+export interface Props {
+  /**
+   * Text or icon to be placed on the upper right corner of the badgeable element.
+   *
+   * TODO-JL: prop naming / typing
+   */
+  badge?: React.ReactNode;
```
jinlee93, line 11: Seeking advice on this prop name and how to type. I'm trying to do something like `badge?: string | Icon` but not sure how I can do that `Icon` part

src/components/Badge/Badge.stories.tsx
dierat: I vote `OverNineThousand`

dierat:
Strangely, in [the chromatic snapshots](https://www.chromatic.com/test?appId=61313967cde49b003ae2a860&id=639d07807509523f96410dc3), the badge is on the top left corner and behind the fpo card, but it looks fine in storybook on chromatic. [Chromatic snapshots use an older version of chrome](https://www.chromatic.com/docs/infrastructure-release-notes)—is it possible something in your styles isn't supported?

large number badge story in chromatic

src/components/Badge/Badge.tsx
booc0mtaco: The tricky part is that you have two similarly-named things doing different jobs: this prop `badge` defines what _functions_ as the badge, while `<Badge>` will wrap some other component that will show a badge. Component names should be nouns generally, but in this case `<Badge>` is a verb, and `badge` is the noun.

If you keep `<Badge>`, a prop name that could work is `with`, so the interface is like:

```jsx
<Badge with={<Icon ...>}><SomeComponent /></Badge>
```

There's also the `<Badgeable>` approach with the props corresponding to all available positions. 

A third option would be to rename the component and add subcomponents similar to how our Headless UI -based components work. 

```jsx
<Badge.Wrapper>
    <Badge.Content><SomeComponent /></Badge.Content>
    <Badge.$Type position="top-right">{appropriate content}</Badge.$Type>
</Badge.Wrapper>
```

Where `$Type` is how you control what children are allowed.

All three can work, but curious to hear your thoughts before I mention any of my preferences

src/components/Badge/__snapshots__/Badge.test.ts.snap
```diff
@@ -0,0 +1,113 @@
...
+`;
+
+exports[`<Badge /> LargeBadgeableObject story renders snapshot 1`] = `
+<span
+  class="badge__wrapper"
+>
+  <div
+    class="badge"
+  >
+    10
```
dierat, line 54: I'm wondering what we should do about screen reader support here. "10" on its own doesn't seem like it would be very meaningful without the context of the badge appearance. We know when we see that styling that the number refers to some kind of update or notification, but what's the right way to communicate that to a screen reader and what API should the component use to encourage that?

src/components/Badge/__snapshots__/Badge.test.ts.snap
```diff
@@ -0,0 +1,113 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`<Badge /> Default story renders snapshot 1`] = `
+<span
+  class="badge__wrapper"
+>
+  <div
+    class="badge badge--empty"
+  />
```
dierat, line 9: This is going to need some sort of screen reader text 🤔

src/components/Badge/__snapshots__/Badge.test.ts.snap
booc0mtaco: @dierat this is a good question. A good approach here is that we provide a facility to `<Badge>` for labeling it with text that captures the meaning of the thing being badged AND the count or meaning of the badge itself. The badge is hidden, so screen readers don't simply say "5". That number usually has a "type" as well. 

Rough example:

```
<Badge badge={5} description="You have 5 unread notifications"><NotificationsComponent /></Badge>
```

jinlee93:
> Strangely, in [the chromatic snapshots](https://www.chromatic.com/test?appId=61313967cde49b003ae2a860&id=639d07807509523f96410dc3), the badge is on the top left corner and behind the fpo card, but it looks fine in storybook on chromatic. [Chromatic snapshots use an older version of chrome](https://www.chromatic.com/docs/infrastructure-release-notes)—is it possible something in your styles isn't supported?
> 
> large number badge story in chromatic

I'm not sure but I'm considering using some zIndexing so that we don't run potentially run into this kind of issue

booc0mtaco:
@jinlee93 let's add a case where the badge grows to be WIDER than the thing it's badging. I think I know how it will look but having a snapshot of that case would be nice.

Also a point of clarification for design; what part of the badge should be above the badged corner? should it be centered, or is it a fixed offset from the corner? will it ever show ellipsis, or can it have any amount of text and we expect to see all of that (things that would be described in figma for sure, but we might add safeguards around)?

jinlee93:
> @jinlee93 let's add a case where the badge grows to be WIDER than the thing it's badging. I think I know how it will look but having a snapshot of that case would be nice.

I'm hesitant to add a case where the badge is wider than the thing that it's badging since the max character count of the badge should be 3 and I don't want to be recommending a case going against guidelines or a really small badgeable object 

> Also a point of clarification for design; what part of the badge should be above the badged corner? should it be centered, or is it a fixed offset from the corner? will it ever show ellipsis, or can it have any amount of text and we expect to see all of that (things that would be described in figma for sure, but we might add safeguards around)?

The center of the left semicircle of the badge should be aligned to the top right corner of the badgeable object, with a minimum of it being a circle (so 22px diameter circle incl the white border). I didn't add any safeguards regarding what can go into the badge. Should we apply safeguards besides possibly type checking as `string` and `Icon` as mentioned before?

booc0mtaco:
> > @jinlee93 let's add a case where the badge grows to be WIDER than the thing it's badging. I think I know how it will look but having a snapshot of that case would be nice.
> 
> I'm hesitant to add a case where the badge is wider than the thing that it's badging since the max character count of the badge should be 3 and I don't want to be recommending a case going against guidelines or a really small badgeable object

Oh interesting. I hadn't been thinking of the storybook tests as recommendations but instead as a way to document what happens. Right now, someone could set this to have >3 characters i think? 

> > Also a point of clarification for design; what part of the badge should be above the badged corner? should it be centered, or is it a fixed offset from the corner? will it ever show ellipsis, or can it have any amount of text and we expect to see all of that (things that would be described in figma for sure, but we might add safeguards around)?
> 
> The center of the left semicircle of the badge should be aligned to the top right corner of the badgeable object, with a minimum of it being a circle (so 22px diameter circle incl the white border). I didn't add any safeguards regarding what can go into the badge. Should we apply safeguards besides possibly type checking as `string` and `Icon` as mentioned before?

In general my opinion is to add checks against any things that show up on the component description with design. If our agreed-upon documentation for the badge include types, or character lengths, we should treat those as test cases and they will enforce it in mocks. 🤝🏾 Naturally, if it is tough to write code that checks a thing, we can feed that back to design and possibly change course.

src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,45 @@
...
+    className,
+  );
+
+  return (
+    <span className={styles['badge__wrapper']}>
+      {children}
+      <div className={componentClassName} {...other}>
+        {badge}
+      </div>
+    </span>
```
booc0mtaco, line 43: Technically, it's a violation to have block-level elements inside inline ones like this, with the practical exception of anchor tags wrapping DIVs, etc. If you need the wrapper to be block-level behavior, you can use a DIV or similar with `display: inline-block`

dierat:
> Oh interesting. I hadn't been thinking of the storybook tests as recommendations but instead as a way to document what happens. Right now, someone could set this to have >3 characters i think?

In the EDS team we use them as regression tests, but people outside the team use them as reference for what's available and how to use them, so we try to limit the stories to recommended usage. You can use the component in other ways, but if you do, we can't guarantee it's going to work well because you're not supposed to do that. 🤷

src/components/Badge/Badge.tsx
jinlee93: I do like the `<Badge.Wrapper>` approach, and pushed a commit to represent, but would need more explanation regarding `$Type` or a sync.

booc0mtaco: yah sure, so tl;dr `$Type` wouldn't be literal – I was thinking it would be replaced with stuff like `Text` or `Icon`. But let's chat thru it and make sure that isn't crazy or cumbersome

src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,109 @@
...
+  children?: React.ReactNode;
+  /**
+   * CSS class names that can be appended to the component.
+   */
+  className?: string;
+  // TODO-JL: appropriate prop name? i.e. variant?: 'empty' or empty?: boolean?
+  /**
+   * Empty variant of the badge with a smaller badge circle and no content.
+   */
+  empty?: boolean;
```
jinlee93, line 20: As a result of converting the common `<Badge>` subcomponent to `<Badge.Icon>` and `<Badge.Text>`, the need for the `empty` variant case rises to support the smaller badge without any content. This can be solved with, and is also supported with `<Badge.Text />` but using the `Text` subcomponent seems unintuitive.

Please confirm whether this variant is necessary, and if so, should use `empty?: boolean` or `variant?: 'empty'`

src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,109 @@
...
+ * <Badge>
+ *   {badgeableObject}
+ *   <Badge.Text>1</Badge.Text>
+ * </Badge>
+ *
+ * <Badge>
+ *   {badgeableObject}
+ *   <Badge.Icon name="alarm"/>
+ * </Badge>
+ * ```
```
jinlee93, line 95: Refactored to use the subcomponent pattern

src/components/Badge/__snapshots__/Badge.test.ts.snap
jinlee93: Added associated `aria-label`

src/components/Badge/__snapshots__/Badge.test.ts.snap
jinlee93: Since the badge itself is `aria-hidden` the object it's attached to needs to have an accessible name for screen reader support. In the stories, I added `aria-label`s but not sure how I can further enforce besides documentation? @dierat @booc0mtaco

src/components/Badge/Badge.tsx
jinlee93: Synced, and to summarize, using a 
`<Badge>` wrapper, and
`<Badge.Icon>` subcomponent or 
`<Badge.Text>` subcomponent

src/components/Badge/Badge.stories.tsx
jinlee93: Changed to `OverNineThousand`

src/components/Badge/__snapshots__/Badge.test.ts.snap
dierat: I was thinking of something similar to @booc0mtaco 's example above where a prop is passed into the badge components, which we can then type as a required prop.

src/components/Badge/Badge.tsx
dierat: I feel like that's a little odd because the other variants pass in a subcomponent but this one would use a prop. Would something like `<Badge.Empty /> or `<Badge.Dot />` be too weird?

booc0mtaco: I'm with @dierat and think something like `.Dot` could actually work nicely. Internally, it could be it's own sub-component, or just `Badge.Text` with a non-breaking or no space in its content (or some other thing). for these "types", they are just a convenience API we offer :)

Then we could do away with the prop.

People can then do something like:

```
{ count > 0 ? (
    <Badge.Text>{count}</Badge.Text>
) : (
    <Badge.Empty />
)}
```

if needed.

src/components/Badge/__snapshots__/Badge.test.ts.snap
jinlee93: As discussed, using a prop would mean having to dig down into the Badgeable object and applying the prop as a description or sr-only text but would run into various complications regarding Fragments, already existing descriptions, which element to attach to, etc, so going forward with strong documentation about needing an accessible name acknowledging the badge

src/components/Badge/Badge.tsx
jinlee93: Created `<Badge.Dot>` as suggested and removed `empty` prop

src/components/Badge/Badge.test.tsx
```diff
@@ -0,0 +1,20 @@
+import { generateSnapshots } from '@chanzuckerberg/story-utils';
+import { render } from '@testing-library/react';
+import React from 'react';
+import { Badge } from './Badge';
+import * as stories from './Badge.stories';
+
+describe('<Badge />', () => {
+  generateSnapshots(stories);
+
+  test('throws an error if Badge.Text length is > 3', () => {
```
booc0mtaco, line 10: ++

src/components/Badge/Badge.stories.tsx
```diff
@@ -0,0 +1,133 @@
...
+          aria-label="Ava has notification"
+          className="fpo flex items-center justify-center h-8 w-8"
+        >
+          Ava
+        </div>
+        <Badge.Text>{''}</Badge.Text>
+      </>
+    ),
+  },
+};
```
booc0mtaco, line 53: Still recommended?

booc0mtaco (approved):
This is a good base to build upon. Having the sub components lets us decide any future refactoring and won't require consumers to make changes to their code (fewer breaking changes!)

src/components/Badge/Badge.stories.tsx
jinlee93: Nah, will remove story, thanks for the call out

PR #1441, opened 2022-12-17
feat(Skeleton): add skeleton components

booc0mtaco:
### Summary:

* add known three states for use
* default to round-rect and document others
* add tests for each and layout

### Test Plan:

- [x] chromatic snapshots
- [x] unit tests for each (with code snapshots)
- [x] toggling "reduce motion" in accessibility settings to make sure it disables the motion

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -26,6 +26,10 @@ export const parameters = {
   backgrounds: {
     values: [
       {
+        name: 'white',
+        value: '#ffffff',
+      },
+      {
         name: 'gray',
         value: '#f3f3f3',
       },
--- a/src/components/Skeleton/Skeleton.module.css
+++ b/src/components/Skeleton/Skeleton.module.css
@@ -1,0 +1,41 @@
+/*------------------------------------*\
+    # SKELETON
+\*------------------------------------*/
+
+@keyframes bg-wipe {
+  0% {
+    background-position: 0% 0%;
+  }
+  100% {
+    background-position: 100% 0%;
+  }
+}
+
+/**
+ * Skeleton
+ */
+.skeleton {
+  border-radius: var(--eds-border-radius-lg);
+  background-color: var(--eds-theme-color-background-neutral-medium);
+  background-image: linear-gradient(
+    to right,
+    var(--eds-theme-color-background-neutral-medium) 50%,
+    var(--eds-theme-color-background-neutral-subtle) 65%,
+    var(--eds-theme-color-background-neutral-medium) 80%
+  );
+  background-size: 500% 100%;
+}
+
+.skeleton--is-animating {
+  animation: bg-wipe 2s ease-out infinite;
+}
+
+.skeleton.skeleton-circle {
+  border-radius: var(--eds-border-radius-round);
+}
+
+@media screen and (prefers-reduced-motion: reduce) {
+  .skeleton--is-animating {
+    animation: none;
+  }
+}
--- a/src/components/Skeleton/Skeleton.stories.tsx
+++ b/src/components/Skeleton/Skeleton.stories.tsx
@@ -1,0 +1,61 @@
+import { BADGE } from '@geometricpanda/storybook-addon-badges';
+import type { StoryObj, Meta } from '@storybook/react';
+import React from 'react';
+
+import { Skeleton } from './Skeleton';
+import PopoverContainer from '../PopoverContainer';
+
+export default {
+  title: 'Atoms/Skeleton',
+  component: Skeleton,
+  subcomponents: {
+    'Skeleton.Circle': Skeleton.Circle,
+    'Skeleton.Rect': Skeleton.Rect,
+    'Skeleton.Text': Skeleton.Text,
+  },
+  parameters: {
+    badges: [BADGE.BETA],
+    layout: 'centered',
+    backgrounds: {
+      default: 'white',
+    },
+  },
+} as Meta<Args>;
+
+type Args = React.ComponentProps<typeof Skeleton>;
+
+export const Default: StoryObj<Args> = {
+  args: {
+    width: 100,
+    height: 100,
+  },
+};
+
+export const Circle: StoryObj<Args> = {
+  render: (args) => {
+    return <Skeleton.Circle width={100} />;
+  },
+};
+
+export const Text: StoryObj<Args> = {
+  render: (args) => {
+    return <Skeleton.Text height="1.5rem" width="30ch" />;
+  },
+};
+
+export const LayoutExample: StoryObj<Args> = {
+  render: (args) => {
+    return (
+      <div aria-label="Loading example profile card" role="status">
+        <PopoverContainer className="p-3 m-2">
+          <Skeleton.Circle className="mb-3" width="50px" />
+          <Skeleton.Text className="mt-2 mb-2" height="4rem" width="15ch" />
+          <Skeleton.Text className="mt-2 mb-2" height="1.5rem" width="30ch" />
+          <Skeleton.Text className="mt-2 mb-2" height="1.5rem" width="25ch" />
+          <Skeleton.Text className="mt-2 mb-8" height="1.5rem" width="10ch" />
+          <Skeleton height={50} width="100%" />
+        </PopoverContainer>
+      </div>
+    );
+  },
+};
--- a/src/components/Skeleton/Skeleton.tsx
+++ b/src/components/Skeleton/Skeleton.tsx
@@ -1,0 +1,119 @@
+import clsx from 'clsx';
+import React from 'react';
+import styles from './Skeleton.module.css';
+
+type BaseProps = {
+  /**
+   * CSS class names to augment the layout of the skeleton component. For skeletons,
+   * this is mostly useful to add layout styles (e.g., tailwind CSS margins, flex, etc.)
+   */
+  className?: string;
+  /**
+   * Determine whether there should be an animation in the skeleton state
+   */
+  isAnimating?: boolean;
+};
+
+type SkeletonProps = BaseProps & {
+  width?: number | string;
+  height?: number | string;
+};
+
+/**
+ * BETA: This component is still a work in progress and is subject to change.
+ *
+ * `import {Skeleton} from "@chanzuckerberg/eds";`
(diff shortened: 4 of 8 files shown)


dierat:
Strangely, in the chromatic snapshots, the badge is on the top left corner and behind the fpo card, but it looks fine in storybook on chromatic. Chromatic snapshots use an older version of chrome—is it possible something in your styles isn't supported?

booc0mtaco:
> Edit: oops wrong PR; plz ignore me. Btw did you mean to add reviewers to this one or no?

Ah not just yet. Waiting on design to give more details on the colors and such. I tossed up this imperfect first attempt to help get some details cleared up.

booc0mtaco:
Sync'd with design; 

* colors and radii have been approved
* animation has been approved

We are good to merge this one from that side

booc0mtaco:
https://user-images.githubusercontent.com/3056447/209011675-ea1ff728-3101-4352-a542-e11315360c88.mov

dierat:
Is it just me or does it feel weird that it's swiping right-to-left instead of left-to-right?

src/components/Skeleton/Skeleton.module.css
```diff
@@ -0,0 +1,35 @@
...
+    to right,
+    var(--eds-theme-color-background-neutral-subtle) 50%,
+    var(--eds-theme-color-background-neutral-default) 65%,
+    var(--eds-theme-color-background-neutral-subtle) 80%
+  );
+  background-size: 500% 100%;
+}
+
+.skeleton--is-animating {
+  animation: bg-wipe 2s ease infinite;
```
dierat, line 30: Is there a particular reason you chose `ease`? We usually use `ease-out`.

booc0mtaco: @dierat not really. lemme see how `ease-out` looks tho (i usually use `ease-in-out` and i have been known to do my own curve when i'm feeling really energized)

src/components/Skeleton/Skeleton.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export default {
+  title: 'Atom/Skeleton',
+  component: Skeleton,
+  subcomponents: {
+    'Skeleton.Circle': Skeleton.Circle,
+    'Skeleton.Rect': Skeleton.Rect,
+    'Skeleton.Text': Skeleton.Text,
+  },
+  parameters: {
```
dierat, line 16: I think we should also made the background white because the gray is the same color as the default background. Here's an example of specifying the background in stories: https://github.com/chanzuckerberg/edu-design-system/blob/e0288ce88d4cccbf5a1e3fc5cd79b5b8ee0ce4fd/.storybook/preview.tsx#L28

src/components/Skeleton/__snapshots__/Skeleton.test.ts.snap
```diff
@@ -0,0 +1,62 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`<Skeleton /> Circle story renders snapshot 1`] = `
+<div
+  aria-hidden="true"
```
dierat, line 5: I wonder if we should have some kind of screen ready-only text like the spinner has because skeleton states are like an alternative to spinners.

src/components/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,119 @@
...
+    <div
+      aria-hidden
+      className={componentClassName}
+      style={{ width, height: lineHeight }}
+    ></div>
+  );
+};
+
+type CirlceProps = BaseProps & {
+  size?: number | string;
```
dierat, line 92: I wonder if `Circle` should use `width` instead of `size` to be consistent with the other shapes.

src/components/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,119 @@
...
+      aria-hidden
+      className={componentClassName}
+      {...other}
+      style={{ width, height }}
+    ></div>
+  );
+};
+
+type TextProps = BaseProps & {
+  lineHeight?: number | string;
```
dierat, line 66: I feel like `height` would be simpler here than `lineHeight` since it is just a rectangle even though it represents text.

src/components/Skeleton/__snapshots__/Skeleton.test.ts.snap
booc0mtaco: I think for these ones, they would only be used on a page's initial load, and shown for a very brief moment, so any text or assistive aids should be handled by whatever caused the navigation, triggering an update to some component with `aria-live`. the loading indicator is similar, but would appear where some content used to appear, or overlaying the whole body of a component? Hard to say since we don't have reference designs just yet :S

src/components/Skeleton/Skeleton.tsx
booc0mtaco: I think this makes sense. I def intended to avoid people having circles that ... werent :)

src/components/Skeleton/Skeleton.tsx
booc0mtaco: i think you're right. before, this was `fontSize`, with the intent that people would use a value similar to the font it replaces. but font size and line height are different things, so i changed it to lineHeight. but that's just height now, so makes sense to document what the intent of height means when using `.Text`

booc0mtaco:
updated with the higher contrast colors

src/components/Skeleton/Skeleton.module.css
```diff
@@ -0,0 +1,35 @@
...
+    var(--eds-theme-color-background-neutral-medium) 50%,
+    var(--eds-theme-color-background-neutral-subtle) 65%,
+    var(--eds-theme-color-background-neutral-medium) 80%
+  );
+  background-size: 500% 100%;
+}
+
+.skeleton--is-animating {
+  animation: bg-wipe 2s ease-out infinite;
+}
```
jinlee93, line 31: Will `prefers-reduced-motion` be determined via JS?

.storybook/preview.tsx
```diff
@@ -25,6 +25,10 @@ export const decorators = [
 export const parameters = {
   backgrounds: {
     values: [
+      {
+        name: 'white',
+        value: '#ffffff',
+      },
```
jinlee93, line 31: nice

src/components/Skeleton/__snapshots__/Skeleton.test.ts.snap
jinlee93: Makes sense with `aria-live`. Am wondering how you're going to communicate such requirement?

src/components/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,119 @@
...
+  return (
+    <div
+      aria-hidden
+      className={componentClassName}
+      style={{ width, height: height }}
+    ></div>
+  );
+};
+
+type CirlceProps = BaseProps & {
```
jeremiah-clothier, line 91:
```suggestion
type CircleProps = BaseProps & {
```

jeremiah-clothier (commented):
Looks cool 👍

src/components/Skeleton/__snapshots__/Skeleton.test.ts.snap
dierat: I thought about it and I think the big difference between the skeleton states and a spinner is there's typically one spinner (well I have seen apps where there are a bunch of spinners but that looks pretty weird) whereas there are more likely to be multiple skeletal state shapes, and we wouldn't want each one of them to have a message announcing that something is loading. I think this might be a similar situation to the `Badge` where we're going to have to leave it to the developer to add something themselves, and maybe the best thing for us to do is just add very visible documentation.

src/components/Skeleton/Skeleton.module.css
dierat: Easiest solution is probably
```
  @media screen and (prefers-reduced-motion) {
    animation: none;
  }
```

src/components/Skeleton/Skeleton.module.css
dierat: I'm not really opinionated by which one we use, but I feel like it's nice to keep them consistent unless there's a specific reason why an animation needs a different timing function.

src/components/Skeleton/Skeleton.module.css
booc0mtaco: Added this, and it can be tested by toggling the setting in the OS

Screen Shot 2023-01-03 at 09 10 34

src/components/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,119 @@
...
+  ...other
+}: SkeletonProps) => {
+  const componentClassName = clsx(
+    className,
+    styles['skeleton'],
+    isAnimating && styles['skeleton--is-animating'],
+  );
+  return (
+    <div
+      aria-hidden
```
ahuth, line 57: Should we have some sort of message for screen readers saying something like "loading..."?

Along does that at https://github.com/chanzuckerberg/along/blob/3215c989afd8b41fc5960952df1fb4991265844a/client/src/ui/Spinner/Spinner.tsx#L13

ahuth (approved):
LGTM 🤘 

I recommend adding screen reader text to the loader (could be `aria-label` or a sr-only element).

src/components/Skeleton/Skeleton.tsx
booc0mtaco: @ahuth We do have such text for our spinner (`LoadingIndicator`), which can be configured if the default isn't sufficient. For this one, opted to leave it out, as a layout could have many skeleton objects mocking the pre-load UI, and the screen reader would see "loading loading loading loading...". In discussion, we reasoned it's best to document how to add a label on the container, or to use `aria-live` on a page-level component:

- user interacts with some control
- page loads some layout showing these skeleton blocks (instead of a spinner, etc.)
- the interaction should then update the app's `aria-live` text, or could have an `aria-label` on the container that is removed once the data is fully loaded (and the skeletons are replaced with actual content)

I'll add an example of this `aria-label` to storybook to reflect this recommendation

PR #1444, opened 2023-01-03
fix(LoadingIndicator): add reduced motion behavior

booc0mtaco:
### Summary:

Define a fallback behavior when reduce motion is enabled. This accessibility behavior is esp. useful in context of many loading indicators on screen at once, which could produce some undesirable optical effects.

### Test Plan:

- [x] add fallback and test by toggling "reduce motion" at OS level
- [x] existing test suite should pass


booc0mtaco:
https://user-images.githubusercontent.com/3056447/210429147-9185c91c-97b0-42fe-8434-f023b2716f3b.mov

ahuth (approved):
LGTM

PR #1448, opened 2023-01-04
Don't ship theming demo fonts in the package

ahuth:
Noticed in https://github.com/FB-PLP/traject/pull/13040 that upgrading EDS to 7.2.0 results in:
1. [Perfbot in Traject thinking the bundles now include **_Along_**'s fonts...](https://github.com/FB-PLP/traject/pull/13040#issuecomment-1370207966)
2. [Some banner font sizing/spacing differences](https://www.chromatic.com/test?appId=614b969675171d003ac6f98a&id=63b49468db3b6bef0e6f7da0)

I believe the first is coming from #1385 and #1388. We've included the Along and Wireframe demo fonts in the EDS fonts. Don't know what's up with the 2nd (may be intentional).

This PR moves the Along and Wireframe demo fonts out of the tier 1 tokens and into `.storybook/`, so they don't get included in the package.


jinlee93 (approved):
Thanks for moving these into `./storybook`

booc0mtaco (approved):
Great catch. This kinda thing is not likely to happen often, but I do wonder if there's something we can document about adding demos (or change where demos go) in the future?

ahuth:
Thanks 🙏

PR #1453, opened 2023-01-05
chore: use recognized React SVG props in demo

booc0mtaco:
### Summary:

- fill-rule => fillRule
- cilp-rule => clipRule

### Test Plan:

- n/a


jinlee93 (approved):
Nice catch

booc0mtaco:
i marked this one as `chore` as it would not have any effect on versioned content (components, tokens, etc)

PR #1454, opened 2023-01-10
chore: remove and audit TODO comments

booc0mtaco:
### Summary:

- remove any comments that have no clear call to action
- create follow up tickets for units of work that can be addressed
- convert some comments to Notes for information purposes

### Test Plan:

- [x] existing tests should pass without any warnings

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -54,7 +54,7 @@ jobs:
         run: yarn build:tokens
       - name: Build Storybook 📚
         run: yarn build:storybook
-      # TODO: Since the storybook builds to root now, no longer need to fiddle with "--storybook-base-dir", "--webpack-stats-json",
+      # Note: Since the storybook builds to root now, no longer need to fiddle with "--storybook-base-dir", "--webpack-stats-json",
       # and other related storybook/chromatic flags, explore turning TurboSnap back on with "--only-changed".
       # https://app.shortcut.com/czi-edu/story/203071
       - name: Publish to Chromatic 🦄
@@ -91,5 +91,4 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Lint Javascript + SCSS 👕
-        # TODO: enable --max-warnings=0
-        run: yarn run lint
+        run: yarn run lint --max-warnings=0
--- a/.storybook/recipes/BaselineCard/BaselineCard.tsx
+++ b/.storybook/recipes/BaselineCard/BaselineCard.tsx
@@ -63,8 +63,10 @@ export type LinkType = {
 /**
  * Recipe for a Card component that displays a common card use case.
  * A link may be added to the card to create a clickable card.
+ *
  * NOTE: when card is clickable, it does not support links or buttons
- * in the contents of the card.
+ * in the contents of the card. Users will not be able to click on these inner
+ * links without an event to stop propagation.
  */
 export const BaselineCard = ({
   className,
@@ -83,11 +85,6 @@ export const BaselineCard = ({
 
   /**
    * Hook up the link functionality on component mount if a link is present
-   *
-   * TODO: If the visible text of the card also contains links, these will
-   * need to have an event listener added to them to stop propagation of the
-   * click event; otherwise, users will not be able to click on these inner
-   * links.
    */
   useEffect(() => {
     /**
--- a/docs/COMPONENTS.md
+++ b/docs/COMPONENTS.md
@@ -6,19 +6,18 @@ This codebase contains all the components, [recipes](https://bradfrost.com/blog/
 
 ### Using components
 
-TODO: define this process
-
 Using EDS components in your React application involves first installing the EDS package as a dependency:
 
-```
-npm install @/EDS
-yarn add @/EDS
+```sh
+npm install @chanzuckerberg/eds
+# or
+yarn add @chanzuckerberg/eds
 ```
 
 Next, import any needed components from the EDS component library into your application like so:
 
 ```js
-import { Button, Grid, GridItem } from '@/EDS';
+import { Button, Grid, GridItem } from '@chanzuckerberg/eds';
 ```
 
 From there, call EDS components in your React application and pass in the desired values into each component's API:
@@ -33,7 +32,7 @@ From there, call EDS components in your React application and pass in the desire
 </Button>
 ```
 
-Each EDS component is documented in Storybook, which surfaces each component's API and provides copy-and-paste code snippets. If you have questions or are experiencing any issues when working with EDS's components, please [reach out for support](#) (TODO: provide support link) and the team will be happy to help.
+Each EDS component is documented in Storybook, which surfaces each component's API and provides copy-and-paste code snippets. If you have questions or are experiencing any issues when working with EDS's components, please reach out for support via slack and the team will be happy to help.
 
 ### Creating a component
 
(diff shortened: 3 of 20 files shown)


.storybook/recipes/BaselineCard/BaselineCard.tsx
```diff
@@ -83,11 +85,6 @@ export const BaselineCard = ({
 
   /**
    * Hook up the link functionality on component mount if a link is present
-   *
-   * TODO: If the visible text of the card also contains links, these will
```
booc0mtaco, line 87: moved details up to `BaselineCard` as a warning about this behavior

docs/COMPONENTS.md
```diff
@@ -99,27 +98,9 @@ Pages
 
 ## Working with recipes
 
-TODO: discuss recipes and EDS candidates; ratify recipe location
```
booc0mtaco, line 102: created documentation ticket

src/components/ButtonDropdown/ButtonDropdown.tsx
```diff
@@ -150,12 +150,11 @@ export const ButtonDropdown = ({
 
   const dropdownMenuTriggerWithProps = React.Children.map(
     dropdownMenuTrigger,
-    // TODO: improve `any` type
```
booc0mtaco, line 153: Removed as we will deprecate (and delete) this component in the coming year

src/components/Layout/Layout.module.css
```diff
@@ -121,7 +121,7 @@
  */
 .layout--sidebar-wide > .layout__section--sidebar {
   @media all and (min-width: $eds-bp-xl) {
-    width: 22rem; /* TODO: Make more dynamic */
```
booc0mtaco, line 124: There's no specific thing to change here except that there may be a better way to define width. created a ticket for the layout components to revisit them

src/components/LayoutContainer/LayoutContainer.stories.tsx
```diff
@@ -8,7 +8,6 @@ export default {
   component: LayoutContainer,
   parameters: {
     axe: {
-      // TODO: re-enable when component is worked on
```
booc0mtaco, line 11: One of the details added to the ticket for re-visiting the layout components; re-enabling the story / AXE tests

src/components/TimelineNav/TimelineNav.module.css
```diff
@@ -87,7 +87,7 @@
     visibility: visible;
     transform: translateX(0);
     padding: var(--eds-size-5) var(--eds-size-8);
-    box-shadow: 0px 2px 8px var(--eds-theme-color-border-neutral-subtle); /* TODO: Tokenize the box shadow */
```
booc0mtaco, line 90: created ticket

.github/workflows/test.yml
```diff
@@ -91,5 +91,5 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Lint Javascript + SCSS 👕
-        # TODO: enable --max-warnings=0
```
jeremiah-clothier, line 94: I interpret this comment as "we should fix all linting violations and then add `--max-warnings=0`". This may be low priority, how many linting violations are there?

.storybook/components/IconGrid/IconGrid.tsx
```diff
@@ -70,6 +70,5 @@ const IconGrid = () => (
   </div>
 );
 
-// TODO: not sure why this is failing the lint check when it has associated story file
```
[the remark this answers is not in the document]
jeremiah-clothier: In the future can you provide links to tickets as well in case others want to "watch" tickets or add comments to them. I'm wondering if the error with this is the story file has a `.ts` extension but the component is `.tsx`?

src/components/FieldsetItems/FieldsetItems.tsx
```diff
@@ -26,10 +26,9 @@ export type FieldsetItemsProps = {
  */
 export const FieldsetItems = ({
   children,
-  as: Component = 'div',
+  as: Component = 'div' as const,
   className,
 }: FieldsetItemsProps) => {
   const componentClassName = clsx(styles['fieldset-items'], className);
-  // @ts-expect-error TODO: investigate error
   return <Component className={componentClassName}>{children}</Component>;
```
jeremiah-clothier, line 33: What was the error here that typescript was flagging?

Although the error was removed, I think the type safety is slightly worse now. Now that the type is `as?: ElementType | string;` a developer could accidentally type `<FieldsetItems as="spna" />` and TS wouldn't complain

.storybook/components/IconGrid/IconGrid.tsx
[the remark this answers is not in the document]
jinlee93: This is mitigated with https://github.com/chanzuckerberg/edu-design-system/pull/1456, where it just places the IconGrid story under Icon

src/components/FieldsetItems/FieldsetItems.tsx
booc0mtaco: @jeremiah-clothier I left off the links as they point to internal trackers in this case, but would generally prefer linking them. Relevant tickets are linked to the ticket name in the branch, for safe keeping.

.github/workflows/test.yml
booc0mtaco: That's a good callout. I will check tomorrow and see where we stand right now (might be at 0!)

src/components/FieldsetItems/FieldsetItems.tsx
booc0mtaco: oops, responded to the wrong comment :/ @jeremiah-clothier that's a good point actually. I saw something similar for our other `as` props, but this could be a problem. I'll double check (just in case react has some internal handling for this param when converting JSX to actual function calls. What I want here is "strings that can be used as the list of supported component names"

src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -191,12 +191,8 @@ export const TimelineNav = ({
    * On KeyDown
    *
    * Find active tab. If there isn't one, do nothing on Keydown.
-   *
-   * TODO: determine why backRef.current?.focus() needs to be in a callback
-   *
-   * TODO: improve `any` type
    */
-  function onKeyDown(e: any) {
+  const onKeyDown: KeyboardEventHandler = function (e) {
```
jinlee93, line 195: Nice

jinlee93 (approved):
Think if the tix are in the respective backlogs, looks good to me!
Thanks for the cleanup

.github/workflows/test.yml
booc0mtaco: @jeremiah-clothier so the actual count was one warning. Found a weird case where we were getting a false positive for an unused var. According [to this](https://typescript-eslint.io/rules/no-unused-vars/), we can enable this linting rule instead of the built in one to check for unused things even in types. Will update, and set the config to allow 0 warnings

src/components/FieldsetItems/FieldsetItems.tsx
booc0mtaco: @jeremiah-clothier here's the full error once i remove the `string` type:

```
'Component' cannot be used as a JSX component.
  Its element type 'Element | Component<any, any, any> | null' is not a valid JSX element.
    Type 'Component<any, any, any>' is not assignable to type 'Element | ElementClass | null'.
      Type 'Component<any, any, any>' is not assignable to type 'ElementClass'.
        The types returned by 'render()' are incompatible between these types.
          Type 'React.ReactNode' is not assignable to type 'import("/Users/aholloway/Development/edu-design-system/node_modules/@types/react-beautiful-dnd/node_modules/@types/react/index").ReactNode'.
            Type '{}' is not assignable to type 'ReactNode'.ts(2786)
```
I have a few ideas how to resolve this but will do some more digging tomorrow. I'm not sure where that `null` is coming from, which I think is the main issue.

booc0mtaco: @jeremiah-clothier added a comment for this, and fixed up the types. unfortunately looks like we may need to keep this expected error for a while longer :( 

`Type 'React.ReactNode' is not assignable to type 'import("/Users/aholloway/Development/edu-design-system/node_modules/@types/react-beautiful-dnd/node_modules/@types/react/index").ReactNode'.` is coming from a conflict between our version of react, and one that this dependency is pulling in via its own node module code :/

booc0mtaco:
Screen Shot 2023-01-12 at 17 36 58

We now have better safety with the `as` prop here, so we catch it via TypeScript. Something we can borrow for other polymorphic components like this

src/components/FieldsetItems/FieldsetItems.tsx
```diff
@@ -24,12 +24,23 @@ export type FieldsetItemsProps = {
...
-export const FieldsetItems = ({
+export const FieldsetItems = <T extends ElementType = 'div'>({
   children,
-  as: Component = 'div',
+  as,
   className,
-}: FieldsetItemsProps) => {
+  ...props
+}: FieldsetItemsProps<T> &
+  Omit<ComponentPropsWithoutRef<T>, keyof FieldsetItemsProps<T>>) => {
```
booc0mtaco, line 33: Here we are saying to use `"div"` as the default element type, and the params are our custom props and React's default props minus our custom ones (to avoid duplicates with any ones we redefine like `className`)

jeremiah-clothier: Thanks for investigating this and figuring it out! I was about to say we can punt on it and leave the `//@ts-expect-error` if needbe

jeremiah-clothier (approved):

PR #1455, opened 2023-01-11
fix(TimelineNav): remove unused props and styles

booc0mtaco:
### Summary:

- update related snapshots

### Test Plan:

- [x] verify that existing tests pass
- [x] update any snapshots

### DIFF
--- a/src/components/TimelineNav/TimelineNav.module.css
+++ b/src/components/TimelineNav/TimelineNav.module.css
@@ -156,9 +156,10 @@
   */
 .timeline-nav__list--ordered .timeline-nav__item {
   &:after {
+    --line-spacing: var(--eds-size-4);
     content: '';
     position: absolute;
-    top: 0;
+    top: var(--line-spacing);
     /**
      * This sets the size of the gap above the vertical line by determining the top of the line relative to the top of the current 
      * list item; a value of 0 would position the line at the top of the current list item; 100% would position it at the bottom of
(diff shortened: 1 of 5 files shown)


ahuth (approved):

src/components/TimelineNav/TimelineNav.module.css
```diff
@@ -280,7 +280,6 @@
   height: fit-content;
   padding: var(--eds-size-half) 0;
   /* Provides white background to fake a visual gap in the vertical line between nav items. */
-  background: var(--eds-theme-color-background-neutral-default);
```
booc0mtaco, line 283: the visual diffs were generated b/c there was an embedded bg around the icons that was inconsistently transparent. Removing the bg fixes this bug
Screen Shot 2023-01-10 at 17 59 56

booc0mtaco: Before:
Screen Shot 2023-01-10 at 18 20 01

jinlee93 (approved):
Thanks for the cleanup!

booc0mtaco:
I noticed some other oddities in the styling and behavior as compared to what's in figma, so expanded on this PR a little bit to simplify the CSS. Do more with less :) (should get a second look tho... right now making final adjustments to align properly and get the right diff in chromatic)

src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -389,16 +371,14 @@ export const TimelineNav = ({
...
-            {
-              [styles['timeline-nav__list--ordered']]: variant === 'ordered',
-            },
           )}
           role="tablist"
         >
           {timelineNavItems().map((tab, i) => {
             const isActive = activeIndexState === i;
-            const itemVariant = variant && tab.props.variant;
+            const itemVariant = tab.props.variant;
```
booc0mtaco, line 381: @jinlee93 the chromatic test got to this line, which now will always have a value. previously, `variant` being undefined would set `itemVariant` to undefined i think

jinlee93: That does make sense

PR #1456, opened 2023-01-11
Flatten and alphabetize stories

jinlee93:
[EDS-784]
### Summary:
- removes the ambiguous folder org of storybook and just places into folders components/recipes/pages
- places icon grid story under icon stories 
### Test Plan:
- no changes besides sb org

[EDS-784]: https://czi-tech.atlassian.net/browse/EDS-784?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Accordion/Accordion.stories.tsx
```diff
@@ -5,7 +5,7 @@ import React from 'react';
 import { Accordion } from './Accordion';
 
 export default {
-  title: 'Organisms/Interactive/Accordion',
+  title: 'Components/Accordion',
```
ahuth, line 8: Could remove the `title` from all stories and let Storybook auto-magically figure out the structure.

Tried it previously in #813 and it works. But pretty sure it took some config to make it work well for our case.

ahuth (approved):
LGTM

src/components/Accordion/Accordion.stories.tsx
jinlee93: I don't want to have the components in the root sb folder (which is default without `title`) or have individual folders for each component. Will update the plop stories template to auto place in `Components/` to ease dev experience

PR #1458, opened 2023-01-11
docs(select): remove mentions of undefined prop

jinlee93:
### Summary:
I don't think `labelText` prop is being used in `Select`, removing references to it in the prop desc
### Test Plan:
n/a


booc0mtaco (approved):
🎣

PR #1459, opened 2023-01-11
refactor: remove default exports from components

jinlee93:
### Summary:
- removes default exports from 5 components
- correctly points affected files to named import
- all other components don't have default exports
- resolves eslint errors that were happening locally 
### Test Plan:
- doesn't break LP
  - tested via [7.3.0-alpha.0](https://www.npmjs.com/package/@chanzuckerberg/eds/v/7.3.0-alpha.0) release and [tests pass](https://app.circleci.com/pipelines/github/FB-PLP/traject?branch=jlee%2FtestNoDefaultComponentExportEDS) in LP test branch
  - manually looked for imports to `@czi/eds/lib/components/{component}/{component}` and didn't find default import instances

### DIFF
--- a/src/components/Heading/Heading.tsx
+++ b/src/components/Heading/Heading.tsx
@@ -110,5 +110,3 @@ export const Heading = forwardRef(
 );
 
 Heading.displayName = 'Heading'; // Satisfy eslint.
-
-export default Heading;
--- a/docs/CODE_GUIDELINES.md
+++ b/docs/CODE_GUIDELINES.md
@@ -393,7 +393,7 @@ The comment should begin with an import example, include a general description o
 
 Example:
 
-````
+````tsx
 /**
  * `import {ButtonGroup} from "@chanzuckerberg/eds";`
  *
@@ -419,7 +419,7 @@ Do not use [jsdoc tags](https://devhints.io/jsdoc) (e.g. `@example`) if possible
 
 Example:
 
-````
+````tsx
 /**
  * The Banner component is deprecated and will be removed in an upcoming release.
  *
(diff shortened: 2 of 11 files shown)


src/components/Heading/Heading.tsx
```diff
@@ -110,5 +110,3 @@ export const Heading = forwardRef(
 );
 
 Heading.displayName = 'Heading'; // Satisfy eslint.
-
-export default Heading;
```
ahuth, line 114: Are we generally expecting to not use default exports for components in EDS now?

ahuth (approved):
LGTM

booc0mtaco (approved):
Given there were only a few of these, I guess not having defaults is the ... default :)

Can you double-check the plop template to make sure it doesn't do a default export, and perhaps add a note about this in our code guidelines (if it's missing)

src/components/Heading/Heading.tsx
jinlee93: The index file for the respective component exports these as default already, but we don't need default exporting (and haven't been) for the component file itself, i.e. `import Heading from '@czi/eds/lib/components/Heading` works and now removing `import Heading from @czi/eds/liib/components/Heading/Heading.tsx`. These are being still being re-exported in `src/index` anyways so general use is still `import {Heading} from @czi/eds`.

jinlee93:
> Given there were only a few of these, I guess not having defaults is the ... default :)
> 
> Can you double-check the plop template to make sure it doesn't do a default export, and perhaps add a note about this in our code guidelines (if it's missing)

Can confirm component plop template doesn't have default export, and added note in `CODE_GUIDELINES.md`

PR #1461, opened 2023-01-12
refactor(Popover)!: remove showArrow from popover

booc0mtaco:
### Summary:

- after discussion with design, will remove the example showing the (already deprecated) `showArrow` prop

### Test Plan:

- update test snapshots

### DIFF
--- a/src/components/Popover/Popover.stories.tsx
+++ b/src/components/Popover/Popover.stories.tsx
@@ -49,20 +49,6 @@ export const Default: StoryObj<PopoverProps> = {
   },
 };
 
-export const Arrow: StoryObj<PopoverProps> = {
-  args: {
-    children: (
-      <>
-        <Popover.Button as={Button}>Open Popover</Popover.Button>
-        <Popover.Content showArrow>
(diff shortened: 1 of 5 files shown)


jeremiah-clothier (approved):

src/components/Popover/Popover.stories.tsx
```diff
@@ -49,20 +49,6 @@ export const Default: StoryObj<PopoverProps> = {
   },
 };
 
-export const Arrow: StoryObj<PopoverProps> = {
-  args: {
-    children: (
-      <>
-        <Popover.Button as={Button}>Open Popover</Popover.Button>
-        <Popover.Content showArrow>
```
jeremiah-clothier, line 57: Should the `showArrow` prop be removed as well?

jinlee93 (approved):

src/components/Popover/Popover.stories.tsx
jinlee93: Honestly, I think it would be ok to axe the `showArrow` prop and the couple usages in the recipes, since it's not being used downstream

PR #1463, opened 2023-01-14
refactor!: remove unused components 2/n

jinlee93:
### Summary:
- removes more unused components:
  - stackedblock: Just a div wrapper wrapping link and text, but takes prop `title` for the link? not sure why that's preferable over just using a regular <Link>
  - navcontainer: another div wrapper with styling specific to <GlobalHeader>, removed and attached styling there
  - **definitionlist**: this is the one i'm most concerned about removing in this PR because there is a definite use for `<dl> <dd> and <dt>` semantically for sure so its existence is a good reminder. However 1: I don't like the implementation, again using `title` prop to pass into `<dt>`, using the term "definition" instead of "description", and just general overload of styling. 2: not knowing the semantic elements don't think should be a cause for having extra components, should be more of an outreach / KT problem

- Regarding GlobalHeader and PrimaryNav, created a [ticket](https://czi-tech.atlassian.net/browse/EDS-797) since header / nav components have use but not in the way it's currently implemented. prob needs to go thru the whole design/dev cycle

### Test Plan:
- minimal, if any, visual regression


.storybook/pages/ProjectOverview/ProjectOverview.tsx
```diff
@@ -125,7 +122,7 @@ export const ProjectOverview = ({ activeIndex = 0 }: Props) => {
             }
           >
             <Card className="!mb-4">
-              <CardBody>
+              <Card.Body>
```
booc0mtaco, line 125: ++

.storybook/recipes/GlobalHeader/GlobalHeader.tsx
```diff
@@ -173,9 +172,11 @@ export const GlobalHeader = ({
         />
       </Button>
 
-      <NavContainer
-        className={styles['global-header__nav-container']}
-        isActive={isActive}
+      <div
```
booc0mtaco, line 175: nit: or `<section>` could work... or `<nav>`?

booc0mtaco (approved):
LGTM. We can make semantic updates to the one page, but not a blocker IMO

.storybook/recipes/GlobalHeader/GlobalHeader.tsx
jinlee93: `section` would not work since [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section) suggests "Sections should always have a heading ... If you are only using the element as a styling wrapper, use a `<div>` instead"

The nested `PrimaryNav` component provides the `<nav>`

PR #1465, opened 2023-01-17
docs: update viewport and background params

booc0mtaco:
### Summary:

- update storybook parameters to use known device viewports
- update names of background colors to match token names

### Test Plan:

- [x] verify no tests regress
- [x] verify options appear properly in storybook UI

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -14,6 +14,8 @@ import './css/styleguide-only.css';
 import type { Story } from '@storybook/react';
 import React from 'react';
 
+import viewports from './util/viewports';
+
 export const decorators = [
   (Story: Story) => (
     <div dir="ltr">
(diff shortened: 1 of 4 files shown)


booc0mtaco:
Screen Shot 2023-01-17 at 14 07 10
Screen Shot 2023-01-17 at 14 07 22
Screen Shot 2023-01-17 at 14 07 39

jeremiah-clothier (approved):

jinlee93:
Just a quick thought is that chromatic uses its own `viewports` parameter and don't think respects the storybook viewport for snapping

booc0mtaco:
@jinlee93 correct, it doesn't. I have a small follow-up task to reuse this viewports utility to export values chromatic config can use as well. That way, we can avoid the magic numbers for those

PR #1466, opened 2023-01-17
test(Skeleton): update stories to use control arg.s

booc0mtaco:
### Summary:

- the `Skeleton` stories weren't using the args, but showing the controls. This meant that one could try and toggle the values, but nothing would happen. Fix this up such that the controls work for the given stories and subcomponents

### Test Plan:

- [ ] make sure there are no updates to any snapshots related to the existing stories


src/components/Skeleton/Skeleton.stories.tsx
```diff
@@ -31,29 +31,60 @@ export const Default: StoryObj<Args> = {
...
+  },
   render: (args) => {
-    return <Skeleton.Text height="1.5rem" width="30ch" />;
+    return <Skeleton.Text {...args} />;
   },
 };
 
 export const LayoutExample: StoryObj<Args> = {
-  render: (args) => {
+  render: ({ isAnimating }) => {
```
jinlee93, line 54: We can use `argTypes` to disable controls for certain props if we want here, i.e.:
```
  argTypes: {
    className: {
      control: false,
    },
    width: {
      control: false,
    },
    height: {
      control: false,
    },
  },
```

booc0mtaco: good call. I thought this was controlled by the type argument to `StoryObj`, but that didn't seem to work

jinlee93 (approved):
Looks good, thanks!

PR #1473, opened 2023-01-24
Side effects

ahuth:
### Summary:

This PR specifies that EDS does not have [sideEffects](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free)
- Let's esbuild and webpack tree shake EDS.
- Meaning apps using EDS only ship code for the components they use, as opposed to all components.
- I actually made this change 1.5 years ago in #554, but it seems to have gotten wiped out in the BM repo changes 🤨 

### Test Plan:

- Before this change: in the Remix test app code for all EDS components is present, despite only using Heading.
- After this change: only the Heading code is present.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "homepage": "https://github.com/chanzuckerberg/edu-design-system",
   "license": "MIT",
   "main": "lib/index.js",
+  "sideEffects": false,
   "publishConfig": {
     "access": "public"
   },


package.json
```diff
@@ -6,6 +6,7 @@
   "homepage": "https://github.com/chanzuckerberg/edu-design-system",
   "license": "MIT",
   "main": "lib/index.js",
+  "sideEffects": false,
```
ahuth, line 9: Setting this implies that there are no "side effects" to importing any EDS modules.

e.g. nothing happens just from importing (such as setting global variables, changing state, firing off network requests, etc).

I think this is valid/reasonable. It is a constraint we'll have to make sure we follow.

anniehu4: oh good catch -- I actually think I was the one who removed it, when we consolidated into a single package 😅  I wasn't sure if the exported token files (e.g. `variables.css`, which get imported at the root of consuming apps) needed to be specified in this array.

ahuth: Good Q.

I think this setting only affects modules imported in the main `index.ts` file.

Something like `lib/tokens/css/variables.css` won't be included anyway unless people explicitly import it into consuming apps.

ahuth:
Thanks for taking a look, and the Qs, @booc0mtaco.

> Curious if someone can explain the difference between specifying that optimization in webpack versus what this does?

For the webpack optimization you do [specify this in the package.json](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free). It's also become quasi-standard, so things like esbuild also take it into account (which is how this affects Remix).

> Also, would our use of CSS modules imply a side-effect on import?

They don't. CSS module imports are basically just like importing a JS module. The "css code" doesn't run when its imported. It "runs" later when the component is rendered.

jinlee93:
>> Also, would our use of CSS modules imply a side-effect on import?

> They don't. CSS module imports are basically just like importing a JS module. The "css code" doesn't run when its imported. It "runs" later when the component is rendered.

If styles are built into a stylesheet, i.e. for EDS build, would usage of that stylesheet be a side-effect?

jeremiah-clothier (approved):
LGTM

Out of curiosity, how did you notice this?
> Before this change: in the Remix test app code for all EDS components is present, despite only using Heading.
> After this change: only the Heading code is present.

booc0mtaco:
@ahuth Ah i got confused by this section, but now makes sense
Screenshot 2023-01-24 at 11 47 43

booc0mtaco (approved):
Makes sense to me. We will keep an eye on future side-effects

ahuth:
@jinlee93 

> If styles are built into a stylesheet, i.e. for EDS build, would usage of that stylesheet be a side-effect?

Negative, those styles still won't be a side effect. It'll be like the `lib/tokens/css/variables.css` case @anniehu4 brought up: either way the styles need to be explicitly imported by consuming apps.

"Side effects" for this repo applies only to modules imported into the main `index.ts`.

ahuth:
@jeremiah-clothier 

> Out of curiosity, how did you notice this?

Randomly noticed a 160kb-ish JS file being downloaded by the Remix test app, which was suspicious.

PR #1474, opened 2023-01-24
docs: update and clarify publishing instructions

booc0mtaco:
### Summary:

- add clarity on the message to use for the PR to `main`
- clarify when not to do what is output from the console

### Test Plan:

- n/a (documentation)

### DIFF
--- a/docs/PUBLISHING.md
+++ b/docs/PUBLISHING.md
@@ -44,7 +44,9 @@ yarn release
 yarn release:major
 ```
 
-We use [standard-version](https://github.com/conventional-changelog/standard-version) to increment the version number in `package.json`, create a git tag for the new release, and update `CHANGELOG.md` based on the commit log. The package is not published, yet. If needed, you can [make changes to the CHANGELOG now](#editing-the-changelog).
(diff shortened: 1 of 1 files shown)


docs/PUBLISHING.md
```diff
@@ -44,7 +44,9 @@ yarn release
 yarn release:major
 ```
 
-We use [standard-version](https://github.com/conventional-changelog/standard-version) to increment the version number in `package.json`, create a git tag for the new release, and update `CHANGELOG.md` based on the commit log. The package is not published, yet. If needed, you can [make changes to the CHANGELOG now](#editing-the-changelog).
+We use [standard-version](https://github.com/conventional-changelog/standard-version) to increment the version number in `package.json`, create a git tag for the new release, and update `CHANGELOG.md` based on the commit log. The package is not published, yet. If needed, you can [make additional changes to the CHANGELOG now](#editing-the-changelog).
+
+**NOTE**: running this command will print a command to run to complete the process. IGNORE THIS. Continue following the directions below.
```
jinlee93, line 49: Nice callout

jinlee93 (approved):

PR #1475, opened 2023-01-24
[EFI-875] chore: remove uneccessary usage of react-uid

jeremiah-clothier:
### Summary:

Removed usage of `useUID` from the three following components: `<Icon />`, `<PrimaryNav />` , and `<Breadcrumbs />`. Note, the `id` prop can still be passed to the components.

Why are we making this change?
To remove `console.error` messages from SSR in remix-run. Rendering these components would log errors like
> Warning: Prop `id` did not match. Server: "1" Client: "2"

### Test Plan:
- [x] Built library locally and validated the `console.error` message no longer appear when rendering these components

Follow-up, there is still a fair bit of usage of `react-uid` in the EDS repo. Those IDs are being used for accessibility and will be dealt with in a separate PR. This PR removes unnecessary usages

### DIFF
--- a/src/components/Icon/Icon.tsx
+++ b/src/components/Icon/Icon.tsx
@@ -1,7 +1,6 @@
 import clsx from 'clsx';
 import type { ReactNode, CSSProperties } from 'react';
 import React, { useEffect } from 'react';
-import { useUID } from 'react-uid';
 import svg4everybody from 'svg4everybody';
 import styles from './Icon.module.css';
 import icons, { type IconName } from '../../icons/spritemap';
@@ -112,8 +111,6 @@ export const Icon = (props: IconProps) => {
     size,
     viewBox,
   } = props;
-  const generatedId = useUID();
-  const idVar = id || generatedId;
 
   useEffect(() => {
     svg4everybody(); // Required to get IE to render icon sprites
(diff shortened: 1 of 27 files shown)


booc0mtaco:
@jeremiah-clothier for the components where we remove react-uid, can you make sure we have a test that confirms `id` can be passed down?

src/components/Icon/Icon.tsx
```diff
@@ -148,7 +145,7 @@ export const Icon = (props: IconProps) => {
   if (purpose === 'informative') {
     return (
       <svg {...svgCommonProps} role="img">
-        <title id={idVar}>{props.title}</title>
+        <title id={id}>{props.title}</title>
```
ahuth, line 148: 👍

ahuth (approved):
LGTM 🤘 

Thanks for the cleanup

jeremiah-clothier:
> @jeremiah-clothier for the components where we remove react-uid, can you make sure we have a test that confirms `id` can be passed down?

Does typescript + eslint count as a test? The `id` prop is destructured from props in each of the components which means if it wasn't being used then eslint would flag the unused variable and CI wouldn't pass

booc0mtaco (approved):

PR #1477, opened 2023-01-26
[feat] Upgrade storybook to use webpack5 manager

jeremiah-clothier:
### Summary:

Motivation: @reach/auto-id uses the ?? operator which was causing build errors. I was deciding between updating the storybook webpack config and just using the webpack5 builder. I think using the webpack5 builder is more simple.

This PR includes two things
* running `yarn lint:fix` locally
* installing dependencies for `webpack5` storybook and updating config

### Test Plan:
- [x] Rely on existing CI and storybook deployment


jinlee93 (approved):
Interesting that there's a chromatic diff.
Think it's an unrelated bug

jeremiah-clothier:
> Interesting that there's a chromatic diff.
> Think it's an unrelated bug

Yeah, I re-ran and the diff disappeared

PR #1478, opened 2023-01-26
test(ClickableStyle): add stories for the utility component

booc0mtaco:
### Summary:

- as we are recommending `ClickableStyle` for use with router-based link components, expose the doc.s for this so that it can be referenced in zeroheight and other places

### Test Plan:

- [x] stories and relevant snapshots are created

### DIFF
--- a/src/components/ClickableStyle/ClickableStyle.stories.tsx
+++ b/src/components/ClickableStyle/ClickableStyle.stories.tsx
@@ -1,0 +1,67 @@
+import type { StoryObj, Meta } from '@storybook/react';
+import React from 'react';
+
+import { ClickableStyle, STATUSES, VARIANTS } from './ClickableStyle';
+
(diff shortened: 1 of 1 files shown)


src/components/ClickableStyle/ClickableStyle.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+export const Default: StoryObj<Args> = {
+  args: {
+    children: 'link',
+  },
+};
+
+type ExampleProps = {
+  to: string;
+};
+const CustomElement = ({ to, ...rest }: ExampleProps) => (
```
booc0mtaco, line 45: This is clear in context, but to document it... the custom element here is arbitrary, and not a specific example of what can be don with `as`. But deliberately mimics what a router link from some library might look like

PR #1480, opened 2023-01-31
Many dep updates

ahuth:
This PR updates a lot of stuff:
- terser and loader-utils to resolve security warnings
- size-limit (and fixes it so it actually runs on most of our prs...)
- husky (which has completely changed 😑) and commitlint
- postcss
- prettier


.github/workflows/size-limit.yml
```diff
@@ -3,6 +3,7 @@ on:
   pull_request:
     branches:
       - main
+      - next
```
ahuth, line 6: Turns out when we changed branching model we never updated this, so size-limit hasn't been running on most PRs (just the release PRs into main).

I was tempted to just delete the size-limit stuff, as I'm not sure it's really providing that much value.

But figured what they hey, for now I'll just configure it to run on PRs into next.

.husky/commit-msg
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+yarn commitlint --edit $1
```
ahuth, line 4: husky now works based on these `.husky` files, instead of package.json entries.

Kind of annoying, but I want to stay up to date.

Tested all these out with various local commits.

.storybook/recipes/CalendarCard/CalendarCard.tsx
```diff
@@ -8,7 +8,7 @@ import type { Variant as TagVariant } from '../../../src/components/Tag/Tag';
 
 export const VARIANTS = ['brand', 'revise', 'success'] as const;
 
-export type Variant = typeof VARIANTS[number];
+export type Variant = (typeof VARIANTS)[number];
```
ahuth, line 11: Prettier changes

package.json
```diff
@@ -40,6 +40,9 @@
     "lint:styles:fix": "yarn run lint:styles --fix",
     "lint:scripts": "eslint --ignore-path .gitignore --ext=js,jsx,ts,tsx .",
     "lint:scripts:fix": "yarn run lint:scripts --fix",
+    "postinstall": "husky install",
+    "postpack": "pinst --enable",
+    "prepack": "pinst --disable",
```
ahuth, line 45: This pinst stuff is unfortunate, but Husky needs it now.

Apparently postinstall runs in consuming apps, as well. We don't want husky to install in other apps, so this disables that behavior or something.

package.json
```diff
@@ -62,14 +65,12 @@
     {
       "path": "lib/**/*([a-zA-Z-_]).js",
       "limit": "140KB",
-      "name": "components",
-      "webpack": false
```
ahuth, line 66: Don't need/want the `webpack` option, since we're checking the size of our own build command's output (vs using webpack to build).

package.json
```diff
@@ -100,11 +101,11 @@
...
     "@chanzuckerberg/story-utils": "^3.0.11",
-    "@commitlint/cli": "^16.3.0",
-    "@commitlint/config-conventional": "^16.2.4",
-    "@commitlint/prompt-cli": "^16.3.0",
+    "@commitlint/cli": "^17.4.2",
+    "@commitlint/config-conventional": "^17.4.2",
+    "@commitlint/prompt-cli": "^17.4.2",
     "@geometricpanda/storybook-addon-badges": "^0.2.2",
-    "@size-limit/preset-small-lib": "^4.12.0",
+    "@size-limit/file": "^8.1.2",
```
ahuth, line 108: Replaced @size-limit/preset-small-lib with @size-limit/file.

The "small-lib" preset uses webpack to do its thing. Based on our config we clearly want to check the result of our own `build` command, and can use @size-limit/file directly.

package.json
```diff
@@ -122,57 +123,51 @@
...
     "@types/lodash.debounce": "^4.0.7",
-    "@types/node": "^17.0.21",
+    "@types/node": "^18.11.18",
     "@types/react": "^17.0.53",
     "@types/react-beautiful-dnd": "^13.1.3",
     "@types/react-dom": "^17.0.18",
     "@types/react-portal": "^4.0.4",
     "@types/svg4everybody": "^2.1.2",
-    "axe-core": "^4.4.3",
+    "axe-core": "4.4.3",
```
ahuth, line 132: Locking axe-core to 4.4.3. There's timeouts on newer versions per 4862f030fb8340221d03661179cbbf04cfaa9525

.github/workflows/size-limit.yml
```diff
@@ -3,13 +3,14 @@ on:
...
 jobs:
   size:
     runs-on: ubuntu-latest
     env:
       CI_JOB_NUMBER: 1
     steps:
       - uses: actions/checkout@v1
-      - uses: chanzuckerberg/size-limit-action@master
+      - uses: andresz1/size-limit-action@v1
```
ahuth, line 14: Trying out using the upstream action instead of our fork.

Seems our fork is doing stuff to prevent more than 1 size-limit comment per PR. Not sure the added complexity of a fork is worth worrying about doing that.

jeremiah-clothier (approved):

ahuth:
Thanks 🙏

PR #1481, opened 2023-01-31
fix(build): add css as having side effects

jinlee93:
### Summary:
- tokens are not being provided since css marked as side effect free https://614b969675171d003ac6f98a-jstsqkayzt.chromatic.com/?path=/story/core-components-projectcard--default
- adds css files as having side effects
### Test Plan:
- LP storybook unbroken
  - install `@chanzuckerberg/eds: 8.1.0-alpha.4` in LP, where `8.1.0-alpha.4` is the latest alpha publish off next
  - in `traject/node_modules/@chanzuckerberg/eds/package.json` change `sideEffects: false` to 
  ```
  "sideEffects": [
    "lib/tokens/css/variables.css",
    "lib/tokens/fonts.css"
  ]
  ```
  - run `npm run storybook` and see that storybook is unbroken
    - chromatic: https://www.chromatic.com/build?appId=614b969675171d003ac6f98a&number=16062
  - alternatively install `8.1.0-alpha.5` which is alpha publish off of this branch
    - chromatic: https://www.chromatic.com/build?appId=614b969675171d003ac6f98a&number=16065
  - chromatic without `lib/tokens/fonts.css` in `sideEffects`: https://www.chromatic.com/build?appId=614b969675171d003ac6f98a&number=16064

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
   "license": "MIT",
   "main": "lib/index.js",
   "sideEffects": [
-    "*.css"
+    "lib/tokens/css/variables.css",
+    "lib/tokens/fonts.css"
   ],
   "publishConfig": {
     "access": "public"
--- a/package.json
+++ b/package.json
@@ -6,7 +6,9 @@
   "homepage": "https://github.com/chanzuckerberg/edu-design-system",
   "license": "MIT",
   "main": "lib/index.js",
-  "sideEffects": false,
+  "sideEffects": [
+    "*.css"
+  ],
   "publishConfig": {
     "access": "public"
   },


anniehu4 (approved):
tyty! could you add to the test plan how you verified the fix with alpha publishes?

anniehu4:
Follow-up to https://github.com/chanzuckerberg/edu-design-system/pull/1473

package.json
```diff
@@ -6,7 +6,10 @@
   "homepage": "https://github.com/chanzuckerberg/edu-design-system",
   "license": "MIT",
   "main": "lib/index.js",
-  "sideEffects": false,
+  "sideEffects": [
+    "lib/tokens/css/variables.css",
+    "lib/tokens/fonts.css"
```
ahuth, line 11: What's happening on the Traject side without this?

When writing #1473, I didn't think this was necessary because Traject should be importing these explicitly.

jinlee93: Fonts and CSS custom properties are being lost  
image

jinlee93:
> tyty! could you add to the test plan how you verified the fix with alpha publishes?

Yup, done

package.json
ahuth: Dang, thanks for digging into this. Looks like you were totally right @anniehu4.

There's still something I don't understand here about how Webpack handles tree shaking (Jin and I confirmed that it does work as we expected in esbuild) 😓 

There are some old docs at https://github.com/webpack/webpack.js.org/pull/1865/files which mention marking css files as side effects. Other than that I haven't found much.

PR #1483, opened 2023-02-01
refactor!: remove unused components 3/3

jinlee93:
[EDS-797]
### Summary:
- this should be the final shifting/removal of components (for now)
- moves `Header` component into `GlobalHeader`, recipe since it is only used in there 
- moves `PrimaryNavItem` subcomponent into `PrimaryNav`
- moves entirety of `PrimaryNav` as a recipe
- Both `Header` and `PrimaryNav` do seem to have some useful behavior, such as stickiness, and semantic HTML for <nav>  but as they stand, seem very specific to LP and is not even used in there
  - having these in recipes will provide a reference point in case we want to explore these in the future
### Test Plan:
- LP doesn't use `<Header>`, `<PrimaryNav>`, `<PrimaryNavItem>` so removing them is fine
  - These are pretty LP specific so shouldn't affect pilot apps (made sure gst not using tho)
- No visual regression from the shifting of these components

[EDS-797]: https://czi-tech.atlassian.net/browse/EDS-797?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


booc0mtaco:
Let's hold on to this one for a little while, until we sort out what should go in our 8.1 (or next) release. Looks good tho

booc0mtaco (approved):

PR #1484, opened 2023-02-01
refactor: shorthand prop={true}

jinlee93:
### Summary:
- Just refactors `<Component prop={true}>` to just `<Component prop>` to match rest of repo 
- Lint warns on `react/jsx-boolean-value` for this 
### Test Plan:
- n/a

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -40,7 +40,8 @@
       }
     ],
     "@typescript-eslint/no-unused-vars": "warn",
-    "no-unused-vars": "off"
+    "no-unused-vars": "off",
+    "react/jsx-boolean-value": "warn"
   },
   "overrides": [
     {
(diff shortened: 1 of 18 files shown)


ahuth (approved):

jeremiah-clothier (approved):
There's also a lint rule to enforce and auto-fix if EDS wants to keep this consistent https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md

jinlee93:
> There's also a lint rule to enforce and auto-fix if EDS wants to keep this consistent https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md

Ohhh thanks for the rule, added

PR #1485, opened 2023-02-03
refactor!: rename inputfield to input and textfield to inputfield

jinlee93:
[EDS-829]

### Summary:
- the old `<InputField>` is a subcomponent of `<SearchBar>` and old `<TextField>` component and shouldn't be used by itself
- renames old `<InputField>` component to `<Input>` which is more appropriate 
- renames old `<TextField>` component to `<InputField>` which matches ZeroHeight and Figma
- should be batched with the other breaking change
### Test Plan:
- no changes in functionality or visual regression, manually tested in storybook stories


src/components/Input/Input.stories.tsx
```diff
@@ -0,0 +1,52 @@
+import { BADGE } from '@geometricpanda/storybook-addon-badges';
```
anniehu4, line 1: Is it worth having stories & tests for this if it's covered by the `InputField` stories? I guess my mental model is this is a subcomponent like `CheckboxInput` (or maybe better example is `PopoverContainer`, which is also used by multiple components). Worry that it's confusing to show stories for a component that's only for internal use

src/components/InputField/InputField.test.ts
```diff
@@ -1,6 +1,6 @@
 import { generateSnapshots } from '@chanzuckerberg/story-utils';
 import * as InputFieldStoryFile from './InputField.stories';
 
-describe('<InputField />', () => {
+describe('<TextField />', () => {
```
anniehu4, line 4: 👀

jinlee93: Oop nice catch

src/components/Input/Input.stories.tsx
jinlee93: That's a good point. I was being cautious about removing tests / stories but do agree that it's not worth having these. Will remove

.storybook/pages/StudentRefinement/StudentRefinement.tsx
```diff
@@ -473,7 +473,7 @@ export const StudentRefinement = () => {
               </div>
               <Layout className="mb-8" variant="50-50">
                 <SearchBar>
-                  <SearchBar.InputField />
+                  <SearchBar.Field />
```
booc0mtaco, line 476: ++

src/components/Input/Input.module.css
```diff
@@ -0,0 +1,13 @@
+@import '../../design-tokens/mixins.css';
+
+/*------------------------------------*\
+    #TEXT INPUT
```
booc0mtaco, line 4: `# INPUT`

src/components/Input/Input.tsx
```diff
@@ -0,0 +1,124 @@
+import clsx from 'clsx';
+import type { ChangeEventHandler } from 'react';
+import React, { forwardRef } from 'react';
+import styles from './Input.module.css';
+
+export type InputProps = React.InputHTMLAttributes<HTMLInputElement> & {
```
booc0mtaco, line 6: I think many of these attributes will be covered by unioning with `HTMLInputElement` already

jinlee93: I removed `aria-describedby` prop but kept the others since they provide helpful hints 
When removed, there's no hint: Screenshot 2023-02-07 at 3 05 30 PM

src/components/Input/Input.module.css
jinlee93: Thanks!

src/components/Input/Input.tsx
booc0mtaco: It's weird; i would have expected any internal docs to show for the attributes ... [but it doesn't seem to](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1848-L1915) 

I guess we can keep our own for now for documentation purposes

booc0mtaco (approved):

PR #1488, opened 2023-02-06
refactor(text): deprecate text passage and spacing variants

jinlee93:
### Summary:
- deprecates the `as="div"` aka Text Passage variant of the Text component since its utility can be easily replicated with utility classes
- deprecates the `spacing` variants of the Text component since its utility can be easily replicated with utility classes, of which there are more options
### Test Plan:
- Text passage stories have the `deprecated` badge
- Text spacing story has the `deprecated` badge
- Text docstrings render fine in storybook

### DIFF
--- a/src/components/Text/Text.tsx
+++ b/src/components/Text/Text.tsx
@@ -46,7 +46,8 @@ export type Props = {
   variant?: Variant;
   size?: Size;
   /**
-   * Adds margin bottom spacing to the <Text> component.
+   * Deprecated. Adds margin bottom spacing to the <Text> component. Use utility classes instead.
+   * @deprecated
    */
   spacing?: 'half' | '1x' | '2x';
   tabIndex?: number;
--- a/src/components/Text/Text.tsx
+++ b/src/components/Text/Text.tsx
@@ -31,9 +31,9 @@ export type Variant =
 
 export type Props = {
   /**
-   * Controls whether to render text inline (defaults to "p");
-   * Use "div" to indicate usage of `<Text>` as a text passage,
-   * (i.e. wraps <p>, <h1>-<h6>, <a>, <ol>, <ul>, <blockquote>, <hr>)
+   * Controls whether to render text inline (defaults to "p").
+   *
+   * "div" variant is deprecated.
    */
   as?: 'p' | 'span' | 'div';
   /**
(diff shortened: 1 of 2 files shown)


booc0mtaco:
I saw we have this `spacing` prop on this component. Do people use this prop instead of specifying classes that apply bottom margins? what does `'1x'` or `'2x'` mean? are they line-height equivalents?

Perhaps we should deprecate this as well? Margin-bottom doesn't apply like one would expect on inline elements, and it could conflict with custom classes or styles applied to the component if it works by adding a bottom margin directly.

```
  /**
   * Adds margin bottom spacing to the <Text> component.
   */
  spacing?: 'half' | '1x' | '2x';
```

jinlee93:
> I saw we have this `spacing` prop on this component. Do people use this prop instead of specifying classes that apply bottom margins? what does `'1x'` or `'2x'` mean? are they line-height equivalents?
> 
> Perhaps we should deprecate this as well? Margin-bottom doesn't apply like one would expect on inline elements, and it could conflict with custom classes or styles applied to the component if it works by adding a bottom margin directly.
> 
> ```
>   /**
>    * Adds margin bottom spacing to the <Text> component.
>    */
>   spacing?: 'half' | '1x' | '2x';
> ```

I am down to deprecate `spacing` as well. They co-relate to our size tokens but def do agree there could be conflict with custom classes. This definitely would be a breaking change for LP but I think the effort required to convert them to TW utility classes is pretty low
image

src/components/Text/Text.tsx
```diff
@@ -46,7 +46,8 @@ export type Props = {
   variant?: Variant;
   size?: Size;
   /**
-   * Adds margin bottom spacing to the <Text> component.
+   * Deprecated. Adds margin bottom spacing to the <Text> component. Use utility classes instead.
```
jeremiah-clothier, line 49: Clarifying question, what do you mean by `utility classes`, are you referring to tailwind classes?

booc0mtaco: It would likely be tailwind, or could be some class created in the product that applies spacing to a container, etc. This one is good to avoid b/c it mixes strategies on how things would get bottom margins outside their component view

booc0mtaco:
@jinlee93 Gotcha re: LP. I do like the pattern of marking it as deprecated now, with the line suggestion on what to do instead. Also `spacing` could be confused with "line spacing" or `line-height` which is what I thought it was doing at first...

jinlee93:
> Let's also update the PR title/commit message to include a note about spacing being deprecated

Will do

PR #1489, opened 2023-02-06
chore: remove usage of svg4everybody

anniehu4:
### Summary:

Removes use of `svg4everybody`. It was brought over from BM to support rendering icon sprites in Internet Explorer, but:
- we don't support IE (and never did ... should have just removed this in the first place 🤦 )
- we also no longer use an svg spritemap for icons due to SSR

I was inspired seeing @jeremiah-clothier's PR for EDS jest config requirements in product apps and remembered we needed some mocking for this package

### Test Plan:
should be no visual changes

### DIFF
--- a/jest/jest.setup.ts
+++ b/jest/jest.setup.ts
@@ -1,10 +1,6 @@
 import '@testing-library/jest-dom';
 import { format } from 'util';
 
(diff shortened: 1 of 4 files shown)


jeremiah-clothier (approved):
🥳

PR #1491, opened 2023-02-08
chore: consolidate usage of viewport values

booc0mtaco:
### Summary:

- use same viewport constants in stories and chromatic config
- move viewports file to src/utils
- add types for future reference

### Test Plan:

- [x] ensure chromatic snapshots do not regress


jeremiah-clothier (approved):

src/components/Breadcrumbs/Breadcrumbs.stories.tsx
```diff
@@ -104,7 +111,7 @@ export const LongTextMenu: StoryObj<Args> = {
     viewport: {
       defaultViewport: 'tablet',
```
jinlee93, line 112: Could you also change this to one of the new viewports?

jinlee93, line 112: Are these `defaultViewport` values not affected?

booc0mtaco: Good call. we have 5 examples of this default viewport and a few of them aren't pointing to anything. Will update them to reference the keys from this viewport config

jinlee93 (approved):
Looks good, thanks!

PR #1493, opened 2023-02-10
feat(TextareaField): add TextArea base component and TextareaField

booc0mtaco:
### Summary:

- create TextArea component
- for now exporting, but we would want to treat this similar to how we do the input base component
- will sync with latest design before merging

### Test Plan:

- [x] add snapshots and base tests
- [x] verify no other snapshots change

### TODO

- [x] create proper exportable component (implementing label, error message, counting when maxlength used, etc.)
- [x] brief discussion on how to (re-)organize such base components (atoms that map directly to HTML tags)

### DIFF
--- a/src/components/TextArea/TextArea.module.css
+++ b/src/components/TextArea/TextArea.module.css
@@ -1,0 +1,16 @@
+@import '../../design-tokens/mixins.css';
+
+/*------------------------------------*\
+    # TEXTAREA 
+\*------------------------------------*/
+
+/**
+ * Default input styles 
+ */
+.textarea {
+  @mixin inputStyles;
+}
+
+.textarea--disabled:focus {
+  outline: none;
+}
--- a/src/components/TextArea/TextArea.tsx
+++ b/src/components/TextArea/TextArea.tsx
@@ -1,0 +1,56 @@
+import clsx from 'clsx';
+import React, { forwardRef } from 'react';
+import styles from './TextArea.module.css';
+
+export interface Props
+  extends React.TextareaHTMLAttributes<HTMLTextAreaElement> {
+  /**
+   * CSS class names that can be appended to the component
+   */
+  className?: string;
+  /**
+   * Text default contents of the field
+   */
+  children?: string;
+  /**
+   * Whether the disabled stat is active
+   */
+  disabled?: boolean;
+  /**
+   * Whether the error state is active
+   */
+  isError?: boolean;
+}
+
+/**
+ * Base component, applying styles to a <textarea> tag
+ */
+export const TextArea = forwardRef<HTMLTextAreaElement, Props>(
+  (
+    {
+      className,
+      children,
+      disabled,
+      defaultValue = '',
+      isError = false,
+      ...other
+    },
+    ref,
+  ) => {
+    const componentClassName = clsx(
+      styles['textarea'],
+      isError && styles['error'],
+      disabled && styles['textarea--disabled'],
+      className,
+    );
+
+    return (
+      <textarea
+        className={componentClassName}
+        defaultValue={children || defaultValue}
+        ref={ref}
+        {...other}
+      ></textarea>
+    );
+  },
+);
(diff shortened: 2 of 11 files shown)


booc0mtaco:
Notes to self:

- structure should mimic `RadioInput` and `CheckboxInput`
- fix up the a11y problems from test
- review with Ryan once designs are finalized

booc0mtaco:
@jinlee93 we can talk thru the states when an input is required and empty. b/c we don't apply `.error` on empty/required, the UI shows no problem. if we were to add `:invalid` to the inputStyles mixin to match .error, any empty fields would be styled as an error when required is true in chromatic

I was thinking to try and use `:invalid` across our inputs, but will save that for another time. I added a TODO for this in the PR for consideration at some point

src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,56 @@
...
+  disabled?: boolean;
+  /**
+   * Whether the error state is active
+   */
+  isError?: boolean;
+}
+
+/**
+ * Base component, applying styles to a <textarea> tag
+ */
```
jinlee93, line 27: I've been keeping the beta comment to align with the storybook tag, and how to import the component, which shows up in the Storybook docs page: 
image

src/components/TextareaField/TextareaField.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+  args: {
+    required: true,
+  },
+};
+
+export const WithADifferentSize: StoryObj<Args> = {
+  args: {
+    rows: 10,
+  },
+};
```
jinlee93, line 83: What do you think about having an interactive story to reflect the ongoing character count, like:
image

src/design-tokens/mixins.css
```diff
@@ -133,13 +133,15 @@
    * Input error state
    */
   &.error {
+    /* TODO: should this color be applied when a field is in an :invalid state */
```
jinlee93, line 136: Can we make input have an `invalid` state manually, or is it interpreted from fields such as the `disabled` attribute?

src/components/TextArea/TextArea.module.css
```diff
@@ -0,0 +1,16 @@
+@import '../../design-tokens/mixins.css';
+
+/*------------------------------------*\
+    # TEXTAREA
```
jinlee93, line 4: nit: is textarea one word or two words loll 🤔 (doesn't actually need to change)

booc0mtaco: @jinlee93 lol, treating it as one word since [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea) does. which may mean the base component name should lower-case the A :)

src/design-tokens/mixins.css
booc0mtaco: `:invalid` is determined by the browser based on other attributes, and the value in the field, so it's not really an attribute unto itself. for `<textarea>`s `:invalid` applies if the value in the field is outside the bounds of what `minlength` and `maxlength` says is allowed (and in the case of `maxlength` only after some user action triggers the check, which is weird)

For `<input>`s, `:invalid` applies if you have a `required` field, and it becomes empty OR if the [`pattern`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#individual_attributes) is violated, along with the above. 

When i added `&:invalid` to the selector list on line 135, it changed some TextField snapshots to an error state b/c the fields didn't have any default value. this means that blank forms might start out with red outlines if marked as required, which we don't want

src/components/TextareaField/TextareaField.stories.tsx
booc0mtaco: not a bad idea. I was planning to create a separate ticket for the text count thing, as there are a few behaviors i want to discuss.

jinlee93 (approved):
Looks good!

src/components/TextareaField/TextareaField.stories.tsx
booc0mtaco: @jinlee93 created ticket and tagged you on that one!

PR #1497, opened 2023-02-16
feat: bump to React 18, user-event v14, remove reach/auto-id

jinlee93:
### Summary:
- with the version locking of EDS to v8 on LP, EDS can bump to React 18 now
- this allows us to use React's `useId` hook instead of depending on `reach/auto-id`
- however this caused some test failures not sure why 
- test failures were mostly resolved via bumping to user-event v14 and using and awaiting the new user methods
- whilst it would have been nice to have 3 separate prs (react 18, user-event v14, and reach/auto-id), react 18 itself won't pass without user-event v14 and reach/auto-id checks for the React useId hook and uses it if present so snaps wouldn't have changed anyways by itself
### Test Plan:

  - converted imports of `userEvent` from `@storybook/testing-library` to `@testing-library/user-event`
  - followed [recommended userEvent setup](https://testing-library.com/docs/user-event/intro#writing-tests-with-userevent) to convert old tests
  - most ci tests pass


.storybook/pages/AlongDemo/AlongDemo.stories.ts
```diff
@@ -1,5 +1,5 @@
 import type { StoryObj, Meta } from '@storybook/react';
-import { within } from '@storybook/testing-library';
+import { userEvent, within } from '@storybook/testing-library';
```
ahuth, line 2: https://github.com/storybookjs/testing-library/issues/26 is tracking supporting user-event 14 in @storybook/testing-library.

.storybook/pages/WireframeDemo/WireframeDemo.tsx
```diff
@@ -1,6 +1,5 @@
-import { useId } from '@reach/auto-id';
 import clsx from 'clsx';
-import React, { useState } from 'react';
+import React, { useId, useState } from 'react';
```
ahuth, line 2: Wow, super convenient that the hook name is the same 🎉

package.json
```diff
@@ -76,12 +76,11 @@
     }
   ],
   "peerDependencies": {
-    "react": ">= 16.8.0"
+    "react": ">= 18"
```
ahuth, line 79: 👍

src/components/Accordion/Accordion.stories.tsx
```diff
@@ -217,7 +217,7 @@ export const UsingRenderProp: StoryObj<Args> = {
   render: () => (
     <Accordion headingAs="h2">
       <Accordion.Row>
-        {({ open }) => (
+        {({ open }: { open: boolean }) => (
```
ahuth, line 220: Render prop isn't getting typed right now?

src/components/Checkbox/__snapshots__/Checkbox.test.tsx.snap
```diff
@@ -14,7 +14,7 @@ exports[`<Checkbox /> Checked story renders snapshot 1`] = `
         checked=""
         class="checkbox__input"
         data-bootstrap-override="checkbox"
-        id="3"
+        id=":r1:"
```
ahuth, line 17: 🤘

src/components/Menu/Menu.tsx
```diff
@@ -66,13 +66,15 @@ const MenuButton = ({ children, className, ...other }: MenuButtonProps) => {
...
       {...other}
     >
-      {children}
-      <Icon
-        className={styles['menu__button--icon']}
-        name="expand-more"
-        purpose="decorative"
-        size="1.25rem"
-      />
+      <>
```
ahuth, line 69: I'm surprised we need a fragment here

src/components/Menu/Menu.tsx
```diff
@@ -114,7 +116,7 @@ const MenuItem = ({
             disabled={disabled}
             icon={icon}
           >
-            {children}
+            {children as React.ReactNode}
```
ahuth, line 119: TS doesn't think `children` is a ReactNode?

ahuth (approved):
LGTM 🤘 

Thanks for updating all the tests

src/components/AccordionRow/AccordionRow.tsx
```diff
@@ -12,7 +12,9 @@ type Props = {
   /**
    * Child node(s) that can be nested inside component.
    */
-  children: React.ReactNode;
+  children:
```
booc0mtaco, line 15: Isn't this what lines 7-9 and 26-31 already doing? ... or supposed to be doing? The types for children appear to be mixed up.

src/components/Checkbox/Checkbox.test.tsx
```diff
@@ -23,11 +23,13 @@ describe('<Checkbox />', () => {
...
 
-  test('should toggle the checkbox with space', () => {
+  // eslint-disable-next-line jest/no-disabled-tests
+  test.skip('should toggle the checkbox with space', async () => {
+    const user = userEvent.setup();
     render(<Default />);
     const checkbox = screen.getByRole('checkbox');
     checkbox.focus();
-    userEvent.keyboard('{space}');
+    await user.keyboard(' ');
```
booc0mtaco, line 32: shouldn't `{space}` and `' '` behave in the same way? why is this one being disabled... flaky?

The doc.s around [`.keyboard`](https://testing-library.com/docs/user-event/keyboard) are a little weak for my tastes; i wonder where the other special keys are documented.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -74,7 +74,7 @@ type DropdownProps = ListboxProps & {
   optionsClassName?: string;
 };
 
-type RenderProp<Arg> = (arg: Arg) => ReactNode;
+type RenderProp<Arg> = (arg: Arg) => ReactElement;
```
booc0mtaco, line 77: I assume some of these changes are b/c of React 18 being slightly different or more strict?

Also in this case `children` is either a `ReactNode` or `ReactElement`. So when we cast on ln285, would this cause issues? is one of these a super-set of the other?

src/components/Menu/Menu.tsx
booc0mtaco: @ahuth this might be b/c headlessUI added warnings when nodes have more children than "they should". We still have some warnings that are false positives, so not sure how much actual value that has given us. :P

src/components/Popover/Popover.tsx
```diff
@@ -182,7 +182,9 @@ const PopoverContent = ({
       as="article"
       className={componentClassName}
     >
-      <PopoverContainer className={bodyClassName}>{children}</PopoverContainer>
+      <PopoverContainer className={bodyClassName}>
+        {children as React.ReactNode}
```
booc0mtaco, line 186: What happens when you don't cast the type here?

src/components/Radio/Radio.test.tsx
```diff
@@ -16,11 +16,13 @@ describe('<Radio />', () => {
     }).toThrow(/must provide a visible label or aria-label/);
   });
 
-  test('should toggle the radio with space', () => {
+  // eslint-disable-next-line jest/no-disabled-tests
```
booc0mtaco, line 19: Why do these tests get skipped? I saw another one earlier

booc0mtaco (commented):
For any of the HeadlessUI components, we should be able to borrow from the types it exports (e.g., `Accordion` => `Disclosure` etc.) to avoid any custom render prop types in a few places

jeremiah-clothier (approved):

timzchang (approved):
Exciting! 🎉

src/components/Menu/Menu.tsx
ahuth: Gotcha, thanks

src/components/Checkbox/Checkbox.test.tsx
jinlee93: It probably should behave in the same way but `' '` worked while `{space}` didnt haha

src/components/AccordionRow/AccordionRow.tsx
jinlee93: Ya you're right, types inherited after removing the duplicate

src/components/Radio/Radio.test.tsx
jinlee93: Found cause of need for skip, no longer skipping (removing throwing error on console error)

src/components/Dropdown/Dropdown.test.tsx
```diff
@@ -86,7 +89,8 @@ describe('<Dropdown />', () => {
     const renderMethod = () => {
       render(dropdownWithChildrenAndButtonText);
     };
-
+    // expect console error from react, suppressed.
+    jest.spyOn(console, 'error').mockImplementation();
```
booc0mtaco, line 93: nit: you can place this in `beforeEach` to apply to each test. also it would be good to add a `afterEach` to reset, so that we don't leave any mocks between tests

https://jestjs.io/docs/jest-object#jestspyonobject-methodname-accesstype

booc0mtaco (approved):
one testing suggestion but otherwise looks good to me

src/components/Dropdown/Dropdown.test.tsx
jinlee93: I didn't place in a `beforeEach` in case an error actually needs to be thrown during testing. However, I did add `mockRestore`s after each use for clean up

PR #1502, opened 2023-02-17
test(badge): suppress expected console error

jinlee93:
### Summary:
- suppresses expected console error that was missed previously

### Test Plan:


booc0mtaco (approved):
for future reference, you can nest `describe` blocks, so if we ever end up with a bunch of tests that require mocking things like this, we can have `describe('on error')` and an appropriate before and after each just for that one 🧰

jinlee93:
> for future reference, you can nest `describe` blocks, so if we ever end up with a bunch of tests that require mocking things like this, we can have `describe('on error')` and an appropriate before and after each just for that one 🧰

ahh makes sense

PR #1503, opened 2023-02-18
feat(slider): create slider

jinlee93:
[EDS-818]
### Summary:
- motivation:
  - GST presented a need for a input component with a moving thumb along a track (aka "Slider")
  - Discussed how to display the value of the slider
    - May have to rely on consumer but will show recipes
      - Could be tooltip
      - Could be labels on ends
- implementation:
  - Usage of semantic `<input type="range">` element
  - Usage of pseudo elements to style the thumb and the track for different browsers
    - Alternatively could use non semantic elements and overlay elements ourselves, but would have to put more time into accessibility considerations

### Test Plan:

Screenshot 2023-02-17 at 4 30 09 PM
Screenshot 2023-02-17 at 4 29 51 PM


src/components/Slider/Slider.module.css
```diff
@@ -0,0 +1,110 @@
...
+  --slider-track-height: var(--eds-size-1);
+  --slider-thumb-size: var(--eds-size-2-and-half);
+
+  appearance: none;
+  background: transparent;
+
+  height: var(--slider-thumb-size);
+}
+.slider__input:focus {
+  outline: none;
```
jinlee93, line 29: This will focus the entire slider rather than the button

src/components/Slider/Slider.module.css
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/**
+ * Slider Input Track
+ */
+/* Firefox */
+.slider__input::-moz-range-track {
```
jinlee93, line 36: Main concern is using these pseudo elements

src/components/Slider/Slider.module.css
```diff
@@ -0,0 +1,110 @@
...
+  /* fills left side of track as a percentage of the input value */
+  background: linear-gradient(
+    /* fill from left to right */ to right,
+    /* fill color */ var(--eds-theme-color-background-brand-primary-strong)
+      /* percent to fill */ calc(var(--ratio) * 100%),
+    /* unfilled color */ var(--eds-theme-color-background-neutral-medium)
+      /* don't blend the colors */ 0
```
jinlee93, line 43: Hopefully these comments are explanatory

jeremiah-clothier (commented):
> May have to rely on consumer but will show recipes
> - Could be tooltip
> - Could be labels on ends

It would be nice to also have markers/tracks with labels on the sliders. I have a few questions about how do recipes typically work for EDS.

1. Are these just stories in `Slider.stories.tsx` are the stored elsewhere?
2. Are recipes exported from EDS or just purely examples to mimic in other repositories

src/components/Slider/Slider.module.css
jeremiah-clothier: Is it worth writing a `play` function that tabs onto the input so the focus is captured by chromatic or is that overkill?

jinlee93:
> > May have to rely on consumer but will show recipes
> > 
> > * Could be tooltip
> > * Could be labels on ends
> 
> It would be nice to also have markers/tracks with labels on the sliders. I have a few questions about how do recipes typically work for EDS.
> 
> 1. Are these just stories in `Slider.stories.tsx` are the stored elsewhere?
> 2. Are recipes exported from EDS or just purely examples to mimic in other repositories

I also agree that it would be nice to provide markers for the slider, but using the native list markers doesn't align with `<datalist>` out of the box, and the aligning the markers with the center of the marker labels would differ depending on the number of markers, width of the marker labels, emoji, etc. Also we would have to consider a case where there are a lot of options (like a volume slider) where a tooltip may be a more useful visual indicator, and thus we tentatively arrived at not having markers out of the box (may change, and hence pr is draft lol)

Regarding recipes:
1. Recipes can be found in the Storybook folder "Recipes", [for example](https://chanzuckerberg.github.io/edu-design-system/?path=/story/recipes-filters--default), which are at the bottom due to alphabetic organization amongst folders (components, pages, recipes). 
2. Recipes are not exported and are code examples to mimic / copypasta

src/components/Slider/Slider.module.css
```diff
@@ -0,0 +1,110 @@
...
+  outline: none;
+}
+
+/**
+ * Slider Input Track
+ */
+/* Firefox */
+.slider__input::-moz-range-track {
+  /* fills left side of track as a percentage of the input value */
+  background: linear-gradient(
```
jinlee93, line 38: Would be nice to use `accent-color` but to my knowledge only colors the filled track and is [too modern](https://developer.mozilla.org/en-US/docs/Web/CSS/accent-color#browser_compatibility) for [our browser policy](https://czi.atlassian.net/wiki/spaces/LP/pages/2057733851/Edu+Tech+Inclusive+Browser+Support+Policy)

src/components/Slider/Slider.module.css
jinlee93: Ya, I think it's worth snapping!
Would it replace the default snap or should I make a new story for focus purpose?

jeremiah-clothier:
fyi @allisonbrownczi I think this PR is relevant for https://czi-tech.atlassian.net/browse/GST-108?

allisonbrownczi:
@jinlee93 I can't seem to respond to specific comments, this is in regards to the tooltip vs markers. 
When using this component would it be possible to add a dataset (ie pass in a `list` attribute) and have us consumers style it ourselves?

jinlee93:
> @jinlee93 I can't seem to respond to specific comments, this is in regards to the tooltip vs markers. When using this component would it be possible to add a dataset (ie pass in a `list` attribute) and have us consumers style it ourselves?

I think in a week or two would have a better answer, but for now:
This current implementation relies on native `<input type="range">` so it does allow for using `list` attribute (for now), but some thoughts:

1) Can't style [native tick marks](https://css-tricks.com/sliding-nightmare-understanding-range-input/#aa-tick-marks-and-labels) that's used with `list` attribute easily and for each browser
2) We're moving towards [a consistent design](https://www.figma.com/file/fqvLTr6OPuhr7G97Bd0nhF/Sliding-Selection?node-id=9388%3A27878&t=oRcKjBo87bI013l4-4) for what the marker labels would look like, feel free to list any concerns here or in the mocks
3) We're thinking consumers would add the marker labels separate from the slider component
4) `<datalist>` options don't display by default for Safari, would have to figure that out 

I can pair / discuss / feel free to bring up any thoughts / comments 
@booc0mtaco @anniehu4

src/components/Slider/Slider.module.css
jinlee93: Sounds good, went with `--slider-track-background`

src/components/Slider/Slider.module.css
jinlee93: Found that the entirety of the slider is interactable, but made the hit box 44px high so it will be AAA compliant. Focus indicator remains on the thumb, however, and is snapped on chromatic

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,116 @@
...
+              key={'slider-option-' + marker}
+            >
+              {marker}
+            </p>
+          ))}
+        </div>
+      )}
+    </div>
+  );
+};
```
jinlee93, line 116: Considered subcomponentizing the input and the markers but both need the css properties of the whole component to be styled properly

src/components/Slider/Slider.module.css
```diff
@@ -0,0 +1,155 @@
...
+  gap: var(--eds-size-1-and-half);
+}
+
+/**
+ * Slider Input
+ */
+.slider__input {
+  /* increases vertical hitbox for target size accessibility */
+  padding-top: 22px;
+  padding-bottom: 22px;
```
jinlee93, line 25: Increased slider hitbox

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,116 @@
...
+   */
+  id?: string;
+  /**
+   * HTML label text
+   */
+  label?: string;
+  /**
+   * List of markers to imply slider value.
+   */
+  markers?: string[];
```
jinlee93, line 34: Currently constrained to strings (which could also incl emoji, for better or for worse) and evenly separates them out via `justify-content: space-between`. Could consider expanding to `ReactNode` if the need arises

jinlee93:
@allisonbrownczi Added `markers` prop (`string[]`) for markers. Didn't go with `<datalist>` b/c Safari really does not want to display that element... Should be fine since screen readers rely on value of the input and the markers are for visual purposes only. It does lose the `autocomplete`(snapping) feature of using `list` attribute but I think it's worth losing for Safari
cc @booc0mtaco

src/components/Slider/Slider.tsx
allisonbrownczi: I think strings should be enough for our usecase at least

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,116 @@
...
+          { '--ratio': (value - min) / (max - min) } as CSSProperties
+        }
+        type="range"
+        value={value}
+        {...other}
+      />
+      {markers && (
+        <div className={styles['slider__markers']}>
+          {markers.map((marker) => (
+            <p
```
booc0mtaco, line 105: hm, why a `<p>`?

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * BETA: This component is still a work in progress and is subject to change.
+ *
+ * `import {Slider} from "@chanzuckerberg/eds";`
+ *
+ * Allows input of a value via dragging a thumb along a track.
+ * Strict: This slider requires a visual indicator of value/markers.
+ * Please check out our recipes for possible ideas.
+ *
+ * TODO: update this comment with a description of the component.
```
booc0mtaco, line 66:
```suggestion

```

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,116 @@
...
+        }
+        type="range"
+        value={value}
+        {...other}
+      />
+      {markers && (
+        <div className={styles['slider__markers']}>
+          {markers.map((marker) => (
+            <p
+              className={styles['slider__option']}
```
booc0mtaco, line 106: `slider__marker` ? Option sorta implies this is the actual value instead of the marker for a value

src/components/Slider/Slider.tsx
booc0mtaco: This is a little too open-ended at the moment. I think the intent is that the markers will be determined by the `min`/`max`/`step` values, and the number of markers is mathematically determined by those three props when the display some arbitrary text (`max - min / step`).

One way to account for this is to add checks to make sure markers is properly constrained. That would be useful whether they are simple strings or other symbols, or `ReactNode[]`. spec tests to verify it works too.

Also, if the common case is that markers are always these numeric sets, we can add a marker utility function to generate the steps. For example, `markers` is either `string[]` or `'numeric'` (a special documented keyword), where we check for `numeric` and generate the array values for consumers.

booc0mtaco (commented):
We should also test what happens when the number of markers doesn't align with the `min`/`max`/`step` values. Unlike using `<datalist>` markers will only allow evenly-spaced marks

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,116 @@
+import clsx from 'clsx';
+import React, {
+  useId,
+  type ChangeEventHandler,
+  type CSSProperties,
+} from 'react';
+import styles from './Slider.module.css';
+import Label from '../Label';
+
+export type Props = {
```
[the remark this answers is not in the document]
jinlee93: That's correct, will add union!

src/components/Slider/Slider.tsx
jinlee93: Will change to `<Text>` unless you were thinking of a more appropriate semantic element?

src/components/Slider/Slider.tsx
jinlee93: Removed!

src/components/Slider/Slider.tsx
jinlee93: Was hesitant between `slider__marker` and `slider__option` since "marker" looks so close to containing "markers" but will change to `slider__marker` since it's more intuitive

jinlee93:
> We should also test what happens when the number of markers doesn't align with the min/max/step values. Unlike using <datalist> markers will only allow evenly-spaced marks

the intention right now is to only have [evenly-spaced marks](https://www.figma.com/file/fqvLTr6OPuhr7G97Bd0nhF/Sliding-Selection?node-id=9182%3A27437&t=5uSeYHVMdtOSVJbi-4)

> This is a little too open-ended at the moment. I think the intent is that the markers will be determined by the min/max/step values, and the number of markers is mathematically determined by those three props when the display some arbitrary text (max - min / step).
>
> One way to account for this is to add checks to make sure markers is properly constrained. That would be useful whether they are simple strings or other symbols, or ReactNode[]. spec tests to verify it works too.
>
> Also, if the common case is that markers are always these numeric sets, we can add a marker utility function to generate the steps. For example, markers is either string[] or 'numeric' (a special documented keyword), where we check for numeric and generate the array values for consumers.

I left it open like this since steps can be smaller than the differences between markers. E.g. a 0-100 control with steps of 1 but using markers to hint at current value

src/components/Slider/Slider.tsx
jeremiah-clothier: > This is a little too open-ended at the moment. I think the intent is that the markers will be determined by the `min`/`max`/`step` values, and the number of markers is mathematically determined by those three props when the display some arbitrary text (`max - min / step`).

I like this idea

Another idea is that we could update markers to be customizable `markers?: Array<{ label?: string, value: number }>` instead of `markers: Array<string>`, but then placing the actual markers may be get complicated

jinlee93: Yea placing the markers according to the value of the marker (like how `<option>` is done) does complicate things and make the prop a little more complex than I want. However, if the need arises for it in a product, will definitely be considered.

Andrew and I did pair on the generation of the markers which is now in this pr

src/components/Slider/Slider.stories.tsx
```diff
@@ -0,0 +1,136 @@
...
+      value={sliderValue}
+    />
+  );
+};
+
+export const Default: StoryObj<Args> = {
+  args: {
+    label: 'Slider Label',
+  },
+  render: (args) => <InteractiveSlider {...args} />,
```
jeremiah-clothier, line 50: nit: it seems like the default export could specify `render` and remove the `render` prop from each of the individual stories

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+const lowestTenMultiplier = (numbers: number[]) => {
+  let multiplier = 1;
+  while (
+    numbers.some((number) => !Number.isInteger((number * multiplier) % 1))
+  ) {
+    multiplier *= 10;
+  }
+  return multiplier;
+};
```
jeremiah-clothier, line 19: Some docstring with an `@example` could be helpful for future developers

jeremiah-clothier (approved):
LGTM

src/components/Slider/Slider.stories.tsx
jinlee93: Ahh nice, didn't think to put the `render` fn in the default export, done

src/components/Slider/Slider.tsx
jinlee93: Added docstring, thanks!

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,182 @@
...
+ */
+const lowestTenMultiplier = (numbers: number[]): number => {
+  let multiplier = 1;
+  while (
+    numbers.some((number) => !Number.isInteger((number * multiplier) % 1))
+  ) {
+    multiplier *= 10;
+  }
+  return multiplier;
+};
```
booc0mtaco, line 28: nit: we can also save this to src/utils and write separate unit tests for it, since this is a generic piece of functionality.  The tests for this would serve as a good set of examples too.

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+  const generatedId = useId();
+  const sliderId = id || generatedId;
+
+  return (
+    <div className={componentClassName}>
+      {label && (
+        <Label className={labelClassName} htmlFor={sliderId} text={label} />
+      )}
+      <input
```
booc0mtaco, line 133: a11y: we might be able to add a meaningful `aria-describedby` which formats a string describing the slider. This could be a fieldNode (which other inputs use), perhaps hidden by default or something. the `fieldNote` would override this when specified.

src/components/Slider/Slider.tsx
jinlee93: Moved to `src/util/` and wrote some tests 👍

src/components/Slider/Slider.tsx
jinlee93: Added a `fieldNote` prop that takes the markers space if present. I don't think the markers itself should describe the slider as the native markers don't:
Screenshot 2023-02-27 at 3 53 30 PM

src/util/findLowestTenMultiplier.test.ts
```diff
@@ -0,0 +1,29 @@
+import { findLowestTenMultiplier } from './findLowestTenMultiplier';
+
+describe('findLowestTenMultiplier', () => {
+  describe('error throws', () => {
```
booc0mtaco, line 4: ++

booc0mtaco (approved):

PR #1506, opened 2023-02-23
fix(Select): sync label design with form fields

booc0mtaco:
### Summary:

- remove little used helper mixin, and apply typography to labels
- add overline container to match label structure for textarea and input fields
- update snapshots
- EDS-841

### Test Plan:

### DIFF
--- a/src/components/Label/Label.module.css
+++ b/src/components/Label/Label.module.css
@@ -10,7 +10,8 @@
  * Labels can sometimes be marked up as legends for field groups (i.e. radio field).
  */
 .label {
-  @mixin labelStyles;
+  @mixin eds-theme-typography-form-label;
+  display: inline-block;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


booc0mtaco:
Screenshot 2023-02-23 at 13 07 31

jinlee93 (approved):
FWIW the input has an overline container since there are two elements in it spaced apart (the label and optional/required)

booc0mtaco:
> FWIW the input has an overline container since there are two elements in it spaced apart (the label and optional/required)

Makes sense. I feel like `Select` should also be able to be marked as required? so this will set us up to implement that after chatting with design

jinlee93:
> > FWIW the input has an overline container since there are two elements in it spaced apart (the label and optional/required)
> 
> Makes sense. I feel like `Select` should also be able to be marked as required? so this will set us up to implement that after chatting with design

Sounds good 👍

PR #1507, opened 2023-02-23
test(Select): add story for selected value

booc0mtaco:
### Summary:

- add a new story for showing a value selected on component load
- update typing in story file for clarity

This is to ensure the behavior of selected values when using the component, as this was not previously covered in our wrapper

### Test Plan:


jinlee93 (approved):
Sweet, thanks for the story

PR #1508, opened 2023-02-24
refactor: lower clickable style specificity by importing css module last

jinlee93:
### Summary:
- found out that css modules get loaded on demand, meaning if a component specific style sheet is imported before a subcomponent, the subcomponent style sheet will be imported into the html farther down than the component style sheet, causing specificity issues
  - we generally want the component styles to have priority over the styles of the consumed component
- added lint rule to ensure css module style sheet is imported after the components
- removed doubling of AccordionButton, Button, Menu Button class selectors in their css files
- added `:where` pseudoselectors to `ClickableStyle.module.css` to keep specificity to (0,1,0) so they can be overriden via class styling
- ran `yarn lint --fix` to push css module imports to the end of imports
### Test Plan:

  - new lint rule 
- [x] CI tests pass
  - no visual regression, despite some styles specificity updates

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -39,7 +39,7 @@
         "groups": ["builtin", "external", "sibling", "parent", "index"],
         "pathGroups": [
           {
-            "pattern": "*.css",
+            "pattern": "*.module.css",
             "group": "index",
             "patternOptions": { "matchBase": true },
             "position": "after"
--- a/.eslintrc
+++ b/.eslintrc
@@ -36,7 +36,15 @@
         "alphabetize": {
           "order": "asc"
         },
-        "groups": ["builtin", "external", "sibling", "parent", "index"]
+        "groups": ["builtin", "external", "sibling", "parent", "index"],
+        "pathGroups": [
+          {
+            "pattern": "*.css",
+            "group": "index",
+            "patternOptions": { "matchBase": true },
+            "position": "after"
+          }
+        ]
       }
     ],
     "@typescript-eslint/no-unused-vars": "warn",
--- a/src/components/AccordionButton/AccordionButton.module.css
+++ b/src/components/AccordionButton/AccordionButton.module.css
@@ -7,7 +7,7 @@
 /**
  * Accordion Button, wraps the heading and open indicator icon.
  */
-.accordion-button.accordion-button {
+.accordion-button {
   display: flex;
   justify-content: space-between;
 
@@ -21,7 +21,7 @@
 /**
  * Small variant.
  */
-.accordion-button--sm.accordion-button--sm {
+.accordion-button--sm {
   padding: var(--eds-size-1);
   height: 2.25rem;
 }
@@ -29,7 +29,7 @@
 /**
  * Outline variant.
  */
-.accordion-button--outline.accordion-button--outline {
+.accordion-button--outline {
   padding: var(--eds-size-2) var(--eds-size-3);
 }
 
@@ -43,7 +43,7 @@
 /**
  * The heading text.
  */
-.accordion-button__heading.accordion-button__heading {
+.accordion-button__heading {
   @mixin eds-theme-typography-body-text-sm;
   color: var(--eds-theme-color-text-neutral-strong);
 }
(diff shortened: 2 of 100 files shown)


.eslintrc
```diff
@@ -36,7 +36,15 @@
...
           "order": "asc"
         },
-        "groups": ["builtin", "external", "sibling", "parent", "index"]
+        "groups": ["builtin", "external", "sibling", "parent", "index"],
+        "pathGroups": [
+          {
+            "pattern": "*.module.css",
+            "group": "index",
+            "patternOptions": { "matchBase": true },
+            "position": "after"
```
jinlee93, line 45: Looks for `module.css` and places after the `index` file, which is placed after imports from parents (where subcomponents live)

src/components/AccordionButton/AccordionButton.module.css
```diff
@@ -7,7 +7,7 @@
 /**
  * Accordion Button, wraps the heading and open indicator icon.
  */
-.accordion-button.accordion-button {
+.accordion-button {
```
jinlee93, line 10: removes doubling selectors in this file

src/components/Button/Button.module.css
```diff
@@ -14,10 +14,8 @@
 
 /**
  * Primary disabled button styles
- *
- * Class is doubled to increase specificity.
  */
-.button--primary.button--primary {
+.button--primary {
```
jinlee93, line 18: removes doubling selectors in this file

src/components/ClickableStyle/ClickableStyle.module.css
```diff
@@ -24,7 +24,7 @@
   transition-duration: var(--eds-anim-fade-quick);
   transition-timing-function: var(--eds-anim-ease);
 
-  svg {
+  :where(svg) {
```
jinlee93, line 27: I added `:where` pseudo selectors to nested `svg` selectors to reduce specificity to (0, 1, 0) to be easily overridden by class selectors

src/components/ClickableStyle/ClickableStyle.module.css
```diff
@@ -61,7 +61,7 @@
 /**
  * Primary brand clickable style
  */
-.clickable-style--primary.clickable-style--brand {
+:where(.clickable-style--primary).clickable-style--brand {
```
jinlee93, line 64: Reduced specificity to these double class selectors to (0,1,0) as well

src/components/Menu/Menu.module.css
```diff
@@ -19,7 +19,7 @@
   width: 100%;
 }
 
-.menu__button.menu__button {
+.menu__button {
```
jinlee93, line 22: removed doubled class selectors

jinlee93:
> ## size-limit report 📦
> Path	Size
> components	118.07 KB (-0.02% 🔽)
> styles	3.1 KB (0%)

?? possibly with how bundling happens

booc0mtaco:
This is a great catch. I didn't realize we weren't consistent in this, but this follows the same order one would use if you managed the CSS (or any dependencies really) in the "classical" way:

- for any component, import its dependencies first
- include the code that may override or augment the functionality of the dependencies
- implement the component, using the dependencies, and any overrides

booc0mtaco:
@jinlee93 would this be a good one to test as an alpha on GST? Logically, this will Just Work (tm)

jinlee93:
> @jinlee93 would this be a good one to test as an alpha on GST? Logically, this will Just Work (tm)

Yea, TW classes have a lower specificity in GST than EDS due to it being imported way earlier in the main app file and EDS components being imported ad hoc, so I imagine there wouldn't be any breaks where TW suddenly has more specificity over EDS components, but will alpha publish and test for sanity 👍

timzchang (approved):
Wow this is a really interesting learning! The change looks reasonable to me 👍🏼

jinlee93:
@booc0mtaco GST [CI tests pass](https://github.com/chanzuckerberg/edu-ps-goal-setting-test/actions/runs/4265052962/) and Chromatic only has the [InputField label changes](https://www.chromatic.com/build?appId=63c98be979fe9ae756d85e0e&number=287)

booc0mtaco (approved):
letsgo!

PR #1509, opened 2023-02-24
Update some deps

ahuth:
This PR updates some deps.


.storybook/pages/AlongDemo/AlongDemo.stories.ts
```diff
@@ -19,4 +19,11 @@ export const WatchPage: StoryObj<Args> = {
     const nextPageButton = await canvas.findByText('Hogwarts');
     userEvent.click(nextPageButton);
   },
+  parameters: {
+    axe: {
+      // Don't run axe on this story. The video seems to cause asset preloading to fail (even when
+      // the video has preload="none").
```
ahuth, line 25: Fails at https://github.com/dequelabs/axe-core/blob/d4522cdd7a90018336098f9b2119e353bd5a71ee/lib/core/utils/preload.js#L105

package.json
```diff
@@ -127,22 +127,22 @@
     "@types/react-beautiful-dnd": "^13.1.3",
     "@types/react-dom": "^18.0.11",
     "@types/react-portal": "^4.0.4",
-    "axe-core": "4.4.3",
-    "chromatic": "^6.17.0",
+    "axe-core": "4.6.3",
```
jinlee93, line 130: `^4.6.3`? 
Iirc we kept it at 4.4.3 due to timeouts but it looks like we're just skipping that. Can we remove `axe-core` from the ignore in the [renovate config](https://github.com/chanzuckerberg/edu-design-system/blob/main/.github/renovate.json5#L24-L25)?

ahuth: 👍

jinlee93 (approved):
Sweet thanks!

package.json
jinlee93: Can we add the caret back (shoulda been more clear lol sorry)

PR #1510, opened 2023-02-25
feat(Avatar): add avatar component

booc0mtaco:
### Summary:

- implement core component
- add in variants, sizes, and shapes
- add tests and stories

### Test Plan:


booc0mtaco:
Pending:

- a11y fixes
- `play` to show focus state
- discuss with design on the different color variants (to account for dynamic treatment, boring avatars, etc.)

src/components/Avatar/Avatar.module.css
```diff
@@ -0,0 +1,49 @@
...
+
+.avatar:focus {
+  outline: var(--eds-border-width-sm) solid var(--eds-theme-color-focus-ring);
+  outline-offset: 2px;
+}
+
+.avatar--sm {
+  @mixin eds-theme-typography-label-md;
+
+  height: 32px;
```
jinlee93, line 29: nit: should these be size tokens or just rems?

src/components/Avatar/Avatar.stories.ts
```diff
@@ -0,0 +1,57 @@
...
+export const Large: StoryObj<Args> = {
+  args: {
+    size: 'lg',
+  },
+};
+
+export const UsingImage: StoryObj<Args> = {
+  args: {
+    variant: 'image',
+    src: 'http://placekitten.com/200/200',
```
jinlee93, line 48: 🐈👍

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+export interface Props {
+  /**
+   * CSS class names that can be appended to the component.
+   */
+  className?: string;
+  /**
+   * The user name associated with this avatar
+   */
+  name?: string;
```
jinlee93, line 14: Maybe specify we want in "Firstname Lastname"?

src/components/Avatar/Avatar.stories.ts
booc0mtaco: I'll have to remove this probably :( the image it returns might not be stable, and we'd be hitting that site a lot. What I may do is include some generic inline svg or something

src/components/Avatar/Avatar.tsx
booc0mtaco: @jinlee93 I'll bring this up in discussion, but `name` could be anything. Right now it assumes fname lname with some parsing to exclude middle names, but I think the ideal state is either some object representing user data (name, login, email, etc.) or renaming this such that it does not imply their "name" name. Will figure all that out today :)

src/components/Avatar/Avatar.module.css
booc0mtaco: Correct, i will test a few things with the sizing. We may have the proper values in tokens already, and their usages here are tier 3

src/components/Avatar/Avatar.tsx
jinlee93: That does sound more robust, thanks!

src/components/Avatar/Avatar.module.css
booc0mtaco: @jinlee93 the sizes here correspond to `eds-size-4` thru `eds-size-6` for each size

src/components/Avatar/Avatar.module.css
```diff
@@ -0,0 +1,48 @@
...
+  align-items: center;
+  justify-content: center;
+
+  overflow: hidden;
+
+  color: var(--eds-theme-color-text-neutral-strong);
+  background-color: var(--eds-theme-color-background-neutral-medium);
+}
+
+.avatar:focus {
```
jinlee93, line 21: Should this be [focus-visible](https://github.com/chanzuckerberg/edu-design-system/blob/main/docs/CODE_GUIDELINES.md?plain=1#L108-L126)?

src/components/Avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+    ),
+  ],
+  play: async ({ canvasElement }) => {
+    if (isChromatic()) {
+      const canvas = within(canvasElement);
+      const avatar = await canvas.findByRole('button');
+      userEvent.click(avatar);
+    }
+  },
+};
```
jinlee93, line 92: Is there value in adding 'square' snaps at least for visual regression?

src/components/Avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+      <div className="p-3">
+        <Story />
+      </div>
+    ),
+  ],
+  play: async ({ canvasElement }) => {
+    if (isChromatic()) {
+      const canvas = within(canvasElement);
+      const avatar = await canvas.findByRole('button');
+      userEvent.click(avatar);
```
jinlee93, line 89: Just commenting: interesting that the focus state is captured in Chromatic but not when I load the story (might be my own machine?)

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,123 @@
...
+   */
+  variant?: 'icon' | 'initials' | 'image';
+}
+
+function getInitials(fromName: string): string {
+  /**
+   * Scenarios:
+   * - User's name is spelled as first and last name: John Smith
+   * - User's name has a middle name or initial: John C. Smith
+   * - User's Name has dashes in it
```
jinlee93, line 60: Names are so complicated lol

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const componentClassName = clsx(
+    styles['avatar'],
+    shape === 'circle' && styles['avatar--circle'],
+    shape === 'square' && styles['avatar--square'],
+    size === 'sm' && styles['avatar--sm'],
+    size === 'md' && styles['avatar--md'],
+    size === 'lg' && styles['avatar--lg'],
+    variant === 'icon' && styles['avatar--icon'],
+    variant === 'initials' && styles['avatar--initials'],
+    variant === 'image' && styles['avatar--image'],
```
jinlee93, line 100: Alternatively ```size && styles[`avatar--${size}`]``` and for variant. Not sure what is more readable tho lol

booc0mtaco: oh i like this. unwritten lines of code never have bugs 🐛

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * BETA: This component is still a work in progress and is subject to change.
+ *
+ * `import {Avatar} from "@chanzuckerberg/eds";`
+ *
+ * Representation of a single, unique user, keyed by the user name
+ */
+export const Avatar = ({
+  ariaLabel,
+  as: TagName = 'div' as const,
```
jeremiah-clothier, line 82: Surprised `as const` is needed when the `as` prop already has the type of `'div' | 'button'`

src/components/Avatar/Avatar.stories.tsx
ahuth: May be due to `focus-visible`? Wouldn't expect to see the focus outline on click in browsers that support it.

Could try changing this to `tab` and seeing if it shows up locally.

src/components/Avatar/Avatar.tsx
ahuth: Probably TS thinks the default value is a general `string`.

src/components/Avatar/Avatar.stories.tsx
jinlee93: I think at the time of this comment, it was before it was `focus` and not `focus-visible`

src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,118 @@
...
+    variant && styles[`avatar--${variant}`],
+    className,
+  );
+
+  const descriptiveLabel =
+    ariaLabel ??
+    `Avatar for ${!user && 'unknown'} user ${user?.fullName || ''}`;
+
+  // when not a button, add some props to allow focus states
+  const a11yProps = TagName === 'button' ? {} : { role: 'button', tabIndex: 0 };
```
ahuth, line 104: What's the scenario for using a div vs button?

Wondering if this should either be:
- Always a div, since I don't see an `onClick` handler or other interaction?
- Always a button, since we're giving it a button role anyway?

src/components/Avatar/Avatar.tsx
booc0mtaco: This doesn't appear to be needed after all. This pattern is used in a few other components implementing `as` here so I based this off of those 🤔 I wonder why those ones use `as const`

src/components/Avatar/Avatar.tsx
booc0mtaco: There are these two cases:

- an avatar can appear as an inert component in things like cards, or profiles (where there is no interaction)
- as an actual button (imagine a profile menu, where clicking the avatar shows a `Menu`'s content below)

Now that I think about it, in the former case it wouldn't need such a role or tab index b/c there's nothing to do ... its essentially an image then. This was initially to ensure focus state was working but should be removed now

booc0mtaco: @ahuth it may be true that it should never be a button, and cases where it is clickable would be wrapped by a button? maybe I should just take this part out for now, and keep it simple

jinlee93: I personally think avatar should beinert, and if interactive functionality is needed, it will be wrapped by an appropriate element.
However, this does complicate focus states, but using <Button> will assume to better fit target size requirements

src/components/Avatar/Avatar.tsx
ahuth: Maybe just copy pasta

src/components/Avatar/Avatar.tsx
ahuth: Not sure of the right answer, but maybe it's best to start with the simpler solution. Can add in a button later if we think it's better.

booc0mtaco: @jinlee93 @ahuth Agreed. I thought about it some more , and i think what's happening here is actually more like "an avatar can be wrapped by an 'invisible' button" where the button itself has the focus ring instead of the avatar... I'm gonna take this out and let the avatar be more image-like.

There are actually 1-2 other things behaving like this in EDS already, and I've seen this variant in other design systems i've worked on, so i will try some things out in a follow-up

PR #1511, opened 2023-02-25
Checkbox

ahuth:
EFI-948

This new feature allows Checkboxes to be un-controlled, so that they work more smoothly with Remix.

### Explain yourself

Instead of needing to explicitly manage a checkbox's checked state:

```ts
// In a Remix app.

const [checked, setChecked] = useState(false) // <- Why do this busy work?

<Checkbox
  checked={checked}
  onChange={setChecked(!checked)}
/>
```

We should let the browser manage that boring stuff for us.

```ts
// In a Remix app.

<Checkbox /> // <- Just let the checkbox do checkbox things
```

Un-controlled components fit in nicely with Remix's philosophy of progressive enhancement and using standard HTML stuff.

### Consequences

There's a major consequence of allowing the component to be uncontrolled, though...

The current implementation renders an `<svg>` element for the custom checkbox, relying on a prop to know which svg to render.

Basically

```tsx
// In Checkbox.tsx

if (checked) {
  <CheckedIcon />
} else {
  <NotCheckedIcon />
}
```

By making the component un-controlled, it won't know which icon to render 😑 

### Options

I considered several options for dealing with this conundrum.

1. **_Manage state inside the EDS component_**, if necessary. Pretty sure this is what MUI does. If `checked` is not provided, we could maintain our own state internally for it.

2. **_Simplify the design and use `accent-color`_**. With [accent-color](https://developer.mozilla.org/en-US/docs/Web/CSS/accent-color) we can tell browsers to use EDS colors for the checkbox. But browsers still control the design. This option is by far the simplest, but I don't really feel like trying to convince everyone to just go with it.

3. **_Use pure CSS to create the custom design_**. There are plenty of examples of using only CSS to create custom checkboxes. For example, [this one from Stephanie Eckles](https://moderncss.dev/pure-css-custom-checkbox-style/) is pretty recent. It makes interesting use of `clip-path: polygon(...)`.

### What I went with

The 3rd option. It's not that complex, and we can actually use `clip-path: path(...)` to re-use the SVG path data we were already using.

Downsides:
- There are some magic values I haven't figured out how to get rid of (more inline).
- We'll have to update our browser support a bit (which won't be a problem).

Positives:
- Simpler, lighter weight
- All the uncontrolled goodness we wanted


src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -24,9 +23,8 @@ export default {
   decorators: [
     (Story) => (
       <div
-        style={{
-          margin: '0.25rem', // Provides spacing to see focus indicator around checkbox.
-        }}
+        // Provides spacing to see focus indicator around checkbox.
+        className="m-1"
```
ahuth, line 27: Added Tailwind classes in the stories to replace css modules and inline styles. That way we can delete an entire css file.

src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -36,47 +34,55 @@ export default {
 
 type Args = React.ComponentProps<typeof Checkbox>;
 
-/**
- * Controlled example to make checked stories interactive.
- */
-function CheckboxExample(args: Args) {
-  const [checked, setChecked] = React.useState<
```
ahuth, line 43: Checkbox can be uncontrolled now, so we don't need CheckboxExample.

src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -36,47 +34,55 @@ export default {
...
+};
+
+export const Large: StoryObj<Args> = {
+  ...Default,
+  args: {
+    size: 'lg',
+  },
+};
+
+export const LargeChecked: StoryObj<Args> = {
```
ahuth, line 68: Added "Large" stories so I could see what they look like.

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -25,76 +24,69 @@ export type CheckboxInputProps = CheckboxHTMLElementProps & {
...
-};
-
 /**
  * Checkbox input element, exported for greater flexibility.
  * You must provide an `id` prop and connect it to a visible label.
  */
 export const CheckboxInput = React.forwardRef<
   HTMLInputElement,
   CheckboxInputProps
->(({ checked = false, className, disabled, ...other }, ref) => {
```
ahuth, line 66: Removing the default prop value lets the component be uncontrolled (because `value` is `undefined`).

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -25,76 +24,69 @@ export type CheckboxInputProps = CheckboxHTMLElementProps & {
...
+>(({ checked, className, disabled, ...other }, ref) => {
   // Make indeterminate checkbox visually match the colors of a
   // checked state, but announce itself as "mixed" to screen readers
   const checkedProps =
     checked === 'indeterminate'
       ? {
           'aria-checked': 'mixed' as const,
-          checked: true,
+          // Indeterminate checkboxes are unchecked when submitted to the server.
+          checked: false,
```
ahuth, line 42: Fixed a logical bug here.

We were counting "indeterminate" checkboxes as "checked", but that's not right. They're supposed to be functionally the same as unchecked.

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -25,76 +24,69 @@ export type CheckboxInputProps = CheckboxHTMLElementProps & {
...
+    />
   );
 });
 
 CheckboxInput.displayName = 'CheckboxInput';
+
+/**
+ * Use a forwarded ref if present, or create one if not.
+ */
+function useForwardedRef<T>(ref: React.ForwardedRef<T>): React.RefObject<T> {
```
ahuth, line 77: TODO: extract this hook to another file, if we end up going with this.

Idea is to use a forwarded ref, which may or may not be provided.

An alternative I've seen (which is basically the same as this, just different semantics) is to come up with a scheme to combine refs:

```ts
const myRef = useRef();
const combinedRef = useCombinedRefs(forwardedRef, myRef);
```

src/components/CheckboxInput/CheckboxInput.tsx
ahuth: TODO: add a test for this behavior.

src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -36,47 +34,55 @@ export default {
...
-export const Default: StoryObj<Args> = {
-  render: (args) => <CheckboxExample {...args} />,
-};
+export const Default: StoryObj<Args> = {};
 
 export const Checked: StoryObj<Args> = {
-  render: (args) => <CheckboxExample {...args} checked />,
+  ...Default,
+  args: {
+    defaultChecked: true,
```
ahuth, line 42: `defaultChecked` makes it start off checked, but is still uncontrolled/interactive.

src/components/Checkbox/Checkbox.stories.tsx
ahuth: Actually I guess "large" is the default. So maybe no need for this.

src/components/CheckboxInput/CheckboxInput.module.css
```diff
@@ -1,71 +1,64 @@
...
 }
-/**
- * The visually hidden input element for the checkbox. The visual checkmark is provided by an svg element.
- */
+
 .checkbox__input {
-  cursor: pointer;
-  /* Removes default margins placed by browser for checkboxes. */
+  height: 18px;
+  width: 18px;
```
ahuth, line 14: TODO: figure out what to do with these hard coded values.

For some reason this implementation needs different height/width to get the same visual size as the previous svg implementation.

src/components/CheckboxInput/CheckboxInput.module.css
```diff
@@ -1,71 +1,64 @@
...
+  place-content: center;
+}
+
+.checkbox__input:checked::before {
+  background-color: currentColor;
+  content: '';
+  height: 24px;
+  width: 24px;
+
+  clip-path: path("M10.6 16.2L17.65 9.15L16.25 7.75L10.6 13.4L7.75 10.55L6.35 11.95L10.6 16.2ZM3 21V3H21V21H3Z");
```
ahuth, line 34: https://developer.mozilla.org/en-US/docs/Web/CSS/clip-path

Will need to slightly update our browser support

src/components/CheckboxInput/CheckboxInput.module.css
```diff
@@ -1,71 +1,64 @@
...
+  /* Place the ::before content right in the middle of the input. */
+  display: grid;
+  place-content: center;
+}
+
+.checkbox__input:checked::before {
+  background-color: currentColor;
+  content: '';
+  height: 24px;
+  width: 24px;
```
ahuth, line 32: I think the height/width basically need to match the svg's viewbox.

src/components/CheckboxInput/CheckboxInput.module.css
```diff
@@ -1,71 +1,67 @@
...
+  width: 18px;
+
+  /* The parent Checkbox component is a flex container. Make sure the input doesn't shrink. */
+  flex-shrink: 0;
+
+  /* Remove the browser's checkbox styles, allowing us to provide our own. */
+  appearance: none;
+
+  /* Magic value to center the checkbox on its label. */
+  margin: 3px;
```
ahuth, line 14: 😐

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -25,76 +24,69 @@ export type CheckboxInputProps = CheckboxHTMLElementProps & {
...
   HTMLInputElement,
   CheckboxInputProps
->(({ checked = false, className, disabled, ...other }, ref) => {
+>(({ checked, className, disabled, ...other }, ref) => {
   // Make indeterminate checkbox visually match the colors of a
   // checked state, but announce itself as "mixed" to screen readers
   const checkedProps =
     checked === 'indeterminate'
       ? {
           'aria-checked': 'mixed' as const,
```
ahuth, line 40: TODO: verify if we need `'aria-checked': 'mixed'` or not.

src/components/CheckboxInput/CheckboxInput.module.css
ahuth: Maybe the magic numbers are fine for now? 😓 

There already some there (the height/width on the svg element), so at the very least I didn't make things worse.

src/components/CheckboxInput/CheckboxInput.module.css
ahuth: Also, kinda tempted to put all these styles behind an `@supports (clip-path: path)` and gracefully fall back to `accent-color`.

booc0mtaco:
We should hold off on merging this one in so we can publish the new components, and avoid any refactoring work for consumers over the next sprint at least. I'll take a deeper look once I finish up work on `Avatar` (the component, not the movie :) )

ahuth:
> avoid any refactoring work for consumers over the next sprint at least.

This change doesn't require refactoring, @booc0mtaco. Anyone using a Checkbox in a controlled manner can keep on doing that.

Also, waiting 2 weeks is... quite a long time.

booc0mtaco:
> > avoid any refactoring work for consumers over the next sprint at least.
> 
> This change doesn't really require refactoring by consumers, @booc0mtaco. Anyone using a Checkbox in a controlled manner can keep on doing that.
> 
> Also, waiting 2 weeks is... quite a long time.

Where is the urgency or request to make this breaking change?

ahuth:
On further review, this isn't a breaking change. Just a new feature.

Any issue shipping a non-breaking change?

src/components/CheckboxInput/CheckboxInput.module.css
```diff
@@ -1,71 +1,64 @@
...
+  content: '';
+
+  /* Clip this element in the shape of a checkbox. The element's background color will be visible
+  anywhere the svg path is solid. Whatever is behind the checkbox will show through wherever the
+  svg is transparent. See https://developer.mozilla.org/en-US/docs/Web/CSS/clip-path */
+  clip-path: path("M10.6 16.2L17.65 9.15L16.25 7.75L10.6 13.4L7.75 10.55L6.35 11.95L10.6 16.2ZM3 21V3H21V21H3Z");
+
+  /* The SVG path's viewbox. */
+  height: 24px;
+  width: 24px;
```
timzchang, line 35: are these `--eds-size-3`? going off your comment above

src/components/CheckboxInput/CheckboxInput.module.css
timzchang: Not a css expert or EDS pundit by a long stretch, so def ok deferring to other opinions. My gut says this is not ideal but fine. Imo this makes more sense to me than the arbitrary values that were here previously.

An alternative I think you mentioned is replacing `18px` and `3px` with eds constants. Obvi changes the sizing of things, but if EDS folks are ok with it then I'd prefer that I think

src/components/CheckboxInput/CheckboxInput.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  it('is not indeterminate when checked is true', () => {
+    render(<CheckboxInput checked id="test" readOnly />);
+    const checkbox = screen.getByRole<HTMLInputElement>('checkbox');
+    expect(checkbox.indeterminate).toEqual(false);
+  });
+
+  it('is not indeterminate when checked is uncontrolled', () => {
+    render(<CheckboxInput id="test" />);
+    const checkbox = screen.getByRole<HTMLInputElement>('checkbox');
```
timzchang, line 26: check for understanding: is a component controlled if props are passed in? I thought it was only controlled if state is managed (this excludes props in my understanding)

src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -102,18 +108,18 @@ export const Disabled: StoryObj<Args> = {
 
 export const WithoutVisibleLabel: StoryObj<Args> = {
   args: {
-    'aria-label': 'a checkbox has no name',
```
timzchang, line 105: a [Game of Thrones](https://www.gameofthronesquote.com/arya-stark/a-girl-has-no-name) reference I believe ⚔️ 👑

src/components/CheckboxInput/CheckboxInput.test.tsx
ahuth: > is a component controlled if props are passed in? I thought it was only controlled if state is managed

Kind of both.

It's controlled if we supply the `value`/`checked`. But doing so probably implies app developers managing the state (or at least someone managing it somewhere).

src/components/CheckboxInput/CheckboxInput.module.css
ahuth: Yes, I suppose it is.

I'll play around with that, think about whether this should be px to match the svg viewbox or not.

I'm not totally sure of the relationship between the height/width here and the svg, but it seems to need to be the viewbox. Which is interesting.

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -25,74 +25,39 @@ export type CheckboxInputProps = CheckboxHTMLElementProps & {
...
 export const CheckboxInput = React.forwardRef<
   HTMLInputElement,
   CheckboxInputProps
->(({ checked = false, className, disabled, ...other }, ref) => {
-  // Make indeterminate checkbox visually match the colors of a
-  // checked state, but announce itself as "mixed" to screen readers
-  const checkedProps =
-    checked === 'indeterminate'
-      ? {
-          'aria-checked': 'mixed' as const,
```
ahuth, line 72: Tested it, and `aria-checked` is not needed.

src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -36,47 +34,55 @@ export default {
...
+
+export const Large: StoryObj<Args> = {
+  ...Default,
+  args: {
+    size: 'lg',
+  },
+};
+
+export const LargeChecked: StoryObj<Args> = {
+  ...Large,
```
jinlee93, line 69: Not sure what spreading all these Stories are doing

src/components/CheckboxInput/CheckboxInput.module.css
jinlee93: If you really wanted, you could probably do something like 
```
.checkbox__input {
  box-sizing: content-box;
  height/width: calc(var(--eds-size-2) - var(--eds-border-width-md));
  ...
  border: var(--eds-border-width-md)) solid currentColor;
}
```

and create a tertiary token for the `3px` for checkbox use, but it gets unwieldy and I think having these values is fine for now?

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -25,74 +25,39 @@ export type CheckboxInputProps = CheckboxHTMLElementProps & {
...
-          checked: true,
-        }
-      : {
-          checked,
-        };
+>(({ checked, className, disabled, ...other }, ref) => {
+  const forwardedRef = useForwardedRef(ref);
+
+  // If the checkbox is "indeterminate", set that property on the node. This can only be done in JS
```
jinlee93, line 38: Strange, but also auto handles `aria-checked=mixed`

src/components/Fieldset/__snapshots__/Fieldset.test.ts.snap
```diff
@@ -148,29 +76,11 @@ exports[`<Fieldset /> Default story renders snapshot 1`] = `
...
-          style="--icon-size: 1.5rem;"
-          viewBox="0 0 24 24"
-          width="1.5rem"
-          xmlns="http://www.w3.org/2000/svg"
-        >
-          <path
-            d="M3 21V3H21V21H3ZM5 19H19V5H5V19Z"
-          />
-        </svg>
-      </span>
```
jinlee93, line 173: Prob more performant without these svgs

src/components/Checkbox/Checkbox.stories.tsx
ahuth: Clones the Large story.

Yeah there's not actually any properties being brought over currently. But makes it clear that this story is based on Large.

src/components/CheckboxInput/CheckboxInput.tsx
ahuth: Yep.

Also, see https://github.com/facebook/react/issues/1798 for discussion on why React won't create an `indeterminate` prop (spoiler: because there isn't a DOM attribute, and adding one wouldn't play nice with SSR I guess).

ahuth: Actually, I'm going to revert doing this the "right" way. Want to handle that in a separate commit/PR.

src/components/CheckboxInput/CheckboxInput.tsx
ahuth: Reverted this change, even though it's correct. I want to handle this in another commit/PR.

src/components/CheckboxInput/CheckboxInput.tsx
ahuth: Reverted this change, so I can handle it in another commit/PR.

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -25,47 +24,22 @@ export type CheckboxInputProps = CheckboxHTMLElementProps & {
...
+>(({ checked, className, disabled, ...other }, ref) => {
+  // Make indeterminate checkbox visually match the colors of a checked state, but announce itself
+  // as "mixed" to screen readers
+  //
+  // However, https://html.spec.whatwg.org/multipage/input.html#checkbox-state-(type=checkbox)
+  // seems to consider `indeterminate` as a separate bit of information than `checked`. They can be
+  // in all 4 combinations (only 3 visual states, though).
+  //
+  // Should we make them separate props?
   const checkedProps =
```
ahuth, line 43: Currently we consider indeterminate checkboxes as checked. That's not right, according to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes.

A prev iteration of this PR fixed that. I reverted it, though. Figured it makes more sense to revisit this in another effort/PR.

src/components/CheckboxInput/CheckboxInput.module.css
booc0mtaco: @ahuth is the clip-path the SVG for the checkmark itself?

src/components/CheckboxInput/CheckboxInput.module.css
```diff
@@ -1,71 +1,64 @@
...
-   */
-  opacity: 0;
+
+.checkbox__input:checked::before {
+  background-color: currentColor;
+  content: '';
+
+  /* Clip this element in the shape of a checkbox. The element's background color will be visible
+  anywhere the svg path is solid. Whatever is behind the checkbox will show through wherever the
+  svg is transparent. See https://developer.mozilla.org/en-US/docs/Web/CSS/clip-path */
```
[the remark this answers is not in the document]
booc0mtaco: Ahh, it is. gotcha (nvm my other comment here)

src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -36,47 +34,55 @@ export default {
...
+  args: {
+    ...Checked.args,
+    ...Large.args,
```
booc0mtaco, line 72: ++ this is more scannable than duplicating args all over the stories file

src/components/CheckboxInput/CheckboxInput.module.css
```diff
@@ -1,71 +1,64 @@
...
-/**
- * The visible checkbox svg icon element.
- */
-.checkbox__icon {
-  color: var(--eds-theme-color-icon-brand-primary);
+.checkbox__input[aria-checked="mixed"]::before {
+  background-color: currentColor;
+  content: '';
+
+  clip-path: path("M7 13H17V11H7V13ZM3 21V3H21V21H3Z");
```
booc0mtaco, line 43: What shape does this clip-path represent?

src/components/CheckboxLabel/CheckboxLabel.module.css
```diff
@@ -1,12 +1,5 @@
-/*------------------------------------*\
-    # CHECKBOX LABEL
-\*------------------------------------*/
-
-/**
- * Checkbox medium variant.
- */
```
booc0mtaco, line 7: I'm guessing you're in favor of these header blobs being stripped out wholesale :) they are in the ploppable templates, and don't really add much, given how we currently use them. I suspect this came from the BM code?

src/components/CheckboxInput/CheckboxInput.module.css
booc0mtaco: I think it's worth following up with design on this. Is this the only thing that is 18px square? and if so, how does that measure fit in with our sizes for other things. 

Setting a variable for just this doesn't help much, esp if it will only be used on this one line 🤔

src/components/CheckboxInput/CheckboxInput.module.css
booc0mtaco: I'd also would have leaned toward using our token here, BUT the sizes for the checkbox are all pixels (18px, 3px, and this 24px) which:

- should agree with the viewports
- should agree with one another
- shouldn't be partially controlled by tokens

If some team overrode eds-size-3, it'd still be using the other values and the combination would likely be off

src/components/CheckboxInput/CheckboxInput.module.css
```diff
@@ -1,71 +1,64 @@
...
+
+  height: 24px;
+  width: 24px;
 }
-.checkbox__input:focus-visible + .checkbox__icon {
+
+.checkbox__input:focus-visible {
   @mixin focus;
-  outline-offset: -2px;
+  outline-offset: 1px;
```
booc0mtaco, line 51: it appears our normal offset is 2px (set in the mixin). Curious why this one is one less, but gets us the look/feel we want. Is there some other thing giving us 1 pixel of space?

booc0mtaco (commented):
Looks much tidier than before. There was one snapshot i was worried about, and will comment there as a check

src/components/CheckboxInput/CheckboxInput.module.css
ahuth: That's right 👍

src/components/CheckboxInput/CheckboxInput.module.css
ahuth: Agreed. This 3px is specific to the SVG. It was always there, but with this implementation is a bit more obvious. Definitely shouldn't make this a var.

src/components/CheckboxInput/CheckboxInput.module.css
ahuth: Indeterminate checkbox. Sort of looks like a `-`.

src/components/CheckboxLabel/CheckboxLabel.module.css
ahuth: Yeah, they came from BM.

Deleted them because they don't add anything to these examples. But maybe I should add them back and worry about them later.

src/components/CheckboxInput/CheckboxInput.module.css
ahuth: 1px makes the new implementation exactly match the old one (which had -2px for some reason).

My guess for why these magic values are needed is:
- In both new/old cases, the input is 24px wide, but visually it's 18px.
- So if we add a focus outline that's 26px (24px + 2px offset) it appears too large compared to the 18px input box thing.

src/components/CheckboxInput/CheckboxInput.module.css
ahuth: Good point. These values need to align with the checkbox path stuff. Wouldn't really expect folks to override them.

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -25,47 +24,22 @@ export type CheckboxInputProps = CheckboxHTMLElementProps & {
...
-  // checked state, but announce itself as "mixed" to screen readers
+>(({ checked, className, disabled, ...other }, ref) => {
+  // Make indeterminate checkbox visually match the colors of a checked state, but announce itself
+  // as "mixed" to screen readers
+  //
+  // However, https://html.spec.whatwg.org/multipage/input.html#checkbox-state-(type=checkbox)
+  // seems to consider `indeterminate` as a separate bit of information than `checked`. They can be
+  // in all 4 combinations (only 3 visual states, though).
+  //
+  // Should we make them separate props?
```
anniehu4, line 42: hmmmm yeah probably 🤔

src/components/CheckboxInput/CheckboxInput.module.css
ahuth: Also these svg paths haven't changed. This is exactly what's always been here.

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -77,22 +51,14 @@ export const CheckboxInput = React.forwardRef<
...
-        disabled={disabled}
-        ref={ref}
-        type="checkbox"
-        {...checkedProps}
-        {...other}
-      />
-      <CheckboxSvg checked={checked} disabled={disabled} />
-    </span>
+    <input
+      className={clsx(className, styles['checkbox__input'])}
```
ahuth, line 55: Not related to the changes, but the class name should be `styles.input`.

The BEM-ing isn't doing anything for us, except making us type more.

It's like we took the `checkedProps` JS variable in this file and instead named it `checkbox__input__checkedProps`.

`</rant>`

src/components/CheckboxInput/CheckboxInput.tsx
ahuth: #1520

jinlee93 (approved):
I think this looks good

ahuth:
Thanks!

Let me know when the GST stuff is shipped, and it's cool to merge this thing 🙏

booc0mtaco:
@ahuth Yep, we were able to publish those yesterday, and fixed up the weirdness that came from GitHub having an outage mid-publish :P so should be good to go

src/components/CheckboxInput/CheckboxInput.module.css
anniehu4: since we control all those px values, I wonder (per [Jin's comment](https://github.com/chanzuckerberg/edu-design-system/pull/1511/files#r1119519886)) if it's worth adding some `calc` math to at least make one of the numbers less magic. not a big deal though, things are well commented

ahuth:
Thanks!

PR #1514, opened 2023-02-28
test(input): specify change event handler type

jinlee93:
### Summary:
- specifies change event handler type in Input for QOL

### Test Plan:

  - build EDS
  - copy / paste `Input` and `InputField` folders into node_modules of gst
  -  following doesn't have type error:
```
<InputField
      className="py-5"
      label={label}
      onChange={(e) => onChange(id, e.target.value)}
      value={selection}
/>
```

image

### DIFF
--- a/src/components/Input/Input.tsx
+++ b/src/components/Input/Input.tsx
@@ -51,7 +51,7 @@ export type InputProps = React.InputHTMLAttributes<HTMLInputElement> & {
   /**
    * Function that fires when field value has changed
    */
-  onChange?: ChangeEventHandler;
+  onChange?: ChangeEventHandler<HTMLInputElement>;
   /**
    * Placeholder attribute for input. Note: placeholder should be used sparingly
    */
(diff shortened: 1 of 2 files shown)


jeremiah-clothier (approved):
💜 thanks for such a quick turnaround!

I wonder if there are other event handler types like this where the typing can be improved

jinlee93:
> 💜 thanks for such a quick turnaround!
> 
> I wonder if there are other event handler types like this where the typing can be improved

At a quick glance, there are 5 other components that have onChange callbacks: Checkbox, Radio, Tabs, Select, and Toggle.

Checkbox and Radio are the closest `<input>` components and already have the appropriate typing:
image
image

Tabs is not an input component and doesn't control the state of the Tabs component and is just a reactive call back in case the parent component needs to do something on change 
image

[Select](https://headlessui.com/react/listbox#listbox) and [Toggle](https://headlessui.com/react/switch#switch) onchange is from Headless and just return their respective values and not the target

PR #1516, opened 2023-02-28
feat(menu): add Menu.PlainButton as a minimally styled Menu button

jinlee93:
[EDS-845]

### Summary:
- [GST card](https://www.figma.com/file/MjFKgcvuFFOGKXuuRg5b0P/Goals-Setting-Test---P0-%22Internal-Alpha%22?node-id=467%3A10708&t=1HHLLx80p5ilbtJL-4) requires enhancement of EDS Menu to allow custom button triggers
- Changes Menu.Button from ClickableStyle to Button component
- Introduces Menu.PlainButton, a minimally styled alternative for the opinionated Menu.Button
  - Adds a story with FPO block
  - Adds MenuWithIconButton recipe to show usage with an Icon

### Test Plan:

[EDS-845]: https://czi-tech.atlassian.net/browse/EDS-845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jinlee93:
> adds buttonText prop to the main <Menu> which styles the button as previous, and opens up <Menu.Button> to be closer to <Button> component for more flexibility

Alternatively, we can check if `typeof children === 'string'` which would be non breaking but makes it more difficult if want to have custom button with just string

src/components/Menu/Menu.tsx
```diff
@@ -14,6 +15,10 @@ import styles from './Menu.module.css';
 
 // Note: added className here to prevent private interface collision within HeadlessUI
 export type MenuProps = ExtractProps<typeof HeadlessMenu> & {
+  /**
+   * Text to be placed within the Button trigger to open the Menu.
+   */
+  buttonText?: React.ReactNode;
```
jinlee93, line 21: Not entirely sold on this prop name, could be buttonNode? buttonFill? suggestions?

src/components/Menu/Menu.tsx
```diff
@@ -49,33 +55,60 @@ export type MenuItemsProps = ExtractProps<typeof HeadlessMenu.Items>;
...
 /**
  * A button that when clicked, shows or hides the Options
+ * Refer to Button documentation for more style variants.
  */
-const MenuButton = ({ children, className, ...other }: MenuButtonProps) => {
-  const buttonClassNames = clsx(styles['menu__button'], className);
+const MenuButton = ({ status, ...other }: MenuButtonProps) => {
   return (
-    <HeadlessMenu.Button
-      as={ClickableStyle}
```
jinlee93, line 64: Originally was `<ClickableStyle>` but changed it to `<Button>` since the alternative (link) is not an appropriate trigger for a menu

src/components/Menu/Menu.tsx
```diff
@@ -49,33 +55,60 @@ export type MenuItemsProps = ExtractProps<typeof HeadlessMenu.Items>;
...
-        {children}
-        <Icon
-          className={styles['menu__button--icon']}
-          name="expand-more"
-          purpose="decorative"
-          size="1.25rem"
-        />
-      </>
-    </HeadlessMenu.Button>
+    <HeadlessMenu.Button as={Button} status={status || 'neutral'} {...other} />
```
jinlee93, line 111: Kept `as={Button}` for accessibility features (such as target size) and EDS variants

src/components/Menu/Menu.stories.tsx
```diff
@@ -20,9 +21,9 @@ export default {
 } as Meta<MenuProps>;
 
 export const Default: StoryObj<MenuProps> = {
-  render: (args) => (
```
jinlee93, line 23: Converted these to `children` in `args` since it's no longer mulitiple React Elements

src/components/Menu/Menu.tsx
```diff
@@ -49,33 +55,60 @@ export type MenuItemsProps = ExtractProps<typeof HeadlessMenu.Items>;
...
+                {buttonText}
+                <Icon
+                  className={styles['menu__button--icon']}
+                  name="expand-more"
+                  purpose="decorative"
+                  size="1.25rem"
+                />
+              </MenuButton>
+            )}
+            {children({ open, close })}
```
jinlee93, line 81: Would prefer to remove `close` here since `Menu` closes on click anyways but TS complains if `close` is not passed
image

src/components/Menu/Menu.stories.tsx
```diff
@@ -44,15 +45,72 @@ export const Default: StoryObj<MenuProps> = {
...
+          Headless UI Docs
+        </Menu.Item>
+        <Menu.Item
+          href="https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu"
+          icon="link"
+        >
+          MDN: Menu
+        </Menu.Item>
+        {/* eslint-disable-next-line no-alert */}
+        <Menu.Item onClick={() => alert('Item clicked')}>
```
ahuth, line 99: Future idea: we could experiment with using argTypes to generate Storybook "actions" for things like this: https://storybook.js.org/docs/react/essentials/actions#action-argtype-annotation.

Would tie into Storybook a little more. But wouldn't be as obvious as an alert. So not sure which is better, or if it really matters.

src/components/Menu/Menu.tsx
ahuth: `buttonContent`?

src/components/Menu/Menu.tsx
```diff
@@ -49,33 +55,60 @@ export type MenuItemsProps = ExtractProps<typeof HeadlessMenu.Items>;
...
+  ...other
+}: MenuProps) => {
   const menuClassNames = clsx(className, styles['menu']);
-  return <HeadlessMenu as="div" className={menuClassNames} {...other} />;
+  if (typeof children === 'function') {
+    return (
+      <HeadlessMenu as="div" className={menuClassNames} {...other}>
+        {({ open, close }) => (
+          <>
+            {buttonText && (
```
ahuth, line 70: Is `buttonText` mostly a more convenient way to provide a custom button?

src/components/Menu/Menu.tsx
ahuth: Dang, Headless's Menu.Button isn't necessarily a button? 😬

jinlee93: It is but I was thinking keeping it EDS themed hence the EDS Button

src/components/Menu/Menu.stories.tsx
jinlee93: That makes sense to me, can add to this pr as well

src/components/Menu/Menu.tsx
jinlee93: I agree, and I think I over complicated this lol.
How about: keeping the changes from ClickableStyle to Button, but also keeping the changes to accept Button props. I can add a conditional prop for the chevron icon. This way, the changes won't be breaking, but the Menu.Button will still have more customizability

src/components/Menu/Menu.tsx
```diff
@@ -39,7 +40,18 @@ export type MenuItemProps = ExtractProps<typeof HeadlessMenu.Item> & {
   onClick?: MouseEventHandler<HTMLAnchorElement>;
 };
 
-export type MenuButtonProps = ExtractProps<typeof HeadlessMenu.Button>;
+export type MenuButtonProps = Omit<
+  ExtractProps<typeof HeadlessMenu.Button>,
+  // we want Button children
```
anniehu4, line 45: nit
```suggestion
  // we want to use the 'children' type from ButtonProps
```

src/components/Menu/Menu.tsx
```diff
@@ -56,24 +68,33 @@ export const Menu = ({ className, ...other }: MenuProps) => {
...
+  ...other
+}: MenuButtonProps) => {
   const buttonClassNames = clsx(styles['menu__button'], className);
   return (
     <HeadlessMenu.Button
-      as={ClickableStyle}
+      as={Button}
       className={buttonClassNames}
-      status="neutral"
+      status={status || 'neutral'}
```
anniehu4, line 85: do you need to pass `status` for the GST use case? I'm not convinced we want to open the color up more, but lmk what other cases you had in mind

src/components/Menu/Menu.test.tsx
```diff
@@ -55,4 +56,25 @@ describe('<Menu />', () => {
 
     expect(screen.queryByTestId('menu-content')).not.toBeInTheDocument();
   });
+
+  it('should allow render prop usage', async () => {
```
anniehu4, line 60: oh neat, pleasantly surprised that this already works without the extra logic. good idea to add a test 👍

src/components/Menu/Menu.stories.tsx
```diff
@@ -20,90 +21,167 @@ export default {
...
+          <Menu.Item disabled href="https://example.org/" icon="warning">
+            Not Possible (disabled)
+          </Menu.Item>
+        </Menu.Items>
+      </>
+    ),
+  },
+};
+
+export const DotsVerticalButton: StoryObj<MenuProps> = {
```
anniehu4, line 118: nit
```suggestion
export const WithVerticalDotsButton: StoryObj<MenuProps> = {
```

src/components/Menu/Menu.stories.tsx
```diff
@@ -20,90 +21,167 @@ export default {
...
+          <Menu.Item disabled href="https://example.org/" icon="warning">
+            Not Possible (disabled)
+          </Menu.Item>
+        </Menu.Items>
+      </>
+    ),
+  },
+};
+
+export const CustomButton: StoryObj<MenuProps> = {
```
anniehu4, line 151: nit
```suggestion
export const WithCustomButton: StoryObj<MenuProps> = {
```

anniehu4 (commented):
nice, I really like this updated solution! clean and nicely leverages existing styling 👍  had 1 question about the `status` prop & wanted a look from @booc0mtaco but otherwise good to approve

src/components/Menu/Menu.tsx
anniehu4: I dig it

src/components/Menu/Menu.stories.tsx
booc0mtaco: @anniehu4 @jinlee93 we don't have a written convention for these but i have been using prefixes `With` if it demonstrates a composition, and `When` if the story exists to show some set of prop values.

Also, all `With` stories might actually be a signal that the story is actually a recipe in the format, "here's how you can combine ______ and _____ to get a _______"

src/components/Menu/Menu.test.tsx
booc0mtaco: this tests HeadlessUI behavior, but does ensure we didn't prevent it from working. Good Test

src/components/Menu/Menu.tsx
```diff
@@ -56,24 +68,33 @@ export const Menu = ({ className, ...other }: MenuProps) => {
...
         {children}
-        <Icon
-          className={styles['menu__button--icon']}
-          name="expand-more"
-          purpose="decorative"
-          size="1.25rem"
-        />
+        {showExpandIcon && (
+          <Icon
+            className={styles['menu__button--icon']}
```
booc0mtaco, line 92: nit: the modifier bit should be verb-like, but `icon` by itself acts like an element. Icon is also generic (this particular one is always the chevron). `with-chevron` comes to mind, but something more like that would tell more about the intent

src/components/Menu/Menu.tsx
booc0mtaco: I think this makes sense. The rationale here was that, since Headless gives you a `<button>`, this composed a styled button in the same way `Button` does internally. If we give it `Button` directly, looks like we get the same result. using `ClickableStyle` composes one less thing I guess

What things can be used as a menu trigger?

src/components/Menu/Menu.tsx
jinlee93: Agree that we don't need to open up the color more... same amount of specificity would be required for custom styling whether it's passed or not, so will just set to "neutral"

booc0mtaco (commented):
I did have an alternative suggestion for how we might compose custom buttons, mostly when this was a breaking change. This seems more straight-forward and extends default behavior!

src/components/Menu/Menu.tsx
jinlee93: So far we have text in the mocks, and the vertical dots icon in GST

src/components/Menu/Menu.tsx
jinlee93: No longer going with the `button{prefix}` prop, resolving

src/components/Menu/Menu.tsx
jinlee93: With alt approach, no longer a concern

src/components/Menu/Menu.tsx
jinlee93: I think `withChevron` sounds good

src/components/Menu/Menu.stories.tsx
jinlee93: WithVerticalDots does sound like a recipe, and could be redundant as the WithCustomButton story. Consider removing?

src/components/Menu/Menu.stories.tsx
jinlee93: Finding it's difficult to add argType for this since it's an argument for the subcomponent `Menu.Item` and the Args for these stories are for `Menu`

src/components/Menu/Menu.stories.tsx
booc0mtaco: @jinlee93 the custom ones using `fpo` stuff make sense per component, as they show where placeholders belong. We can add the recipe showing how to use the vertical dots as/with a custom button and keep both

jinlee93: Moved vertical dots story to its own recipe, since there already is a story using fpo block for button

anniehu4: makes sense to me! would be good to document that convention somewhere

src/components/Menu/Menu.tsx
```diff
@@ -56,24 +68,32 @@ export const Menu = ({ className, ...other }: MenuProps) => {
 
 /**
  * A button that when clicked, shows or hides the Options
+ * Refer to Button documentation for more style variants.
```
anniehu4, line 71: Could you also add documentation for the new `Menu.Button` functionality to the top-level `Menu` comment so it shows up readily in the Storybook Docs tab?

anniehu4 (dismissed):

src/components/Menu/Menu.stories.tsx
```diff
@@ -20,90 +20,129 @@ export default {
...
+  },
+};
+
+export const WithCustomButton: StoryObj<MenuProps> = {
+  args: {
+    children: (
+      <>
+        <Menu.Button showExpandIcon={false} status="neutral" variant="icon">
+          <div className="fpo !py-0">Menu Button</div>
+        </Menu.Button>
```
booc0mtaco, line 123: I do believe there is an abstraction that can further simplify this, but I havent had the time to think thru it properly. As a hint tho, since we have control over what `Menu.Button` is, we could also specify additional `.Button` subcomponents that apply the desired props when used, so that consumers don't have to remember these prop values, class overrides, etc.

For example, something like:

```jsx
<Menu.PlainButton>
    <div className="fpo">... or an icon, or an avatar component, etc.</div>
</Menu.PlainButton>
```

In this case, whatever is in `PlainButton` (coming from a properly accessible design) takes over all the duties, and this essentially exposes the untreated HeadlessUI button.

This wouldn't block the current PR. If such an approach makes sense, it is additive and would mean we could later deprecate anything it would replace.

.storybook/recipes/MenuWithIconButton.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+  args: {
+    children: (
+      <>
+        <Menu.Button
+          className="!border-transparent-white-0 !bg-transparent-white-0 hover:!bg-button-secondary-neutral-background-hover"
+          showExpandIcon={false}
+          variant="icon"
+        >
+          <Icon name="dots-vertical" purpose="informative" title="show more" />
+        </Menu.Button>
```
booc0mtaco, line 27: Related to that other comment, how might we create a sub-component that allows for this bit of code to be much briefer, and specify the proper prop values such that we avoid the boilerplate of lines 22-24

src/components/Menu/Menu.stories.tsx
anniehu4: ooh interesting. I also just realized that the Button variants (especially `"icon"`) don't work properly because the menu's built-in button styles are overriding it. Ideally we want to make this "plain" styling case really easy -- Andrew's suggestion would work, or something like `<Menu.Button variant="plain">`.

Using the EDS Button underneath still makes sense, but I'm now questioning if we should allow consumers to pass through all Button props ... may be better/less confusing for us to control the few cases at the Menu.Button level.

.storybook/recipes/MenuWithIconButton.stories.tsx
anniehu4: we should especially not need people to pass extra styling to get this case working 😬

src/components/Menu/Menu.stories.tsx
jinlee93: I'm liking the PlainButton api which just ships the Headless Menu Button. Not sure about the button variants as well. We can just go back to not accepting button props and if customization is wanted, consumers can use the PlainButton?

.storybook/recipes/MenuWithIconButton.stories.tsx
jinlee93: Only needs the hover styling now with PlainButton

src/components/Menu/Menu.module.css
```diff
@@ -28,8 +28,20 @@
...
+  min-width: 42px;
+}
+.menu__plain-button:focus-visible {
+  @mixin focus;
+}
+@supports not selector(:focus-visible) {
+  .menu__plain-button:focus {
+    @mixin focus;
+  }
+}
```
jinlee93, line 42: Just some accessibility styling for the plain button

src/components/Menu/Menu.stories.tsx
booc0mtaco: One other benefit of having a `"plain"` variant on `Button`/`ClickableStyle` is that it can manage implementing the focus ring. Ex:

```jsx
<Button variant="plain">
   <Avatar user={...} />
</Button>
```

would automatically allow a properly-spaced focus ring clamped around the shape of the `children` (or could have Tailwind utilities provided to change the wrapper shape to get a circle). `PlainButton` could allow `as` so the consumer can implement the entire button if we ship the headless wrapper. I'm not sure I follow the button prop bit, but we can chat thru that if it still is relevant

src/components/Menu/Menu.module.css
booc0mtaco: @jinlee93 the designs *should* apply the accessible size, and specifying a minimum will be a visual hint that something is off if the input is too small.

We might also apply some centering in this case?

src/components/Menu/Menu.stories.tsx
jinlee93: Think you commented this just as I pushed a commit with `PlainButton` haha. I do like the `as` prop.
The button prop was allowing MenuButton to accept Button props for further styling, but I removed that since that's possible with PlainButton now.

src/components/Menu/Menu.module.css
jinlee93: I think it's centered by default 
image

Hesitant to add any more styling on top since TW classes don't override EDS styles in usage in edu-stack

.storybook/recipes/MenuWithIconButton.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+      options: Object.keys(icons),
+      defaultValue: 'dots-vertical',
+    },
+  },
+} as Meta<MenuProps>;
+
+export const WithVerticalDotsButton: StoryObj<MenuProps> = {
+  render: ({ iconName, args }) => (
+    <Menu {...args}>
+      <Menu.PlainButton className="hover:bg-button-secondary-neutral-background-hover">
```
anniehu4, line 28: hmm I feel like it'd be worth adding a hover color & border radius to the plain button styling. those don't feel too opinionated and reduce a bit of developer hassle.

src/components/Menu/Menu.module.css
```diff
@@ -28,8 +28,20 @@
   font-weight: var(--eds-font-weight-light);
 }
 
-/* Needed to create higher specificity than ClickableStyle. */
-.menu__button--icon {
+.menu__plain-button {
+  min-height: 44px;
+  min-width: 44px;
```
anniehu4, line 33: not for the future: might be useful to extract this pattern

src/components/Menu/Menu.module.css
```diff
@@ -28,8 +28,20 @@
   font-weight: var(--eds-font-weight-light);
 }
 
-/* Needed to create higher specificity than ClickableStyle. */
-.menu__button--icon {
+.menu__plain-button {
```
anniehu4, line 31: would be good to add a comment that this is for accessibility. note for the future: might be useful to extract this pattern

.storybook/recipes/MenuWithIconButton.stories.tsx
jinlee93: Ooh agree with border radius, but not sure about hover color since it's a transparent background, will ask design

jinlee93: Ohhh I agree with border radius but will ask design about hover bg color since the default bg is transparent and the hover tokens we have are solid colors

src/components/Menu/Menu.module.css
jinlee93: Added!

jinlee93:
> @jinlee93 nudge to update the PR title if it's not accurate anymore :)

Ya will update summary too

PR #1520, opened 2023-03-02
feat!: split apart checkbox checked and indeterminate props

ahuth:
EFI-949

### Summary:

Breaking change! (although I'm not aware of any consumers affected by it)

- Give the Checkbox component separate `checked` and `indeterminate` props.
- Also makes `indeterminate` checkboxes officially indeterminate in the DOM.

### Why:

Thought about this while working on #1511.

https://html.spec.whatwg.org/multipage/input.html#checkbox-state-(type=checkbox) calls out `indeterminate` as a separate piece of state than `checked`.

> The control is never a true tri-state control, even if the element's [indeterminate](https://html.spec.whatwg.org/multipage/input.html#dom-input-indeterminate) IDL attribute is set to true. The [indeterminate](https://html.spec.whatwg.org/multipage/input.html#dom-input-indeterminate) IDL attribute only gives the appearance of a third state.

Practically this means a checkbox can be in all 4 combinations of `checked` and `indeterminate` (although there are only 3 visual states, because both indeterminates look the same).

Verified this is how browsers behave with https://codepen.io/ahuth/pen/MWqmwyv.

### Test Plan:

  - Storybook

### DIFF
--- a/src/components/Checkbox/Checkbox.stories.tsx
+++ b/src/components/Checkbox/Checkbox.stories.tsx
@@ -13,13 +13,6 @@ export default {
   title: 'Components/Checkbox',
   component: Checkbox,
   args: defaultArgs,
-  argTypes: {
-    // For some reason Storybook does not infer all `checked` correctly;
-    checked: {
-      control: 'radio',
-      options: [true, false, 'indeterminate'],
-    },
-  },
   decorators: [
     (Story) => (
       <div
@@ -75,8 +68,7 @@ export const LargeChecked: StoryObj<Args> = {
 
 export const Indeterminate: StoryObj<Args> = {
   args: {
-    checked: 'indeterminate',
-    readOnly: true, // Prevent console warning about the field being read only
+    indeterminate: true,
   },
 };
 
@@ -84,18 +76,33 @@ export const Disabled: StoryObj<Args> = {
   render: () => (
     <table className="border-spacing-8">
       <tbody>
-        {[false, true, 'indeterminate' as const].map((checked, i) => (
-          // FIXME
-          // eslint-disable-next-line react/no-array-index-key
-          <tr key={i}>
-            <td>
-              <Checkbox checked={checked} disabled label="Disabled" />
-            </td>
-            <td>
-              <Checkbox checked={checked} label="Default" readOnly />
-            </td>
-          </tr>
-        ))}
+        {/* Un-checked */}
+        <tr>
+          <td>
+            <Checkbox checked={false} disabled label="Disabled" />
+          </td>
+          <td>
+            <Checkbox checked={false} label="Default" readOnly />
+          </td>
+        </tr>
+        {/* Checked */}
+        <tr>
+          <td>
+            <Checkbox checked disabled label="Disabled" />
+          </td>
+          <td>
+            <Checkbox checked label="Default" readOnly />
+          </td>
+        </tr>
+        {/* Indeterminate */}
+        <tr>
+          <td>
+            <Checkbox disabled indeterminate label="Disabled" />
+          </td>
+          <td>
+            <Checkbox indeterminate label="Default" readOnly />
+          </td>
+        </tr>
       </tbody>
     </table>
   ),
@@ -115,10 +122,10 @@ export const WithoutVisibleLabel: StoryObj<Args> = {
     <div className="flex flex-col gap-2">
       <Checkbox {...args} readOnly />
       <Checkbox {...args} checked readOnly />
-      <Checkbox {...args} checked="indeterminate" readOnly />
+      <Checkbox {...args} indeterminate />
       <Checkbox {...args} disabled />
       <Checkbox {...args} checked disabled />
-      <Checkbox {...args} checked="indeterminate" disabled />
+      <Checkbox {...args} disabled indeterminate />
     </div>
   ),
 };
--- a/src/components/Checkbox/__snapshots__/Checkbox.test.tsx.snap
+++ b/src/components/Checkbox/__snapshots__/Checkbox.test.tsx.snap
@@ -137,8 +137,6 @@ exports[`<Checkbox /> Disabled story renders snapshot 1`] = `
             class="checkbox"
           >
             <input
-              aria-checked="mixed"
-              checked=""
               class="checkbox__input"
               disabled=""
               id=":rb:"
@@ -157,8 +155,6 @@ exports[`<Checkbox /> Disabled story renders snapshot 1`] = `
             class="checkbox"
           >
             <input
-              aria-checked="mixed"
-              checked=""
               class="checkbox__input"
               id=":rc:"
               readonly=""
@@ -205,11 +201,8 @@ exports[`<Checkbox /> Indeterminate story renders snapshot 1`] = `
     class="checkbox"
   >
     <input
-      aria-checked="mixed"
-      checked=""
       class="checkbox__input"
       id=":r6:"
-      readonly=""
       type="checkbox"
     />
     <label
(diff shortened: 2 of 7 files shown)


src/components/Checkbox/__snapshots__/Checkbox.test.tsx.snap
```diff
@@ -137,8 +137,6 @@ exports[`<Checkbox /> Disabled story renders snapshot 1`] = `
             class="checkbox"
           >
             <input
-              aria-checked="mixed"
```
ahuth, line 140: Don't need `aria-checked` when we set `indeterminate` on the DOM node.

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -31,32 +35,24 @@ export type CheckboxInputProps = CheckboxHTMLElementProps & {
...
-  //
-  // However, https://html.spec.whatwg.org/multipage/input.html#checkbox-state-(type=checkbox)
-  // seems to consider `indeterminate` as a separate bit of information than `checked`. They can be
-  // in all 4 combinations (only 3 visual states, though).
-  //
-  // Should we make them separate props?
-  const checkedProps =
-    checked === 'indeterminate'
-      ? {
-          'aria-checked': 'mixed' as const,
```
ahuth, line 46: Previously we set `aria-checked` so screen readers knew about it, but the DOM itself didn't consider the checkbox to be indeterminate.

Here we fix that (which unfortunately requires JS and jumping through a hoop).

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -31,32 +35,24 @@ export type CheckboxInputProps = CheckboxHTMLElementProps & {
...
-          checked: true,
```
ahuth, line 47: Previously we considered indeterminate checkboxes as "checked".

That's not correct, according to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes.

Indeterminate checkboxes should be considered un-checked.

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -31,32 +35,24 @@ export type CheckboxInputProps = CheckboxHTMLElementProps & {
...
-        }
-      : {
-          checked,
-        };
+>(({ checked, className, disabled, indeterminate, ...other }, ref) => {
+  const forwardedRef = useForwardedRef(ref);
+
+  // Make this checkbox indeterminate. Can only be done with JS for some reason.
```
ahuth, line 41: See https://github.com/facebook/react/issues/1798 for why React hasn't added an `indeterminate` prop to input checkboxes

src/components/CheckboxInput/CheckboxInput.tsx
jinlee93: Think `'indeterminate'` should be removed from the argtypes `checked` options

ahuth:
How code coverage decrease when I added tests? 🤔

jinlee93:
> How test coverage decrease when I added tests? 🤔

I think coverage is calculated by Hits / Lines and your new tests covered 11 / 12 lines (91.66%) which technically brought the average down LOL

ahuth:
@chanzuckerberg/edu-frontend-infra

booc0mtaco:
@ahuth can you update the PR title to be semver? looks like github made the checkbox PR commit [non-compliant](https://github.com/chanzuckerberg/edu-design-system/commit/6fc50f53b02301789af2aa578e27e31ec970bc06) on merge 

It's annoying that it defaults to the title if there are multiple commits on a PR branch...

ahuth:
Oh, the PR title needs to adhere to conventional-commits? Guess I assumed that the defaults for merges were fine. Maybe that's why we used to only allow rebase merges (since commit messages are enforced).

booc0mtaco:
I might be missing what thing(s) break with this change? is it the place where we try and handle `checked="indeterminate"`? Most of it looks kinda additive to my eye

booc0mtaco:
> Oh, the PR title needs to adhere to conventional-commits? Of course!
> 
> Guess I assumed that the defaults for merges were fine. Maybe that's why we used to only allow rebase merges (since commit messages are enforced) 🤔

Yah, probably. It's a REALLY annoying "feature" in GitHub where it does the Right Thing (tm) if the branch has one commit to squash and merge, but this behavior when it has multiple ones. I'd honestly prefer if it left the title field blank when ambiguous but 🤷. That code is good, so we can update changelog manually if we want; no biggie

ahuth:
Updated the title.

> what thing(s) break with this change?

The `checked` prop changed from `true | false | `indeterminate` to just `true | false`. (And added an indeterminate boolean, but that's not breaking).

Any usages using `checked="indeterminate"` would need to be updated... although I don't believe there are any in the wild.

src/components/CheckboxInput/CheckboxInput.tsx
```diff
@@ -9,11 +10,9 @@ type CheckboxHTMLElementProps = Omit<
 
 export type CheckboxInputProps = CheckboxHTMLElementProps & {
   /**
-   * Whether checkbox is checked. Defaults to false.
-   * "indeterminate" can be used when a checkbox visually represents
-   * a list of checkboxes that is "partially" checked.
+   * Whether checkbox is checked.
    */
-  checked?: boolean | 'indeterminate';
+  checked?: boolean;
```
booc0mtaco, line 15: The breaking change => reducing the available options

booc0mtaco (approved):
Thanks for patching this up! I think this PR will have enough details for anyone to sort out what changes they should make in the off chance they were trying to use indeterminate directly.

src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -115,10 +122,10 @@ export const WithoutVisibleLabel: StoryObj<Args> = {
     <div className="flex flex-col gap-2">
       <Checkbox {...args} readOnly />
       <Checkbox {...args} checked readOnly />
-      <Checkbox {...args} checked="indeterminate" readOnly />
+      <Checkbox {...args} indeterminate />
       <Checkbox {...args} disabled />
       <Checkbox {...args} checked disabled />
-      <Checkbox {...args} checked="indeterminate" disabled />
+      <Checkbox {...args} disabled indeterminate />
```
jinlee93, line 128: I like how simple this looks

src/components/CheckboxInput/CheckboxInput.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+    render(<CheckboxInput id="test" indeterminate={false} />);
+    const checkbox = screen.getByRole<HTMLInputElement>('checkbox');
+    expect(checkbox.indeterminate).toEqual(false);
+  });
+
+  it('is not indeterminate when "indeterminate" is not present', () => {
+    render(<CheckboxInput id="test" />);
+    const checkbox = screen.getByRole<HTMLInputElement>('checkbox');
+    expect(checkbox.indeterminate).toEqual(false);
+  });
```
jinlee93, line 28: Just also manually tested <Checkbox indeterminate> and then checking it to true and seeing that the input.indeterminate === false

src/util/useForwardedRef.ts
```diff
@@ -0,0 +1,21 @@
+import { useRef, useEffect, type ForwardedRef, type RefObject } from 'react';
+
+/**
+ * Safely access a forwarded ref, which may or may not be present.
+ */
```
jinlee93, line 5: nice

jinlee93 (approved):

src/components/CheckboxInput/CheckboxInput.test.tsx
ahuth: Sweet, thanks!

ahuth:
Thanks!

PR #1457, opened 2023-01-11
Use czi/prettier-config-edu

jinlee93:
### Summary:
- adds prettier config for Tailwind for linted classname orders
- recommends TW vscode extension in docs
- fixes lint issues from aforementioned prettier config
### Test Plan:
- no associated lint errors

PR #1524, opened 2023-03-03
Use our shared prettier config

ahuth:
### Summary:

Not sure if this was intended from #1457 or not, but it turns out we aren't actually using our shared prettier config. More deets inline.

This PR
- Fixes that so we're using prettier-config-edu.
- Introduces an override to avoid a large diff. We may want to remove this override at some point, though.

### Test Plan:

  - Introduced prettier violations
  - Observed prettier complaining/fixing

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,0 @@
-{
-  "trailingComma": "all",
-  "singleQuote": true,
-  "plugins": ["@chanzuckerberg/prettier-config-edu"]
-}
--- a/prettier.config.js
+++ b/prettier.config.js
@@ -1,0 +1,9 @@
+const config = require('@chanzuckerberg/prettier-config-edu');
+
+module.exports = {
+  ...config,
+
+  // Override of our shared config. Should we remove this and use the same config as our other
+  // repos? Doing so would result in a large diff (every file would be updated).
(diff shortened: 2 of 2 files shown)


.prettierrc
```diff
@@ -1,5 +0,0 @@
-{
-  "trailingComma": "all",
-  "singleQuote": true,
-  "plugins": ["@chanzuckerberg/prettier-config-edu"]
```
ahuth, line 4: prettier-config-edu isn't a plugin, so adding it here wasn't actually using it. I think 🤔 

It at least wasn't using it for the rules, but may have been for the Tailwind plugin?

.prettierrc
```diff
@@ -1,5 +0,0 @@
-{
-  "trailingComma": "all",
-  "singleQuote": true,
```
ahuth, line 3: These overrides were a clue that our shared config wasn't being used, since the shared config has the same values.

jeremiah-clothier (approved):
So the end result was the same because the overrides, but now we're actually using the correct config?

ahuth:
> So the end result was the same because the overrides, but now we're actually using the correct config?

That's right

jinlee93 (approved):

PR #1527, opened 2023-03-07
chore: catchup missing merge commits

booc0mtaco:
### Summary:

There is a quirk in the existing release process that resulted in different merge commits ending up in `next` compared to `main`. this means that updates to the branches had equivalent merge commits with different hashes. This makes `next` look like its behind `main` which increments by 1 every time the process is followed. 

This catches up the merge commits. The result should have the following outcomes:

- No files are changed (as the merge commits pulled from `main` should not contain any actual diffs
- if a release is created from this branch, it should not contain any old changelog entries
- subsequent merges from `main` into `next` should show the proper diff (containing just the commits made on the release branch)

### Test Plan:

- [x] prepare a fake release from this branch to test the above output to changelog using `--dry-run`
- [x] examine diff for any changes


booc0mtaco:
### Test Branch

- created a local test branch via `git checkout -b test-branch`
- pulled in the merge commits from main via `git merge main`
- checked diff (shows updates made from release commit directly to `next`)
- test what release thinks is available via `yarn release --dry-run`

Worth noting it skips the docs and chore commits, which maybe it always has?

Both the github diff and running `git diff next...main` show the changelog and package.json updates which ... are already on `next`? Not sure why that's happening, but it doesn't apply a duplicate of that info in the requisite files from what I am seeing.

jinlee93:
If this works, we can change https://github.com/chanzuckerberg/edu-design-system/blob/main/docs/PUBLISHING.md?plain=1#L66 to merge `main` into `next` instead?

booc0mtaco:
> If this works, we can change https://github.com/chanzuckerberg/edu-design-system/blob/main/docs/PUBLISHING.md?plain=1#L66 to merge `main` into `next` instead?

@jinlee93 I had to think about it for a bit and realized this won't fully fix it like i was hoping. It's probably easier to talk thru what's happening and why we we wind up with duplicate merge commits *sometimes*. If we merged from `main` in the github UI, it would still produce a new merge commit **of the merge commit to `main`** which is possibly more confusing.

jinlee93 (approved):

booc0mtaco:
I also solved the mystery of why one particular thing is happening. The first commit we use when creating a new release is the merge commit from the previous release. That's why `main` usually shows two merge commits from the old release branch :S 

example:

Screenshot 2023-03-07 at 14 46 41

What this PR is doing is bringing those `main`-specific merges into `next` so far. This will keep happening so long as we use "merge pull request" strategy :/

timzchang:
> This will keep happening so long as we use "merge pull request" strategy :/

Does this imply another strategy will fix this? It feels like kind of a hassle to have to repeatedly catch up on merge commits

booc0mtaco:
@timzchang i'm brainstorming still, but one way to fix git-flow is by unprotecting `next`. that would let us back-merge the exact commit from `main` avoiding the duplicate merge commit created by GitHub UI. There are also other release flows that arent based on git-flow. Pros and cons abound :) 

If we stick with git flow we would want to catch up just to clean up, so that the output during release stops being so confusing. Git flow assumes there are no protected branches by design

booc0mtaco:
Abandoning this to resolve in a separate process

PR #1530, opened 2023-03-08
chore(deps): bump headless, eslint, @types/node

jinlee93:
### Summary:
- previous attempt at eslint and @types/node [bump](https://github.com/chanzuckerberg/edu-design-system/actions/runs/4346592609/jobs/7592846297) caused [axe error to occur with a headless component
](https://github.com/chanzuckerberg/edu-design-system/actions/runs/4346592609/jobs/7592846297) and had to revert
- headless update resolves this type error
- headless was previously blocked by react 18 upgrade, but now can bump

- headless bump caused some types to be more specific, had to update those
- headless now uses [ResizeObserver and not IntersectionObserver for dialog resize listening](https://github.com/tailwindlabs/headlessui/commit/9ecd8dd9261e3f474b3961f2a2f03d43c3197180#diff-887b16418d08f2a0244dbe2c80593ed1c359205992256b024534f4345be79397R312)
  - removes intersection-observer polyfill 
### Test Plan:


.github/renovate.json5
```diff
@@ -20,11 +20,6 @@
     fetchReleaseNotes: true,
   },
 
-  ignoreDeps: [
-    // headlessui starting v1.7.6 causes some downstream feature tests to fail
-    '@headlessui/react',
-  ],
-
```
jinlee93, line 27: Allows renovate to auto bump headless again

.storybook/recipes/MenuWithIconButton.stories.tsx
```diff
@@ -20,11 +20,11 @@ export default {
       defaultValue: 'dots-vertical',
     },
   },
-} as Meta<MenuProps>;
```
jinlee93, line 23: Point of story is to have different icons as menu buttons. This previously should have caused type error and now does, changed to iconName arg only

package.json
```diff
@@ -131,15 +131,14 @@
...
-    "eslint": "^8.34.0",
+    "eslint": "^8.35.0",
     "eslint-config-prettier": "^8.6.0",
     "eslint-plugin-jest": "^27.2.1",
     "eslint-plugin-prettier": "^4.2.1",
     "eslint-plugin-storybook": "^0.6.11",
     "eslint-plugin-testing-library": "^5.10.2",
     "husky": "^8.0.3",
     "identity-obj-proxy": "^3.0.0",
-    "intersection-observer": "^0.12.2",
```
jinlee93, line 142: Headless now uses ResizeObserver instead of intersection-observer. Could have used similar ResizeObserver polyfill but they were ancient and just decided to assign it in the test file

src/components/Dropdown/__snapshots__/Dropdown.test.tsx.snap
```diff
@@ -16,7 +16,7 @@ exports[`<Dropdown /> Compact story renders snapshot 1`] = `
   <button
     aria-controls="headlessui-listbox-options-:re:"
     aria-expanded="true"
-    aria-haspopup="true"
+    aria-haspopup="listbox"
```
jinlee93, line 19: This is probably the more correct aria property

src/components/Menu/Menu.tsx
```diff
@@ -65,29 +72,24 @@ export const Menu = ({ className, ...other }: MenuProps) => {
...
   return (
-    <HeadlessMenu.Button
-      as={Button}
-      className={buttonClassNames}
-      status="neutral"
-      {...other}
-    >
-      <>
+    <HeadlessMenu.Button as={React.Fragment}>
+      <Button className={buttonClassNames} status="neutral" {...other}>
```
jinlee93, line 76: This is how Headless recommends it in their [docs](https://headlessui.com/react/menu#rendering-a-different-element-for-a-component) (second code block).

Throws type error if not done this way now since it no longer merges the `as` types

src/components/Modal/Modal.test.tsx
```diff
@@ -9,8 +9,11 @@ import '../../../jest/helpers/removeModalTransitionStylesJestSerializer';
 
 const { DefaultInteractive } = composeStories(stories);
 
-// Required because the modal uses react portal under the hood.
-require('intersection-observer');
+window.ResizeObserver = class FakeResizeObserver {
+  observe() {}
+  unobserve() {}
+  disconnect() {}
+};
```
jinlee93, line 16: How headless tests as well https://github.com/tailwindlabs/headlessui/commit/9ecd8dd9261e3f474b3961f2a2f03d43c3197180#diff-7cd642ada4942212b7b7b991e9f456e7c522e21afe805fedefa4966acf5d1ae4R47

src/components/Popover/Popover.tsx
```diff
@@ -41,6 +41,11 @@ export type PopoverProps = ExtractProps<typeof HeadlessPopover> & {
    * If your trigger element is in a fixed container, use the fixed strategy.
    */
   strategy?: PopperJSOptions['strategy'];
+  /**
+   * Callback ran after Popper positions the element the first time.
+   * Refer to https://popper.js.org/docs/v2/lifecycle/#hook-into-the-lifecycle.
+   */
+  onFirstUpdate?: PopperJSOptions['onFirstUpdate'];
```
jinlee93, line 48: This in [use in the component](https://github.com/chanzuckerberg/edu-design-system/blob/main/src/components/Popover/Popover.tsx#L83), not sure why it wasn't throwing Type error previously

src/components/Toggle/Toggle.tsx
```diff
@@ -75,8 +76,12 @@ const ToggleButton = ({ className, checked, ...other }: ToggleButtonProps) => (
 /**
  * In order to use the individual <Toggle.Label> and <Toggle.Button> subcomponents for customization,
  * they need to be wrapped in this <ToggleWrapper> component for some functionality HeadlessUI provides.
+ *
+ * Instead of direct assignment to ToggleWrapper, wrapping into a custom component prevents TypeScript erroring of using private Headless Switch Group.
  */
-const ToggleWrapper = Switch.Group;
+const ToggleWrapper = (props: ExtractProps<typeof Switch.Group>) => (
+  <Switch.Group {...props} />
+);
```
jinlee93, line 84: As stated in the comments, this wrapping prevents type erroring that occurs with direct assignment and re-export, similar as https://github.com/chanzuckerberg/edu-design-system/blob/main/src/components/Popover/Popover.tsx#L106-L111

ahuth: Weird

ahuth (approved):
LGTM 🤘

src/components/Modal/Modal.test.tsx
booc0mtaco: - We could probably run this thru Jest, which would set up any spies and such in case we wanted to assert any calls to the functions ([see here](https://jestjs.io/docs/es6-class-mocks)
- Usually when overriding objects on `window` it's good to use the set/reset pattern, where we place the previous value in a temp variable, and use `beforeEach`/`afterEach` to restore

```js

const origResizeObserver = window.ResizeObserver;
window.ResizeObserver = class Fake {...}

// ... and later reset the value back to originalResizeObserver
```

jinlee93: - I didn't want to set spies as ResizeObserver is being used in Headless internally, and would test Headless functionality rather than behavior.
- Thought tear down wouldn't be necessary since a new JSDom is created for every test file and felt weird reassigning `window.ResizeObserver` to `undefined`

PR #1531, opened 2023-03-08
docs(recipe): remove BaselineCard recipe

booc0mtaco:
### Summary:

- this recipe, as-is, only shows how one might compose a stardard card, so it is redundant

### Testing

- n/a


timzchang (approved):

jinlee93 (approved):

booc0mtaco:
Thanks, all!

PR #1532, opened 2023-03-08
docs(recipe): remove DataSummaryCard recipe

booc0mtaco:
### Summary:

- this recipe, as-is, only shows how one might compose a stardard card, so it is redundant
- the layout in question is mostly specific to a single project, so not general purpose
- embedding in the example page in EDS for now

### Test Plan:

- n/a


booc0mtaco:
@anniehu4 @jinlee93 It appears this recipe is imported into the Student summary page. I reckon that should probably be dropped as well? By name alone, it appears too specific to be in EDS

timzchang (approved):
👋🏼

jinlee93:
> @anniehu4 @jinlee93 It appears this recipe is imported into the Student summary page. I reckon that should probably be dropped as well? By name alone, it appears too specific to be in EDS

I am ok with dropping that page but would check with design. Think the point of that page where the table breaks into cards in smaller viewports is captured with the `StackedCardsToTable` recipe anyways

booc0mtaco:
@jinlee93 yah, i opted to keep the page for now, and port over the summary cards just to keep the diff small. We can bring up pages with design over the next pairing session

jinlee93:
@booc0mtaco Can you place the DataSummaryCard in the page itself then, since it's the only use

booc0mtaco:
@jinlee93 done :>

jinlee93 (approved):

jinlee93:
@booc0mtaco sweet thanks :D

PR #1533, opened 2023-03-08
docs(recipe): remove GlobalHeader recipe

booc0mtaco:
### Summary:

- the layout in question is mostly specific to a single project, so not general purpose

### Test Plan:

- n/a


jinlee93:
> @jinlee93 have a peek at the chromatic snaps to see the change there. instead of the global header, it uses one of our `fpo` blocks to demonstrate the menu/sidebar container on those pages

Ohh I like it a lot 👍 What's actually pretty funny is when BM was working on this they had it like that with the FPO block first, and then filled it with the header, and now we're back to FPO block

jinlee93 (approved):

PR #1534, opened 2023-03-09
refactor(breadcrumbs): use useEffect over useLayoutEffect

jinlee93:
### Summary:
- with react 18, [useEffect fires synchronously before layout](https://github.com/reactjs/reactjs.org/blob/d14cbdca2445cd676526c4c52e1e106342ff7bb3/content/docs/hooks-reference.md?plain=1#L155) so can use useEffect over useLayoutEffect

### Test Plan:

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.tsx
+++ b/src/components/Breadcrumbs/Breadcrumbs.tsx
@@ -43,11 +43,7 @@ export const Breadcrumbs = ({
 
   const ref = React.useRef<HTMLUListElement>(null);
 
-  /**
-   * Needs useLayoutEffect over useEffect since it needs to be run before paint.
-   * Note: with React 18, might be able to use useEffect https://github.com/reactjs/reactjs.org/blob/d14cbdca2445cd676526c4c52e1e106342ff7bb3/content/docs/hooks-reference.md?plain=1#L155
-   */
-  React.useLayoutEffect(() => {
+  React.useEffect(() => {
     const updateShouldTruncate = () => {
       const willOverflow = ref.current
         ? ref.current.clientWidth < ref.current.scrollWidth


src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -43,11 +43,7 @@ export const Breadcrumbs = ({
 
   const ref = React.useRef<HTMLUListElement>(null);
 
-  /**
-   * Needs useLayoutEffect over useEffect since it needs to be run before paint.
-   * Note: with React 18, might be able to use useEffect https://github.com/reactjs/reactjs.org/blob/d14cbdca2445cd676526c4c52e1e106342ff7bb3/content/docs/hooks-reference.md?plain=1#L155
-   */
-  React.useLayoutEffect(() => {
+  React.useEffect(() => {
```
ahuth, line 46: Does this solve the flakiness? The change seems fine regardless, but I'm a little skeptical.

The linked docs say effects run before paint only on user input or `flushSync`. Otherwise I think it's still after the browser paints.

ahuth (approved):

src/components/Breadcrumbs/Breadcrumbs.tsx
booc0mtaco: There's a case for changing the implementation here from `debounce` to `throttle`, and adjusting the timeout. In the storybook UI, switching viewports causes the window to ease in to the new size, and `debounce` waits the whole time to do any layout updates. I imagine the flakiness is still possible since the overall behavior on resize is still the same.

booc0mtaco (approved):
@jinlee93 you might trigger a re-run of the CI jobs, to also check if the chromatic infrastructure updates don't wreck anything :)

src/components/Breadcrumbs/Breadcrumbs.tsx
jinlee93: Y'all are right, this won't solve the flakiness. Flakiness is happening in the FF snap as a result of the resize handler, which is fixed by adding a leading call to the debounce

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -56,7 +52,9 @@ export const Breadcrumbs = ({
       setShouldTruncate(willOverflow);
     };
 
-    const debouncedUpdateShouldTruncate = debounce(updateShouldTruncate, 200);
+    const debouncedUpdateShouldTruncate = debounce(updateShouldTruncate, 200, {
+      leading: true,
+    });
```
jinlee93, line 57: The leading call

https://user-images.githubusercontent.com/86632227/224142032-204f2292-e55d-485b-91c5-7f3802d6bf8b.mov

jinlee93: Actually gonna revert this since chrome now fails, and ff might be iframe edge case... not that many consumers of Breadcrumbs right now anyways

PR #1535, opened 2023-03-09
refactor(tailwind): configure tw config for export

jinlee93:
[EDS-813]

### Summary:
- dedupes words for TW utility classes
image
- in screenshot: 
  - can see it's no longer `border-border-link-neutral` and simply `border-link-neutral`
  - `bg-link-neutral` is not associated with any styles because `link-neutral` is border specific
  - `bg-icon-neutral-default` is still applicable since `icon-neutral-default` is not bg/border/text specific
- grabs tokens from one file in `lib` instead of `src` since `lib` is part of EDS package

- purpose is so pilots can import eds tw config and use out of the box with only adding `content` property to config
- does not provide primary color tokens for use 

### Test Plan:

[EDS-813]: https://czi-tech.atlassian.net/browse/EDS-813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   "scripts": {
     "build": "yarn build:clean && yarn build:tokens && yarn build:js && yarn build:styles && yarn copy-fonts-to-lib",
     "build:clean": "rm -rf lib/",
-    "build:tokens": "rm -rf src/tokens-dist/ && node ./style-dictionary.config.js && yarn prettier-tokens-dist && yarn copy-tokens-to-lib",
+    "build:tokens": "rm -rf src/tokens-dist/ && node ./style-dictionary.config.js && yarn copy-tokens-to-lib && yarn prettier-tokens-dist",
     "build:js": "tsc --project tsconfig.build.json",
     "build:storybook": "build-storybook -o storybook-static -s src/design-tokens/tier-1-definitions/fonts",
     "build:styles": "postcss \"src/components/**/*.css\" --dir lib/ --base src/ --verbose",
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -23,12 +23,15 @@ module.exports = {
       ...colorTokens,
     },
     backgroundColor: {
+      ...colorTokens,
       ...backgroundColorTokens,
     },
     borderColor: {
+      ...colorTokens,
       ...borderColorTokens,
     },
     textColor: {
+      ...colorTokens,
       ...textColorTokens,
     },
     fontSize: {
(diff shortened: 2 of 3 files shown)


jinlee93:
Not sure if breaking since no one using this config right now

jinlee93:
TBD: alpha publish and use in GST

tailwind.config.js
```diff
@@ -1,5 +1,11 @@
-const variableTokens = require('./src/tokens-dist/json/css-variables-nested.json');
-const staticTokens = require('./src/tokens-dist/json/variables-nested.json');
+const edsTokens = require('./lib/tokens/json/css-variables-nested.json').eds;
```
jinlee93, line 1: This is needed b/c `lib` is available in module but not `src`, but this causes CI tests to fail since `lib` is not being created in CI. How can I fix the CI tests? Add `src/tokens-dist/json/css-variables-nested.json` to files and require that? Build `lib` in CI? Alternatives?

booc0mtaco:
> Not sure if breaking since no one using this config right now except internally

If it wasn't documented as an available interface, we can assume it wasn't used and wouldn't be breaking.

tailwind.config.js
jinlee93: Mostly already built in the ci runs, tokens just now has to be built before lint job since tw config references it and its absence triggers lint error

package.json
```diff
@@ -27,7 +27,7 @@
   "scripts": {
     "build": "yarn build:clean && yarn build:tokens && yarn build:js && yarn build:styles && yarn copy-fonts-to-lib",
     "build:clean": "rm -rf lib/",
-    "build:tokens": "rm -rf src/tokens-dist/ && node ./style-dictionary.config.js && yarn prettier-tokens-dist && yarn copy-tokens-to-lib",
+    "build:tokens": "rm -rf src/tokens-dist/ && node ./style-dictionary.config.js && yarn copy-tokens-to-lib && yarn prettier-tokens-dist",
```
jinlee93, line 30: tokens has to move to lib first before prettier or else import error occurs

booc0mtaco (approved):
I think this makes sense to me. We can write up some docs showing how to get from a token ref in figma to the proper token name

jinlee93:
> I think this makes sense to me. We can write up some docs showing how to get from a token ref in figma to the proper token name

Was going to include the docs changes in this pr but will separate out into another pr

PR #1536, opened 2023-03-09
fix: markdown story styling

jinlee93:
### Summary:
- due to removing default browser styling via TW preflight, the docs also [lost styles and looks not great
](https://chanzuckerberg.github.io/edu-design-system/?path=/story/gettingstarted--getting-started)- seems a lot of the markdown files had their own tsx file, which was unnecessary, so dried those into one `Documentation.tsx`
- readded lost browser styling into the `Documentation` component
- also moves these markdown docs into `Documentation` folder 

### Test Plan:

  - Docs look pretty close to [pre-preflight docs](https://61313967cde49b003ae2a860-ytawirrrmk.chromatic.com/?path=/story/documentation-guidelines--icons)


.storybook/main.js
```diff
@@ -8,7 +8,6 @@
  */
 module.exports = {
   stories: [
-    './components/GettingStarted/GettingStarted.stories.tsx',
```
jinlee93, line 11: now in `../docs`

.storybook/components/Docs/Guildelines/Documents.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+      skip: true,
+    },
+  },
+  decorators: [
+    (Story) => (
+      <Documentation>
+        <Story />
+      </Documentation>
+    ),
+  ],
```
booc0mtaco, line 26: this could be a JSON in Docs/ that gets imported into each of the subsequent documentation pages, so that they all share the same grammar. DRY-ing the story config some

.storybook/components/Docs/Documentation.module.css
```diff
@@ -0,0 +1,105 @@
+.documentation {
```
booc0mtaco, line 1: Another possible name is `.markdown` and the component is `MarkDownWrapper` or something... 🤔?

booc0mtaco (approved):
Looks good to me. We might have places where we can tighten things even more, but we can always revisit that too

.storybook/components/Docs/Documentation.module.css
jinlee93: I'm down for `MarkDownWrapper` and `.markdown`

.storybook/components/Docs/Guildelines/Documents.stories.tsx
jinlee93: Oh good idea. Placed it in the `MarkdownWrapper.tsx` file instead tho since it needs a tsx extension for the decorator

booc0mtaco (approved):

.storybook/components/Docs/Guildelines/Documents.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+import ComponentDocs from '../../../../docs/COMPONENTS.md';
+import IconDocs from '../../../../docs/ICONS.md';
+import LayoutDocs from '../../../../docs/LAYOUT.md';
+import TokenDocs from '../../../../docs/TOKENS.md';
+import TypographyDocs from '../../../../docs/TYPOGRAPHY.md';
+import { markdownStorybookOptions } from '../MarkdownWrapper';
+
+export default {
+  title: 'Documentation/Guidelines',
+  ...markdownStorybookOptions,
```
anniehu4, line 13: ooh this is nice 😍

.storybook/components/Docs/MarkdownWrapper.module.css
```diff
@@ -0,0 +1,105 @@
+.markdown {
```
anniehu4, line 1: late to review! but we should add a file comment explaining
- purpose of this file
- where the styles come from
- why we don't care that it's all magic numbers

PR #1539, opened 2023-03-10
ci: bump gh action checkout and commitlint

jinlee93:
### Summary:
Was getting warnings about deprecations in ci jobs:
image
Bumps actions/checkout to v3 and commitlint-github-action to v5

### Test Plan:

  - Deprecation messages now gone
  - CI tests run without fail
  - [x] size limit still works


booc0mtaco (approved):
Good catch 🎣

PR #1540, opened 2023-03-10
feat(LoadingIndicator): extract and use SVG animation directly

booc0mtaco:
### Summary:

- we only have one particular loading indicator we want to use at this time
- apply styles and parameters to match previous implementation's output

### Test Plan:

  - update the snapshots as needed. verify results are otherwise the same
  - verify all existing tests pass

### DIFF
--- a/src/components/LoadingIndicator/LoadingIndicator.tsx
+++ b/src/components/LoadingIndicator/LoadingIndicator.tsx
@@ -1,7 +1,6 @@
 import clsx from 'clsx';
 import React from 'react';
 
-import { Oval } from 'react-loader-spinner';
 import styles from './LoadingIndicator.module.css';
 
 export type Props = {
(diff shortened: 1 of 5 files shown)


booc0mtaco:
@ahuth something is still depending on styled-components looks like. kinda weird...

jinlee93:
> @ahuth something is still depending on styled-components looks like. kinda weird...

Looks like the `storybook-addon-badges` package
image
But it looks like with its [latest release](https://github.com/geometricpanda/storybook-addon-badges/blob/v1.1.1/package.json) no longer uses `styled-components`

src/components/LoadingIndicator/LoadingIndicator.tsx
```diff
@@ -18,17 +17,11 @@ export type Props = {
...
    */
   ariaLabel?: string;
 
   /**
    * Whether the oval is visible. Default is true.
-   * (Shadowed from OvalProps)
-   *
-   * See: https://mhnpd.github.io/react-loader-spinner/docs/components/oval
    */
   visible?: boolean;
```
jinlee93, line 26: What's the point of having an invisible loading spinner?

booc0mtaco: `visible` for loaders allows for those cases where you want to preserve content when the loading indicator would otherwise be added and removed from the UI in quick succession. We can use this prop to handle transitions or other behaviors

jinlee93 (approved):
Thanks!

PR #1544, opened 2023-03-13
fix(Avatar): loosen props for avatar aria-label component

booc0mtaco:
### Summary:

- mark ariaLabel prop as optional
- ensure labels are always generated via additional tests

### Test Plan:


jinlee93 (approved):
🛠️

booc0mtaco:
@jeremiah-clothier this also fixes the display of "false" that was appearing in the default `aria-label` text

PR #1545, opened 2023-03-13
docs: add paper release version to each component

booc0mtaco:
### Summary:

Use the /paper version release information to each component using the existing badge storybook add-on. This makes allows us to display information that links back to the relevant design-side paper versions as they are released. Before, there may have been some expectation that the published versions of EDS aligned with the paper released version. 

Screenshot 2023-03-13 at 18 16 32

Screenshot 2023-03-13 at 18 17 08

**Note to reviewers**: if any versions look WAY off, let me know :) 

### Test Plan:

- n/a (documentation changes)

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -24,6 +24,23 @@ export const decorators = [
   ),
 ];
 
+function createPaperReleaseConfig(usingLabel: string) {
+  return {
+    [usingLabel]: {
+      styles: {
+        backgroundColor: '#ffffff',
+        borderColor: '#000000',
+        color: '#000000',
+      },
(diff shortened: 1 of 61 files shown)


jinlee93:
Good idea, was design looking for something like this?

booc0mtaco:
@jinlee93 no this didn't come up from their end, but was missing in general. Since our versions wouldn't match what's released in the UI kit, there was no good way to map when things became available

This was part of the cookbook project and in the RFC, but I saw that it was pretty easy to complete now :)

jinlee93 (approved):
Looks good, could run by design if they had any thoughts on what could go in the tooltip as well

PR #1547, opened 2023-03-15
docs(token): update tailwind config usage and show utility class e.g.s

jinlee93:
### Summary:
- with updated tailwind config, needed docs update on usage
- updated readme on how to use tailwind config file
- adds table of contents to tokens markdown file
- adds tailwind class tokens sections in tokens markdown file, with examples showing usage

### Test Plan:

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,21 @@ html {
 }
 ```
 
+### Tailwind Config
+
+The EDS Tailwind theme provides EDS color [tokens](./docs/TOKENS.md) and screens. Import the tailwind config into the app's tailwind config and supply the [content](https://tailwindcss.com/docs/content-configuration) property for use:
+
+```js
+const edsConfig.theme = require('@chanzuckerberg/eds/tailwind.config');
+
+module.exports = {
+  content: ['./app/**/*.{ts,tsx,jsx,js}'],
+  theme: edsConfig.theme,
+};
+```
+
+Refer to the [tokens tailwind section](./docs/TOKENS.md#tailwind-class-tokens) for usage guide.
+
 ## Usage
 
 Import any of the components from this package
--- a/docs/TOKENS.md
+++ b/docs/TOKENS.md
@@ -2,6 +2,26 @@
 
 EDS is a [themeable design system](https://bradfrost.com/blog/post/creating-themeable-design-systems/) in order to support different brands, design language generations, and product families. [Design tokens](https://css-tricks.com/what-are-design-tokens/) are a core part of building a EDS theme. Design tokens define low-level design values and apply those values to UI applications in order to manage the brand's look and feel separately from structural component styles.
 
+- [Style Dictionary](#style-dictionary)
+- [Design Token Architecture](#design-token-architecture)
+- [Tier 1 Tokens](#tier-1-tokens)
+  - [Animation](#tier-1-animation)
+  - [Borders](#tier-1-borders)
+  - [Breakpoints](#tier-1-breakpoints)
+  - [Colors](#tier-1-colors)
+  - [Layout](#tier-1-layout)
+  - [Shadows](#tier-1-shadows)
+  - [Size](#tier-1-size)
+  - [Typography](#tier-1-typography)
+  - [Z-index](#tier-1-z-index)
+- [Tier 2 Usage Tokens](#tier-2-usage-tokens)
+  - [Borders](#tier-2-borders)
+  - [Colors](#tier-2-colors)
+  - [Forms](#tier-2-forms)
+  - [Shadows](#tier-2-shadows)
+- [Tier 3 Component Variables](#tier-3-component-variables)
+- [Tailwind Class Tokens](#tailwind-class-tokens)
+
 ## Style Dictionary
 
 EDS uses [Style Dictionary](https://amzn.github.io/style-dictionary/#/) to manage design tokens. Style Dictionary transforms design token source files into various formats, including CSS variables, Sass variables, JSON, and native app-specific formats.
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -32,6 +32,21 @@ html {
...
+```js
+const edsConfig.theme = require('@chanzuckerberg/eds/tailwind.config');
+
+module.exports = {
+  content: ['./app/**/*.{ts,tsx,jsx,js}'],
+  theme: edsConfig.theme,
+};
+```
+
+Refer to the [tokens tailwind section](./docs/TOKENS.md#tailwind-class-tokens) for usage guide.
```
jinlee93, line 48: `CODE_GUIDELINES.md` also has a tailwind blurb, but that doc is for usage / contribution to the EDS repo, not consumers

docs/TOKENS.md
```diff
@@ -29,48 +49,48 @@ EDS's token architecture and nomenclature is as follows:
 
 Tier 1 tokens define the brand’s design values. They can be thought of as a brand style guide converted into variables. Tier1 token values should not imply any UI usage, and should not be used directly by components. Instead, Tier 1 values should be mapped to Tier 2 values.
 
-### Animation
+### Animation <a name="tier-1-animation"></a>
```
jinlee93, line 52: Had to add all these anchors for the table of contents to link properly

tailwind.config.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
   ],
   theme: {
     colors: {
+      transparent: 'transparent',
```
jinlee93, line 23: Sneaking this one in here since it's useful

docs/TOKENS.md
```diff
@@ -162,14 +182,33 @@ Additional branded values ("secondary", "tertiary", "accent", "supports", etc) c
...
 
 - `box-shadow` defines the shadow value that gets applied by default to buttons and other "elevated" components like cards.
 
 ## Tier 3 component variables
 
 Tier 3 component variables represent each component's specific values.
+
+## Tailwind Class Tokens
+
+If the EDS tailwind config theme is being used, Tier 2 and tier 3 color tokens are available as a part of tailwind utility classes. Background, border, and text colors will only be available for themselves specifically. e.g.:
```
jinlee93, line 199: Not sure how to phrase this better, but hopefully following examples provide enough context

docs/TOKENS.md
```diff
@@ -162,14 +182,33 @@ Additional branded values ("secondary", "tertiary", "accent", "supports", etc) c
...
+
+```jsx
```
booc0mtaco, line 201: Some might be confused with mixing classes that work or do not work (recommendations versus warnings). I'd recommend a "do/don't" set up here showing a div with allowed classes, and rationale, and ones that don't work and why (possibly as a header)

jeremiah-clothier (commented):
Is there an existing way for developers to easily know "which EDS colors/variables I can use" if someone doesn't understand the distinction between tier1 and tier2 then it seems like they could be quite lost

docs/TOKENS.md
booc0mtaco: I think this makes sense. You could tweak it to say "tokens apply to specific attributes of a component" and list the prefixes perhaps.

booc0mtaco (commented):
Looks good. a couple of small ideas for clarity to see what you think

tailwind.config.js
jinlee93: It already exists as `eds-theme-color-transparent-black-0` (as `rgb(0,0,0,0)` but same same) but felt like using `transparent-black-0` for common usage felt unwieldy, can revert tho

docs/TOKENS.md
```diff
@@ -162,14 +182,33 @@ Additional branded values ("secondary", "tertiary", "accent", "supports", etc) c
...
+<div
+  className="
+    bg-brand-grape-100 // will NOT work since tier 1 colors are not available
+    bg-button-icon-brand // will work since button colors are available across all color utility classes
+    bg-link-neutral // will NOT work since border colors are available for only border color utility classes
+    border-link-neutral // will work since border colors are available for border color utility classes
```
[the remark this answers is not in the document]
jinlee93: `border-link-neutral` is the token name and I'm trying to say that since it's border specific, won't work to apply it as a background color...
Lemme know if the wording doesn't seem intuitive and am super open to suggestions regarding

tailwind.config.js
booc0mtaco: I wouldn't be opposed to having a token using the keyword as well? "transparent-black-0" is ... decidedly weird; its technically black but not practically. whereas `transparent` is just that. Much clearer AND exists like all our other tokens do, which i prefer

jinlee93:
> Is there an existing way for developers to easily know "which EDS colors/variables I can use" if someone doesn't understand the distinction between tier1 and tier2 then it seems like they could be quite lost

There's another [docs page](https://chanzuckerberg.github.io/edu-design-system/?path=/story/design-tokens-tier-2-usage--colors) for tier 2 tokens. I can update this to include a tailwind suffix or something
@jeremiah-clothier

docs/TOKENS.md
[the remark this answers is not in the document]
jeremiah-clothier: Hmm, I'm still confused since I though `bg-link-neutral` meant "background link neutral" but then the text mentions border and not background, "// will NOT work since border"

jeremiah-clothier:
> There's another [docs page](https://chanzuckerberg.github.io/edu-design-system/?path=/story/design-tokens-tier-2-usage--colors) for tier 2 tokens. I can update this to include a tailwind suffix or something
> @jeremiah-clothier

Okay, I wouldn't worry about this right now. The tailwind vscode plugin has some level of autocomplete to help with this if you know what the class name starts with

tailwind.config.js
booc0mtaco: @jinlee93 i think we can say the following: "tier 1 tokens are sometimes immutable and define fixed values."

We have a few like this already, like many of the z-index.json values, some of the typography values, etc. when the token name INCLUDES the value, it is immutable... maybe also a hint that its an anti-pattern but that's for a later discussion)

docs/TOKENS.md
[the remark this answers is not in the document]
jinlee93: Lemme know if the new comment seems clearer

tailwind.config.js
jinlee93: @booc0mtaco I am going to remove this `transparent` line from this pr and add its a token in anther pr since it seems to be a need as a t(n) token in the system itself

docs/TOKENS.md
jinlee93: Let me punt that into another PR, have to think a little more about finagling tables haha

booc0mtaco (approved):

PR #1548, opened 2023-03-16
refactor: remove unused unnecessary resets

jinlee93:
[EDS-835]

### Summary:
- removes reset mixin which is duplicative of tailwind preflight 
- also removes resetAll mixin which was never used

### Test Plan:

  - No visual regression expected
  - Checked on storybook of changed stories and saw that the styles were being applied from preflight after removing

[EDS-835]: https://czi-tech.atlassian.net/browse/EDS-835?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.storybook/recipes/PrimaryNav/PrimaryNav.stories.tsx
+++ b/.storybook/recipes/PrimaryNav/PrimaryNav.stories.tsx
@@ -4,7 +4,7 @@ import React from 'react';
 import { PrimaryNav } from './PrimaryNav';
 
 export default {
-  title: 'Components/PrimaryNav',
+  title: 'Recipes/PrimaryNav',
   component: PrimaryNav,
   subcomponents: { 'PrimaryNav.Item': PrimaryNav.Item },
   decorators: [
--- a/src/components/Breadcrumbs/Breadcrumbs.module.css
+++ b/src/components/Breadcrumbs/Breadcrumbs.module.css
@@ -1,5 +1,3 @@
-@import '../../design-tokens/mixins.css';
-
 /*------------------------------------*\
     # BREADCRUMBS
 \*------------------------------------*/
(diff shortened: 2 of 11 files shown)


.storybook/recipes/PrimaryNav/PrimaryNav.stories.tsx
```diff
@@ -4,7 +4,7 @@ import React from 'react';
 import { PrimaryNav } from './PrimaryNav';
 
 export default {
-  title: 'Components/PrimaryNav',
+  title: 'Recipes/PrimaryNav',
```
jinlee93, line 7: @booc0mtaco snuck this one in here since it's in the `recipes` folder... there's prob low value in having this as a recipe but didn't want to remove yet

booc0mtaco: oh good catch. while you're in there, see if any others in the recipe folder are showing up where they shouldn't!

src/components/Breadcrumbs/Breadcrumbs.module.css
```diff
@@ -1,5 +1,3 @@
-@import '../../design-tokens/mixins.css';
```
booc0mtaco, line 1: It's interesting that mixins.css was in design-tokens. it doesn't have any associated tokens, right?

booc0mtaco (approved):
LGTM 🚢

src/components/Breadcrumbs/Breadcrumbs.module.css
jinlee93: Nope, no associated tokens

.storybook/recipes/PrimaryNav/PrimaryNav.stories.tsx
jinlee93: Not that I saw, but I did leave some comments in the component audit spreadsheet 👍

PR #1552, opened 2023-03-20
feat(Button): use LoadingIndicator for loading state

booc0mtaco:
### Summary:

- replace the inline icon (loading) with our <LoadingIndicator />
- remove custom animation as well
- update snapshots to account for this
- don't use the visible prop, as this breaks ARIA for buttons
- mark `loading` as deprecated (we will direct people and designers to compose instead, like we do with icons)

### Test Plan:

- update snapshots

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -3,7 +3,8 @@ import type { ReactNode } from 'react';
 import React, { forwardRef } from 'react';
 import ClickableStyle from '../ClickableStyle';
 import type { ClickableStyleProps, VariantStatus } from '../ClickableStyle';
-import Icon from '../Icon';
+import LoadingIndicator from '../LoadingIndicator';
+
 import styles from './Button.module.css';
 
 type ButtonHTMLElementProps = React.ButtonHTMLAttributes<HTMLButtonElement>;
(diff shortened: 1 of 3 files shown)


src/components/Button/Button.tsx
```diff
@@ -93,16 +95,7 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
...
-          <Icon
-            className={styles['eds-is-loading']}
-            name="spinner"
-            purpose="informative"
-            size="1.5em"
-            title="loading"
-          />
-        )}
-
+        {loading && <LoadingIndicator className="eds-is-loading" size="sm" />}
```
jinlee93, line 98: When do you think this variant will be removed? Just not a fan of this `eds-is-loading` classnaming haha should be more BEM

jinlee93 (approved):
Looks good, just one probably unrelated to landing q

src/components/Button/Button.tsx
booc0mtaco: Let's chat about it this week. we could do whenever, but removing is technically breaking, so didn't want to commit to a v12 just yet :) 

agreed on that class and this implementation 🪓

booc0mtaco:
@jinlee93 Thanks! we can chat this week to see what the coming week has in store w.r.t. publishable changes

PR #1553, opened 2023-03-21
chore: use Node 18.15.0 for local dev

ahuth:
### Summary:

I'm updating all frontend infra repos to use the same Node version: v18.15.0.

Not that it really matters, but there's no harm in upgrading every once in a while to get bugfixes, perf improvements, etc. But we don't need to rush to always be on the absolute latest version or anything.

Once this merges you can run `nodenv install` and be good to go.

### Test Plan:

  - Tests


jeremiah-clothier (approved):

jinlee93 (approved):

booc0mtaco (approved):
@ahuth cool. This could be a cool dogfooding opportunity for `dev-doctor` as one of its version rules. Not ready for that yet, but will keep in mind (plus, I don't think we know of bugs caused by an old node version just yet)

PR #1554, opened 2023-03-22
feat: add utility types to enforce either label or aria-label 

booc0mtaco:
### Summary:

- we have a chunk of boilerplate code for checking labels on some components. since the text was the same from component to component, there was some confusion on which component was causing the error. use types to assert the needed props (1 of 2 labeling props as necessary), and document this in the intellisense popups as a component note
 
### Test Plan:

  - existing test suite with language updates

### DIFF
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -1,5 +1,6 @@
 import clsx from 'clsx';
 import React, { useId } from 'react';
+import type { EitherInclusive } from '../../util/utility-types';
 import type { CheckboxInputProps } from '../CheckboxInput';
 import CheckboxInput from '../CheckboxInput';
 import type { CheckboxLabelProps } from '../CheckboxLabel';
@@ -15,14 +16,23 @@ export type CheckboxProps = Omit<CheckboxInputProps, 'id'> & {
    */
   id?: string;
   /**
-   * Visible text label for the checkbox.
-   */
-  label?: React.ReactNode;
-  /**
    * Size of the checkbox label.
    */
   size?: CheckboxLabelProps['size'];
-};
+} & EitherInclusive<
+    {
+      /**
+       * Visible text label for the component.
+       */
+      label: React.ReactNode;
+    },
+    {
+      /**
+       * Aria-label to provide an accesible name for the text input if no visible label is provided.
+       */
+      'aria-label': string;
+    }
+  >;
 
 /**
  * `import {Checkbox} from "@chanzuckerberg/eds";`
@@ -31,29 +41,13 @@ export type CheckboxProps = Omit<CheckboxInputProps, 'id'> & {
  *
  * Checkbox control indicating if something is selected or unselected.
  *
- * Requires either a visible label or an accessible name.
- *
- * NOTE: usually, we would re-export subcomponents with
- *   Checkbox.Input = CheckboxInput;
- *   Checkbox.Label = CheckboxLabel;
- * but this does not compile with Typescript since Checkbox itself is a
- * forwarded ref component. Thus Input and Label must be imported separately.
+ * NOTE: Requires either a visible label or `aria-label` prop.
  */
 export const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
   (props, ref) => {
     // All remaining props are passed to the `input` element
     const { className, id, label, size = 'lg', disabled, ...other } = props;
 
-    // When possible, use a visible label through the `label` prop instead.
-    // In rare cases where there's no visible label, you must provide an
-    // `aria-label` for screen readers.
-    if (
-      process.env.NODE_ENV !== 'production' &&
-      !label &&
-      !props['aria-label']
-    ) {
-      throw new Error('You must provide a visible label or aria-label');
-    }
     const generatedId = useId();
     const checkboxId = id || generatedId;
 
@@ -75,4 +69,11 @@ export const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
   },
 );
 
+/**
+ * NOTE: usually, we would re-export subcomponents with
+ *   Checkbox.Input = CheckboxInput;
+ *   Checkbox.Label = CheckboxLabel;
+ * but this does not compile with Typescript since Checkbox itself is a
+ * forwarded ref component. Thus Input and Label must be imported separately.
+ */
 Checkbox.displayName = 'Checkbox';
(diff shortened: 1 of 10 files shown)


src/util/assertAccessible.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Assertion function for enforcing the inclusion of a label on the component
+ *
+ * @param componentName Name of the component which needs a label (.displayName)
+ * @param labelProps list of props to check for a label value
+ */
+export function assertHasLabel(
+  componentName: string,
+  ...labelProps: (undefined | string)[]
+): void {
+  if (process.env.NODE_ENV !== 'production') {
```
jeremiah-clothier, line 11: I still kinda like wrapping the calling of `assertHasLabel` in the `if (process.env.NODE_ENV !== 'production') {...}` block so the code and import gets stripped from production builds. But this is a very small distinction and I understand the tradeoff of not wanting to repeat having to type of `if (process.env.NODE_ENV !== 'production')`

src/util/assertAccessible.ts
```diff
@@ -0,0 +1,18 @@
...
+    if (!labelProps.some((labelProp) => !!labelProp)) {
+      throw new Error(
+        `You must provide at least one label for component: ${componentName}`,
```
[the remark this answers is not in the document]
booc0mtaco: all valid points. It's hard to make the error message detailed without passing more info to the function, and the fixes could vary per component in theory, so the error message could be more unique.

I've actually changed my mind about this PR... maybe a better way to deal with this is thru typescript. I'll try that out to see how it feels

src/components/Toggle/Toggle.tsx
```diff
@@ -35,21 +35,27 @@ type ToggleButtonProps = {
...
-   * Visible text label for the toggle.
-   */
-  label?: React.ReactNode;
-};
+} & EitherBoth<
+    {
+      /**
+       * Visible text label for the toggle.
+       */
+
```
booc0mtaco, line 47: nit: trim white space

src/util/utility-types.ts
```diff
@@ -4,3 +4,23 @@
  * Required since React.ComponentProps doesn't think it's a JSXElementConstructor/ReactElement. https://github.com/FB-PLP/traject/pull/11929/files#r948531375
  */
 export type ExtractProps<T> = T extends React.ComponentType<infer P> ? P : T;
+
+/**
+ * Given two types, only allow one of them to be valid
+ */
```
jinlee93, line 10: Could you add a blurb about if the types are missing, will the ts error message will say it'll prefer the first type?

jinlee93 (approved):
Definitely liking the typing approach over the assertion function

jeremiah-clothier (approved):

src/components/Checkbox/Checkbox.tsx
```diff
@@ -31,29 +41,13 @@ export type CheckboxProps = Omit<CheckboxInputProps, 'id'> & {
...
 
-    // When possible, use a visible label through the `label` prop instead.
-    // In rare cases where there's no visible label, you must provide an
-    // `aria-label` for screen readers.
-    if (
-      process.env.NODE_ENV !== 'production' &&
-      !label &&
-      !props['aria-label']
-    ) {
-      throw new Error('You must provide a visible label or aria-label');
```
ahuth, line 55: It's great removing these runtime checks 👏

src/util/utility-types.ts
```diff
@@ -4,3 +4,23 @@
...
+  [P in keyof T]: T[P];
+} & {
+  [P in keyof U]?: never;
+};
+
+/**
+ * Given two types, specify that only one of them can be in
+ * the structure at a given time
+ */
+export type Either<T, U> = Only<T, U> | Only<U, T>;
```
ahuth, line 21: Also, how is this different from `Only`?

Both read to me that they're basically `XOR`

src/util/utility-types.ts
```diff
@@ -4,3 +4,23 @@
...
+
+/**
+ * Given two types, do the same as Either type, OR possibly both
+ */
+export type EitherBoth<T, U> = Either<T, U> | (T & U);
```
ahuth, line 26: Naming is hard. This conceptually seems similar to inclusive-or, and the others are exclusive-or.

booc0mtaco: `Either` is currently XOR, and `EitherBoth` is OR. Pretty CS-y names. I think those above could work, more consistent with each other for sure

booc0mtaco:
one funny observation. the complex `EitherBoth` type let's us define a preferred prop, while using `T | U` picks the one to show in the error alphabetically :/ We do prefer `label`, but if both are missing the type error will complain about `aria-label` missing

src/util/utility-types.ts
```diff
@@ -4,3 +4,23 @@
...
 export type ExtractProps<T> = T extends React.ComponentType<infer P> ? P : T;
+
+/**
+ * Given two types, only allow one of them to be valid
+ */
+export type Only<T, U> = {
+  [P in keyof T]: T[P];
+} & {
+  [P in keyof U]?: never;
+};
```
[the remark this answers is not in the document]
booc0mtaco: EDIT: tested `T | U` and it had an error actually :'( it lost the type for `label` somehow on each component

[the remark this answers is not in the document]
ahuth: Oh, I see. `Only` isn't useful on its own.

But as part of `Either` you get:

```
| { keyA: string, keyB?: never }
| { keyB: string, keyA?: never }
```

AKA one or the other.

So the way to read this is for `Only<A, B>` you can only have `A`.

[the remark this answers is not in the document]
ahuth: > We do have a few places that can be XOR, and maybe these component props should be XOR?

Yea, seems reasonable.

src/util/utility-types.ts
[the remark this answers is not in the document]
ahuth: nvm, now that I understand what `Only` is doing

ahuth: nvm

src/util/utility-types.ts
ahuth, line 10:
```suggestion
/**
 * Given two object types, only the first is valid.
 */
```

src/util/utility-types.ts
ahuth: Right, exactly. EitherBoth is inclusive, and Either is exclusive.

I like the idea of using more CS-y names. Makes it more clear from the name what to expect.

PR #1557, opened 2023-03-22
refactor: remove circular dependencies

jinlee93:
[EDS-883]

### Summary:
- removes circular dependencies by 
  - importing components from component folder rather than `src/index`
  - moving Accordion subcomponents within Accordion file itself (could have been fixed via finagling exports/imports but this needed to be done anyways to follow subcomponent within main component file pattern)
- required to resolve [rollup warnings](https://github.com/chanzuckerberg/edu-design-system/pull/1555)

### Test Plan:

- [x] CI tests / new tests are not applicable
- [x] When cherry picked onto rollup branch and running `yarn build:js` no more circular deps warning

[EDS-883]: https://czi-tech.atlassian.net/browse/EDS-883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1558, opened 2023-03-23
build(rollup): mark dependencies as external

jinlee93:
[EDS-884]

### Summary:
- uses [`@rollup/plugin-node-resolve`](https://github.com/rollup/plugins/tree/master/packages/node-resolve#rollupplugin-node-resolve) as listed in rollup docs to mark external deps as [external](https://rollupjs.org/configuration-options/#external)
- alternatively [could import package json](https://github.com/chanzuckerberg/edu-design-system/commit/0a9d46ec86f7350cf0f4e5be607d8a02506156ac) and do this ourselves to prevent a devdep to keep track of but not sure if importing `package.json` is good practice (never seen it before lol)

### Test Plan:

  - external deps warning is gone when running `yarn build:js`

[EDS-884]: https://czi-tech.atlassian.net/browse/EDS-884?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jeremiah-clothier:
Can't figure out why some of the UI Tests have diffs. I swear they look the same :/

jinlee93:
> Can't figure out why some of the UI Tests have diffs. I swear they look the same :/

I think something in Chromatic changed since even pr's with [unrelated changes](https://github.com/chanzuckerberg/edu-design-system/pull/1557) had the same visual diffs across all the stories. Rebased and the diffs are gone

rollup.config.mjs
```diff
@@ -9,8 +10,9 @@ export default {
     preserveModules: true,
     preserveModulesRoot: 'src',
   },
-  external: ['react-children-by-type'],
+  external: [/node_modules/],
   plugins: [
+    nodeResolve(),
```
ahuth, line 15: What problem were you seeing? Node modules were included in the rollup build?

ahuth: Okay. Wasn't clear before, but I think I got it:

- By default rollup only includes relative modules in the bundle
- The node-resolve plugin lets us include node_modules in the bundle

With this new config, is React and/or Headless being included in the bundle? If so, should those be explicitly marked as external (instead of included in the bundle)?

ahuth (approved):
LGTM 🤘

rollup.config.mjs
ahuth: nvm, Jin and I looked into this more and all's good with it.

The plugin just tells rollup how to resolve paths to node_modules.

jinlee93: Added a comment explaining what the external option does

PR #1568, opened 2023-03-29
chore: Tailwind 3.3 and TS config file

ahuth:
### Summary:

This PR updates Tailwind to 3.3: https://tailwindcss.com/blog/tailwindcss-v3-3

Lots of cool stuff in this release, but the main reason I want to update it because we can now write our Tailwind config files in TS.
- Will make future config updates (if needed) easier
- Makes it easier to use TS config files in Pilot apps (since they will be importing EDS's config)

### Test Plan:

  - CI + Chromatic

### DIFF
--- a/.storybook/pages/AlongDemo/AlongDemo.tsx
+++ b/.storybook/pages/AlongDemo/AlongDemo.tsx
@@ -330,7 +330,7 @@ const WatchPage = ({ onLogout }: { onLogout: () => void }) => {
               </a>
             </video>
 
-            <div className="mt-4 ml-2 flex gap-4">
+            <div className="ml-2 mt-4 flex gap-4">
               {ReplyIcon}
               <Text>
                 My greatest accomplishment this year was in math, because I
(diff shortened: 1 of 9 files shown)


.storybook/pages/AlongDemo/AlongDemo.tsx
```diff
@@ -330,7 +330,7 @@ const WatchPage = ({ onLogout }: { onLogout: () => void }) => {
               </a>
             </video>
 
-            <div className="mt-4 ml-2 flex gap-4">
+            <div className="ml-2 mt-4 flex gap-4">
```
ahuth, line 333: I guess the class ordering changed with 3.3. Seems more alphabetical now.

jinlee93 (approved):
Thanks for the update!

.storybook/pages/AlongDemo/AlongDemo.tsx
booc0mtaco: Looks like it used to follow the order of the actual CSS shortcut values for `margin` (top, left, bottom, right)...

PR #1569, opened 2023-03-29
docs(Table): add recipe showing input with split label

booc0mtaco:
### Summary:

- create a recipe showing how to one can use `Table`, `Input`, and `Label` to create a tabular form input structure

### Test Plan:

### DIFF
--- a/.storybook/recipes/Sliders.stories.tsx
+++ b/.storybook/recipes/Sliders.stories.tsx
@@ -160,7 +160,7 @@ export const WithVisualLabel: StoryObj<Args> = {
             id="slider-label"
             text="Mood Slider"
           />
-          <div className="w-1/2 py-4 text-center text-h1">
+          <div className="text-h1 w-1/2 py-4 text-center">
             {moodData.map((mood) => {
               return sliderValue === mood.value && <>{mood.emoji}</>;
             })}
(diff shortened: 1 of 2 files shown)


.storybook/recipes/Sliders.stories.tsx
```diff
@@ -160,7 +160,7 @@ export const WithVisualLabel: StoryObj<Args> = {
             id="slider-label"
             text="Mood Slider"
           />
-          <div className="w-1/2 py-4 text-center text-h1">
+          <div className="text-h1 w-1/2 py-4 text-center">
```
booc0mtaco, line 163: ..................

timzchang: Sorry, I don't get it 😅 I think there's something about ordering I don't understand

timzchang (approved):

.storybook/recipes/Sliders.stories.tsx
booc0mtaco: I think this might be from the update to tailwind.. i'm not sure why it wants this class to be first, but prettier complained and re-saving the file made this change. Very bizarre

booc0mtaco: @timzchang my guess is that it places it first b/c it's not like the other classes. this one comes from our EDS-specific tailwind config, and doesn't map as cleanly to some CSS property: value combo? I tried looking for something in the release notes, but didn't see anything

booc0mtaco:
Thanks!

PR #1570, opened 2023-03-29
docs: add profile card recipe using skeleton state

booc0mtaco:
### Summary:

- move card example from skeleton state docs
- create separate recipe with example controls

### Test Plan:

  - add separate recipe snapshots


jinlee93 (approved):
Looks good!

booc0mtaco:
@jinlee93 yah, re: the animation, this is probably b/c it doesn't actually start until like 2-3 seconds after the element renders. if the animation started immediately, there'd be some flake :)

PR #1571, opened 2023-03-30
feat(slider): add tooltip

jinlee93:
### Summary:
- adds tooltip feature to slider
Screenshot 2023-03-29 at 8 44 01 PM
- tooltip spawns outside of the slider hit box:
Screenshot 2023-03-29 at 7 50 48 PM

### Test Plan:

  - Play function with hover to snap in Chromatic

### DIFF
--- a/.storybook/recipes/Sliders.stories.tsx
+++ b/.storybook/recipes/Sliders.stories.tsx
@@ -160,7 +160,7 @@ export const WithVisualLabel: StoryObj<Args> = {
             id="slider-label"
             text="Mood Slider"
           />
-          <div className="w-1/2 py-4 text-center text-h1">
+          <div className="text-h1 w-1/2 py-4 text-center">
             {moodData.map((mood) => {
               return sliderValue === mood.value && <>{mood.emoji}</>;
             })}
@@ -195,7 +195,7 @@ export const WithMultipleVisualLabels: StoryObj<Args> = {
             id="slider-label"
             text="Mood Slider"
           />
-          <div className="flex w-1/2 justify-between py-4 text-center text-h1">
+          <div className="text-h1 flex w-1/2 justify-between py-4 text-center">
             {moodData.map((mood, index) => {
               return (
                 <div
--- a/src/components/Slider/Slider.module.css
+++ b/src/components/Slider/Slider.module.css
@@ -181,7 +181,7 @@
 }
 
 /**
- * Centers the tooltip arrow to the tooltip horizontally.
+ * Horizontally centers the tooltip arrow respective to the tooltip.
  */
 .slider :global(.tippy-arrow) {
   transform: none !important;
--- a/src/components/Slider/Slider.module.css
+++ b/src/components/Slider/Slider.module.css
@@ -182,7 +182,7 @@
  * Horizontally centers the tooltip arrow respective to the tooltip.
  */
 .slider :global(.tippy-arrow) {
-  transform: none !important;
+  transform: initial !important;
   margin-left: auto;
   margin-right: auto;
   left: 0;
--- a/src/components/Slider/Slider.module.css
+++ b/src/components/Slider/Slider.module.css
@@ -179,3 +179,14 @@
 .slider__marker:not(:first-child):not(:last-child) {
   text-align: center;
 }
+
+/**
+ * Centers the tooltip arrow to the tooltip horizontally.
+ */
+.slider :global(.tippy-arrow) {
+  transform: none !important;
+  margin-left: auto;
+  margin-right: auto;
+  left: 0;
+  right: 0;
+}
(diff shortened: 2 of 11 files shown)


src/components/Slider/Slider.module.css
```diff
@@ -179,3 +179,14 @@
 .slider__marker:not(:first-child):not(:last-child) {
   text-align: center;
 }
+
+/**
+ * Centers the tooltip arrow to the tooltip horizontally.
+ */
+.slider :global(.tippy-arrow) {
```
jinlee93, line 186: Had to add a global selector since we can't reference the shadow dom thumb, not ideal but seems stable, easy to catch if there are changes, easy to fix.

src/components/Slider/Slider.tsx
```diff
@@ -135,15 +147,30 @@ export const Slider = ({
...
+          hideOnClick={false}
+          offset={({ reference }) => {
+            // offsets the tooltip relative to the position of the thumb
+            // the magic number 24 is the slider thumb size (see relevant css module)
+            return [(ratio - 0.5) * (reference.width - 24), 0];
+          }}
+          reference={ref}
+          text={tooltip}
+        />
+      )}
```
jinlee93, line 173: Ideally we wouldn't have Tooltip functionality baked into the Slider component like this, but the native `<input type="range">` makes the thumb a shadow dom element and unselectable for the Tooltip.

I did explore having an invisible, absolutely positioned div on top of the thumb, but ran into a couple issues:
1. The div on top would need to pass mouse functionality to the thumb below, but with `pointer-events: none` removes the hover on it to spawn the tooltip
2. If we forward a ref to the tooltip, I feel like it should be on the input element and not the invisible div. That can be somewhat resolved with `useImperativeHandle` to have two refs but that means moving two refs around and also increases complexity too much for this, and seems like a lot of boilerplate to be writing for the consumer

jinlee93: Some thoughts to mitigate would be to no longer use the semantic element and handle the various accessibility requirements ourselves and hence have a referencable thumb... but that's outside current scope

.storybook/recipes/Sliders.stories.tsx
```diff
@@ -195,7 +195,7 @@ export const WithMultipleVisualLabels: StoryObj<Args> = {
             id="slider-label"
             text="Mood Slider"
           />
-          <div className="flex w-1/2 justify-between py-4 text-center text-h1">
+          <div className="text-h1 flex w-1/2 justify-between py-4 text-center">
```
jinlee93, line 198: This is just reordering the class names so it doesn't throw lint errors... @booc0mtaco

booc0mtaco: Hm, I had to do this on several PRs too. I think that minor version of tailwind contained this breaking change :(

booc0mtaco:
@jinlee93 Does the tooltip appear on hover on the thumb, or when sliding begins? I kinda assumed it would only appear when the thumb is being dragged but that behavior requirement is undefined

jinlee93:
> @jinlee93 Does the tooltip appear on hover on the thumb, or when sliding begins? I kinda assumed it would only appear when the thumb is being dragged but that behavior requirement is undefined

Tooltip appears on hover but dragged on mobile, which I think is fine since larger breakpoints have more space and can be greedier with the tooltip

src/components/Slider/Slider.tsx
```diff
@@ -135,15 +147,30 @@ export const Slider = ({
...
       />
+      {tooltip && (
+        <Tooltip
+          align="top"
+          appendTo="parent"
+          hideOnClick={false}
+          offset={({ reference }) => {
+            // offsets the tooltip relative to the position of the thumb
+            // the magic number 24 is the slider thumb size (see relevant css module)
+            return [(ratio - 0.5) * (reference.width - 24), 0];
```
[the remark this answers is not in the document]
jinlee93: Created a tier 3 token and also compiled it for use here

src/components/Slider/Slider.module.css
```diff
@@ -179,3 +177,14 @@
 .slider__marker:not(:first-child):not(:last-child) {
   text-align: center;
 }
+
+/**
+ * Horizontally centers the tooltip arrow respective to the tooltip.
+ */
+.slider :global(.tippy-arrow) {
+  transform: none !important;
```
booc0mtaco, line 185: would [`unset`](https://developer.mozilla.org/en-US/docs/Web/CSS/unset) work here? or `initial`?

booc0mtaco (approved):
Looks good to me. We maybe need to give more space between the tooltip and the thumb, but if that didn't come up in design feedback, then all is good.

src/components/Slider/Slider.module.css
jinlee93: Both would work, which would be `none` for transform in this case, but will change to `initial` in case a parent wants to set transform

jinlee93:
> Looks good to me. We maybe need to give more space between the tooltip and the thumb, but if that didn't come up in design feedback, then all is good.

Design was fine with it, and in second screenshot of pr summary shows that the tooltip is outside of the hitbox of the slider

PR #1572, opened 2023-03-31
build!: generate cjs and type declarations

jinlee93:
[EDS-886]

### Summary:
- builds both cjs and esm with rollup 
  - cjs built as `*.cjs` files to `/lib`
  - esm built as `*.js` files to `lib`
  - sourcemaps also turned on
- mitigated by not building declarations with rollup, and using tsc to build declarations separately into `lib/types`
- now looks like: 
image

### Test Plan:

  - built and copied folder into edu_stack `node_modules/@chanzuckerberg/eds`
  - added a `debugger` line to one of the used components in the `es` folder to make sure the esm build was being used:
Screenshot 2023-03-31 at 12 55 10 AM
  - also saw that sourcemaps being built correctly
  - adding `debugger` into a component not being used did not trigger the browser debugger, suggesting tree shaking is working 
  - [x] running `npm test` in edu-stack passes even with`transformIgnorePatterns: ['/node_modules/(?!(@chanzuckerberg/eds))']` removed
    - passes now without `tslib` since that was a peerdep of `@rollup/plugin-typescript`

[EDS-886]: https://czi-tech.atlassian.net/browse/EDS-886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tsconfig.json
```diff
@@ -4,13 +4,13 @@
...
     "allowSyntheticDefaultImports": true,
     "declaration": true,
+    "declarationDir": "lib/types",
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
     "jsx": "react",
     "lib": ["es5", "es6", "dom"],
     "moduleResolution": "node",
     "noUnusedLocals": true,
-    "outDir": "lib",
```
jinlee93, line 13: No longer using `tsc` to build

tsconfig.json
```diff
@@ -4,13 +4,13 @@
     "allowJs": false,
     "allowSyntheticDefaultImports": true,
     "declaration": true,
+    "declarationDir": "lib/types",
```
jinlee93, line 7: builds the declaration files to this folder

rollup.config.mjs
```diff
@@ -24,6 +37,15 @@ export default {
...
       tsconfig: 'tsconfig.build.json',
+      compilerOptions: {
+        /**
+         * Rollup wants declarations in the same directory as the output folder,
+         * but since we ship both cjs and esm in separate folders, this will throw an error.
+         * Declarations are hence built separately using tsc into lib/types
+         */
+        declaration: false,
+        declarationDir: undefined,
+      },
```
jinlee93, line 48: There's already an issue raised in the plugin repo https://github.com/rollup/plugins/issues/1230#issuecomment-1324735742

rollup.config.mjs
```diff
@@ -2,14 +2,27 @@ import { nodeResolve } from '@rollup/plugin-node-resolve';
...
+      preserveModules: true,
+      preserveModulesRoot: 'src',
+      sourcemap: true,
+    },
+    {
+      dir: 'lib/cjs',
+      format: 'cjs',
+      preserveModules: true,
+      preserveModulesRoot: 'src',
+      sourcemap: true,
```
jinlee93, line 23: We can turn sourcemaps off, not sure if anyone will take a look but also wouldn't hurt to have

package.json
```diff
@@ -90,8 +93,6 @@
     "react-popper": "^2.3.0",
     "react-portal": "^4.2.2"
   },
-  "entry": "lib/index.js",
```
jinlee93, line 93: Not sure what this line did?

ahuth: No idea. Probs something in the BM template that doesn't do anything.

package.json
```diff
@@ -5,7 +5,9 @@
   "author": "CZI <edu-frontend-infra@chanzuckerberg.com>",
   "homepage": "https://github.com/chanzuckerberg/edu-design-system",
   "license": "MIT",
-  "main": "lib/index.js",
+  "main": "lib/cjs/index.js",
+  "module": "lib/es/index.js",
```
ahuth, line 9: These will certainly work 👍 

Although in the future we may want to switch over to [conditional exports](https://nodejs.org/api/packages.html#conditional-exports).

Per https://nodejs.org/api/packages.html#dual-commonjses-module-packages (and other places in the docs), I think `exports` is preferred now (although main/module is still common).

ahuth (approved):
Nice 🤘

package.json
```diff
@@ -5,7 +5,24 @@
...
+      "require": "./lib/index.css"
+    },
+    "./fonts.css": {
+      "import": "./lib/tokens/fonts.css",
+      "require": "./lib/tokens/fonts.css"
+    },
+    "./tailwind.config.mjs": {
+      "import": "./tailwind.config.mjs"
+    }
+  },
```
jinlee93, line 24: @ahuth Not sure if I'm using conditional exports right but seems that we have to explicitly state which files are allowed to be exported

rollup.config.mjs
```diff
@@ -2,14 +2,28 @@ import { nodeResolve } from '@rollup/plugin-node-resolve';
...
+    },
+    {
+      dir: 'lib',
+      format: 'cjs',
+      preserveModules: true,
+      preserveModulesRoot: 'src',
+      sourcemap: true,
+      entryFileNames: '[name].cjs',
```
jinlee93, line 24: This builds the cjs files using `.cjs` extension into the same directories, no longer requiring separate `es` and `cjs` directories... I prefer it this way but we can always go back to separate directories

tailwind.config.mjs
```diff
@@ -64,4 +66,4 @@ export default {
       xxl: '1400px',
     },
   },
-} satisfies Config;
+};
```
jinlee93, line 69: @ahuth I converted this into an `mjs` file so that it can be used without processing in pilots

package.json
```diff
@@ -5,7 +5,24 @@
   "author": "CZI <edu-frontend-infra@chanzuckerberg.com>",
   "homepage": "https://github.com/chanzuckerberg/edu-design-system",
   "license": "MIT",
-  "main": "lib/index.js",
+  "exports": {
+    ".": {
+      "import": "./lib/index.js",
+      "require": "./lib/index.cjs"
+    },
+    "./index.css": {
```
jinlee93, line 13: I short handed some of these styles exports... it IS breaking but the whole build feature will be breaking so might as well include it... lmk if we shouldn't break this

package.json
```diff
@@ -5,7 +5,24 @@
...
+      "import": "./lib/index.css",
+      "require": "./lib/index.css"
+    },
+    "./fonts.css": {
+      "import": "./lib/tokens/fonts.css",
+      "require": "./lib/tokens/fonts.css"
+    },
```
jinlee93, line 20: I short handed some of these styles exports so it's less unwieldy (i.e. `import '@chanzuckerberg/eds/fonts.css` and not `import @chanzuckerberg/eds/lib/tokens/fonts.css`) ... it IS breaking but the whole build feature will be breaking so might as well include it... lmk if we shouldn't break this since cost is pretty low

jinlee93:
Still getting the test errors in edu-stack 
image
with conditional exports

package.json
ahuth: Yeah, that's true. I kind of forgot that there was more than just the index.js files (such as fonts, css, tailwind, etc).

Does this seem like it's too cumbersome, or is it okay?

rollup.config.mjs
ahuth: Either way seems reasonable, so let's go with your preference.

tailwind.config.mjs
ahuth: I think it's better to keep TS, so that it's easier for the pilots to write their tailwind files in TS.

Meaning, the pilot swill have to process their tailwind.config.ts anyway, so it's fine for that TS file to import another TS file (the EDS tailwind config).

package.json
jinlee93: It's fine unless consumers want to use some of the internal components or the utils, which we haven't come across yet

tailwind.config.mjs
jinlee93: Makes sense to me, reverted to TS

tsconfig.json
```diff
@@ -4,17 +4,17 @@
...
     "lib": ["es5", "es6", "dom"],
     "moduleResolution": "node",
     "noUnusedLocals": true,
-    "outDir": "lib",
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": true,
     "suppressImplicitAnyIndexErrors": true,
-    "target": "es6"
+    "target": "es2018"
```
jinlee93, line 18: Target is being bumped to remove `tslib` usage in running rollup build with `@rollup/plugin-typescript`. tslib was used for `__rest` utility which is the spread object functionality that comes with `es2018`. `tslib` is being grepped for as a build test

.github/workflows/test.yml
```diff
@@ -41,6 +41,14 @@ jobs:
       - name: Build Dist 🔧
         run: yarn run build
 
+      - name: Check for "tslib" in build # tslib is peer dep of rollup ts plugin, but we're not including tslib due to lack of need
+        run: |
+          if grep -r '--include=*.'{cjs,js} 'tslib' ./lib -q
+          then
+            echo "tslib found in lib"
+            exit 1
+          fi
```
jinlee93, line 50: tested failing with `es2017` https://github.com/chanzuckerberg/edu-design-system/actions/runs/4600946862/jobs/8128192461

ahuth (approved):

PR #1576, opened 2023-04-01
fix: include tailwind config in package

ahuth:
### Summary:

In #1568 I converted the `tailwind.config.js` to typescript, but forgot to update the `files` entry to include the TS version in the package 😐.

So now consuming apps can no longer access the EDS tailwind config to get theme variables.

This PR fixes that.

### Test Plan:

  - `npm publish --dry-run`

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   },
   "files": [
     "/lib",
-    "tailwind.config.js"
+    "tailwind.config.*"
   ],
   "scripts": {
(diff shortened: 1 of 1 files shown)


anniehu4 (approved):

ahuth:
> since the previous change was in v11.1.0, I think we should do a patch release to include this fix? i can do that in a bit

Patch release is right. Thanks! Let me know if there's anything I can do to support.

PR #1579, opened 2023-04-03
feat: export some subcomponents

jinlee93:
[EDS-893]

### Summary:
- with upcoming conditional exports of EDS, we have to be explicit about what we want to export
- this adds subcomponents of `InputField`, `TextareaField`, and `Checkbox` to their respective parent components as object values
- adds `ForwardedRefComponent` type, which is the return type of `forwardRef` to make this possible
### Test Plan:

[EDS-893]: https://czi-tech.atlassian.net/browse/EDS-893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -1,6 +1,9 @@
 import clsx from 'clsx';
 import React, { useId } from 'react';
-import type { EitherInclusive } from '../../util/utility-types';
+import type {
+  EitherInclusive,
+  ForwardedRefComponent,
+} from '../../util/utility-types';
 import type { CheckboxInputProps } from '../CheckboxInput';
 import CheckboxInput from '../CheckboxInput';
 import type { CheckboxLabelProps } from '../CheckboxLabel';
@@ -34,6 +37,11 @@ export type CheckboxProps = Omit<CheckboxInputProps, 'id'> & {
     }
   >;
 
+type CheckboxType = ForwardedRefComponent<HTMLInputElement, CheckboxProps> & {
+  Input?: typeof CheckboxInput;
+  Label?: typeof CheckboxLabel;
+};
+
 /**
  * `import {Checkbox} from "@chanzuckerberg/eds";`
  *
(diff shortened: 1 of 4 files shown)


src/components/Checkbox/Checkbox.tsx
```diff
@@ -43,37 +51,25 @@ export type CheckboxProps = Omit<CheckboxInputProps, 'id'> & {
...
-/**
- * NOTE: usually, we would re-export subcomponents with
- *   Checkbox.Input = CheckboxInput;
- *   Checkbox.Label = CheckboxLabel;
- * but this does not compile with Typescript since Checkbox itself is a
- * forwarded ref component. Thus Input and Label must be imported separately.
- */
 Checkbox.displayName = 'Checkbox';
+Checkbox.Input = CheckboxInput;
+Checkbox.Label = CheckboxLabel;
```
booc0mtaco, line 75: I'm not sure if we will need to export these parts, as they should all be using the already-exported `Label` component. Tho, I suppose we do have a few counter examples...

src/components/Checkbox/Checkbox.tsx
```diff
@@ -43,37 +51,25 @@ export type CheckboxProps = Omit<CheckboxInputProps, 'id'> & {
  *
  * NOTE: Requires either a visible label or `aria-label` prop.
  */
-export const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
-  (props, ref) => {
-    // All remaining props are passed to the `input` element
-    const { className, id, label, size = 'lg', disabled, ...other } = props;
+export const Checkbox: CheckboxType = React.forwardRef((props, ref) => {
```
booc0mtaco, line 54: nit: usually we have an import of `forwardRef` directly, so we could tidy up the imports here

src/components/InputField/InputField.tsx
```diff
@@ -121,12 +124,17 @@ export type Props = React.InputHTMLAttributes<HTMLInputElement> & {
...
 
+type InputFieldType = ForwardedRefComponent<HTMLInputElement, Props> & {
+  Input?: typeof Input;
+  Label?: typeof Label;
+};
+
 /**
  * `import {InputField} from "@chanzuckerberg/eds";`
  *
  * An input with optional labels and error messaging built-in.
```
booc0mtaco, line 135: While here, can you add similar guidance about how the field must be labeled ([example](https://github.com/chanzuckerberg/edu-design-system/blob/next/src/components/TextareaField/TextareaField.tsx#L58))

src/components/Checkbox/Checkbox.tsx
jinlee93: Yea I can revert this one, since `<CheckboxInput>` usage pattern has already been established as well

src/components/InputField/InputField.tsx
jinlee93: Yup, done

src/components/Checkbox/Checkbox.tsx
jinlee93: Dried!

booc0mtaco (approved):

PR #1580, opened 2023-04-04
feat(TextareaField): add character length counter

booc0mtaco:
### Summary:

- only shows up if max length is specified
- turns red if the default value is greater than allowed
- updates as you type
- currently no warning color as you approach limit
- by default textareas do not allow entry above maxlength
- but you can delete if somehow above maxlength
- field not in error state when above max length

### Test Plan:

  - review snapshot / chromatic changes

### DIFF
--- a/src/components/TextareaField/TextareaField.module.css
+++ b/src/components/TextareaField/TextareaField.module.css
@@ -1,3 +1,5 @@
+@import '../../design-tokens/mixins.css';
+
 /*------------------------------------*\
     # TEXTAREA FIELD
 \*------------------------------------*/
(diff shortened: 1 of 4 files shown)


booc0mtaco:
> I don't see a story that includes the character count, would it be worth capturing that state?

We apply character count when there's a `maxlength` specified, so stories [like this one](https://61313967cde49b003ae2a860-mtsuzqeqoy.chromatic.com/?path=/story/components-textareafield--with-a-max-length) will do the trick

jeremiah-clothier:
> > I don't see a story that includes the character count, would it be worth capturing that state?
> 
> @jeremiah-clothier We apply character count when there's a `maxlength` specified, so stories [like this one](https://61313967cde49b003ae2a860-mtsuzqeqoy.chromatic.com/?path=/story/components-textareafield--with-a-max-length) will do the trick

Oh, I guess I was looking at the wrong chromatic build. I didn't realize their had been a few for the PR and the stories with that charater limit displayed were already accepted

jeremiah-clothier (approved):

jinlee93 (approved):
Any thoughts on having character count validation either as a recipe or a wrapping component to also be used on InputField?

booc0mtaco:
@jinlee93 having the counter exist as a separate composable component isn't a bad idea. We can confer with design if any other fields like `InputField` could use this. I'm inclined to keep it as an integral part of `TextareaField` for now, since that's the only place in Figma that it lives currently.

jinlee93:
> @jinlee93 having the counter exist as a separate composable component isn't a bad idea. We can confer with design if any other fields like `InputField` could use this. I'm inclined to keep it as an integral part of `TextareaField` for now, since that's the only place in Figma that it lives currently.

Makes sense to me, thanks

PR #1581, opened 2023-04-04
build!: remove some tokens from build

jinlee93:
[EDS-894]

### Summary:
- with the move to explicit conditional exports, I realized that all the individual tokens also need to be explicitly exported
- instead of this just removing them from the build since tokens as css custom properties are already included as `index.css` and don't see a need for a js use of them
- style-dictionary will still build tokens into `src` for use within the repo
- style-dictionary will build the tw tokens directly into `lib`
- removing the copy of tokens from `src/tokens-dist/` to `lib/tokens`

### Test Plan:

  - styles unaffected as shown by chromatic

[EDS-894]: https://czi-tech.atlassian.net/browse/EDS-894?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Icon/Icon.tsx
```diff
@@ -29,8 +29,7 @@ interface IconPropsBase {
    * The SVG Color, expects a valid css color (hex, rgb, etc.).
    *
    * Recommendation: if `currentColor` isn't sufficient,
-   * use `EdsThemeColor` tokens from
-   * `@chanzuckerberg/eds/lib/tokens/colors.ts`
+   * style the fill with Tailwind: https://tailwindcss.com/docs/fill
    */
   color?: string;
```
jinlee93, line 34: We can also just choose to remove this prop

booc0mtaco (approved):
Was this JSON ever used??

I think this makes sense to me... I'm still a bit curious how all of this will look once complete

jinlee93:
> Was this JSON ever used??
> 
> I think this makes sense to me... I'm still a bit curious how all of this will look once complete

`css-variables-nested.json` is used with tw config... some of the `variables-nested.json` was not used iirc and `variables.css` has been absorbed into `@czi/eds/index.css`

image
image

PR #1585, opened 2023-04-05
fix: restore check for undefined any types

booc0mtaco:
### Summary:

- fix up places that were relying on that rule
- do this now, as a future version of TS will ignore this override and start breaking everything

### Test Plan:

  - make sure all tests pass

### DIFF
--- a/src/components/Heading/Heading.tsx
+++ b/src/components/Heading/Heading.tsx
@@ -17,21 +17,44 @@ export const VARIANTS = [
    */ 'info',
 ] as const;
 export type Variant = (typeof VARIANTS)[number];
-export type HeadingElement = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+export type HeadingElement =
+  | 'h1'
+  | 'h2'
+  | 'h3'
+  | 'h4'
+  | 'h5'
+  | /** @deprecated */ 'h6';
+export type TokenSize =
+  | 'headline-lg'
+  | 'headline-md'
+  | 'headline-sm'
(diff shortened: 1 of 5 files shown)


src/components/Heading/Heading.tsx
```diff
@@ -17,21 +17,44 @@ export const VARIANTS = [
...
 ] as const;
 export type Variant = (typeof VARIANTS)[number];
-export type HeadingElement = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+export type HeadingElement =
+  | 'h1'
+  | 'h2'
+  | 'h3'
+  | 'h4'
+  | 'h5'
+  | /** @deprecated */ 'h6';
```
jeremiah-clothier, line 26: :o are <h6> not recommended anymore? I don't think I've ever actually used an <h6> just wondering where this is coming from?

jinlee93 (approved):
Thanks for fixing, looks good pending an explanation comment on the tsconfig option

tsconfig.json
```diff
@@ -14,7 +14,6 @@
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": true,
-    "suppressImplicitAnyIndexErrors": true,
```
booc0mtaco, line 17: [Documentation]([documentation suggests](https://www.typescriptlang.org/tsconfig#suppressImplicitAnyIndexErrors)) - we were mass ignoring places where `any` was set by TS

Similar to this: https://github.com/TypeStrong/ts-loader/pull/1559

PR #1587, opened 2023-04-06
docs: hide non-functional children controls in stories

booc0mtaco:
### Summary:

Storybook is showing stringified versions of some of the components' `children` and other props. Trying to edit any of these will break the storybook UI and require a page refresh. This disables those ones in the controls, which appears not to be a global option.

Screenshot 2023-04-06 at 14 07 09
Screenshot 2023-04-06 at 14 07 26

### Test Plan:

- n/a (only affects storybook)

### DIFF
--- a/.storybook/recipes/CardWithNotification/CardWithNotification.stories.tsx
+++ b/.storybook/recipes/CardWithNotification/CardWithNotification.stories.tsx
@@ -18,6 +18,13 @@ export default {
       </div>
     ),
   ],
+  argTypes: {
+    children: {
+      control: {
+        type: null,
+      },
+    },
+  },
 } as Meta<Args>;
 
 type Args = React.ComponentProps<typeof CardWithNotification>;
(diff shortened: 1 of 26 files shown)


jeremiah-clothier (commented):
In general, how does one decide which `controls` to disable and not disable?

It seems like this is the result of maybe not taking `controls` into consideration and could continue to happen in the future

booc0mtaco:
@jeremiah-clothier tl;dr this hides the controls that will break storybook if you try to edit them in the UI. 

for the most part, the ones hidden here are any props where the type is a `React.Node` type. I couldn't find anything in the storybook docs on how better to handle this, and in the screenshot above, you see this complex, `toString` version of the node in the controls? clicking into that or trying to type any content breaks storybook and requires a refresh :( 

It's possible there is a better solution, like preventing storybook from seeing these types? The doc.s are sparse on this...

booc0mtaco:
added some more clarity to the ticket and description above

jeremiah-clothier (approved):

PR #1594, opened 2023-04-17
chore: update CODEOWNERS with new team designation

booc0mtaco:
### Summary:

- update the codeowners file with the proper team name
- update the author in package.json with the same handle

### Test Plan:

  - use visual codeowners check file in GH
  - check email alias in google groups

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@chanzuckerberg/eds",
   "version": "12.0.0",
   "description": "The React-powered design system library for Chan Zuckerberg Initiative education web applications",
-  "author": "CZI <edu-dev-foundations@chanzuckerberg.com>",
+  "author": "CZI <edu-frontend-infra@chanzuckerberg.com>",
   "homepage": "https://github.com/chanzuckerberg/edu-design-system",
   "license": "MIT",
   "exports": {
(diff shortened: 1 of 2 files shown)


jeremiah-clothier (approved):

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@chanzuckerberg/eds",
   "version": "12.0.0",
   "description": "The React-powered design system library for Chan Zuckerberg Initiative education web applications",
-  "author": "CZI <edu-frontend-infra@chanzuckerberg.com>",
+  "author": "CZI <edu-dev-foundations@chanzuckerberg.com>",
```
jeremiah-clothier, line 5: Actually, I don't think this email exists yet

jeremiah-clothier (changes requested):

jeremiah-clothier:
> * check email alias in google groups

Where can you check this? I can't find the team (at least when sending email invites)

package.json
booc0mtaco, line 5: Checked and this does not exist yet.

```suggestion
  "author": "CZI <edu-frontend-infra@chanzuckerberg.com>",
```

booc0mtaco: @jeremiah-clothier when logged in, I hit groups.google.com and searched for the team alias, but it wasn't defined. I'll revert using that suggestion above

PR #1595, opened 2023-04-21
fix: polyfill for useid for react <18

jinlee93:
### Summary:
- attempts to polyfill useid for react <18 with incremental numbers
  - uses default React.useId from react 18 if exists
  - polyfill is not ssr safe since it's generated upon usage, which is render timing dependent
  - prefixes with `eds-` to avoid collision with other number id's if number id's are used elsewhere

### Test Plan:

    - `12.0.1-alpha.1` created and tested in https://github.com/chanzuckerberg/along/pull/5942

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     }
   ],
   "peerDependencies": {
-    "react": ">= 18"
+    "react": ">= 17"
   },
   "dependencies": {
     "@headlessui/react": "^1.7.14",
--- a/src/util/useId.ts
+++ b/src/util/useId.ts
@@ -5,9 +5,10 @@ function genId() {
   return ++id;
 }
 
-// Warning: probably not ssr friendly if using React <18
+console.warn(
+  'EDS useId is not ssr friendly if using React <18 as it uses the native React.useId hook if available https://react.dev/reference/react/useId.',
+);
 export const useId =
-  // Prefer React's `useId` if it's available.
   React.useId ??
   function useId() {
     const id = 'eds-' + genId();
--- a/src/util/useId.ts
+++ b/src/util/useId.ts
@@ -1,0 +1,15 @@
+import React from 'react';
+
+let id = 0;
+function genId() {
+  return ++id;
+}
+
+// Warning: probably not ssr friendly if using React <18
+export const useId =
+  // Prefer React's `useId` if it's available.
+  React.useId ??
+  function useId() {
+    const id = genId();
+    return id;
+  };
--- a/src/util/useId.ts
+++ b/src/util/useId.ts
@@ -10,6 +10,6 @@ export const useId =
   // Prefer React's `useId` if it's available.
   React.useId ??
   function useId() {
-    const id = genId();
+    const id = 'eds-' + genId();
     return id;
   };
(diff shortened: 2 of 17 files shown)


jinlee93:
> might clash with other number id's if number id's are used elsewhere

anyway we can resolve this with some kind of hashing?

src/util/useId.ts
jeremiah-clothier, line 1: Rather than just an incremented number, is there any value to having a common prefix? Can't think of any great examples so maybe not, just wandering if something like `prefix:${++id}` has any more value

jeremiah-clothier: Ooh, actually maybe this helps with your question about id conflicts

jinlee93: Oh snap I like the prefixing idea, will add with `eds:`

src/util/useId.ts
```diff
@@ -0,0 +1,15 @@
...
+function genId() {
+  return ++id;
+}
+
+// Warning: probably not ssr friendly if using React <18
+export const useId =
+  // Prefer React's `useId` if it's available.
+  React.useId ??
+  function useId() {
+    const id = 'eds-' + genId();
```
booc0mtaco, line 13: We could actually add a `console.warning` here as a reminder to anyone pulling this in. that could include any details we want to suggest, links to `useId` doc.s, etc.

jinlee93: Sounds good, converted the two comments into one console warn with a link to the useId hook doc

package.json
```diff
@@ -84,7 +84,7 @@
     }
   ],
   "peerDependencies": {
-    "react": ">= 18"
+    "react": ">= 17"
```
jinlee93, line 87: One other thing to watch out for is with the [Breadcrumbs component we use `useEffect`](https://github.com/chanzuckerberg/edu-design-system/blob/9998586a11ed82da7af2861417663e5961d12f01/src/components/Breadcrumbs/Breadcrumbs.tsx#L46) after the React 18 bump. Don't think it would have much problems with React 17 but something to keep an eye on

src/util/useId.ts
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+let id = 0;
+function genId() {
+  return ++id;
+}
+
+console.warn(
```
ahuth, line 8: I'd put the `console.warn` behind a `process.env.NODE_ENV !== 'production'` check.

Would need to verify that rollup is removing it at build time.

package.json
ahuth, line 87: Could consider adding a CI check that installs React 17 and runs the test suites.

Kind of a pain, but may be worthwhile if we need to support 17 and 18 in consumers.

booc0mtaco: @ahuth do we have an example of this somewhere? I can dig in... sounds like we'd want to configure a test run for each react version and both should pass

ahuth: No examples that I'm aware of.

Would need to jump through some hoops to do it (maybe another job that runs `npm install` and then `npm install react@v17.x.x`?)

booc0mtaco: @ahuth that makes sense. I'll create a ticket to implement this in our checks. It'll be a good one to have to prevent accidental coupling.

It's likely some other open source projects might do the same too, so I can look externally for some kinda set up.

booc0mtaco (approved):

PR #1597, opened 2023-04-22
feat: add table of token mapping

jinlee93:
### Summary:
- adds a table of token mapping so it's easier to see which tier > 1 tokens map to which tier 1 tokens
  - filterable with search term
  - should support a case where a t3 token maps to a t2 token which maps to a t1 token via some recursion
  - uses style-dictionary to spit out a flat json with the mapped token value, not the primary value
### Test Plan:

image


.storybook/components/DesignTokens/TokensTable.tsx
```diff
@@ -0,0 +1,93 @@
...
+    Object.values(map) as {
+      token: string;
+      value: string;
+      mappedTokens: string[];
+    }[]
+  ).filter(
+    (item) =>
+      // filter based on if token name includes search term or if one of the mapped token names includes search term
+      item.token.includes(searchTerm) ||
+      item.mappedTokens.join('|').includes(searchTerm),
```
jinlee93, line 34: Some cheating with a delimiter here... think it's easier to read than looping thru the array

booc0mtaco:
Let's make this a `chore` or `docs` since it doesn't emit any new features (unless we will now provide the JSON for the mapping... which maybe we should not?)

booc0mtaco:
A few suggestions for this one. I think it seems like this can replace the existing breakdowns with a few tweaks.

* formatting the token names so that they are `code` like on the other pages
* modify "Mapped Tokens" to specify they are tier2 or tier3 as appropriate
* treat the mapped tokens as an unordered list, so each falls on a separate line
* fix the column widths so that, during search, the individual columns don't change size
* consider using [`<input type="color">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color) for the color tokens along with the text of the value (include `disabled`)
* consider adding a "type" column that corresponds to the tier categories (colors, animation, borders, etc.)

jinlee93:
> A few suggestions for this one. I think it seems like this can replace the existing breakdowns with a few tweaks.
> 
> * formatting the token names so that they are `code` like on the other pages
- changed to 'code ' but don't have the orange font since that's just a hex value in the other pages. can apply if wanted
> * modify "Mapped Tokens" to specify they are tier2 or tier3 as appropriate
- done
> * treat the mapped tokens as an unordered list, so each falls on a separate line
- done
> * fix the column widths so that, during search, the individual columns don't change size
- done, made them 25% each
> * if we make the search field `type="search"` it'll add an X to clear the field, and hitting escape will as well
- done
> * consider using [`<input type="color">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color) for the color tokens along with the text of the value (include `disabled`)
- done
> * consider adding a "type" column that corresponds to the tier categories (colors, animation, borders, etc.)
- hesitant to add this since we don't have predefined categories in the docs?

@booc0mtaco 
Logic prob needs some cleaning up and had to do some typecasting if you could provide some eyes on that as well

.storybook/components/DesignTokens/TokensTable.stories.ts
```diff
@@ -0,0 +1,13 @@
...
+import { TokensTable as TokensTableComponent } from './TokensTable';
+
+export default {
+  title: 'Design Tokens/Tokens Table',
+  component: TokensTableComponent,
+  axe: {
+    // For documentation purposes only
+    skip: true,
+  },
+};
```
ahuth, line 11: Should be nested under `parameters`?

```suggestion
  parameters: {
    axe: {
      // For documentation purposes only
      skip: true,
    },
  },
};
```

Also wonder if we should skip Chromatic, as well.

booc0mtaco: yah, having chromatic for these will not be as useful since this particular thing is more of an interactive utility. Any actually useful changes detected here will also show up in other ones we capture currently.

booc0mtaco:
When we have time, let's chat thru a few use cases to test the UX. I think the finished version of this would be good to present to folks either as a recorded or live demo, which we can link to. Maybe put on a relevant wiki page or in zeroheight

booc0mtaco:
* For the tailwind docs use case
  * we only extend color, font-size and font-weight
  * should probably optimize to search anything that people would reference in design and need to know the value for
  * if making a recipe, can find the appropriate tokens they need or custom work

booc0mtaco:
Will revisit this in another change later. Closing for now.

All tokens should be searchable, and components should show which tokens they use in a useful way

PR #1600, opened 2023-04-25
fix: restore check for undefined any types

booc0mtaco:
### Summary:

- restore check to verify any types aren't making it into components

### Test Plan:

- [X] Manually tested my changes, and here are the details:
  - CI/CD build

### DIFF
--- a/src/components/Icon/Icon.stories.tsx
+++ b/src/components/Icon/Icon.stories.tsx
@@ -34,8 +34,8 @@ type Args = React.ComponentProps<typeof Icon>;
 export const Default: StoryObj<Args> = {
   render: ({ name, color, ...rest }) => {
     // ESlint can't tell if ColorTokens[color] is valid or not, since it's computed at runtime.
-    /* eslint-disable-next-line import/namespace */
-    const computedColor = color && ColorTokens[color];
+    // @ts-expect-error cannot check type compliance on runtime imports
+    const computedColor = color && ColorTokens[color]; /* eslint-disable-line */
     return <Icon {...rest} color={computedColor} name={name} />;
   },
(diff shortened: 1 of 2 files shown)


src/components/Icon/Icon.stories.tsx
```diff
@@ -34,8 +34,10 @@ type Args = React.ComponentProps<typeof Icon>;
 export const Default: StoryObj<Args> = {
   render: ({ name, color, ...rest }) => {
     // ESlint can't tell if ColorTokens[color] is valid or not, since it's computed at runtime.
-    /* eslint-disable-next-line import/namespace */
+    /* eslint-disable import/namespace */
+    // @ts-expect-error cannot check type compliance on runtime imports
     const computedColor = color && ColorTokens[color];
+    /* eslint-enable import/namespace */
```
anniehu4, line 40: I find this eslint pattern really hard to deal with in case of refactors 😅 would an inline suppression comment on line 39 work instead?

anniehu4 (approved):

src/components/Icon/Icon.stories.tsx
booc0mtaco: yah, it's a weird case b/c of a combination of things:

* we have these two exceptions stacking b/c of how those tokens are exported
* the expect error is for the typescript error and the eslint thing needs to apply to the same line
* when you put the full enlint-disable text on the line, prettier splits it into two lines, thus breaking the expect-error :lolsob: 

I changed it to just disable the one line

anniehu4: ahh of course, prettier 🥲 thanks for explaining

PR #1601, opened 2023-04-25
fix(Typography): update type ramp to sync with design 

booc0mtaco:
### Summary:

- add additional precision to the coded type ramp, to help pixel sizes resolve to whole numbers in the CSS engine (prevents browsers from resolving 2- and 3-digit sizes as partial pixel values like 31.98px vs 32px)
- sync up all font-size and line-height values between design and coded typography ramp
- flag any items that need further attention
- this is 1/n ... there may be future updates as we find any other discrepancies

### Test Plan:
  - Review with design

### DIFF
--- a/.storybook/components/DesignTokens/Tier1/TypographyPresets.jsx
+++ b/.storybook/components/DesignTokens/Tier1/TypographyPresets.jsx
@@ -5,50 +5,62 @@ import styles from './TypographyPresets.module.css';
 
 export const PRESET_SIZE_MAP = {
   '001': {
+    weights: ['regular', 'bold'],
     fontSize: 40,
     lineHeight: 48,
   },
   '002': {
+    weights: ['light', 'regular', 'bold'],
     fontSize: 32,
     lineHeight: 40,
   },
   '003': {
+    weights: ['regular', 'bold'],
     fontSize: 24,
     lineHeight: 32,
   },
   '004': {
+    weights: ['regular', 'bold'],
     fontSize: 18,
     lineHeight: 24,
   },
   '005': {
+    weights: ['light', 'regular', 'bold'],
     fontSize: 16,
     lineHeight: 24,
   },
   '006': {
+    weights: ['light', 'regular', 'bold'],
     fontSize: 14,
-    lineHeight: 24,
+    lineHeight: 22,
   },
   '007': {
+    weights: ['regular', 'bold'],
     fontSize: 14,
     lineHeight: 18,
   },
   '008': {
+    weights: ['light', 'bold'],
     fontSize: 12,
     lineHeight: 20,
   },
   '009': {
+    weights: ['regular', 'bold'],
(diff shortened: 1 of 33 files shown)


src/design-tokens/tier-1-definitions/typography-presets.css
```diff
@@ -133,7 +133,7 @@
   font-family: var(--eds-font-family-primary) $important;
   font-weight: var(--eds-font-weight-medium) $important;
   font-size: var(--eds-font-size-14) $important;
-  line-height: 1.57 $important;
```
booc0mtaco, line 136: This particular preset is very confused. In code, it is always 14/22. In design, it is sometimes 14/22 (14/24 for regular, and 14/22 for light and bold). In storybook it is always 14/24 (but 006-light is missing from the page). currently trying to clarify

booc0mtaco: Fixing up to be based on 14/22 across the board

booc0mtaco:
Screenshot 2023-04-26 at 11 06 39

jeremiah-clothier (approved):

booc0mtaco:
Waiting on some feedback for the build errors. 004-light is used in a few places, but is simultaneously not defined in the type ramp

src/design-tokens/tier-1-definitions/typography.json
```diff
@@ -41,7 +41,7 @@
         "value": "0.75rem"
       },
       "11": {
-        "value": "0.688rem"
+        "value": "0.6875rem"
```
booc0mtaco, line 44: Adding precision so that the pixel size doesn't include any partial pixels

src/design-tokens/tier-2-usage/typography-usage.css
```diff
@@ -40,31 +40,32 @@
   @mixin eds-typography-preset-009-bold $important;
 }
 
-@define-mixin eds-theme-typography-body-text-lg $important {
+@define-mixin eds-theme-typography-body-lg $important {
```
booc0mtaco, line 43: removing `-text` from the mixin names to match design names in figma

.storybook/components/DesignTokens/Tier1/TypographyPresets.jsx
booc0mtaco, line 1: adding mappings to each weight instead of special code at the bottom. This object will come in use later to generate all the typography stuff, since there are several chunks that are duplicative

booc0mtaco:
@jeremiah-clothier this one grew a bit in scope. lemme know if you spot anything that needs clarification here

src/design-tokens/tier-1-definitions/typography.json
booc0mtaco: Somehow, our process won't take changes like this, and generate other diffs to be committed (re: #1608)

PR #1603, opened 2023-04-27
chore: add ability to publish to GitHub registry

booc0mtaco:
### Summary:

- add in config to publish to github registry upon release
- will test after config is validated

### Test Plan:

  - test run on PR to see publish results


.github/workflows/release.yml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: '.node-version'
+          # Configure NPM to point to Github's package registry.
+          registry-url: https://npm.pkg.github.com/
```
ahuth, line 26:
```suggestion
          node-version-file: '.node-version'
          cache: 'yarn',
          # Configure NPM to point to Github's package registry.
          registry-url: https://npm.pkg.github.com/
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Install deps
+        uses: bahmutov/npm-install@v1
```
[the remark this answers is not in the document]
booc0mtaco: @ahuth I'll restore that pull_request option and test these out

.github/workflows/release.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Install Dependencies 🧶
+        run: yarn install --immutable
+
+      - name: Build EDS 🛠️
+        run: yarn build
+
+      # Publish the package on Github's package registry for internal CZI use. You should still
+      # publish to the npm registry separately from this job.
+      - name: Publish package to the Github registry
+        run: npm publish --tag alpha --foreground-scripts true
```
timzchang, line 39: checking my understanding: the `--tag alpha` means every version of eds we publish to the GH registry will be an alpha version, is that right? 

Could you explain why if this is intended? thanks!

booc0mtaco:
got the publish to work by simply removing `postpack`. Noted that in the husky docs, when using `npm`, you only specify `prepare` but the setup for `yarn` is more complex. Also, the output from the script does seem to indicate `postpack` runs before publish, so it appears that it, in fact, WAS running before publish.

Will try `postpublish` instead of `postpack`.

booc0mtaco:
Trying this configuration: https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages

andremalan (approved):

.github/workflows/test.yml
```diff
@@ -1,6 +1,11 @@
 name: Build & Test CI
 
-on: push
+on:
+  push:
+  workflow_call:
+    secrets:
+      CHROMATIC_PROJECT_TOKEN:
+        required: true
```
jeremiah-clothier, line 8: What does adding the `required` here do? Just error out sooner in ci?

booc0mtaco: i think so. before we were reusing this config to run all tests for releases. Andrew and I found that was redundant given how our lifecycle is set up now. 

We could remove this, but if we ever need to reuse this workflow in another workflow, we are set up to fail if the keys aren't passed over

PR #1606, opened 2023-05-02
fix: suppress warnings from useId imports and prop usage

booc0mtaco:
### Summary:

- move the `console.warn` inside the function usage ~(TODO: consider deduping its output)~
- use proper camelCase prop name in textarea tests

### Test Plan:

### DIFF
--- a/src/components/TextareaField/TextareaField.stories.tsx
+++ b/src/components/TextareaField/TextareaField.stories.tsx
@@ -15,7 +15,7 @@ export default {
     label: 'Textarea Field',
     rows: 5,
     fieldNote: 'Longer Field description',
-    spellcheck: false,
+    spellCheck: false,
   },
   parameters: {
     badges: ['1.3'],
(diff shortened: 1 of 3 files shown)


src/components/TextareaField/TextareaField.stories.tsx
```diff
@@ -15,7 +15,7 @@ export default {
     label: 'Textarea Field',
     rows: 5,
     fieldNote: 'Longer Field description',
-    spellcheck: false,
```
jeremiah-clothier, line 18: surprised this wasn't a typescript error before. Would is have been if you changed `as Meta<Args>` to `satisfies Meta<Args>`?

jeremiah-clothier (approved):
So this message will potentially be logged a lottt more now? Is that a concern

src/components/TextareaField/TextareaField.stories.tsx
booc0mtaco: it's possible. I can test that out. It's also possible typescript might allow both, and only warns at runtime... which react does for some kinds of things (the lower-case prop actually does what it says it should)

booc0mtaco:
> So this message will potentially be logged a lottt more now? Is that a concern

Thinking about the browser case, the messages would collapse under a counted set instead of printing all on separate rows if it printed multiple times. I can add a boolean toggle at the higher scope to print this exactly once, since it could become obnoxious.

PR #1608, opened 2023-05-02
fix: add missing token updates

booc0mtaco:
### Summary:

- somehow these values weren't part of any commit
- add them in now
- (note: all build releases run agains a build, including these changes)

### Test Plan:


booc0mtaco:
I think what happened on that other PR is that I made a known change without re-running storybook. Storybook triggers a build before running, so that will catch almost all cases of build changes.

andremalan (approved):

booc0mtaco:
Adding a task to prevent this edge case.

PR #1613, opened 2023-05-11
feat: apply advanced popover handling to select components

booc0mtaco:
### Summary:

- allow use of PopperJS to enhance popover handling for several components
  - Select
  - Menu
  - Popover
- consolidate types
- define documentation for each, demonstrating new behavior options
- add tests where needed

### Test Plan:


booc0mtaco:
https://github.com/chanzuckerberg/edu-design-system/assets/3056447/6e177efd-fb7b-43a9-be78-8817713066ae

booc0mtaco:
https://github.com/chanzuckerberg/edu-design-system/assets/3056447/70930a99-4c0e-4323-8cf9-541727067cdd

src/components/Popover/Popover.tsx
```diff
@@ -119,10 +80,12 @@ const PopoverGroup = (props: ExtractProps<typeof HeadlessPopover.Group>) => (
   <HeadlessPopover.Group {...props} />
 );
 
+// TODO-AH: consolidate renderprop types between all components
```
booc0mtaco, line 83: There are a few items with this comment format to tidy up still, or i may just pull those into a separate QoL ticket

src/components/Menu/Menu.tsx
```diff
@@ -98,13 +150,21 @@ const MenuPlainButton = ({ className, ...other }: MenuPlainButtonProps) => {
...
+  const optionProps = {
+    as: PopoverContainer,
+    ref: setPopperElement,
+    style: popperStyles,
+    ...props,
+    ...popperAttributes,
+  };
+
+  return (
+    <>{createPortal(<HeadlessMenu.Items {...optionProps} />, document.body)}</>
```
jinlee93, line 165: Does this mean that the headlesss menu items will always be in the document body?

booc0mtaco: @jinlee93 Correct. by using portal on `<body>` we can make sure the z-index is always correct for any popover. There are a few side-effects: some of the snapshots will change b/c the dom content won't be adjacent to the trigger anymore, and the size of some chromatic tests changed b/c they were based on relative positioning of content. since the popover is not directly next to the trigger, the visual diffs had to change a bit to show the menu.

jinlee93: Could this potentially cause any tabbing / keyboard accessibility issues?

booc0mtaco: > Could this potentially cause any tabbing / keyboard accessibility issues?

Valid concern. HeadlessUI manages the focus trap from the trigger to the popover, accounting for cases like this. So, whether the popover is adjacent to the trigger in the DOM, or portal'd somewhere else, it will work and return to the trigger if <kbd>ESC</kbd> is hit, for example

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@chanzuckerberg/eds",
-  "version": "12.0.2",
+  "version": "12.1.0-alpha.0",
```
jinlee93, line 3: Are we pushing alpha versions?

src/components/Menu/Menu.tsx
```diff
@@ -1,24 +1,33 @@
 import { Menu as HeadlessMenu } from '@headlessui/react';
 import clsx from 'clsx';
-import React from 'react';
 import type { MouseEventHandler } from 'react';
+import { useContext, useState } from 'react';
+import React from 'react';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { createPortal } from 'react-dom';
```
jinlee93, line 7: Should we have `react-dom` as a regular dependency and not a devdep?

src/components/Menu/Menu.tsx
jinlee93: Not sure if this is relevant, but what about having a persistent `<div>` as a child of the document body to place the popper elements? ref: https://popper.js.org/docs/v2/performance/#attaching-elements-to-the-dom

package.json
booc0mtaco: i pushed one for testing this on consuming projects, and will restore this to the right thing after. Also realized we need to document the step to get this to GH registry, which i will add to this PR

src/components/Menu/Menu.tsx
booc0mtaco: I thought about adding a container somewhere that this will target, but it's tricky to make it generic. We wouldn't want people to have to add some particular container whenever they use a popover-like component. What if they forget to? or change their mind?

`document.body` is guaranteed to exist and using portal here should prevent any issues with react itself (this will put it outside the app container on that page)

One way to solve for this is if we introduce some kind of "EDS app" component which we'd encourage people to use around their whole app. We could then enforce such a container. Otherwise we could put the target next to the trigger, but that would reintroduce the overflow issues ...

I'll keep that in mind when I see how some consuming apps work with a published alpha release

src/components/Menu/Menu.tsx
booc0mtaco: ah yeah forgot to flag those for removal. need to revisit what the error was

src/components/Menu/Menu.tsx
jinlee93: Makes sense to me, thanks for the explanation

booc0mtaco:
Once #1615 merges, will rebase with that to fix the UI tests

booc0mtaco:
Will add the `diffThreshold` setting to bump above 0.063 for these tests, as there is some sub-pixel shenanigans on some chromatic snapshots

booc0mtaco:
@jinlee93 Verified manually by installing an alpha with the new functionality in a pilot project example and re-running the tests there. Also manual verification on a page using `Select`

Will revert the version string change in this branch to prepare for merge.

jinlee93 (approved):
Thanks for the changes!

jinlee93 (approved):
Thanks for the changes!

PR #1614, opened 2023-05-15
chore(deps): update all non-major dependencies

czi-developer-foundations:
This PR contains the following updates:

| Package | Type | Change |
|---|---|---|
| [@rollup/plugin-typescript](https://togithub.com/rollup/plugins/tree/master/packages/typescript/#readme) ([source](https://togithub.com/rollup/plugins)) | devDependencies | [`^11.1.0` -> `^11.1.1`](https://renovatebot.com/diffs/npm/@rollup%2fplugin-typescript/11.1.0/11.1.1) |
| [@types/node](https://togithub.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/node) ([source](https://togithub.com/DefinitelyTyped/DefinitelyTyped)) | devDependencies | [`^18.16.4` -> `^18.16.8`](https://renovatebot.com/diffs/npm/@types%2fnode/18.16.5/18.16.8) |
| [@types/react](https://togithub.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react) ([source](https://togithub.com/DefinitelyTyped/DefinitelyTyped)) | devDependencies | [`^18.2.5` -> `^18.2.6`](https://renovatebot.com/diffs/npm/@types%2freact/18.2.6/18.2.6) |
| [@types/react-dom](https://togithub.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-dom) ([source](https://togithub.com/DefinitelyTyped/DefinitelyTyped)) | devDependencies | [`^18.2.3` -> `^18.2.4`](https://renovatebot.com/diffs/npm/@types%2freact-dom/18.2.4/18.2.4) |
| [eslint](https://eslint.org) ([source](https://togithub.com/eslint/eslint)) | devDependencies | [`^8.39.0` -> `^8.40.0`](https://renovatebot.com/diffs/npm/eslint/8.39.0/8.40.0) |
| [eslint-plugin-testing-library](https://togithub.com/testing-library/eslint-plugin-testing-library) | devDependencies | [`^5.10.3` -> `^5.11.0`](https://renovatebot.com/diffs/npm/eslint-plugin-testing-library/5.10.3/5.11.0) |
| [rollup](https://rollupjs.org/) ([source](https://togithub.com/rollup/rollup)) | devDependencies | [`^3.21.5` -> `^3.21.6`](https://renovatebot.com/diffs/npm/rollup/3.21.5/3.21.6) |

---

 - [ ] If you want to rebase/retry this PR, check this box

---

[...]

PR #1615, opened 2023-05-16
fix: group all font imports together

booc0mtaco:
### Summary:

- removing the font import from mixins caused Open Sans to stop appearing in storybook
- similar to Graphik, we should import the fonts from our fonts.css file
- existing instructions won't need changing, as they specify using this file already (README, storybook, etc.)

I think our Chromatic setting to do TurboSnaps meant that it didn't detect this change as a snapshot diff, so there was no clue the font had disappeared :( 

Busted snapshots start appearing later: #1613, #1614 

More long term, we should revisit [getting fonts from google anyway](https://wpspeedmatters.com/self-host-google-fonts/)

### Test Plan:

  - verify lib/ contents are correct:
  - no import lines in index.css
  - import for google font in fonts.css
  - no snapshot updates for typography page

### DIFF
--- a/src/design-tokens/tier-1-definitions/fonts.css
+++ b/src/design-tokens/tier-1-definitions/fonts.css
@@ -1,3 +1,5 @@
+@import url('https://fonts.googleapis.com/css2?family=Open+Sans:wght@400;500;600&display=swap');
+
 @font-face {
   font-family: 'Graphik';
   font-weight: 400;
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,6 +1,3 @@
-// Ensure @import rules for fonts are before anything else. Otherwise build tools can show warnings
-// in consuming apps.
-import './design-tokens/external-fonts.css';
 // This brings the style tokens that are css custom variables into the built stylesheet so only one stylesheet for EDS has to be imported.
 import './tokens-dist/css/variables.css';
 
--- a/src/design-tokens/external-fonts.css
+++ b/src/design-tokens/external-fonts.css
@@ -1,1 +1,0 @@
-@import url('https://fonts.googleapis.com/css2?family=Open+Sans:wght@400;500;600&display=swap');


booc0mtaco:
Before (12.0.2) and after (`next`):
Screenshot 2023-05-16 at 09 36 05
Screenshot 2023-05-16 at 09 35 41

src/design-tokens/tier-1-definitions/fonts.css
```diff
@@ -1,3 +1,5 @@
+@import url('https://fonts.googleapis.com/css2?family=Open+Sans:wght@400;500;600&display=swap');
```
jinlee93, line 1: This used to be in the mixins file but not opposed to having it here... any reasons why?

src/index.ts
```diff
@@ -1,6 +1,3 @@
-// Ensure @import rules for fonts are before anything else. Otherwise build tools can show warnings
-// in consuming apps.
```
jinlee93, line 2: Will the warnings come back?

booc0mtaco: the warnings were happening b/c the mixin caused at-import statements to appear throughout the index.css file. By moving it to fonts.css, the warning won't return, b/c index.css won't try to import font files anymore

src/design-tokens/tier-1-definitions/fonts.css
booc0mtaco: I think it was in the mixin file before probably b/c of how we let apps handle bundling. it'd do the right thing as it dynamically built stuff. 

I think it probably should have always been adjacent to the other font file stuff, from what I'm seeing?

jinlee93: Makes sense to me!

jinlee93 (approved):
Thanks!

PR #1617, opened 2023-05-16
test(Tooltip): add testing thresholds to tooltip snaps

booc0mtaco:
### Summary:

From https://www.chromatic.com/docs/threshold, allow for more variance in the snapshots for `Tooltip` instead of disabling the tests. Also apply to any snapshots that use render directly. This should reduce the flakiness of some of these tests (and let us re-enable the tests that were turned off before)

### Test Plan:

(n/a) Verify whether this makes any difference to tests by re-running tests a few times


jinlee93 (approved):

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -133,62 +108,32 @@ export const LongText: StoryObj<Args> = {
...
+    <Tooltip align="top" childNotInteractive text={defaultArgs.text} visible>
+      <span>Tooltip trigger</span>
+    </Tooltip>
   ),
+  parameters: {
+    chromatic: {
+      diffThreshold,
+      diffIncludeAntiAliasing: false,
+    },
+  },
```
ahuth, line 128: Should inherit this from the default export, so we can probably remove these parameters

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -200,29 +145,26 @@ export const Interactive: StoryObj<Args> = {
...
-    </div>
+      <div className="fpo  p-1">&bull;</div>
+    </Tooltip>
   ),
+  parameters: {
+    chromatic: {
+      diffThreshold,
+      diffIncludeAntiAliasing: false,
+    },
+  },
```
ahuth, line 167: Should also be able to remove these, since they're already defined at the component level (this file's default export)

src/components/Tooltip/Tooltip.stories.tsx
ahuth: Ah, well, `args` and `parameters` should be handled exactly the same.

Specifying either at the top level should cascade down to individual stories.

booc0mtaco: @ahuth yep my bad, i meant that for some of those `render` examples, we write the test like 

```jsx

render: () => <SomeComponent />
```
in which case we ignore `args` even tho its available if we put `(args)` instead and spread those onto the right component

ahuth: OH, I see 👍 

Makes sense, carry on

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -14,11 +14,7 @@ const defaultArgs = {
       efficitur mauris ac leo semper accumsan.
     </span>
   ),
-  children: (
-    <Button className={clsx(styles['trigger--spacing'])} variant="primary">
-      Tooltip trigger
-    </Button>
-  ),
+  children: <div className="fpo w-3 p-1">&bull;</div>,
```
jinlee93, line 17: Sorry to re-dig this up, why did we make the trigger a bullet? @booc0mtaco

PR #1618, opened 2023-05-17
fix: correct misspelled color token names

booc0mtaco:
### Summary:
- for menu, it was falling back to an incorrect, tho similar, color value
- for select, it was inheriting the correct hex value b/c the specific theme value was not defined

Following the naming conventions defined in [this file][variables-link]

### Test Plan:

[variables-link]: https://github.com/chanzuckerberg/edu-design-system/blob/main/src/tokens-dist/css/variables.css#L203

### DIFF
--- a/src/components/Menu/Menu.module.css
+++ b/src/components/Menu/Menu.module.css
@@ -22,7 +22,7 @@
 .menu__button {
   @mixin eds-theme-typography-body-md;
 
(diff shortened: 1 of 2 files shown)


andremalan (approved):

booc0mtaco:
Thanks!

PR #1620, opened 2023-05-17
test: call decorator story functions

jinlee93:
### Summary:
Slack context: https://czi-edu.slack.com/archives/CTFV79JH4/p1684258065353129
Calling `Story()` rather than using it as a React fn component seems to fix most of the "No Display Name" error when clicking "Show Code" in the Docs tab.
This is fine according to [SB docs](https://storybook.js.org/docs/react/writing-stories/decorators#snippet-your-component-with-decorator)

image
image

### Test Plan:

  - visual regression and jest tests pass

### DIFF
--- a/.storybook/recipes/MenuWithAvatarButton.stories.tsx
+++ b/.storybook/recipes/MenuWithAvatarButton.stories.tsx
@@ -13,41 +13,37 @@ export default {
     badges: [BADGE.BETA],
     layout: 'centered',
   },
-  decorators: [
-    (Story) => (
-      <div className="p-5">
-        <Story />
-      </div>
+  decorators: [(Story) => <div className="p-5">{Story()}</div>],
+  args: {
+    children: (
+      <>
+        <Menu.PlainButton>
+          <Avatar user={{ fullName: 'Josie Sandberg' }} />
+        </Menu.PlainButton>
+        <Menu.Items data-testid="menu-content">
+          <Menu.Item
+            href="https://headlessui.com/react/menu#menu-button"
+            icon="link"
+          >
+            Headless UI Docs
+          </Menu.Item>
+          <Menu.Item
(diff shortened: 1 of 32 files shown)


.storybook/recipes/MenuWithAvatarButton.stories.tsx
```diff
@@ -13,41 +13,37 @@ export default {
...
-        <Menu.Item onClick={() => alert('Item clicked')}>
-          Trigger Action
-        </Menu.Item>
-        <Menu.Item disabled href="https://example.org/" icon="warning">
-          Not Possible (disabled)
-        </Menu.Item>
-      </Menu.Items>
-    </Menu>
-  ),
+  },
```
jinlee93, line 46: Also if args are not spread into the render function, "Show Code" breaks as well, so snuck the conversion of using `children` here

timzchang (approved):
Do we know what causes the issue?

jeremiah-clothier (approved):
Weird, I think there a few tickets related to 'No Display Name' that can be closed if this merges

booc0mtaco:
> Do we know what causes the issue?

It's ... weird https://github.com/storybookjs/storybook/issues/12596#issuecomment-723440097

@jinlee93 I wonder if that `excludeDecorators` option fixes this w/o having to do `{Story()}`?

jinlee93:
> > Do we know what causes the issue?
> 
> It's ... weird [storybookjs/storybook#12596 (comment)](https://github.com/storybookjs/storybook/issues/12596#issuecomment-723440097)
> 
> @jinlee93 I wonder if that `excludeDecorators` option fixes this w/o having to do `{Story()}`?

I tried the `excludeDecorators` but didn't work :(

jinlee93:
> Do we know what causes the issue?

Didn't do too full of a deep dive but seems to be the storybook docs add-on related and how some properties are passed on via args

PR #1621, opened 2023-05-18
fix(drawer): handle clickoutside on slower active boolean

jinlee93:
### Summary:
- drawer was broken since the clickoutside event handler would close the Drawer when the button to open the drawer is clicked since the button is outside the drawer
  - fix by relying on internal active status rather than external active status since the internal active status sets to true after the clickoutside handler is called
### Test Plan:

- [] Manually tested my changes, but I want to keep the details secret
  - Filter popover and popover works


src/components/Drawer/Drawer.tsx
```diff
@@ -142,7 +142,7 @@ export const Drawer = ({
   useEffect(() => {
     function handleOnClickOutside(e: MouseEvent) {
       if (
-        isActive &&
```
timzchang, line 145: for my understanding: did this not work because when "handleOnClickOutside" is triggered, `isActive` is `false`, so the if is bypassed?

timzchang (approved):
I think this makes sense to me 👍🏼

src/components/Drawer/Drawer.tsx
jinlee93: When the button to open the drawer is clicked, `isActive` is set to `true` to open the drawer, then the `handleOnClickOutside` event occurs since the button is outside and immediately closes the drawer after... `activeFocus` is toggled to `true` after `handleOnClickOutside` is triggered so won't close the drawer.
previously:
1. `isActive` is set to true
2. since `isActive` is true, `handleOnClickOutside` closes the drawer

now:
1. `isActive` is set to true
2. `activeFocus` is false for now
3. `handleOnClickOutside` doesn't close drawer
4. `activeFocus` is set to true

booc0mtaco:
@jinlee93 there should be a way to test this. we could add classes to mark the current state, or data attributes, then observe how they change after taking actions. are there existing behavior tests?

jinlee93:
> @jinlee93 there should be a way to test this. we could add classes to mark the current state, or data attributes, then observe how they change after taking actions. are there existing behavior tests?

There are no existing behavior tests, will add some

jinlee93:
Added a test for the opening of the drawer, since that's what this pr is fixing. Am hesitant to add more tests since Drawer is going thru some design changes and possibly rewritten using Headless so we don't have to handle logic such as focus trapping internally @booc0mtaco

booc0mtaco (approved):

PR #1625, opened 2023-05-20
chore: expand codecov ignore values

booc0mtaco:
### Summary:

- jest making sure the regex was doing the right thing

### Test Plan:

- n/a


booc0mtaco:
@timzchang I broke the rules out as separate again, as I don't think codecov supports real regex. The percentage went back to ~91%, and breaking them out re-ignores the story files.

jeremiah-clothier (approved):

PR #1628, opened 2023-05-23
feat: add tier-3 tokens to DataBar and ProgressBar

booc0mtaco:
### Summary:

- define tokens for background and border for these two components
- these point to tier 1 values currently
- use new tokens in specified components

### Test Plan:

  - check updated snapshots


src/components/DataBar/DataBar.stories.tsx
```diff
@@ -19,18 +19,11 @@ export default {
   },
   parameters: {
     badges: ['1.0'],
+    backgrounds: {
```
booc0mtaco, line 22: The background parameter lets you change the default background color on individual stories. This is better than doing the same using a special decorator with inline styles.

Added to both databar and progress bar to align snapshots

jinlee93: Wow, pretty cool

jinlee93 (approved):

src/components/DataBar/DataBar.stories.tsx
booc0mtaco: the values we can use are defined here: https://github.com/chanzuckerberg/edu-design-system/blob/main/.storybook/preview.tsx#L48-L63

PR #1629, opened 2023-05-23
fix: use token name from design

booc0mtaco:
### Summary:

- eds-theme-color-border-brand-primary => eds-theme-color-border-brand-primary-default
- deprecate old name, and point it to the new, better name

### Test Plan:

- verify snapshots don't change


.storybook/pages/WireframeDemo/GlobalStyles.module.css
```diff
@@ -69,7 +69,7 @@
   --eds-theme-color-body-background: var(--wireframe-gray-a);
   --eds-theme-color-border-brand-primary-strong: var(--wireframe-gray-c);
   --eds-theme-color-border-brand-primary-subtle: var(--wireframe-gray-a);
-  --eds-theme-color-border-brand-primary: var(--wireframe-gray-b);
```
jinlee93, line 72: Think we need to start marking these token name changes as breaking since it will break theming for potential consumers

booc0mtaco: @jinlee93 yah lets chat thru that. In this example, i inverted the naming which will introduce regressions potentially. What I should have done is introduce the new token, and make its value default to the old token. 

Then we can have the breaking change later where we fully deprecate the old things and provide steps on how to remedy. I'll do a pr with detailed notes on rationale

PR #1630, opened 2023-05-24
fix!: align token name with design

jinlee93:
[EDS-1001]

### Summary:
- more token naming changes: color token `text-brand-primary` becomes `text-brand-default`

### Test Plan:

  - no visual regressions


jeremiah-clothier (approved):
LGTM! Was there consideration to:
- keeping `eds-theme-color-text-brand-primary`, somehow mark it as deprecated or log a warning
- adding `eds-theme-color-text-brand-default`

It may be more work on EDS side, but the library could delay a major version bump for a bit. Either way, LGTM

jinlee93:
> LGTM! Was there consideration to:
> 
> * keeping `eds-theme-color-text-brand-primary`, somehow mark it as deprecated or log a warning
> * adding `eds-theme-color-text-brand-default`
> 
> It may be more work on EDS side, but the library could delay a major version bump for a bit. Either way, LGTM

Oh good question... lemme bring it up to EDS team @chanzuckerberg/edu-design-system

booc0mtaco:
Yep @jeremiah-clothier, and this relates to #1629. 

* You can add "comment" kv pair to a token which will document it as deprecated in the output files
* You can set a new token to take its value from the old token, e.g., `{some.token.path}` in the value
* We can bundle all the deletions at a later point and provide instructions on how to migrate as part of the release notes (or as a codemod)

We should go down this path, so we don't cause too much disruption. I need to go back into that PR above and make the new read from the old, since i did the opposite 😅

src/design-tokens/tier-2-usage/colors-text.json
```diff
@@ -41,8 +41,12 @@
             "value": "{eds.color.neutral.300}"
           },
           "brand": {
-            "primary": {
+            "default": {
               "value": "{eds.color.brand.grape.700}"
+            },
+            "primary": {
+              "value": "{eds.theme.color.text.brand.default}",
+              "comment": "deprecated"
```
jinlee93, line 49: @jeremiah-clothier @booc0mtaco re-added to prevent breaking change

booc0mtaco:
@jinlee93 can you rebase so that the `fix!` is not in the commit message? Thx

src/tokens-dist/ts/colors.ts
```diff
@@ -146,7 +146,8 @@ export const EdsThemeColorTextGradeComplete = '#007249';
 export const EdsThemeColorTextGradeRevise = '#21272D';
 export const EdsThemeColorTextGradeStop = '#BD0044';
 export const EdsThemeColorTextDisabled = '#C0C4C8';
-export const EdsThemeColorTextBrandPrimary = '#5751D2';
+export const EdsThemeColorTextBrandDefault = '#5751D2';
+export const EdsThemeColorTextBrandPrimary = '#5751D2'; // deprecated
```
jeremiah-clothier, line 150: nit questioning comment position for VSCode interaction

If you change this
```diff
- export const EdsThemeColorTextBrandPrimary = '#5751D2'; // deprecated
+ /** @deprecated will be removed in the next major upgrade? please use ... instead */
- export const EdsThemeColorTextBrandPrimary = '#5751D2';
```

Does the import statement look different and include the ~strikethrough~ text?

jinlee93: Using the `@deprecated` tag does cause the token to have the ~~strikethrough~~ text but the comment is automatically added via `style-dictionary` from https://github.com/chanzuckerberg/edu-design-system/pull/1630/files#diff-845bd8692b0759a8fd359257750969e02a0c0591b4ab4eca7d6402fefc8e13bcR49 and will have to be re-added each time the tokens are built.

This `colors.ts` file is actually only used internally anyways for some stories and we should prob move away from this as well...

PR #1633, opened 2023-05-25
fix(build): build type declarations separately

jinlee93:
[EFI-1103]

### Summary:
- builds declaration files separately since with the typescript rollup config the files exporting only types get pruned 

### Test Plan:

    - `v12.0.4-alpha.1`: Screenshot 2023-05-24 at 11 18 33 PM shows the `onChange` type being inherited properly and can navigate `node_modules/eds` to find the missing utility types in `edu-stack`
      - https://github.com/chanzuckerberg/edu-stack/tree/jlee/edsAlpha14.0.4

[EFI-1103]: https://czi-tech.atlassian.net/browse/EFI-1103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,8 +34,9 @@
     "tailwind.config.*"
   ],
   "scripts": {
-    "build": "yarn build:clean && yarn build:tokens && yarn build:js && yarn copy-fonts-to-lib",
+    "build": "yarn build:clean && yarn build:tokens && yarn build:declarations && yarn build:js && yarn copy-fonts-to-lib",
     "build:clean": "rm -rf lib/",
+    "build:declarations": "tsc --emitDeclarationOnly --project tsconfig.build.json",
     "build:tokens": "rm -rf src/tokens-dist/ && node ./style-dictionary.config.js && yarn prettier-tokens-dist",
     "build:js": "rollup --config",
     "build:storybook": "build-storybook -o storybook-static -s src/design-tokens/tier-1-definitions/fonts",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@chanzuckerberg/eds",
-  "version": "12.0.3",
+  "version": "12.0.4-alpha.1",
   "description": "The React-powered design system library for Chan Zuckerberg Initiative education web applications",
   "author": "CZI <edu-frontend-infra@chanzuckerberg.com>",
   "homepage": "https://github.com/chanzuckerberg/edu-design-system",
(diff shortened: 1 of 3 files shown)


jinlee93:
@ahuth Lmk if this helps with your ts issue at all

src/components/Input/Input.tsx
```diff
@@ -37,13 +37,13 @@ export type InputProps = React.InputHTMLAttributes<HTMLInputElement> & {
    */
   isError?: boolean;
   /**
-   * Maximum number the input can take. When this number equals the input value, the plus button becomes disabled.
+   * Maximum number the input can take.
    */
-  max?: number;
+  max?: number | string | undefined;
```
booc0mtaco, line 42: `?:` might handle the `undefined` case already

booc0mtaco (commented):
Can you say a bit more about what the package content looks like with this change applied?

src/components/Input/Input.tsx
jinlee93: I just copy pasted `number | string | undefined` from `type HTMLInputElement` without seeing the `?:` haha will just simplify to `number | string`

package.json
```diff
@@ -34,8 +34,9 @@
     "tailwind.config.*"
   ],
   "scripts": {
-    "build": "yarn build:clean && yarn build:tokens && yarn build:js && yarn copy-fonts-to-lib",
+    "build": "yarn build:clean && yarn build:tokens && yarn build:declarations && yarn build:js && yarn copy-fonts-to-lib",
     "build:clean": "rm -rf lib/",
+    "build:declarations": "tsc --emitDeclarationOnly --project tsconfig.build.json",
```
jeremiah-clothier, line 39: Does this also work with `--noEmit` be added to`build:js` portion?

jinlee93:
> Can you say a bit more about what the package content looks like with this change applied?

Yup, looking at the file differences between building `next` and this branch, only adds `utility-types.d.ts`, which follows my intention of building definition files from `.ts` files that only export types. I tried converting `utility-types.ts` to the definition file but still gets pruned by rollup

package.json
jinlee93: Rollup used in `build:js` also relies on `tsconfig.build.json` and having `noEmit` in there causes a logical error with building declarations `TS5053: Option 'emitDeclarationOnly' cannot be specified with option 'noEmit'.`

ahuth (approved):
LGTM

PR #1644, opened 2023-05-31
test(TextareaField): add better custom event handler tests

booc0mtaco:
### Summary:

- remove storybook test which was printing a console statement
- instead, add a jest test for the component with content and without
- use expect statements to make sure the mock function is called

### Test Plan:

### DIFF
--- a/src/components/TextareaField/TextareaField.stories.tsx
+++ b/src/components/TextareaField/TextareaField.stories.tsx
@@ -1,5 +1,4 @@
 import type { StoryObj, Meta } from '@storybook/react';
-import { userEvent } from '@storybook/testing-library';
 import React from 'react';
 
 import { TextareaField } from './TextareaField';
@@ -91,15 +90,4 @@ export const WithAMaxLength: StoryObj<Args> = {
     required: true,
   },
   render: (args) => <TextareaField {...args} />,
-};
-
-export const AfterDelete: StoryObj<Args> = {
-  args: {
-    maxLength: 209,
-    onChange: () => console.info('ensure custom events are handled'),
-  },
-  play: () => {
-    userEvent.tab();
-    userEvent.keyboard('{arrowdown}{delete}');
-  },
 };
--- a/src/components/TextareaField/TextareaField.test.tsx
+++ b/src/components/TextareaField/TextareaField.test.tsx
@@ -1,7 +1,6 @@
 import { generateSnapshots } from '@chanzuckerberg/story-utils';
-import { screen } from '@storybook/testing-library';
 
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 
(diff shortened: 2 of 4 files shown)


booc0mtaco:
@jinlee93 This will remove that console statement that was appearing in the test runner, and better covers cases when the function should and should not run

src/components/TextareaField/TextareaField.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+import React from 'react';
+import { TextareaField } from './TextareaField';
+import * as stories from './TextareaField.stories';
+
+describe('<TextareaField />', () => {
+  generateSnapshots(stories);
+
+  it('can handle custom events when there is content', async () => {
+    const onChangeFn = jest.fn(() => console.log('test'));
```
ahuth, line 15: Don't think this needs the `console.log` anymore, since we can directly assert that the spy has been called.

IMO, console logging here just clutters up the console for little gain.

booc0mtaco: Oh yah, I left that in when i was trying to verify the mock was being triggered. Meant to take that back out!

this PR was to remove the very thing you're talking about lol

src/components/TextareaField/TextareaField.stories.tsx
```diff
@@ -92,14 +91,3 @@ export const WithAMaxLength: StoryObj<Args> = {
   },
   render: (args) => <TextareaField {...args} />,
 };
-
-export const AfterDelete: StoryObj<Args> = {
```
booc0mtaco, line 96: Main goal of this PR is to trim that one console message from appearing in the test run, and do a better and more deliberate job of testing the event handler on `TextareaField`.

src/components/TextareaField/TextareaField.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+    await userEvent.tab();
+
+    expect(field).toHaveFocus();
+
+    await user.keyboard('{arrowdown}{delete}');
+
+    expect(onChangeFn).toHaveBeenCalled();
+  });
+
+  it('will not fire a custom event when there is no content', async () => {
```
booc0mtaco, line 36: By default, no change event is triggered if an empty field is edited. This test ensures we don't break that rule at some point in the future.

src/components/TextareaField/TextareaField.test.tsx
```diff
@@ -0,0 +1,52 @@
+import { generateSnapshots } from '@chanzuckerberg/story-utils';
+
+import { render, screen } from '@testing-library/react';
```
booc0mtaco, line 3: Reminder that we have access to both `@testing-library/react` and the similar export from a storybook sub-package. When writing tests here, we don't want the storybook exports, b/c they are wrapped with additional functionality that may throw warnings.

PR #1645, opened 2023-06-01
refactor: use select over deprecated dropdown in demos

jinlee93:
### Summary:
- uses Select component over old deprecated Dropdown component in the Along demo and Wireframe demo pages
  - this removes usage of the old Dropdown component in the repo (except in its respective story file)
    - not to be confused with the ButtonDropdown component which is tied to the DropdownMenu component (confusing, I know)

### Test Plan:

  -  select is interact-ible and looks similar to the Dropdown it's replacing in the story

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -302,7 +302,7 @@ const SelectOption = function (props: SelectOptionProps) {
   );
 };
 
-type SelecButtonProps = {
+type SelectButtonProps = {
   /**
    * Optional className for additional styling.
    */
@@ -322,7 +322,7 @@ type SelecButtonProps = {
  */
 export const SelectButton = React.forwardRef<
   HTMLButtonElement,
-  SelecButtonProps
+  SelectButtonProps
 >(({ children, className, isOpen, ...other }, ref) => {
   const componentClassName = clsx(styles['select-button'], className);
   const iconClassName = clsx(
--- a/.storybook/pages/AlongDemo/AlongDemo.module.css
+++ b/.storybook/pages/AlongDemo/AlongDemo.module.css
@@ -177,12 +177,8 @@
   padding-left: 5.5rem;
 }
 
-/**
- * THEMING-NOTE: Dropdown doesn't allow customization, so we have
- * to wrap the dropdown in a div with a class and drill down to the
- * Dropdown contents.
- */
-.watch-page__dropdown button {
+.watch-page__dropdown-button {
+  color: inherit;
   border-color: var(--along-color-space-500);
   background-color: transparent;
   justify-content: center;
--- a/.storybook/pages/AlongDemo/AlongDemo.tsx
+++ b/.storybook/pages/AlongDemo/AlongDemo.tsx
@@ -3,11 +3,11 @@ import React, { useState } from 'react';
 import {
   Button,
   ButtonGroup,
-  Dropdown,
   Heading,
   Icon,
   Label,
   Link,
+  Select,
   Tabs,
   Tab,
   Text,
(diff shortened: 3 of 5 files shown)


src/components/Select/Select.tsx
```diff
@@ -302,7 +302,7 @@ const SelectOption = function (props: SelectOptionProps) {
   );
 };
 
-type SelecButtonProps = {
+type SelectButtonProps = {
```
jinlee93, line 305: snuck this typo fix

jinlee93:
Unless I'm using Select wrong @booc0mtaco, found a possible bug: in https://github.com/chanzuckerberg/edu-design-system/blob/b74b046a36b2d3d47a31f35b1837eb8075d6f1cc/src/components/Select/Select.tsx#L287 the render prop `selected` seems to stay false when selecting another option. This is present in the [main storybook](https://chanzuckerberg.github.io/edu-design-system/?path=/story/pages-theming-alongdemo--watch-page) before this PR so shouldn't block landing.

https://github.com/chanzuckerberg/edu-design-system/assets/86632227/0204385e-8652-4ed4-9183-f5fd9e5e3c01

src/components/Select/Select.tsx
booc0mtaco: ++

booc0mtaco:
@jinlee93 this seems to handle the checkbox as expected in `Select` itself: https://chanzuckerberg.github.io/edu-design-system/?path=/story/components-select--default so I'm thinking there may be something going on with the along demo itself, and not `Select`? Let's file a bug for the Along demo page for either case

booc0mtaco:
@jinlee93 For the chromatic diff, here's how to find the needed value for the threshold:

- We can use the [interactive diff tool](https://6262c53f521620003ac2ff49-ukmsdlppcb.chromatic.com/?path=/story/stories-diff-threshold-check--test-yours-out) to tell when this threshold results in no diff
- we grab each image by inspecting the element in chromatic diff screen and saving the image
- Then we upload each to this tool, and use the slider until there is no diff showing

From that we can see that it shows a diff still at ~.5 but goes away when closer to .6

Screenshot 2023-06-01 at 10 24 46
Screenshot 2023-06-01 at 10 24 30

I'll create a PR to bump to that level. This should detect if something causes the whole menu to disappear, or for the text to change, but not trigger if the content shifts left/right one pixel

jinlee93:
> @jinlee93 this seems to handle the field as expected in `Select` itself: https://chanzuckerberg.github.io/edu-design-system/?path=/story/components-select--default so I'm thinking there may be something going on with the along demo itself, and not `Select`? Let's file a bug for the Along demo page for either case

Filed as EDS-1027, but not could be lower pri if we decide to make the demo pages less Along specific

jinlee93:
> @jinlee93 For the chromatic diff, here's how to find the needed value for the threshold:
> 
> * We can use the [interactive diff tool](https://6262c53f521620003ac2ff49-ukmsdlppcb.chromatic.com/?path=/story/stories-diff-threshold-check--test-yours-out) to tell when this threshold results in no diff
> * we grab each image by inspecting the element in chromatic diff screen and saving the image
> * Then we upload each to this tool, and use the slider until there is no diff showing
> 
> From that we can see that it shows a diff still at ~.5 but goes away when closer to .6
> 
> Screenshot 2023-06-01 at 10 24 46 Screenshot 2023-06-01 at 10 24 30
> 
> ~I'll create a PR to bump to that level. This should detect if something causes the whole menu to disappear, or for the text to change, but not trigger if the content shifts left/right one pixel~
> 
> Orrrr, actually, can you do that in this PR, since this actually relates to the `Select` component? Select.stories.tsx:237 to 0.65

Bumped to 0.6 @booc0mtaco, thanks for the cool tool

timzchang (approved):
I think this looks good. Really cool chromatic tool Andrew!

booc0mtaco (approved):

PR #1646, opened 2023-06-01
refactor(breadcrumbs): use menu over deprecated buttondropdown

jinlee93:
[EFI-1002]

### Summary:
- uses the supported Menu component over the deprecated ButtonDropdown/DropdownMenu component
- #1647 removes old ButtonDropdown use in the recipes/pages
Screenshot 2023-06-01 at 4 06 59 PM

### Test Plan:

  - Ellipsis breadcrumbs works with the links in Storybook

[EDS-1002]: https://czi-tech.atlassian.net/browse/EDS-1002?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[EFI-1002]: https://czi-tech.atlassian.net/browse/EFI-1002?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Breadcrumbs/Breadcrumbs.stories.tsx
+++ b/src/components/Breadcrumbs/Breadcrumbs.stories.tsx
@@ -133,5 +133,6 @@ export const LongTextMenu: StoryObj<Args> = {
     const canvas = within(canvasElement);
     const dropdownMenuTrigger = await canvas.findByRole('button');
     userEvent.click(dropdownMenuTrigger);
+    // TODO: capture open menu in Chromatic
   },
 };
--- a/src/components/Breadcrumbs/Breadcrumbs.stories.tsx
+++ b/src/components/Breadcrumbs/Breadcrumbs.stories.tsx
@@ -117,6 +117,7 @@ export const LongTextMenu: StoryObj<Args> = {
   args: {
     ...LongText.args,
   },
+  decorators: [(Story) => <div className="pb-28">{Story()}</div>],
   parameters: {
     viewport: {
       defaultViewport: 'ipadMini',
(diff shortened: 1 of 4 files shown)


src/components/Breadcrumbs/Breadcrumbs.stories.tsx
```diff
@@ -133,5 +133,6 @@ export const LongTextMenu: StoryObj<Args> = {
     const canvas = within(canvasElement);
     const dropdownMenuTrigger = await canvas.findByRole('button');
     userEvent.click(dropdownMenuTrigger);
+    // TODO: capture open menu in Chromatic
```
jinlee93, line 136: For some reason this change broke the capturing of the open Menu state in Chromatic, tried the following to avail:
- `await new Promise((resolve) => setTimeout(resolve, 300));` in the play function to wait for loading
- `delay: 500` in the story chromatic properties to wait before snapping

booc0mtaco:
@jinlee93 on the chromatic diffs, b/c the popout menu is not attached to the directly-adjacent DOM of the breadcrumb menu, it is clipping the menu from the snapshot. We can add a decorator which includes some bottom padding to show the menu

jinlee93:
> @jinlee93 on the chromatic diffs, b/c the popout menu is not attached to the directly-adjacent DOM of the breadcrumb menu, it is clipping the menu from the snapshot. We can add a decorator which includes some bottom padding to show the menu

Looks like that was it. Will add some padding to ensure chromatic captures this... Think that resolves the need for a jest test since we're not trying to test the Menu functionality itself @booc0mtaco

src/components/Breadcrumbs/Breadcrumbs.stories.tsx
jinlee93: Got it snapped with the decorator padding

booc0mtaco (approved):
Nice work 🚢

PR #1648, opened 2023-06-03
fix: allow all valid input types

ahuth:
### Summary:

Allow all valid input type attributes.

Needed so the type `conform.input` returns matches EDS's. Without this, we run into type errors with the latest version of React's types -> https://github.com/chanzuckerberg/edu-stack/pull/129/files#r1215506643

### Test Plan:

- [x] Manually tested my changes, but I want to keep the details secret


andremalan (approved):

jinlee93 (approved):

ahuth:
I'm guessing the React type is new, so previously we didn't have something shared/general to use. Haven't confirmed that, though.

PR #1651, opened 2023-06-05
chore: add test command to update omlet on release

booc0mtaco:
### Summary:

- add the config now, in prep for doing a release later this week
- use the newly-added token to test against `init`'d analysis

### Test Plan:

- n/a

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,4 +42,10 @@ jobs:
       - name: Publish package to the Github registry
         run: npm publish --access public
         env:
-          NODE_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
+          NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      
+      - name: Update Omlet 🍳
+        run: npx omlet analyze
+        env:
+          OMLET_TOKEN: ${{ secrets.OMLET_TOKEN }}
+      


timzchang (approved):
🍳

.github/workflows/release.yml
```diff
@@ -42,4 +42,10 @@ jobs:
       - name: Publish package to the Github registry
         run: npm publish --access public
         env:
-          NODE_AUTH_TOKEN: ${{secrets.GITHUB_TOKEN}}
+          NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      
+      - name: Update Omlet 🍳
+        run: npx omlet analyze
+        env:
+          OMLET_TOKEN: ${{ secrets.OMLET_TOKEN }}
```
jeremiah-clothier, line 50: nit: because managing repository secrets is an extra step and then they usually expire

Wondering if modifying the default permissions granted to `GITHUB_TOKEN` is any better. See: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

booc0mtaco: > nit: because managing repository secrets is an extra step and then they usually expire
> 
> Wondering if modifying the default permissions granted to `GITHUB_TOKEN` is any better. See: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

@jeremiah-clothier I think this is probably worth investigating. Right now, this integration is in an evaluation state, just to verify its function. The token used here is generated from Omlet, so it might have different expiry behavior than our other tokens. I'm going to put an item into the epic to follow up on this once we get full validation

PR #1660, opened 2023-06-15
fix: make menu, popover, select ssr friendly

jinlee93:
### Summary:
- wraps `createportal` used in the Menu, Popover, and Select components in a conditional checking for `window`
- tested working with v12.2.0-alpha.2 release on running edu-stack locally
- tried wrapping `document.body` in `useEffect` but still got ssr errors

### Test Plan:

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -159,7 +159,7 @@ const MenuItems = (props: MenuItemsProps) => {
     ...props,
     ...popperAttributes,
   };
-  if (typeof window === 'object') {
+  if (typeof document !== 'undefined') {
     return (
       <>
         {createPortal(<HeadlessMenu.Items {...optionProps} />, document.body)}
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -159,10 +159,14 @@ const MenuItems = (props: MenuItemsProps) => {
     ...props,
     ...popperAttributes,
   };
-
-  return (
-    <>{createPortal(<HeadlessMenu.Items {...optionProps} />, document.body)}</>
-  );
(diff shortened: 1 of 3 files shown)


src/components/Menu/Menu.tsx
```diff
@@ -159,10 +159,14 @@ const MenuItems = (props: MenuItemsProps) => {
...
-    <>{createPortal(<HeadlessMenu.Items {...optionProps} />, document.body)}</>
-  );
+  if (typeof window === 'object') {
+    return (
+      <>
+        {createPortal(<HeadlessMenu.Items {...optionProps} />, document.body)}
+      </>
+    );
+  }
+  return null;
```
jinlee93, line 169: Tried something like:

```tsx
  const [container] = useState(() => {
    const el = document.createElement('div')
    return el
  })

  useEffect(() => {
    document.body.appendChild(container);
    return () => {
      document.body.removeChild(container)
    }
  }, [])

  return (
    <>{createPortal(<HeadlessMenu.Items {...optionProps} />, container)}</>
  );
```

but still got error along the lines of document is not defined

jinlee93: Rather than finagling with this more I felt like checking for `window` was simpler and easier to understand

jeremiah-clothier (approved):
This approach seems reasonable and is along the lines of what the remix docs call out https://remix.run/docs/en/1.17.1/pages/gotchas#typeof-window-checks

jinlee93:
> This approach seems reasonable and is along the lines of what the remix docs call out https://remix.run/docs/en/1.17.1/pages/gotchas#typeof-window-checks

Used `typeof document !== 'undefined'` as recommended by docs, and tested working with alpha v12.2.0-alpha.3

PR #1663, opened 2023-06-22
feat(breadcrumbs): allow custom separators

jinlee93:
[EDS-922]
### Summary:
- allows separator string to be passed to parent <Breadcrumb> wrapper for <BreadcrumbItems>
  - follows similar code pattern in Modal, Popover, etc of using React Context to pass parent prop to children components
  - no longer references uses to <BreadcrumbItems> children components and instead to <Breadcrumb.Items> which has the modified <BreadcrumbItems> with the context

image
image

smallest breakpoint no separator, so stays same:
image

### Test Plan:

  - new story with 3 viewports for visual regression testing

[EDS-922]: https://czi-tech.atlassian.net/browse/EDS-922?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.storybook/pages/ProjectOverview/ProjectOverview.tsx
+++ b/.storybook/pages/ProjectOverview/ProjectOverview.tsx
@@ -2,7 +2,6 @@ import React from 'react';
 
 import {
   Breadcrumbs,
-  BreadcrumbsItem,
   Button,
   Card,
   Grid,
@@ -34,8 +33,8 @@ export const ProjectOverview = ({ activeIndex = 0 }: Props) => {
   return (
     <PageShell className="body--alternate" mentoringIsActive>
       <Breadcrumbs className="!mb-4">
-        <BreadcrumbsItem href="#" text="My Courses" />
-        <BreadcrumbsItem href="#" text="Disciplinary Science 7" />
+        <Breadcrumbs.Item href="#" text="My Courses" />
+        <Breadcrumbs.Item href="#" text="Disciplinary Science 7" />
       </Breadcrumbs>
       <PageHeader
         headingSize="h3"
--- a/src/components/Breadcrumbs/Breadcrumbs.tsx
+++ b/src/components/Breadcrumbs/Breadcrumbs.tsx
@@ -27,11 +27,11 @@ type Props = {
    * Custom string separator between individual breadcrumbs
    * Defaults to '/'
    */
-  separator?: string;
+  separator?: '|' | '>' | '/';
 };
 
 type Context = {
-  separator?: string;
+  separator?: '|' | '>' | '/';
 };
 
 const BreadcrumbsContext = createContext<Context>({});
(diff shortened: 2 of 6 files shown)


src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,42 +1,53 @@
...
-
   /**
    * HTML id for the component
    */
   id?: string;
+  /**
+   * Custom string separator between individual breadcrumbs
+   * Defaults to '/'
+   */
+  separator?: string;
```
jinlee93, line 30: Hesitant to allow separators other than strings right now

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -152,9 +166,16 @@ const flattenBreadcrumbsItems = (children: React.ReactNode) => {
     },
   );
   if (process.env.NODE_ENV !== 'production' && shouldThrowError) {
-    throw 'Only <Breadcrumbs.Item>, <BreadcrumbsItem>, or React.Fragment of aforementioned components allowed';
+    throw 'Only <Breadcrumbs.Item> or React.Fragment of aforementioned components allowed';
```
booc0mtaco, line 169: We should use an `Error` object when throwing, for consistency. I think all the other `throw`s do. Or should...

.storybook/pages/ProjectOverview/ProjectOverview.tsx
```diff
@@ -34,8 +33,8 @@ export const ProjectOverview = ({ activeIndex = 0 }: Props) => {
   return (
     <PageShell className="body--alternate" mentoringIsActive>
       <Breadcrumbs className="!mb-4">
-        <BreadcrumbsItem href="#" text="My Courses" />
-        <BreadcrumbsItem href="#" text="Disciplinary Science 7" />
+        <Breadcrumbs.Item href="#" text="My Courses" />
```
booc0mtaco, line 36: ++

src/components/Breadcrumbs/Breadcrumbs.tsx
booc0mtaco: Fair. In fact, I think we can make separator a subset of possible strings: 

* "|"
* "/"
* ">"

Specifying exact string values lets us also constrain the length without having to write `.length` checks in code.

In the future, also maybe icon? can cross that bridge when we get to it. This would be a good followup with design... figma lets you create lists in the UI that could let designers select from a set of values too, so good parity

jinlee93: I'm down to make it a subset of a few possible strings and not having icon for now. Will follow up with design

jinlee93: Followed up with design, this subset looks good

booc0mtaco (approved):

src/components/Breadcrumbs/Breadcrumbs.tsx
jinlee93: Changed to throwing `new Error`

PR #1664, opened 2023-06-22
feat(typography): support typography theming

booc0mtaco:
### Summary:

Support the ability to theme typography presets that match each consuming project. This has the added benefit of also decoupling us from having to use the postCSS mixin stuff. More functionality with less complexity.

- use font property instead of fixed mixins and property templates
- add in tokens for each preset and usage
- ensure mapping and snapshots match without change

### Test Plan:


src/design-tokens/tier-1-definitions/typography.json
```diff
@@ -59,6 +59,139 @@
...
+            "value": "var(--eds-font-weight-medium) var(--eds-font-size-40)/1.2 var(--eds-font-family-primary)"
+          },
+          "light": {
+            "value": "var(--eds-font-weight-light) var(--eds-font-size-40)/1.2 var(--eds-font-family-primary)"
+          },
+          "bold": {
+            "value": "var(--eds-font-weight-bold) var(--eds-font-size-40)/1.2 var(--eds-font-family-primary)"
+          },
+          "mobile": {
+            "value": "var(--eds-font-weight-bold) var(--eds-font-size-32)/1.25 var(--eds-font-family-primary)"
```
jinlee93, line 76: Now that this is separated, the component itself would have to handle breaks?

booc0mtaco: correct. I marked a TODO for this, but I think that's probably the better way to do this anyway. Tokens shouldn't have logic like that IMO

jinlee93: Makes sense to me

booc0mtaco:
Changes to note in this PR: 

- removing legacy tokens used in pages and recipes (removing `body-XX-bold` as they are legacy only)

.storybook/pages/CoursePlannerEdit/ProjectCard/ProjectCard.tsx
```diff
@@ -110,7 +110,7 @@ export const ProjectCard = ({
         <Heading
           as={headingAs}
           className={styles['project-card__title']}
-          size="body-sm"
+          size="title-xs"
```
booc0mtaco, line 113: `body-sm` may have changed in design while this was under way, so picked a near neighbor for this demo.

.storybook/pages/StudentRefinement/StudentRefinement.module.css
```diff
@@ -82,7 +82,8 @@
  * The title communicates card organization.
  */
 .data-summary-card__title {
-  @mixin eds-theme-typography-overline-sm;
+  font: var(--eds-theme-typography-overline-sm);
+  text-transform: uppercase;
```
booc0mtaco, line 86: There are a few examples of adding in `text-transform` to some cases, because `font` only supports CSS 2.1 values which doesn't include `all-caps` sadly

booc0mtaco:
This fiddle shows how mixing `:root` and other classes of the same specificity doesn't apply font family or other overrides as you might expect. In this example, the family used in the shortcut `font:` rule value doesn't apply even tho chrome indicates it should

https://jsfiddle.net/booc0mtaco/q1h6s5dk/

.storybook/pages/AlongDemo/AlongDemo.module.css
booc0mtaco, line 1: Removed the along demo as along itself is now going to use EDS, has the updated theme, and keeping this as-is would show:

- an outdated theme 
- improper overrides for font family values (manifesting from a limitation in how storybook fetches per-component styles and variables)
- page structure that may change in the future

This avoids us having to keep it in sync with any product code/branding.

.storybook/components/DesignTokens/Tier2/TypographyUsage.module.css
```diff
@@ -5,137 +5,114 @@
...
 .typography-usage--overline {
-  @mixin eds-typography-preset-011-bold;
+  font: var(--eds-typography-preset-011);
+  text-transform: uppercase;
 }
 
 .typography-usage--overline-sm {
-  @mixin eds-typography-preset-012-bold;
+  font: var(--eds-typography-preset-012);
+  text-transform: uppercase;
```
jinlee93, line 100: Are we removing the larger weight treatment on the overline tokens?

.storybook/components/Docs/Guildelines/Typography.stories.mdx
```diff
@@ -6,50 +6,79 @@ Typography in EDS follows [the same structure as design tokens](./?path=/story/d
...
+        "headline": {
+          "lg": {
+            "value": "var(--eds-typography-preset-001-bold)"
+          }
+        }
+      }
+    }
+  }
+}
+```
```
jinlee93, line 46: Nice updates

.storybook/data/tokens.json
```diff
@@ -119,6 +119,41 @@
...
+  "eds-typography-preset-009": "var(--eds-font-weight-medium) var(--eds-font-size-12)/1.333333333 var(--eds-font-family-primary)",
+  "eds-typography-preset-009-bold": "var(--eds-font-weight-bold) var(--eds-font-size-12)/1.333333333 var(--eds-font-family-primary)",
+  "eds-typography-preset-010": "var(--eds-font-weight-medium) var(--eds-font-size-11)/1.272727272 var(--eds-font-family-primary)",
+  "eds-typography-preset-010-bold": "var(--eds-font-weight-bold) var(--eds-font-size-11)/1.272727272 var(--eds-font-family-primary)",
+  "eds-typography-preset-011": "var(--eds-font-weight-medium) var(--eds-font-size-12)/1.333333333 var(--eds-font-family-primary)",
+  "eds-typography-preset-011-bold": "var(--eds-font-weight-medium) var(--eds-font-size-12)/1.333333333 var(--eds-font-family-primary)",
+  "eds-typography-preset-012": "var(--eds-font-weight-medium) var(--eds-font-size-11)/1.272727272 var(--eds-font-family-primary)",
+  "eds-typography-preset-012-bold": "var(--eds-font-weight-medium) var(--eds-font-size-11)/1.272727272 var(--eds-font-family-primary)",
```
jinlee93, line 156: Although there are more tokens, I am pretty happy this exists now since it will ACTUALLy align our typography tokens closer to Figma and the designer's idea of type tokens

.storybook/pages/AlongDemo/AlongDemo.module.css
jinlee93: Think the wireframe page should be enough now?

.storybook/pages/CoursePlannerEdit/ProjectCard/ProjectCard.tsx
jinlee93: Makes sense and `body-sm` prob being removed as part of Heading ticket anyways

.storybook/components/DesignTokens/Tier2/TypographyUsage.module.css
jinlee93: Actually, food for thought, perhaps overline can be a `Text` component variant only and not a token itself since it also has `text-transform`

booc0mtaco: This is the confusing thing... as I was working thru things, i was just applying the values in our designs, and it says we should have had this?

Screenshot 2023-06-28 at 18 13 10

I don't know if this had been out of sync for a while, or was a recent change (pls add some kinda diffing to Figma)

jinlee93 (approved):
I think it looks good, great work! I have some thoughts about overline, but don't think it should be blocking

.storybook/components/DesignTokens/Tier2/TypographyUsage.module.css
jinlee93: Think it's been out of sync for a while and never got noticed since this token prob has low usage... could consider even removing these

.storybook/pages/AlongDemo/AlongDemo.module.css
booc0mtaco: yah. eventually we will invest in a proper wireframe theme (possibly as the default), and use this demo to show how to prototype with EDS

jinlee93: u m b r e l l a b r a n d

.storybook/components/DesignTokens/Tier2/TypographyUsage.module.css
booc0mtaco: @jinlee93 Ohh, yeah, all usages could be run thru `Text` for clarity and simplicity. This would give us a nice way to avoid exposing this `text-transform` thing.

Then we can greatly simplify these typography token pages and probably remove the css modules for those

PR #1669, opened 2023-06-27
feat(icons): make icons rounded outline and add send, mail

jinlee93:
[EDS-1043]

### Summary:
- for Along we're changing material icons from sharp to rounded outline
- also adding send and mail icons since these have utility non specific to Along and we already have similar icons such as print
- need one sync with design to finalize some q's

### Test Plan:
- icon set matches mocks
- components with swapped icons look fine in snaps

  - updated snaps

[EDS-1043]: https://czi-tech.atlassian.net/browse/EDS-1043?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


booc0mtaco (approved):
Looks like a straightforward swap. Interesting and expected that the rounded icons have more complex paths given all the radii at the edges.

PR #1671, opened 2023-07-01
feat(Typography): force sync and simplify typography

booc0mtaco:
### Summary:

- Add in ALL the missing and out of sync typography stuff
- Delete all the unnecessary and manual typography code and replace with functions
- Simplify documentation around tier 1 and 2 tokens
- Delete now unused typography page styles
- Create utility function for usage pages
- Other useful simplifications

### Test Plan:

  - verified snapshots applied to proper components, and checked with design on values for each. Synced with figma


.storybook/components/DesignTokens/Tier2/TypographyUsage.jsx
```diff
@@ -1,89 +1,40 @@
...
-      return (
-        <TokenSpecimen
-          comment={`Ref: preset-${preset}`}
-          name={name}
-          specimenClassName={styles[`typography-usage--${size}`]}
-          value={`${fontSize} / ${lineHeight}`}
-          variant="typography-body"
-        />
+    const values = {};
+    forEach(at(usages, 'eds.theme.typography')[0], (_, key) => {
```
booc0mtaco, line 12: Given the values in the typography file at the path `eds.theme.typography`, create a flat map of each one such that stuff like `{ a: { b: 'c' } }` becomes `{ a-b: 'c' }` at any nesting level.

.storybook/components/DesignTokens/Tier2/TypographyUsage.jsx
```diff
@@ -1,89 +1,40 @@
...
+      values[key] = flatten(
+        at(usages, `eds.theme.typography.${key}`)[0],
+        `eds-theme-typography-${key}-`,
       );
-    };
-
-    return (
-      <div>
```
booc0mtaco, line 29: No more need to manually map the tier two values to the presets. This is the value it has now.

.storybook/components/DesignTokens/Tier2/TypographyUsage.jsx
```diff
@@ -1,89 +1,40 @@
...
+    const contents = Object.keys(values).map((section) => {
+      return (
+        <Section key={section} title={capitalize(section)}>
           <Grid>
-            {renderTypeToken('overline', '011')}
-            {renderTypeToken('overline-sm', '012')}
+            {Object.keys(values[section]).map((usage) => {
+              return (
+                <TokenSpecimen
+                  inlineStyles={{ font: values[section][usage] }}
```
booc0mtaco, line 25: using inline styles avoids the need to map custom class names to the declaration

.storybook/data/tokens.json
```diff
@@ -103,13 +103,14 @@
   "eds-size-1-and-half": "0.75rem",
   "eds-size-2-and-half": "1.25rem",
   "eds-font-family-primary": "'Graphik', sans-serif",
-  "eds-font-family-secondary": "'Open Sans', sans-serif",
+  "eds-font-family-secondary": "'Mint Grotesk V1.1', sans-serif",
```
booc0mtaco, line 106: Confirmed with design that we don't want to use Open Sans anywhere.

.storybook/util/flattenToken.js
```diff
@@ -0,0 +1,27 @@
+import { forEach, isObject } from 'lodash';
+
+export default function flatten(obj, themePrefix = '') {
```
booc0mtaco, line 3: TODO: document and add tests

Given an object of an arbitrary depth, flatten the names using dashes (configure?), and add to any provided suffix. Append the suffix as you recurse.

When it reaches an `@`, treat that as the root name.

package.json
```diff
@@ -89,7 +89,9 @@
     "@headlessui/react": "^1.7.15",
     "@popperjs/core": "^2.11.8",
     "@tippyjs/react": "^4.2.6",
+    "@types/lodash": "^4.14.195",
     "clsx": "^1.2.1",
+    "lodash": "^4.17.21",
```
booc0mtaco, line 94: TODO: remove lodash.debounce in favor of the full-fat lodash import

ahuth (approved):
Nice 🤘

booc0mtaco:
Adding in missing variants and weights for Mint Grotesk, since i noticed they were missing. Parity exists for the primary and secondary fonts now

PR #1675, opened 2023-07-05
feat(focus): create t1 outline and t3 focus tokens

jinlee93:
[EDS-1029]

### Summary:
- create t3 tokens for focus outline
- since no outline t1 tokens, that ramp is also added here (syncec with design)
- checkbox custom offset is removed

### Test Plan:

[EDS-1029]: https://czi-tech.atlassian.net/browse/EDS-1029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/design-tokens/tier-3-component/focus-ring.json
```diff
@@ -10,6 +10,18 @@
             "value": "{eds.color.neutral.white}"
           }
         }
+      },
```
booc0mtaco, line 13: for later: perhaps these should be part of the forms.json, as it will apply to all form elements (and maybe form should actually be the individual form elements?) Just stirring the pot but food for thought.

src/design-tokens/tier-3-component/focus-ring.json
```diff
@@ -10,6 +10,18 @@
...
+      "box": {
+        "focus-ring": {
+          "outline": {
+            "width": {
+              "value": "{eds.outline.width.sm}"
```
booc0mtaco, line 18: by using a static definition with the `{...}` syntax, if someone redefines the tier one values, they would also need to update these two tokens. If we use the css `var` syntax, overriding the base value(s) will propagate to these due to cascading

src/design-tokens/mixins.css
```diff
@@ -65,24 +65,21 @@
...
-  outline-offset: 2px;
+  outline: var(--eds-theme-box-focus-ring-outline-width) solid
+    var(--eds-theme-color-focus-ring);
+  outline-offset: var(--eds-theme-box-focus-ring-outline-offset);
 }
 
 @define-mixin focusInside {
   @mixin focus;
-  outline-offset: -2px;
+  outline-offset: calc(-1 * var(--eds-theme-box-focus-ring-outline-offset));
```
booc0mtaco, line 75: ++

src/design-tokens/mixins.css
```diff
@@ -65,24 +65,21 @@
...
 }
 
 @define-mixin focusInverted {
   @mixin focus;
   outline-color: var(--eds-theme-color-focus-ring-inverted);
 }
 
-@define-mixin focusTransparent {
```
booc0mtaco, line 82: ++

src/design-tokens/tier-1-definitions/outline.json
```diff
@@ -0,0 +1,17 @@
+{
```
booc0mtaco, line 1: box.json?

.storybook/data/tokens.json
```diff
@@ -82,6 +82,9 @@
   "eds-l-max-width": "71.25rem",
   "eds-l-sidebar-width": "13.5rem",
   "eds-l-linelength-width": "36rem",
+  "eds-outline-width-sm": "1px",
+  "eds-outline-width-md": "2px",
+  "eds-outline-width-lg": "4px",
```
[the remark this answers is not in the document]
jinlee93: These are built off the t1 tokens page and based them similarly to border which is on lines 8-11 of this file. Also [referenced](https://chanzuckerberg.github.io/edu-design-system/?path=/story/documentation-guidelines-tokens--page#borders-a-nametier-1-bordersa) to border doesn't have `-box` which I think might be t3 specific for us? I can add `-box` but then should also be added to t1 `border`

src/design-tokens/mixins.css
jinlee93: Yea removed this one since it wasn't being used anywhere idk why we had this

src/design-tokens/tier-1-definitions/outline.json
jinlee93: Like comment above, we can consider consolidating this with `borders` but would be a separate issue

src/design-tokens/tier-3-component/focus-ring.json
jinlee93: Yea there's some discussion to be had around nomenclature around tokens and possibly condensing the amount of possibilities

src/design-tokens/tier-3-component/focus-ring.json
jinlee93: Will switch to the css custom property

booc0mtaco: we can put that on the table for consideration. I think we will probably want to use references in many/most cases, but there may be unthemed things where we should just use the static references. That can work as a brief heuristic for now until we consider some pros/cons.

src/design-tokens/mixins.css
booc0mtaco: if focus were implemented with borders or something that changes layout, you'd want a transparent in place, so that stuff doesn't slip around (and cause a dom repaint). it's possible an older version of this used to not have `outline` but they didn't clean that stuff out.

.storybook/data/tokens.json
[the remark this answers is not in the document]
booc0mtaco: Ah good point. let's save that for some later thinking

booc0mtaco (approved):

PR #1679, opened 2023-07-10
fix(Avatar): add responsive image handling to component

booc0mtaco:
### Summary:

* EDS storybook and component structure implements a global responsive style for `width: 100%` which doesn't get sent along with the component
* Add this to the component using semantic class name

### Test Plan:

  - update the snapshots


jinlee93 (approved):
Makes sense to me

PR #1681, opened 2023-07-12
feat(timelinenav)!: remove top-level sub-component

jinlee93:
### Summary:
- merge TimelineNavPanel into TimelineNav directly

### Test Plan:

  - No snap changes but also LP specific so low risk


src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -51,18 +49,79 @@ export interface Props {
...
-export interface TimelineNavItem {
-  props: {
-    'aria-label': string;
-    children: ReactNode;
-    completed?: boolean;
-    title?: string;
-    titleAfter?: ReactNode;
-    variant?: TimelineNavPanelVariant;
-  };
-}
```
jinlee93, line 65: Never actually used 🤷

src/components/TimelineNav/TimelineNav.tsx
```diff
@@ -51,18 +49,79 @@ export interface Props {
...
+type TimelineNavPanelVariant =
+  | 'bullet'
+  | 'complete'
+  | 'error'
+  | 'incomplete'
+  | 'number'
+  | 'success'
+  | 'warning'
+  | undefined;
```
booc0mtaco, line 62: This is a little funny; if a prop is set to `undefined` react will ignore it. Adding undefined to a type is a bit ambiguous. Instead it's better to use `?` for marking that.

I'm guessing this is inherited tho, so not requesting a change.

booc0mtaco (approved):
LGTM

src/components/TimelineNav/TimelineNav.tsx
jinlee93: Will change to `?` as using `undefined` is dumb 👍

PR #1682, opened 2023-07-12
refactor: set token sizes to be relative to base unit

booc0mtaco:
### Summary:

Since all the discrete values are based on the parent value token, set the others based on that so two things are possible: overriding all the token values is MUCH easier, and to prevent a specific class of bugs with calculating the other values.

### Test Plan:

  - review snapshots and make sure they don't change


jinlee93:
I like that this is a step in making all the size values relative to one source of truth but also half wondering / need to be reminded what the purpose of these size tokens are?

booc0mtaco:
> I like that this is a step in making all the size values relative to one source of truth but also half wondering / need to be reminded what the purpose of these size tokens are?

Normally you'd expect them to be used for internal or external spacing, sorta like token based utility classes. In EDS, not sure if we are doing that consistently. For us, we should use these in components' margins and paddings instead of hard-coded values), but not in recipes or pages (use Tailwind there). 

Another approach would be to use just the base value, and use `calc()` for all the rest so there are fewer total tokens. We can come back to that one :)

jinlee93:
The resulting Storybook [Size Tokens page](https://61313967cde49b003ae2a860-ltlmpkdfnx.chromatic.com/?path=/story/design-tokens-tier-1-definitions--sizes) looks like :
image

I wonder if there's a way for it to show the calculated values as well or at least put `--eds-size-base-unit` on top so that readers can reference it faster

booc0mtaco:
@jinlee93 that could work. We do define the base value in the list, and could make that one first. The ideal is that people don't have to even know what all the sizes really mean, but knowing the base is pretty useful

booc0mtaco:
Hm, something about this is messing up how tippy works :( https://61313967cde49b003ae2a860-ltlmpkdfnx.chromatic.com/?path=/story/components-slider--tooltip

jinlee93:
> Hm, something about this is messing up how tippy works :( https://61313967cde49b003ae2a860-ltlmpkdfnx.chromatic.com/?path=/story/components-slider--tooltip

Think this is due to how Tippy offset is applied via a callback in https://github.com/chanzuckerberg/edu-design-system/blob/4d32194/src/components/Slider/Slider.tsx#L166-L180 and so the offset has to be calculated javascript-ly. 

```
parseFloat(
  computedBodyStyles.getPropertyValue(
    '--eds-theme-size-slider-thumb',
  ),
)
```
now returns `NaN` instead of `24` and hence breaks the Tooltip.

Some possible options:

1. Parse the calc correctly
2. Try to place the tooltip via CSS
3. Use another library for Tippy (I know this is v v stretch but this is also a backlog issue since it's currently not SSR friendly)
4. Punt this

booc0mtaco:
@jinlee93 added clarity on the base size (and what these sizes actually are lol), and fixed the issue with the slider tooltip by adding a local proxy component (since the thumb is hidden by shadow DOM :S)

src/components/Slider/Slider.tsx
```diff
@@ -220,6 +216,7 @@ export const Slider = ({
           ))}
         </div>
       )}
+      <div className={styles['slider__thumb-proxy']} ref={proxyRef}></div>
```
jinlee93, line 219: nit: can you also comment here about why this element has to exist?

jinlee93 (approved):
Sweet, thanks for dealing with the Slider as well.
Looks good pending one nit

PR #1684, opened 2023-07-12
feat!: Merge and remove top-level components

booc0mtaco:
### Summary:

This feature branch contains the remaining work to combine all subcomponents into their parents when the child can ONLY be used within the parent. This is a breaking changes, so we want to annotate what any consumers should do to mitigate. These steps can be part of the release notes or the EDS Wiki (12-to-13 page).

We can merge this in using a merge commit instead of squashing, as the contents of this PR are a bunch of separate PRs that mention specific components and decisions.

### Test Plan:

- n/a


timzchang (approved):
🚀

jeremiah-clothier:
😮 So much work from @booc0mtaco and @jinlee93 to make this happen

PR #1689, opened 2023-07-14
feat(Modal)!: remove top-level sub-component(s)

booc0mtaco:
### Summary:

- merge in the code and styles, and remove the exports

### Test Plan:

  - verify there are no snapshot changes


src/components/Modal/Modal.tsx
```diff
@@ -40,13 +35,14 @@ type ModalContentProps = {
...
    *
-   * @example
+   * ```
    * const inputFieldRef = useRef();
    *
    * <Modal initialFocus={inputFieldRef}>
    *   ...
    *   <InputField ref={inputFieldRef} />
    * </Modal>
+   * ```
```
jinlee93, line 45: nice

jinlee93 (approved):

src/components/Modal/Modal.tsx
booc0mtaco: @anniehu4 Instead of `@example` we can use code blocks in prop doc blocks , which render in storybook. This should fix that glitch we noticed earlier this week

src/components/Modal/Modal.tsx
```diff
@@ -86,6 +82,96 @@ type ModalProps = ModalContentProps & {
...
 
+type ModalTitleProps = Omit<ExtractProps<typeof Heading>, 'size'> & {
+  /**
+   * Text for the modal title.
+   */
+  children: ReactNode;
+  /**
+   * CSS class names that can be appended to the component.
+   */
+  className?: string;
```
booc0mtaco, line 93: There are gonna be some opportunities to consolidate and share common types once we get all of these reduced down

PR #1690, opened 2023-07-15
feat(radio)!: remove top level subcomponents

jinlee93:
[EFI-1231]

### Summary:
- removes radio subcomponents

### Test Plan:

  - no visual regressions

[EFI-1231]: https://czi-tech.atlassian.net/browse/EFI-1231?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Radio/Radio.module.css
```diff
@@ -9,3 +9,88 @@
...
+}
+/**
+ * The visually hidden input element for the radio. The visual radio is provided by an svg element.
+ */
+.radio__input {
+  /* Removes default margins placed by browser for radioes. */
+  margin: 0;
+  /* Remove the radio from the page flow, positioning it on top of the SVG. */
+  position: absolute;
+  left: 0.25rem;
```
booc0mtaco, line 32: nit: this should be `--eds-size-half` i think but I'm guessing this was just brought over from the other file

src/components/Radio/Radio.module.css
```diff
@@ -9,3 +9,88 @@
...
+/**
+ * Radio label size variants.
+ * Used for centering the label with the radio input button.
+ */
+.radio__label--md {
+  top: 0.125rem;
+}
+.radio__label--lg {
+  top: 0.0625rem;
+}
```
booc0mtaco, line 96: Not a fan of this, but again this was just brought over :)

booc0mtaco: 0.125rem corresponds to `--eds-size-quarter` which does exist, but `--eds-size-eighth` or what have you does not...and probably shouldn't. the sizes tokens are kinda weird. I wonder if there are other cases of such a small relative spacing that would indicate we should add this token?

jinlee93: image
There are a some instances where `0.0625rem` is used; still tho `eighth` is kind of small?

booc0mtaco: very small... one pixel in fact :/ 

This seems to exist to vertically center the text with the radio button. there may be a way to do the same via flex and avoid this kinda thing altogether. So as not to over-complicate this PR, we can make a comment on why such a precise value is being used, then leave it be for now

jinlee93: That makes sense to me, will do

PR #1696, opened 2023-07-17
feat(HorizontalStepper)!: remove top-level sub-component(s)

booc0mtaco:
### Summary:

- rename child component to comport with naming rules better

### Test Plan:

  - no unexpected snapshot changes


src/components/HorizontalStepper/HorizontalStepper.stories.tsx
```diff
@@ -87,13 +86,17 @@ export const Interactive: StoryObj<Args> = {
 export const HorizontalSteps: StoryObj<Args> = {
   render: () => (
     <ol className="p-0">
-      <HorizontalStep
+      <HorizontalStepper.Step
```
booc0mtaco, line 89: Not thrilled with the component name here. `Horizontal` is actually a variant embedded into the name. There could be a vertical one at some point, and switching to it could be a responsive behavior even on the horizontal thing. In fact, `TimelineNav` and this thing have a few items in common...

Renaming for now, but we should revisit this component with design at some point

jinlee93 (approved):

PR #1697, opened 2023-07-18
feat(dragdrop)!: remove top level subcomponents

jinlee93:
[EFI-1225]

### Summary:
- removes top level drag and drop components
- prob needs some revisiting to see if too LP specific and how we can supply drag and drop more generally

### Test Plan:

  -  no visual regression

[EFI-1225]: https://czi-tech.atlassian.net/browse/EFI-1225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


booc0mtaco:
@jinlee93 IMO dragdrop is an outlier b/c most of what it does is behavior, with a little bit of EDS reuse. What we'd likely want to do is leave this up to product teams, or offer a separate frontend lib which has EDS as a peer dependency, and any specific wrapping to make building an interface like this easier

I'm pretty sure DragDrop doesn't appear in our Figma... but the stuff used in it does

booc0mtaco (approved):

PR #1698, opened 2023-07-18
fix: make sure all var properties exist on one line

booc0mtaco:
### Summary:

- when a `var()` property spans multiple lines, it will insert that directly into the CSS, which won't read the token
- for some components (mainly `ClickableStyle`) this caused bugs in the values used in the styling
- update the code to remove these, and add a few `/* prettier-ignore */` declarations to lines that are too long to fit into the prettier rules

### Test Plan:

  - Validate any snapshot changes

### DIFF
--- a/.storybook/pages/WireframeDemo/GlobalStyles.module.css
+++ b/.storybook/pages/WireframeDemo/GlobalStyles.module.css
@@ -45,9 +45,7 @@
   --eds-font-family-primary: var(--wireframe-font-family-primary);
   --eds-font-family-secondary: var(--wireframe-font-family-secondary);
   --eds-theme-color-background-brand-primary-default: var(--wireframe-white);
-  --eds-theme-color-background-brand-primary-strong-hover: var(
-    --wireframe-gray-e
-  );
+  --eds-theme-color-background-brand-primary-strong-hover: var(--wireframe-gray-e);
   --eds-theme-color-background-brand-primary-strong: var(--wireframe-gray-b);
   --eds-theme-color-background-disabled: var(--wireframe-gray-b);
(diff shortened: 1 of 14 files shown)


booc0mtaco:
Screenshot 2023-07-18 at 14 06 38

example showing how a variable doesn't get selected normally for `background-color`

booc0mtaco:
This actually appears to be more of a problem with the browser chrome itself. Chrome doesn't show that a value is being set, and doesn't allow the color token to be clicked. Safari shows a variable is available with the "=" icon, but doesn't show the var inline. Firefox clearly shows the text token name and applied value with or without spaces

Here's how it looks in safari, chrome, and firefox. [fiddle](https://jsfiddle.net/booc0mtaco/amy6rvs9/)
Screenshot 2023-07-18 at 15 27 49
Screenshot 2023-07-18 at 15 28 29
Screenshot 2023-07-18 at 15 28 08

Will still add this change, as it makes token values more easily debugged (when there are no spaces, the token names are clickable)

jinlee93 (approved):
Thanks for the hotfix, seems theres only 3 `prettier-ignore` added here... how did the extra lines / whitespace start to begin with?

booc0mtaco:
Filed a chromium bug :3 https://bugs.chromium.org/p/chromium/issues/detail?id=1465958

booc0mtaco:
> Thanks for the hotfix, seems theres only 3 `prettier-ignore` added here... how did the extra lines / whitespace start to begin with?

I'm actually not sure why prettier allows what it allows in this case. It's possible something in prettier may have changed, but i didn't see anything listed in [the changelog](https://github.com/prettier/prettier/blob/main/CHANGELOG.md)

PR #1701, opened 2023-07-19
refactor(textareafield): use inputlabel and re-export as subcomp

jinlee93:
[EFI-1250]

### Summary:
- moves `TextArea` subcomponent into `TextareaField`
- uses InputLabel component over Label component for TextareaField be in line with Checkbox and Radio and InputField
- aligns closer to the mocks for font and text color tokens
- adds disabled color for InputField

### Test Plan:

  - chromatic visual changes only in disabled states for TextareaField and InputField

[EFI-1250]: https://czi-tech.atlassian.net/browse/EFI-1250?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/TextareaField/TextareaField.tsx
```diff
@@ -114,9 +178,13 @@ export const TextareaField: TextareaFieldType = forwardRef(
...
           <div className={overlineClassName}>
-            {label && <Label htmlFor={idVar} text={label} />}
+            {label && (
+              <InputLabel className={labelClassName} htmlFor={idVar}>
+                {label}
+              </InputLabel>
+            )}
             {required && (
-              <Text as="p" size="sm">
+              <Text as="p" className={requiredTextClassName} size="sm">
```
booc0mtaco, line 187: it's interesting that we choose the `P` tag here instead of a relatively unstyled `SPAN`. do you remember anything about this choice?

booc0mtaco (approved):
Looks good. I think we are way more internally consistent with these changes in place now.

src/components/TextareaField/TextareaField.tsx
jinlee93: Uhhh nothing that I recall 😅

PR #1703, opened 2023-07-19
feat(Grid)!: remove top-level sub-component(s)

booc0mtaco:
### Summary:

`GridItem` => `Grid.Item`

### Test Plan:

  - verify snapshots do not change


jinlee93 (approved):
In the near-ish future we should try to deprecate Grid since there are similar grid / width TW utility classes that allow for this kind of functionality and should follow our philosophy to how we deal with `Layout`

booc0mtaco:
> In the near-ish future we should try to deprecate Grid since there are similar grid / width TW utility classes that allow for this kind of functionality and should follow our philosophy to how we deal with `Layout`

@jinlee93 Agreed. In fact, since it's not in figma currently, and has the same kinda properties and usage as `Layout`, I'll just mark them right now

.storybook/components/DesignTokens/Tier2/Forms.jsx
```diff
@@ -8,42 +8,36 @@ export class Tier2Forms extends Component {
...
-                    borderStyle: 'solid',
-                    borderColor: 'black',
-                  }}
-                  name={listItem.name}
-                  value={listItem.value}
-                />
-              </GridItem>
-            );
-          })}
+          {filterTokens(`eds-theme-form-border-width`).map((listItem) => (
```
booc0mtaco, line 11: Cleaned up this one b/c the formatting was getting garbled by prettier in some weird way. All of these could be converted into MDX files instead of having the class component wrapper and storybook file in the end

src/components/Grid/Grid.tsx
```diff
@@ -38,11 +37,28 @@ export interface Props {
...
+  /**
+   * CSS class names that can be appended to the component.
+   */
+  className?: string;
+}
+
 /**
+ * The Grid component is deprecated and will be removed in an upcoming release.
+ * Instead, please make use of utility libraries, like Tailwind CSS:
+ * * https://tailwindcss.com/docs/display#grid
```
booc0mtaco, line 54: Linking to TW docs upon deprecation, as we want to focus EDS on presentation layer, not page layout layer. (we can document guidance around page layout in zeroheight, however, and include code examples using some code sandbox software like codepen, etc.)

PR #1711, opened 2023-07-26
docs(colors)!: remove old colors and convert to input and show figma token

jinlee93:
[EDS-1063]

### Summary:
[mocks](https://www.figma.com/file/IfMWAtWqNrQprImlco4pi0/EDS-SLP-v1.3-%7C-UI-Foundations?type=design&node-id=58-4827&mode=design&t=Yn6yGBwsW0TaECrt-4)
impetus:
- we will soon be pointing ZH colors to the Storybook page as it stays up to date more automatically than Figma styles
execution:
- remove t1 deprecated and legacy colors
  - these are no longer in the mocks
- reordered color list to match mocks order
- updated storybook component `ColorList`:
  - show figma token in the colors page so designers can quick glance it
  - convert the large color swatch to `<input>` so it can be interact-able with browser functionality
upcoming:
- t2 and t3 to use `ColorList` component to show the figma token and the tailwind class
### Test Plan:

  - looks similar enough to current storybook t1 colors page
  - shows the figma token
  - color swatch is interact-able 

image

[EDS-1063]: https://czi-tech.atlassian.net/browse/EDS-1063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.storybook/components/ColorList/ColorList.module.css
```diff
@@ -0,0 +1,26 @@
...
+.color-list__input::-webkit-color-swatch {
+  border-color: transparent;
+}
+.color-list__input::-webkit-color-swatch-wrapper {
+  padding: 0;
+}
+
+.color-list__input::-moz-color-swatch {
+  border-color: transparent;
+}
```
jinlee93, line 26: This removes some shadow dom browser bordering 
image
Not averse to keeping the border which could indicate color selection functionality but I think the border itself could be confusing.
Also using psuedo element selectors here is fine since docs only

.storybook/components/ColorList/ColorList.tsx
```diff
@@ -1,33 +1,42 @@
...
-import { TokenSpecimen } from '../TokenSpecimen/TokenSpecimen';
+import React from 'react';
+import styles from './ColorList.module.css';
 
 type ListItem = {
   name: string;
-  value: any;
+  value: string;
+  figmaToken?: string;
+  tailwindClass?: string;
```
jinlee93, line 8: prepping to show t2 and t3 color tailwind utility classes

.storybook/components/TokenSpecimen/TokenSpecimen.css
```diff
@@ -1,5 +1,3 @@
-@import '../../../src/design-tokens/mixins.css';
```
jinlee93, line 1: unrelated but unneeded

.storybook/components/TokenSpecimen/TokenSpecimen.tsx
jinlee93, line 1: This page edit was also unnecessary but came across it while thinking about the t1 colors page so just converted it to react FC

jinlee93:
> 🥳
> 
> What's an example of a "figma token" in this context?

It's how it shows up in Figma 
image
Any suggestions on what it's actually called?

PR #1712, opened 2023-07-27
docs: show more code in component stories docs

jinlee93:
[EDS-1074]

### Summary:
- found some [documentation](https://storybook.js.org/docs/react/api/doc-block-source#type) on how to show the code more properly with the docs addon: 
  - `parameters: {docs: {source: 'dynamic'}}` renders the story source with dynamically updated arg values, allowing stories with `render` functions to show the code
  - `parameters: {docs: {code: '*code string*'}}` allows custom code to be shown
- addresses Along want for "Show Code" working for Sortable Interactive story for the Table component
- Makes "Show Code" working for some more components
  - stories used for visual regression only were not addressed

### Test Plan:

  - "Show Code" in docs addon shows code
image

[EDS-1074]: https://czi-tech.atlassian.net/browse/EDS-1074?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Accordion/Accordion.stories.tsx
+++ b/src/components/Accordion/Accordion.stories.tsx
@@ -230,5 +230,21 @@ export const UsingRenderProp: StoryObj<Args> = {
   ),
   parameters: {
     chromatic: { disableSnapshot: true },
+    docs: {
+      source: {
+        code: `<Accordion headingAs="h2">
+  <Accordion.Row>
+    {({ open }) => (
+      <>
+        <Accordion.Button data-testid="accordion-button">
+          Accordion Button {(open && 'open') || 'closed'}
+        </Accordion.Button>
+        <Accordion.Panel>Accordion Panel</Accordion.Panel>
+      </>
+    )}
+  </Accordion.Row>
+</Accordion>`,
+      },
+    },
   },
 };
--- a/src/components/DataBar/DataBar.stories.tsx
+++ b/src/components/DataBar/DataBar.stories.tsx
@@ -62,70 +62,75 @@ export const NoSegments: StoryObj<Args> = {
   },
 };
 
-const Interactive = () => {
-  const [max, setMax] = React.useState(100);
-  const [segments, setSegments] = React.useState([
-    { value: 90, text: 'Segment value 90' },
-    { value: 1, text: 'Segment value 1' },
-    { value: 1, text: 'Segment value 1' },
-  ]);
-
-  const [segmentValue, setSegmentValue] = React.useState(1);
-
-  const onPush = () => {
-    segments.push({
-      value: segmentValue || 1,
-      text: `Segment value ${segmentValue}`,
-    });
-    setSegments([...segments]);
-  };
-  const onPop = () => {
-    if (segments.length > 0) {
-      segments.pop();
-      setSegments([...segments]);
-    }
-  };
-  return (
-    <div>
-      <DataBar label="Interactive data bar" max={max} segments={segments} />
-      <br />
-      <label htmlFor="segment-value-input">Value add: </label>
-      <input
-        id="segment-value-input"
-        onChange={(e) => {
-          setSegmentValue(Number(e?.target?.value));
-        }}
-        value={segmentValue}
-      ></input>
-      <br />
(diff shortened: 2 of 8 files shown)


src/components/Accordion/Accordion.stories.tsx
```diff
@@ -230,5 +230,21 @@ export const UsingRenderProp: StoryObj<Args> = {
...
+        <Accordion.Button data-testid="accordion-button">
+          Accordion Button {(open && 'open') || 'closed'}
+        </Accordion.Button>
+        <Accordion.Panel>Accordion Panel</Accordion.Panel>
+      </>
+    )}
+  </Accordion.Row>
+</Accordion>`,
+      },
+    },
```
jinlee93, line 248: custom code string is required since type "dynamic" is useless as example:
image

src/components/DataBar/DataBar.stories.tsx
```diff
@@ -62,70 +62,75 @@ export const NoSegments: StoryObj<Args> = {
   },
 };
 
-const Interactive = () => {
```
jinlee93, line 65: These interactive components have to live within the render function for the code to show properly

src/components/Select/Select.stories.tsx
```diff
@@ -165,18 +165,76 @@ const selectCat: StoryObj['play'] = async (playOptions) => {
...
 };
 
 export const Disabled: StoryObj = {
+  parameters: {
+    docs: {
+      source: {
+        code: '// refer to "Default" story code',
+      },
+    },
+  },
```
jinlee93, line 222: I got lazy with some of these but didn't want to explicit state props here if the story props were to change in the future, then would have to remember to change here as well... also the code example should be enough

src/components/Select/Select.stories.tsx
```diff
@@ -165,18 +165,76 @@ const selectCat: StoryObj['play'] = async (playOptions) => {
...
   ),
 };
 
 export const DefaultWithVisibleLabel: StoryObj = {
+  parameters: {
+    docs: {
+      source: {
+        code: `
+// refer to "Default" story code with:
+// props: {aria-label: 'Favorite Animal', labelComponent: <Select.Label>Favorite Animal</Select.Label>}`,
```
jinlee93, line 234: An example where props is explicitly stated here to show usage of the `labelComponent` prop, but if this story were to change in the future, this might also need to be updated

src/components/Accordion/Accordion.stories.tsx
booc0mtaco: Good discoveries. There's space to add a brief description of these learnings to https://chanzuckerberg.github.io/edu-design-system/?path=/story/documentation-guidelines-code-guidelines--page (could be a separate follow-up task)

* Writing stories
* How to generate "show code" when `render` is used in a story
* etc.

src/components/Select/Select.stories.tsx
```diff
@@ -186,6 +244,13 @@ export const DefaultWithVisibleLabel: StoryObj = {
 };
 
 export const Compact: StoryObj = {
+  parameters: {
+    docs: {
+      source: {
+        code: '// refer to "Default" story code',
+      },
+    },
+  },
```
booc0mtaco, line 253: nit: could be set to a variable as well

src/components/Select/Select.stories.tsx
```diff
@@ -248,6 +346,62 @@ export const UsingChildrenProp: StoryObj = {
...
+                  <Icon
+                    className="ml-4"
+                    name="filter-list"
+                    purpose="decorative"
+                  />
+                </button>
+              )}
+            </Select.Button>
+            <Select.Options>
+              {exampleOptions.map((option) => (
```
booc0mtaco, line 390: in https://chanzuckerberg.github.io/edu-design-system/?path=/story/documentation-guidelines-code-guidelines--page we could advise people use this storybook config when a component has rend props, for example

booc0mtaco (approved):

PR #1713, opened 2023-07-28
chore: remove canonical recipes from EDS

booc0mtaco:
### Summary:

- this removes recipes that are product-specific
- this is prep work to sync up with updated (stricter) recipe guidelines in the doc.s

### Test Plan:

- n/a (only removing stories/pages)


booc0mtaco:
Tagged https://github.com/chanzuckerberg/edu-design-system/releases/tag/remove-lp-pages in case we want to refer to these at some point in the future.

jinlee93 (approved):
🗑️

PR #1715, opened 2023-07-31
docs: nest component recipes under components

booc0mtaco:
### Summary:

- move component recipes (now referred to as implementation examples) under related components
- label them as such in the storybook UI (using a custom badge)
- note: some recipes are used in stories, which breaks convention
- update snapshots for relevant decorator and component changes.
- remove recipe and page references so new ones don't start appearing

Recipes used in other recipes and pages are pending removal. As a reminder to the bigger context... **canonical recipes** are those which are specific to a product and will live within the product's storybook/documentation. **implementation examples** are sandboxed code examples that are composed of a set of other EDS components which might describe a recommended layout. As examples, they are not exported, so we move them alongside other examples in storybook to reduce confusion.

### Test Plan:

- [x] verify updated snapshots and chromatic diff.s


booc0mtaco:
Holding merge on this one until #1713 has merged (so we can remove the recipes used in the soon-to-be-deleted pages)

jinlee93:
Does this mean we won't have the `recipes` SB folder anymore? What if a new recipe involves more than one EDS component?

booc0mtaco:
> Does this mean we won't have the `recipes` SB folder anymore? What if a new recipe involves more than one EDS component?

Correct. After looking thru things, having the recipes broken out as we were doing created two kinds of confusion:

* [Recipes](https://bradfrost.com/blog/post/design-system-components-recipes-and-snowflakes/) are, by definition, compositions that are product specific. But stuff in storybook should be available to be imported?
* We already had some stories that were combining multiple components, and based on feedback and questions (and the fact that the code examples show what's going on) the intent is clear. Why were some broken out separately?

For these implementation examples, there's usually a "main component" and subordinate components used to demonstrate how to use it. This gets us to a spot where we can be more clear on what the word "recipe" means outside of consuming projects. Happy to chat more about this, as it seemed MOSTLY clear to me as I sorted thru approaches!

jinlee93 (approved):
Thanks for the explanation!

.storybook/preview.tsx
```diff
@@ -66,5 +66,17 @@ export const parameters = {
...
+    implementationExample: {
+      styles: {
+        backgroundColor: '#ffffff',
+        borderColor: '#000000',
+        color: '#000000',
+      },
+      title: 'Implementation Example',
+      tooltip: {
+        title: `About Implementation Examples`,
+        desc: 'Implementation examples show how you might combine existing EDS components to create custom recipes.',
```
anniehu4, line 78: should we avoid using the word "recipes" here? maybe "custom layouts" or end the sentence after "EDS components"?

src/components/Card/Card.stories.tsx
```diff
@@ -59,3 +74,181 @@ export const Dragging: StoryObj<Args> = {
     elevation: 'dragging',
   },
 };
+
+export const ButtonActionCalloutCard: StoryObj<Args> = {
```
jinlee93, line 78: Wait what about putting the recipes under {component}/recipes/ via `title: 'Components/Card/recipes` or do you think the badge does the job?

anniehu4 (approved):
love this

src/components/Card/Card.stories.tsx
anniehu4: could also consider always naming the stories "xx Example"

src/components/Card/Card.stories.tsx
```diff
@@ -59,3 +74,181 @@ export const Dragging: StoryObj<Args> = {
...
+          <div>
+            <Button variant="primary">Preview</Button>
+          </div>
+        </div>
+      </Card.Body>
+    </Card>
+  ),
+};
+
+export const CardWithNotification: StoryObj<Args> = {
```
anniehu4, line 110: fyi some files in the "Recipes" section still show up. not sure if you meant to delete them all here

Screenshot 2023-08-02 at 2 33 58 PM

src/components/Card/Card.stories.tsx
anniehu4: could also consider always naming the stories "xx Example"

booc0mtaco: These could work, actually. I wanted to lean away from using that Recipe word in EDS, so we didn't create ambiguity on what a recipe is. but it may make sense to separate out these stories from the more normal stories. 

Let's have a chat about that in our next sync :) I can create a ticket with some options to prep discussion.

.storybook/preview.tsx
booc0mtaco: agreed, good catch. we might s/combine/compose and chop the end of that sentence off

booc0mtaco:
Gonna expand this one just a touch to include the now-unused recipes along with the generators for new recipes and pages to close the loop

PR #1716, opened 2023-07-31
fix(Avatar): support display names using emoji and multi-byte

booc0mtaco:
### Summary:

- add new prop for display name to allow for avatar text not based on the user's real name
- add tests for sensible fallbacks when using this component

### Test Plan:

### DIFF
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -13,6 +13,10 @@ export type UserData = {
    */
   id?: string | number;
   /**
+   * The display shortcut for the user name. Can be initials, emoji, or other text symbols (recommended max: 2)
+   */
+  displayName?: string;
+  /**
    * Additional data for an attached user (email, etc.)
    */
   [k: string]: string | number | boolean | undefined;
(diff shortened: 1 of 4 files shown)


jinlee93 (approved):
Think this makes sense to me, thanks!

src/components/Avatar/Avatar.tsx
```diff
@@ -56,7 +60,7 @@ function getInitials(fromName: string): string {
    * - User's name has a middle name or initial: John C. Smith
    * - User's Name has dashes in it
    */
-  return fromName
+  const initials = fromName
     .split(' ')
     .map((part) => part[0])
```
booc0mtaco, line 65: In terms of how the emoji get messed up, THIS line is the culprit. It will split an emoji into separate parts, leaving some unable to be rendered. Pragmatically, not touching this for now, as the use case for parsing a name with emoji IN the name is pretty niche. We should allow a display name in case people want to use non-name text for the user

ahuth: Maybe a regex is better for finding initials

ahuth (approved):

src/components/Avatar/Avatar.tsx
booc0mtaco: Agreed. I found a regexp that could identify emoji as well that i didn't like AT ALL, so avoided going down that path. It included a bunch of specific escape characters that catch many (not all, maybe most) emoji.

But doing some additional digging yield this new thing i did not know about!

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape

I'll test this out separately.

PR #1717, opened 2023-08-01
docs: clarify hotfix release process

booc0mtaco:
### Summary:

it's possible to hotfix an older version of EDS. in those cases, we don't necessarily want to merge back into `main`. add a note explaining this situation (which should be rare). Also diminish the hotfix instructions so that the document is easier to follow at a glance.

### Test Plan:

- n/a (documentation only changes)

### DIFF
--- a/docs/PUBLISHING.md
+++ b/docs/PUBLISHING.md
@@ -10,8 +10,6 @@ Look to [this helpful document](https://designsystem.morningstar.com/getting-sta
 
 We currently use [standard-version](https://github.com/conventional-changelog/standard-version) to increment the version number in `package.json`, create a git tag for the new release, and update `CHANGELOG.md` based on the commit log.
 
-**NOTE**: `standard-version` is deprecated. Soon, we may move to an alternative script to handle releases.
-
 ---
 
 ## Releasing a new version of EDS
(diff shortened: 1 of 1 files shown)


docs/PUBLISHING.md
```diff
@@ -63,6 +63,8 @@ git push --follow-tags origin <branch> # this will also push tags
 
 For the commit message, use the new version's content in the [CHANGELOG.md](../CHANGELOG.md) (e.g., all the changes for version 1.2.3). Review the content in the changelog to make sure the notes, and the from-version and to-version are correct. Note the link to the storybook on this PR CI/CD. You will add this link to the GitHub release notes later.
 
+**NOTE**: if this is a hotfix release for an older version of EDS (not latest), skip to step 9.
```
jinlee93, line 66: What about the scenario where the hotfix is for an older release of EDS but is still relevant to the latest version of EDS? Maybe as non urgent?

booc0mtaco: I was tempted to dig into that, but it is a corner case. For now, I was thinking we could consider cherry-picking the change back into `next` or just re-implementing it? Pragmatically thinking to cross that bridge when we come to it :)

jinlee93: That sounds good to me

jinlee93 (approved):

booc0mtaco:
thanks !

PR #1719, opened 2023-08-07
docs: split t2 colors stories and remove t2 stories aggregator

jinlee93:
### Summary:
- removes aggregating story file Tier2Tokens.stories.jsx and jsut converts the individual componets to story files
  - similar work will be done for T1 and T3 tokens
- splits T2 colors as requested by design and I also agree as part of survey feedback to not have to scroll that much on ZH
- adds suggested TW class and design token string to the relevant color
  - follow up pr will attempt to differentiate these (is follow up since it'll also affect the other tokens pages)
- converts some jsx to tsx

### Test Plan:

  - no stories changes except for t2 colors
  - adds the tw class and figma design token next to the colors
Screenshot 2023-08-07 at 11 49 03 AM


jinlee93:
Interesting that chromatic flags these as new stories even tho they're structured and look exactly the same
image

booc0mtaco:
@jinlee93 looks good. where do the token names in figma come from?

jinlee93:
> @jinlee93 looks good. where do the token names in figma come from?

It comes from the UI foundations [Figma file](https://www.figma.com/file/IfMWAtWqNrQprImlco4pi0/EDS-SLP-v1.3-%7C-UI-Foundations?type=design&node-id=249-6074&mode=design&t=itky0zeqvRVxFkIF-4) 
image

booc0mtaco (approved):
The names differ slightly in format from figma to code, so we transform the code token names into the figma equivalents for reference. LGTM

PR #1720, opened 2023-08-07
docs: split t3 colors stories and remove t3 stories aggregator 

jinlee93:
### Summary:
- removes aggregating story file Tier3Tokens.stories.jsx and just converts the individual components to story files
- splits T3 colors as requested by design and I also agree as part of survey feedback to not have to scroll that much on ZH
- adds suggested TW class and design token string to the relevant color
- follow up pr will attempt to differentiate these (is follow up since it'll also affect the other tokens pages)
- converts some jsx to tsx
- recurses the T3 and T2 tokens to raw value 
  - this is fine since the Figma style references the respective T1/T2 css property and not the raw value

### Test Plan:

  - T3 color tokens docs pages are affected and they show the suggested TW class and the Figma style
  - T2 color tokens that reference T1 colors recurse to the raw value 
image


.storybook/util/filterTokens.ts
```diff
@@ -1,10 +1,19 @@
...
 
+const recurseToPrimaryValue: (value: string) => string = (value) => {
+  if (value.startsWith('var(--eds')) {
+    return recurseToPrimaryValue(
+      tokens[value.slice(6, -1) as keyof typeof tokens],
+    );
+  }
+  return value;
+};
+
```
jinlee93, line 11: this util function recurses to the raw value

.storybook/util/filterTokens.ts
```diff
@@ -1,10 +1,19 @@
 import tokens from '../data/tokens.json';
 
+const recurseToPrimaryValue: (value: string) => string = (value) => {
```
jinlee93, line 3: not sure why typing like this is necessary 🤔 
image

booc0mtaco:
Now that we have quite a few different values in here, I think labeling them in the UI would be helpful. We could widen the text area for these, and put them in a table or something. 

* CSS Variable name
* Figma Token name
* Tailwind Class name
* Hexadecimal value

jinlee93:
> Now that we have quite a few different values in here, I think labeling them in the UI would be helpful. We could widen the text area for these, and put them in a table or something.
> 
> * CSS Variable name
> * Figma Token name
> * Tailwind Class name
> * Hexadecimal value

Updated to have a Table which also affected the T1 and T2 colors appropriately
Think that sticky header could be useful here haha
@booc0mtaco

booc0mtaco (approved):
@jinlee93 Looks great to me, and good call on sticky header :P

PR #1722, opened 2023-08-14
fix: update token exports

booc0mtaco:
### Summary:

update exported values for link tokens

### Test Plan:

  - updated snapshots 
  - verify no diff is generated when running `yarn start`


jeremiah-clothier (approved):
LGTM! You mentioned this generate any diffs, so the color is the same but the variable referenced is just different?

booc0mtaco:
> LGTM! You mentioned this generate any diffs, so the color is the same but the variable referenced is just different?

yah, the values are committed to a set of JSON files, and the build command writes the output to these files. There's a ticket to make PRs check for cases where running `yarn build` generates a diff to catch when a change to token files results in these differences.

PR #1725, opened 2023-08-15
docs: add clearer warning and recommendation for useId

booc0mtaco:
### Summary:

This updates the emitted warning message when using the built-in fallback for `useId` in react < 18 projects. Includes a recommendation as well.

### Test Plan:

- n/a (documentation only change)

### DIFF
--- a/src/util/useId.ts
+++ b/src/util/useId.ts
@@ -12,7 +12,7 @@ export const useId =
   function useId() {
     if (process.env.NODE_ENV !== 'production' && showWarning) {
       console.warn(
-        'EDS useId is not ssr friendly if using React <18 as it uses the native React.useId hook if available https://react.dev/reference/react/useId.',
+        "React < 18: EDS's `useId` is not SSR friendly. We recommend applying an ID package like `uniqueId` or using React 18's built-in `React.useId` https://react.dev/reference/react/useId",
       );
       showWarning = false;
     }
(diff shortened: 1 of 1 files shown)


src/util/useId.ts
```diff
@@ -12,7 +12,7 @@ export const useId =
   function useId() {
     if (process.env.NODE_ENV !== 'production' && showWarning) {
       console.warn(
-        'EDS useId is not ssr friendly if using React <18 as it uses the native React.useId hook if available https://react.dev/reference/react/useId.',
+        "React < 18: EDS's `useId` is not SSR friendly. We recommend applying an ID package like `uniqueId` or using React 18's built-in `React.useId` https://react.dev/reference/react/useId",
```
jeremiah-clothier, line 15: 1) Is there a way this warning could talk about the "how". For example, if I see this message how would I apply uniqueId to resolve the warning?

2) Will this message ever show if using React >= 18? The message seem's to indicate some opt-in behavior

booc0mtaco: @jeremiah-clothier 

1. that's fair. I didn't want to go into too much detail, as this would have only affected along, and they just moved IDs onto `uniqueId`. I can rephrase to say someone should import and use uniqueId or similar to set the ID prop for safety
2. once they move to React 18, they can keep using `uniqueId` OR, allow this to fall back to react's SSR-safe `.useId()`. This message only appears when someone is using React <= 17 AND isn't setting IDs on form fields themselves.

For context, this naive implementation of an ID generator can lead to race conditions in rare cases, or cause different IDs to appear when rendered on the server vs the client browser.

jeremiah-clothier (approved):

PR #1726, opened 2023-08-17
chore: upgrade to storybook 7.x

booc0mtaco:
### Summary:

- remove extraneous / outdated storybook plugins
- upgrade packages to the latest available for storybook
- amend documentation pages to support MDX2 format
- fix remaining bugs
- update test files to support latest story-utils and other dependencies
- liberally use `act()` around test setup actions

### Test Plan:

  - no snapshot changes and no failing/flaky tests introduced


package.json
```diff
@@ -86,6 +86,7 @@
   "dependencies": {
     "@headlessui/react": "^1.7.16",
     "@popperjs/core": "^2.11.8",
+    "@storybook/addon-styling": "^1.3.6",
```
jeremiah-clothier, line 89: Should this be a `devDependencies`?

.babelrc.json
jeremiah-clothier, line 1: Thoughts on putting this babel config inside `main.ts` so we know "it's only used for storybook". For example, https://github.com/chanzuckerberg/edu-stack/blob/main/.storybook/main.ts#L29C3-L40

src/components/Accordion/Accordion.test.tsx
```diff
@@ -9,16 +10,22 @@ import * as stories from './Accordion.stories';
 const { Default } = composeStories(stories);
 
 describe('<Accordion />', () => {
-  generateSnapshots(stories);
+  generateSnapshots(stories as StoryFile);
```
jeremiah-clothier, line 13: I notice some usages of `generateSnapshots` require `as StoryFile` and others don't, do you know what's happening here?

src/components/Accordion/Accordion.test.tsx
```diff
@@ -9,16 +10,22 @@ import * as stories from './Accordion.stories';
...
   it('should open and close Accordion panel clicking Accordion button', async () => {
     const user = userEvent.setup();
     render(<Default />);
     expect(screen.queryByTestId('accordion-panel')).not.toBeInTheDocument();
     const accordionButton = screen.getByTestId('accordion-button');
-    await user.click(accordionButton);
+
+    await act(async () => {
+      await user.click(accordionButton);
+    });
```
jeremiah-clothier, line 23: Why do these events have to be wrapped in `act`, how'd you come to realize this?

package.json
booc0mtaco: ah, good catch, yep, it should be

src/components/Accordion/Accordion.test.tsx
booc0mtaco: not exactly sure, but it could be the type changes related to some of the storybook libraries. The first sweep was to satisfy typescript to restore parity on `yarn test`, and we can see if something systematic is causing certain story files to need a type

src/components/Accordion/Accordion.test.tsx
booc0mtaco: after upgrading dependencies, `yarn test` would output gigantic errors in console. what's strange is that for some of these user events, if you add the `act()` around them, it will complain that it's redundant 👎 

For whatever reason, not all `user.` calls need this, and on older versions of the libraries, none did. It doesnt make obvious sense which is which

src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -7,7 +7,7 @@ const defaultArgs = {
   label: 'Checkbox',
 };
 
-export default {
+const meta: Meta<typeof Checkbox> = {
```
booc0mtaco, line 10: In cases like this, there were type errors that prevented tests from passing. Following the docs from storybook, this is the recommended typing to apply. What's weird is that types didn't break down on ALL tests, so not sure when types would get lost yet.

It's likely we should make these consistent in one way or the other, and favor aligning with this pattern since it mirrors [the docs](https://storybook.js.org/docs/react/writing-stories/introduction#default-export)

jeremiah-clothier: > it's likely we should make these consistent in one way or the other, and favor aligning with this pattern since it mirrors the docs

:+1:

I did something similiar when upgrading the Learning platform to Storybook v7. I went though and changed every `as Meta<Args>` to `satisfies Meta<Args>` which is a bit better than `as`. Not saying `satifies` is better than `const meta: Meta <...>` just pointing out I ran into similar weirdness

jeremiah-clothier (approved):

jeremiah-clothier:
How does testing for the deployed storybook or zeroheight work?

booc0mtaco:
> How does testing for the deployed storybook or zeroheight work?

@jeremiah-clothier I might not be answering your question but we can chat about it. For ZH, the link between it and storybook is either by embedded iframe, OR by the ZH integration, which are keyed on the individual story URLs/names respectively.

The testing for deployed storybook itself is pretty straightforward; the github action must be able to pass all the existing checks before it can be published.

----

I'll add a backlog item for making the test signatures consistent, and see if that resolves the need for the `act()` call

src/components/Checkbox/Checkbox.stories.tsx
booc0mtaco: Yep yep, i recall that one, and tested it on a few places. I think there may be more issues with the story files since that didn't have an effect in some cases. It's a little bit mysterious...

PR #1727, opened 2023-08-17
fix(Avatar): improve support for generating initials

booc0mtaco:
### Summary:

use [Graphemer](https://github.com/flmnt/graphemer) to:
- add handling for unicode surrogate pairs (mark1)
- handling for extended BMP (Basic Multilingual Plane) characters

### Test Plan:

### DIFF
--- a/src/components/Avatar/Avatar.test.ts
+++ b/src/components/Avatar/Avatar.test.ts
@@ -1,6 +1,37 @@
 import { generateSnapshots } from '@chanzuckerberg/story-utils';
+import { getInitials } from './Avatar';
 import * as stories from './Avatar.stories';
 
 describe('<Avatar />', () => {
   generateSnapshots(stories);
+
+  // Testing handling of surrogate pairs, et al
+  // https://javascript.info/unicode#surrogate-pairs
+  describe('.getInitials', () => {
+    it.each`
+      givenName                     | expectedInitials
+      ${'John Smith'}               | ${'JS'}
+      ${'Jenny C. Sallow'}          | ${'JS'}
+      ${'Thomas Stroughton-Felton'} | ${'TS'}
+      ${'Young Yarn Lad'}           | ${'YL'}
+      ${'Tim 肇廷'}                 | ${'T肇'}
+      ${'Ãndrew 🙂'}                | ${'Ã🙂'}
+      ${'你好世界'}                 | ${'你'}
+      ${'🧶👦🏽'}                     | ${'🧶'}
+      ${'𩷶'}                       | ${'𩷶'}
+      ${'𝒳'}                        | ${'𝒳'}
+      ${'☹️ 🙂'}                    | ${'☹️🙂'}
+      ${'👦🏽🧶'}                     | ${'👦🏽'}
+      ${'🧙‍♂️'}                       | ${'🧙‍♂️'}
+      ${''}                         | ${''}
+      ${' Leading Space'}           | ${'LS'}
+      ${'Trailing Space '}          | ${'TS'}
+      ${'Z͑ͫ̓ͪ̂ͫ̽͏̴̙̤̞͉͚̯̞̠͍A̴̵̜̰͔ͫ͗͢L̠ͨͧͩ͘G̴̻͈͍͔̹̑͗̎̅͛́Ǫ̵̹̻̝̳͂̌̌͘'}                    | ${'Z͑ͫ̓ͪ̂ͫ̽͏̴̙̤̞͉͚̯̞̠͍'}
+    `(
+      'generates initials $expectedInitials when given $givenName',
+      ({ givenName, expectedInitials }) => {
+        expect(getInitials(givenName)).toEqual(expectedInitials);
+      },
+    );
+  });
 });
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -1,4 +1,6 @@
 import clsx from 'clsx';
+import Graphemer from 'graphemer';
+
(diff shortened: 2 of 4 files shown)


booc0mtaco:
Little bit of context from SO: https://stackoverflow.com/questions/4877326/how-can-i-tell-if-a-string-contains-multibyte-characters-in-javascript

src/components/Avatar/Avatar.test.ts
```diff
@@ -1,6 +1,31 @@
...
 
 describe('<Avatar />', () => {
   generateSnapshots(stories);
+
+  // Testing handling of surrogate pairs, et al
+  // https://javascript.info/unicode#surrogate-pairs
+  describe('.getInitials', () => {
+    // ${'🧙‍♂️'}             | ${'🧙‍♂️'}
+    // ${'👦🏽🧶'}           | ${'👦🏽'}
+    // ${'☹️ 🙂'}          | ${'☹️🙂'}
```
ahuth, line 13: Are these commented out because they're composite emoji (or whatever the right terminology is... e.g. are made up of multiple graphemes)?

src/components/Avatar/Avatar.test.ts
```diff
@@ -1,6 +1,31 @@
...
+      ${'John Smith'}     | ${'JS'}
+      ${'Young Yarn Lad'} | ${'YL'}
+      ${'你好世界'}       | ${'你'}
+      ${'🧶👦🏽'}           | ${'🧶'}
+      ${'𩷶'}             | ${'𩷶'}
+      ${'𝒳'}              | ${'𝒳'}
+      ${'Ãndrew 🙂'}      | ${'Ã🙂'}
+      ${'Tim 肇廷'}       | ${'T肇'}
+    `(
+      'Given tne name $givenName, it generates initials $expectedInitials',
```
ahuth, line 25:
```suggestion
      'Given the name $givenName, it generates initials $expectedInitials',
```

or

```suggestion
      'generates $expectedInitials for $givenName',
```

src/components/Avatar/Avatar.tsx
```diff
@@ -53,7 +53,28 @@ export interface Props {
   variant?: 'icon' | 'initials' | 'image';
 }
 
-function getInitials(fromName: string): string {
+/**
+ * Algo for names with special characters in:
+ * - Split name parts by space
+ * - for a given part,
+ *  - if that part has emoji or extended characters, {Action?}
```
ahuth, line 60: Is `{Action?}` a placeholder for a future edit to the comment?

src/components/Avatar/Avatar.tsx
```diff
@@ -53,7 +53,28 @@ export interface Props {
...
+ *  - if that part does not, then use the first ASCII character
+ * @param str
+ * @returns
+ */
+function produceAbbreviation(str: string): string {
+  // unicode character range for surrogate pairs (first and second halves)
+  if (str.match(/([\uD800-\uDFFF])*/g)?.filter((match) => !!match).length) {
```
ahuth, line 67: Could probably use use unicode emoji character class thing to match emoji:

```js
/\p{Emoji}/u.test(str)
```

which may be more readable

src/components/Avatar/Avatar.tsx
```diff
@@ -53,7 +53,28 @@ export interface Props {
...
+    // grab first part if string has a surrogate chunk after the first character
+    // this assumes that the first part of the surrogate pair is legible
+    // TODO: See sample cases where matching splits surrogates improperly
+    return str.slice(0, 2);
```
ahuth, line 71: After some internet research, splitting emoji seems hard.

We could try https://github.com/flmnt/graphemer, see if it passes the compound test cases.

src/components/Avatar/Avatar.tsx
```diff
@@ -53,7 +53,28 @@ export interface Props {
...
 
-function getInitials(fromName: string): string {
+/**
+ * Algo for names with special characters in:
+ * - Split name parts by space
+ * - for a given part,
+ *  - if that part has emoji or extended characters, {Action?}
+ *  - if that part does not, then use the first ASCII character
+ * @param str
+ * @returns
```
ahuth, line 63: May not need the jsdoc comments for @param and @returns, bcuz typescript

ahuth (approved):

src/components/Avatar/Avatar.test.ts
booc0mtaco: Yah the wording around this is so hard :( But, correct; in those cases, there's 2+ parts to each. like the Wizard emoji is composed from the wizard AND the BMP for gender. when running the test on that one, it fails, but the expected and actual LOOK identical :'( 

The others behave incorrectly in different ways... i meant to take this off of review until i got a bit more time to see if there were any reasonable fixes to account for these cases, or maybe just discourage people from having emoji in their names and defer to the `displayName` prop

src/components/Avatar/Avatar.tsx
booc0mtaco: Yep. Gonna clean this up after digging in some more later today

src/components/Avatar/Avatar.tsx
booc0mtaco: [This sequence alone](https://github.com/flmnt/graphemer/blob/master/src/Graphemer.ts#L113-L122) tells me i probably should stop trying to solve this myself lol

booc0mtaco:
@ahuth thanks for the lib. reference. I didn't have the right vocabulary so didn't find anything yesterday when i first looked 😅

PR #1733, opened 2023-08-23
fix: update deploy docs for storybook 7.x

booc0mtaco:
### Summary:

Ref: https://github.com/storybookjs/storybook/discussions/20932

- move away from deprecated deploy package and instead rely directly on GH Actions
- [ ] remove dependency 
- [ ] update package.json scripts
- [ ] test deploy to gh-pages then revert branch to read

### Test Plan:


booc0mtaco:
**Note**: completed publish of v13.1.0 since this fix also did a deploy of what's on next ... which as a result of the release is the same as what WAS released. no need to immediately do another release

ahuth (approved):

.github/workflows/docs.yml
```diff
@@ -1,29 +1,35 @@
...
+      run: yarn build:storybook
 
-      - name: Deploy Storybook Docs 🚀
-        run: yarn run deploy:docs
-        env:
-          GH_TOKEN: '${{ github.actor }}:${{ secrets.GITHUB_TOKEN }}'
+    - name: Deploy Storybook
+      uses: peaceiris/actions-gh-pages@v3
+      with:
+        publish_dir: ./storybook-static
```
booc0mtaco, line 34: the naming is a bit awkward here... this should be read as `dir_to_publish` or similar; this *isn't* where the docs will be published to, and won't end up in the path

PR #1737, opened 2023-08-31
Generate valid tab ids

ahuth:
### Summary:

Noted in https://czi-edu.slack.com/archives/CTFV79JH4/p1693505461086869, we're generating invalid ids for tabs. Here we fix that.

### Test Plan:


andremalan (approved):

booc0mtaco (approved):
thanks for that quick patch

PR #1738, opened 2023-08-31
Implement EDS Theming tooling

booc0mtaco:
### Summary:

This adds some utility scripts to EDS to support setting up, applying, and updating token definitions, using style-dictionary. Further enhancements will be to support reading outputs from figma into a format EDS understands, and flagging theming errors where possible. Final phase will include support for diffing against EDS updates, so that newly-added tokens are available, or deprecated tokens are flagged.

This will also include usage instructions in-repo, as this is developer tooling. 

- [x] init script to set up theming in a project
- [x] apply script to convert JSON into CSS tokens to read in code
- [x] updated config to package the correct files (bin/, theme files, etc.)
- [x] instructions on usage (how to set up config, options, gotchas, etc.)
- [x] FAQ location in README
- [x] Remove TODOs, and test integrations (config to theme EDS, etc.)

### Test Plan:

  - Create an [alpha publish](https://github.com/chanzuckerberg/edu-design-system/blob/main/docs/PUBLISHING.md#alpha-release) and try out in `along`


yarn.lock
```diff
@@ -2005,12 +2007,16 @@ __metadata:
...
+    svg4everybody: ^2.1.9
     tailwindcss: ^3.3.3
     ts-jest: ^29.1.1
     typescript: ^4.9.5
   peerDependencies:
     react: ">= 17"
     react-dom: ">= 17"
+  bin:
+    eds-apply-theme: bin/eds-apply-theme.js
+    eds-init: bin/eds-init.js
```
booc0mtaco, line 2019: These are the commands consumers will have access to. e.g., 

`npx eds-init`
`npx eds-apply-theme`

booc0mtaco:
Created alpa version to test out on a few projects `chanzuckerberg/eds@13.2.0-alpha.0`

booc0mtaco:
Two tools generate expected outputs! will add in some additional checks for reading config, then post for review

bin/eds-apply-theme.js
```diff
@@ -0,0 +1,95 @@
...
+  const { getConfig } = require('./_util');
+
+  let packageRootPath;
+  try {
+    packageRootPath =
+      path.dirname(require.resolve('@chanzuckerberg/eds')) + '/tokens/json/';
+  } catch (e) {
+    console.error('EDS package not installed. Using local path...');
+    packageRootPath =
+      path.dirname(require.main.path) + '/src/tokens-dist/json/';
```
timzchang, line 15: what do we expect to be at this directory if eds is not installed?

bin/eds-init.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  // read in the config from config file, package json "eds", etc.
+  const config = await getConfig();
+
+  // take the packaged token file and place a copy in the project's 'json' specified path
+  if (config) {
+    try {
+      fs.copyFileSync(
+        packageRootPath + 'theme-base.json',
```
timzchang, line 24: where  is theme-base created?

.storybook/components/Docs/Guidelines/Theming.stories.mdx
```diff
@@ -10,7 +10,91 @@ import { Canvas, Meta } from '@storybook/blocks';
...
+```json
+"eds": {
+  "json": "src/components/",
+  "css": "src/components/"
+},
+```
+
+`json` determines where the core theme file will be stored, and `css` determines where the resulting css token file will be stored.
+
+### eds-init-theme
```
timzchang, line 31: nit: can we name the files generated by `eds-init-theme` and specify which one to modify?

timzchang (approved):
This makes sense to me! A couple questions and a nit in the readme only

bin/eds-init.js
booc0mtaco: `theme-base.json` is generated along side the other style-dictionary output files (e.g., variables.css, et al), and gets shipped with EDS. For consumers it will live **somewhere** in `node_modules/@chanzuckerberg/eds`, and this line will match that location

Each released version of EDS will include an updated version of this file for two uses:

* new projects will start off with a fresh copy that includes all the available tokens
* existing projects can do a diff against their copy and this one to add/remove tokens that changes from EDS version X to Y

That second thing is another reason we have people editing JSON, as it is MUCH easier to build tooling around diffing JSON object as opposed to CSS files or some other content.

bin/eds-apply-theme.js
booc0mtaco: This little bit is essentially for testing WITHIN EDS. I tested most of the functionality by dog-fooding this in the repo. I can update the error message to be `info` or something, and make this say something closer to that

.storybook/components/Docs/Guidelines/Theming.stories.mdx
booc0mtaco: 👍🏾! added in file names in appropriate spots and better sequencing in the doc.s

booc0mtaco:
FYI going to merge this one as a merge commit instead of a squash to preserve the tag and commit for the alpha I tested with. We will have 1-2 follow up alpha releases to test TS interop and make sure rollup still provides something that works with consumer projects

PR #1742, opened 2023-09-06
[HOLD] feat(InlineNotification): add neutral variant to InlineNotification

jeremiah-clothier:
### Summary:

Adding a new variant `neutral` to the InlineNotification component. The tokens already existed in the system so I just referenced existing ones.

### Test Plan:

- [x] There are new stories that got generated for this variant and changes can be observed in chromatic

PR #1744, opened 2023-09-06
chore: make `.module.css` class names human readable in local development

jeremiah-clothier:
### Summary:

I noticed that classnames are hashed while working on #1742. This makes finding the origin of styles in the codebase harder and is straightforward to toggle off via webpack.

### Test Plan:

- [x] Manual testing, observe the classnames before and after

**Before:**
image

**After:**
image

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -104,7 +104,7 @@ function updateCSSLoaderPlugin(config: Configuration): Configuration {
                 // @ts-expect-error css-loader doesn't accept "string" options
                 // and will either be an object or undefined
                 ...ruleSetRule.options?.modules,
-                localIdentName: '[name]__[local]',
+                localIdentName: '[name]__[local]--[hash:base64:5]',
               },
             };
           }
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -1,3 +1,6 @@
+import type { StorybookConfig } from '@storybook/react-webpack5';
+import type { Configuration } from 'webpack';
+
 /**
  * Although `--static-dir` is marked as deprecated. The The Chromatic CLI
  * currently pulls the staticDir from the build script, but does not support
@@ -6,7 +9,7 @@
  * We should refrain from using the staticDirs option in this configuration until
  * https://github.com/chromaui/chromatic-cli/issues/462 is resolved.
  */
-module.exports = {
+const config: StorybookConfig = {
   stories: [
     './components/**/*.stories.mdx',
     './components/**/*.stories.@(js|jsx|ts|tsx)',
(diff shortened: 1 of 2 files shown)


.storybook/main.ts
```diff
@@ -70,4 +73,47 @@ module.exports = {
...
+                // and will either be an object or undefined
+                ...ruleSetRule.options?.modules,
+                localIdentName: '[name]__[local]',
+              },
+            };
+          }
+        });
+      }
+    }
+  });
```
jeremiah-clothier, line 114: So many typescript types to narrow check against 😮‍💨

.storybook/main.ts
jeremiah-clothier, line 1: Decided to convert this to `.ts` as well so the logic for updating the `css-loader` plugin is a bit more safe using types

.storybook/main.ts
```diff
@@ -70,4 +73,47 @@ module.exports = {
...
+          ) {
+            ruleSetRule.options = {
+              // @ts-expect-error css-loader doesn't accept "string" options
+              // and will either be an object or undefined
+              ...ruleSetRule.options,
+              modules: {
+                // @ts-expect-error css-loader doesn't accept "string" options
+                // and will either be an object or undefined
+                ...ruleSetRule.options?.modules,
+                localIdentName: '[name]__[local]',
```
ahuth, line 107: I wonder if we should either:
- Keep the hash at the end of the name (e.g. [name]_[local]_[hash])
- Or include the path (e.g. [path]_[name]_[local])

That way the classes are a bit more guaranteed to be unique, like they will be in production.

Probably low chance of conflict as-is, but it could be nice to make double extra sure.

.storybook/main.ts
```diff
@@ -70,4 +73,47 @@ module.exports = {
...
 };
+
+/**
+ * Updates the `css-loader` webpack plugin to make class names human readable.
+ *
+ * NOTE: This should only be used for local development.
+ */
+function updateCSSLoaderPlugin(config: Configuration): Configuration {
+  config.module?.rules?.forEach((rule) => {
+    if (rule && typeof rule === 'object' && Array.isArray(rule.use)) {
```
ahuth, line 91: An alternative could be to use webpack-merge's [mergeWithRules](https://github.com/survivejs/webpack-merge#mergewithrules).

Not sure if that'll make this simpler or not, but could be worth trying out.

.storybook/main.ts
booc0mtaco: the old behavior before the storybook 7 upgrade [included the hash at the end](https://61313967cde49b003ae2a860-lwbwxnvynb.chromatic.com/?path=/story/components-accordion--default) so we could go back to that

jeremiah-clothier: I'll add the `hash`. I think adding the `path` is too verbose for class names given the depth and structure of some directories

.storybook/main.ts
jeremiah-clothier: I'm going to leave this as is for now. I tried it out but merging is kind of weird because the value of `loader` provided from the Storybook webpack config is an absolute path to the `css-loader` library file. 

Specifically,

```ts
const ourConfig: Configuration = {
    module: {
      rules: [
        {
          test: /\.css$/,
          use: [
            {
              loader: 'css-loader',
              options: {
                modules: {
                  localIdentName: '[name]__[local]--[hash:base64:5]',
                },
              },
            },
          ],
        },
      ],
    },
  };

  return mergeWithRules({
    module: {
      rules: {
        test: 'match',
        use: {
          loader: 'match',
          options: 'merge',
        },
      },
    },
  })(theirConfig, ourConfig);
```

does not merge the `css-loader` correctly 
image

Hardcoding the path to `node_modules/css-loader/dist/cjs.js` felt a bit weird so decided to not do it

.storybook/main.ts
jeremiah-clothier: Classes with [hash]

image

PR #1745, opened 2023-09-07
fix(Avatar): add fallback workaround to graphemer import class

booc0mtaco:
### Summary:

Related to https://github.com/flmnt/graphemer/issues/11, we have an issue in some consumers where the commonJS module is imported in such a way that we get the wrong object format. (instead of `[Graphemer class]` we get `{default: [Graphemer class]}`, which causes the code to fail).

Check for double-`default` wrapping as a workaround until graphemer is fixed. This will also unblock work on the theming tooling

### Test Plan:

  - ensure tests pass locally
  - test with project that has the import problem, to ensure tests / storybook work properly

### DIFF
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -55,12 +55,14 @@ export interface Props {
    */
   variant?: 'icon' | 'initials' | 'image';
 }
-
 /**
  * Use graphemer to take a name part, and select the first grapheme (emoji, surrogate pair, ASCII character)
  */
 function produceAbbreviation(fromName: string): string {
-  const splitter = new Graphemer();
+  // @see https://github.com/flmnt/graphemer/issues/11
+  // @ts-expect-error handles case where this library adds .default to the export inappropriately in CJS context
+  const G = Graphemer['default'] ? Graphemer.default : Graphemer;
+  const splitter = new G();
   return fromName ? splitter.splitGraphemes(fromName)[0] : '?';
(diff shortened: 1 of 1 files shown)


src/components/Avatar/Avatar.tsx
```diff
@@ -55,12 +55,14 @@ export interface Props {
...
 }
-
 /**
  * Use graphemer to take a name part, and select the first grapheme (emoji, surrogate pair, ASCII character)
  */
 function produceAbbreviation(fromName: string): string {
-  const splitter = new Graphemer();
+  // @see https://github.com/flmnt/graphemer/issues/11
+  // @ts-expect-error handles case where this library adds .default to the export inappropriately in CJS context
+  const G = Graphemer['default'] ? Graphemer.default : Graphemer;
```
ahuth, line 64: Nit: might be able to use `in` and avoid the TS error

```suggestion
  const G = 'default' in Graphemer ? Graphemer.default : Graphemer;
```

Although I can't imagine any of this makes TS happy 😆

ahuth (approved):

src/components/Avatar/Avatar.tsx
booc0mtaco: ### TypeScript HATES this one simple trick

😅

booc0mtaco:
We don't need this workaround anymore, as setting `interop: auto` in rollup did the trick 👍 

cc @ahuth

PR #1748, opened 2023-09-07
chore: add plugin for intellisense of css variables

jeremiah-clothier:
### Summary:

There are soo many CSS variables in EDS it's hard to know what the values are when looking at a `.module.css` file. I'm adding the "CSS Var Complete" extension to help get some intellisense when working in `.module.css` files.

### Test Plan:

- [x] Manually testing experience with this plugin (before and after)

**Before**
image
image

**After**
image
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,7 +32,6 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 
-/.vscode
 *.css.map
 !DesignTokens.css
 !Utilities.css
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,3 @@
+{
+  "recommendations": ["dbaeumer.vscode-eslint", "phoenisx.cssvar"]
+}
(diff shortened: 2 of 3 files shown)


jeremiah-clothier:
@booc0mtaco It doesn't work great with nested values 
image

booc0mtaco:
@jeremiah-clothier This is great! I think something like this would also be useful in our stack and various projects. I will think about where and how to document that (if nothing else, can be part of the stack and patched into current point solutions)

jeremiah-clothier:
> @jeremiah-clothier This is great! I think something like this would also be useful in our stack and various projects. I will think about where and how to document that (if nothing else, can be part of the stack and patched into current point solutions)

We had this in the Learning Platform, here's where I documented that. Although documentation closer to the repo may be better. https://czi.atlassian.net/wiki/spaces/ETE/pages/2689335723/Visual+Studio+Code+Setup#EDS-Intellisense

booc0mtaco:
@jeremiah-clothier good link, thanks :) I'll add it to the dev docs here, as anyone outside the org would also benefit from such setup info

booc0mtaco (approved):
@jeremiah-clothier for using this plugin, are there any manual steps an engineer would need to follow? if so, can you include a little text for EDS local setup in CONTRIBUTING.md?

PR #1753, opened 2023-09-12
chore: add launch.json config for debugging

booc0mtaco:
### Summary:

Found that i had this config for running the debugger when using `jest`, which enables stepping thru code to discover the problems. add as an option for anyone working on EDS.

### Test Plan:

- n/a (developer-only config update)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,20 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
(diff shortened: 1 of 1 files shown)


booc0mtaco:
This may or may not be all that useful, but it was easy enough to create a PR for it :) Let me know if there are questions on this one (this might also be useful for the stack)

ahuth (approved):

jeremiah-clothier:
> This may or may not be all that useful, but it was easy enough to create a PR for it :) Let me know if there are questions on this one (this might also be useful for the stack)

There's no instructions needed right, VS Code just picks this launch profile up in settings from what I remember?

booc0mtaco:
> > This may or may not be all that useful, but it was easy enough to create a PR for it :) Let me know if there are questions on this one (this might also be useful for the stack)
> 
> There's no instructions needed right, VS Code just picks this launch profile up in settings from what I remember?

That's correct. The settings will appear in the debug panel

PR #1756, opened 2023-09-15
feat(Select): add warning when name is missing from component props

booc0mtaco:
### Summary:

There have been a few recent cases where `Select` presented problems for consumers. Namely, certain frameworks expect a form field to appear so that data can be submitted (no pun intended). This is an important behavior, and luckily HeadlessUI supports this in the underlying [`Listbox`](https://headlessui.com/react/listbox#listbox) component.

Add some documentation to explain this, and a dev-time warning to double-down on the recommendation. In the future, `name` may be a required prop when using any/all form elements.

### Test Plan:

- [x] ~Wrote~ updated [automated tests](https://czi.atlassian.net/wiki/x/Hbl1H)

### DIFF
--- a/src/components/Select/Select.stories.tsx
+++ b/src/components/Select/Select.stories.tsx
@@ -67,6 +67,7 @@ function InteractiveExampleUsingChildren(props: Props) {
         className={clsx(!compact && 'w-60')}
         data-testid="dropdown"
         disabled={disabled}
+        name="interactive-select"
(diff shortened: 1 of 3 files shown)


booc0mtaco:
Screenshot 2023-09-15 at 13 36 41

jeremiah-clothier (approved):
LGTM!

PR #1759, opened 2023-09-18
feat: separate ClickableStyles from core components

booc0mtaco:
### Summary:

- unlink ClickableStyle from Button
- unlink ClickableStyle from Link
- remove code that can no longer be triggered for each component
- mark desired component API changes as deprecated where appropriate
- clean up documentation and technical guidance
- update component snapshots

This will allow Remix and other consuming tools to deal with fewer indirections and couplings when using this base component, and enforce the correct element when building UIs. In the short term, this introduces some duplication of styles/code, as we preserve most of the components' ability to mimic each other. At the next major, this will be simplified to allow the proper props given the component (e.g., remove non-button variants from `Button`, and similar for `Link`).

### Test Plan:

  - Will verify snapshots and chromatic diffs are correct when done

### Additional testing

- **Validated snapshots**. We expect all tags to remain the same, class names used to change, and components consuming `Link` and/or `Button` to change as well. 
- **Validate chromatic**. No snapshots should change appearance. There will be new snapshots for the now-separate `ClickableStyles` test (which were previously coupled to `Button` and `Link`.


src/components/ClickableStyle/__snapshots__/ClickableStyle.test.ts.snap
```diff
@@ -0,0 +1,177 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
booc0mtaco, line 1: Maintaining `ClickableStyles` for now, as the main reason it gets exported is to facilitate some behavior in Remix, and some possible edge cases that aren't buttons or links, but should be clickable.

src/components/Button/Button.tsx
```diff
@@ -1,14 +1,21 @@
...
+  IconStatus,
+  LinkStatus,
+} from '../ClickableStyle/ClickableStyle';
 import LoadingIndicator from '../LoadingIndicator';
 
 import styles from './Button.module.css';
 
 type ButtonHTMLElementProps = React.ButtonHTMLAttributes<HTMLButtonElement>;
 
+type ButtonVariants = IconStatus | PrimaryStatus | SecondaryStatus | LinkStatus;
```
booc0mtaco, line 17: Eventually, we won't need the variant/status types to be this granular and coupled in `Button`/`Link`. At next major, this will be simplified to be the individual statuses and variants, and we can then drop some of this complexity (or sequester it so just `ClickableStyle`

src/components/Button/__snapshots__/Button.test.tsx.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`<Button /> Default story renders snapshot 1`] = `
 <button
-  class="clickable-style clickable-style--lg clickable-style--secondary clickable-style--brand button button--secondary"
+  class="button button--lg button--secondary button--brand"
```
booc0mtaco, line 5: Note for reviewers: if **any** snapshot has a tag changing, this is probably a mistake! The changes should almost entirely be replacing `"chickable-style"` with either `"button"` or `"link"` for now.

src/components/Button/Button.tsx
```diff
@@ -18,46 +25,44 @@ export type ButtonProps = ButtonHTMLElementProps & {
...
+   * The size of the button on screen
+   */
+  size?: Extract<Size, 'sm' | 'md' | 'lg'>;
   /**
    * Disables the field and prevents editing the contents
    */
   disabled?: boolean;
   /**
    * Loading state passed down from higher level used to trigger loader and text change
-   * @deprecated - This will be removed in a future release
```
timzchang, line 29: no longer deprecated?

src/components/Button/__snapshots__/Button.test.tsx.snap
timzchang: thanks for the heads up! was going to scroll thru this file tbh

src/components/Link/Link.stories.tsx
```diff
@@ -1,7 +1,7 @@
 import type { StoryObj, Meta } from '@storybook/react';
 import React from 'react';
-import { Link } from './Link';
-import { SIZES, STATUSES, VARIANTS } from '../ClickableStyle';
+import { Link, type LinkProps } from './Link';
+import { SIZES } from '../ClickableStyle';
```
timzchang, line 4: should Sizes be removed or nah?

timzchang (approved):

src/components/Button/Button.tsx
booc0mtaco: Yah a few things might need to be re-discussed in the new year. Right now, we don't have an alternative way to mark a button as "disabled b/c the app/page is doing something" so not very nice to remove without that :)

src/components/Link/Link.stories.tsx
booc0mtaco: oh good call. will tidy up what storybook shows for this one, and remove/hide that.

PR #1761, opened 2023-09-20
feat(Icons): allow component icon usages to be headless

booc0mtaco:
### Summary:

- part one: using existing icons with current selection as default
  - this will let teams choose from the possible set of icons when overrides are needed
  - some cases only allow one icon for now but abstract this for future use
  - rename any icon prop names to `icon` for consistency and backwards compatibility
- part two will enable using a ReactNode or some render prop, so teams can use custom icons with their own types

### Test Plan:


src/components/Accordion/Accordion.tsx
```diff
@@ -40,6 +40,10 @@ type AccordionButtonProps = {
    * Additional classnames passed in for styling
    */
   className?: string;
+  /**
+   * Icon override for component. Default is 'expand-more'
+   */
+  icon?: Extract<IconName, 'expand-more'>;
```
booc0mtaco, line 46: This might not look all that useful but hear me out :)

Firstly, this is a first step in creating a new union utility type that replaces `expand-more` with `T` along with some other functionality. that will allow for injecting custom icons with some constraints. There is some architecture work to do for that final state so replacing the cases now in anticipation of that functionality. Using [`Extract`](https://www.typescriptlang.org/docs/handbook/utility-types.html#extracttype-union) like this adds some ergonomics for when the the available icons change in the future, or when teams provide their own Icon sets (and thus their own `IconName` potentially)

* Allow for a generic Icon component wrapper with useful constraints (control of viewport, enforce accessibility, etc.) in the type.
* Implement utility type, likely in `Icons.tsx` as a new export
* Find/replace all usages of this construction with that utility type where needed (searching "Extract<IconName," makes that easy)

src/components/DragDrop/DragDrop.module.css
```diff
@@ -75,16 +75,16 @@
   display: flex;
   flex-shrink: 0;
   flex-direction: column;
-  margin-left: var(--eds-size-2);
+  margin-right: var(--eds-size-2);
```
booc0mtaco, line 78: Originally, it was adding margins to the front side of every drag-drop section, and a pseudo-class to chop that margin from the first thing. Flip that around so the margins get added to all but the last section, so that we don't need an override in the most common case.

For future work... this whole component will likely be moved out of EDS with all its core pieces remaining as exported atoms/molecules.

booc0mtaco:
This also includes some storybook tweaks to make the Drag Drop component more generic.

Screenshot 2023-09-22 at 17 39 14

src/components/Accordion/Accordion.tsx
ahuth: Eventually this will allow some subset of icons?

Probably easier to allow any icon name be used (at the expense of control).

ahuth (approved):

src/components/Accordion/Accordion.tsx
booc0mtaco: This is one of the cases where it'd either be this specific built-in icon, or some custom icon that team wants to use (and not another EDS icon) when I work on the second part.

In the next week or two, hoping to have a chat with some design folks so I can start on the second part of this one, and possibly simplify parts of this further

PR #1765, opened 2023-09-25
chore: remove peacock color settings from settings.json

booc0mtaco:
### Summary:

- until https://github.com/microsoft/vscode/issues/15909 is resolved, remove these settings
- once fixed, we can .gitignore such settings while also committing useful global ones

### Test Plan:

- n/a (dev. config only change)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -12,25 +12,5 @@
   "typescript.validate.enable": true,
   "javascript.validate.enable": true,
   "editor.formatOnSave": true,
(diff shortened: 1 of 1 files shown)


booc0mtaco:
Screenshot 2023-09-25 at 12 46 06

We will have to keep an eye on EDS so that no errant settings get committed (since we do want to commit the current settings.json file)

ahuth (approved):

jeremiah-clothier (approved):

PR #1767, opened 2023-09-26
feat(Accordion): allow empty or hidden accordion rows

booc0mtaco:
### Summary:

- adding an `isExpandable={false}` to an `AccordianRow` will hide the feature that afford clicking
- add tests to verify this behavior
- add new snapshots

### Test Plan:

### DIFF
--- a/src/components/Accordion/Accordion.test.tsx
+++ b/src/components/Accordion/Accordion.test.tsx
@@ -95,12 +95,39 @@ describe('<Accordion />', () => {
         </Accordion.Row>
       </Accordion>,
     );
-    const accordionButton = screen.getByTestId('accordion-button');
+    const accordionButton = screen.getByRole('button');
 
     await act(async () => {
       await user.click(accordionButton);
     });
     expect(onOpen).toHaveBeenCalledTimes(1);
+    expect(onClose).not.toHaveBeenCalled();
+  });
+
+  it('should not call onOpen callback when accordion opens on an empty row', async () => {
+    const user = userEvent.setup();
+    const onClose = jest.fn();
+    const onOpen = jest.fn();
+    render(
+      <Accordion headingAs="h2">
+        <Accordion.Row isExpandable={false}>
+          <Accordion.Button
+            data-testid="accordion-button"
+            onClose={onClose}
+            onOpen={onOpen}
+          >
+            Accordion Button
+          </Accordion.Button>
+          <Accordion.Panel>Accordion Panel</Accordion.Panel>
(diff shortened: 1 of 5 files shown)


booc0mtaco:
Screenshot 2023-09-26 at 17 30 16

booc0mtaco:
TBD:

- including the content on a row (`.Panel`) if marked as empty?
- form handling recommendations
- clearer documentation
- sync with design

src/components/Accordion/Accordion.test.tsx
```diff
@@ -103,4 +103,31 @@ describe('<Accordion />', () => {
...
+            onClose={onClose}
+            onOpen={onOpen}
+          >
+            Accordion Button
+          </Accordion.Button>
+          <Accordion.Panel>Accordion Panel</Accordion.Panel>
+        </Accordion.Row>
+      </Accordion>,
+    );
+    const accordionButton = screen.getByTestId('accordion-button');
```
ahuth, line 125: Nit: per [testing libraries query priorities](https://testing-library.com/docs/queries/about/#priority), `getByRole` is preferred if it's possible to use.

src/components/Accordion/Accordion.test.tsx
```diff
@@ -103,4 +103,31 @@ describe('<Accordion />', () => {
...
+
+    await act(async () => {
```
ahuth, line 127: Does anything bad happen if we remove the `act`?

`user.click` is supposed to have that baked into it.

ahuth (approved):

src/components/Accordion/Accordion.test.tsx
booc0mtaco: good callout. will see if there's anything blocking use of "role"

src/components/Accordion/Accordion.test.tsx
booc0mtaco: it generates a gross warning when `act()` is not present:
```
Warning: An update to DisclosureFn inside a test was not wrapped in act(...).

    When testing, code that causes React state updates should be wrapped into act(...):

    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */
```

src/components/Accordion/Accordion.test.tsx
booc0mtaco: Ohh, i remember why I deliberately used testId for these ones. at one point, data attributes were not making it to the DOM, and I added some tests to show that it was working. That's mostly a given, so the tests were only so useful in the long term

PR #1768, opened 2023-09-28
feat(Tabs): add ability to customize tab button headers

booc0mtaco:
### Summary:

As part of the "Headless" strat, allow for flexibility of the structure and content of the tab headers via a new subcomponent `Tab.Button`, which will take the place of the `title` prop on `Tab` if present, and allow for control based on tab state, with intelligent fallbacks.

- use intelligent fallback as backwards-compatibility
- allow composing of custom tab contents
- add new snapshots and stories illustrating new behavior
- use a render prop to pass in title, active, and other future data

### Test Plan:


booc0mtaco:
Screenshot 2023-09-28 at 15 02 58

booc0mtaco:
https://github.com/chanzuckerberg/edu-design-system/assets/3056447/96143d7c-8e20-404b-b049-3eac28bbcff1

src/components/Tab/Tab.tsx
```diff
@@ -28,21 +34,24 @@ export interface Props {
...
  *
  * Individual tab within the Tabs component.
+ * - Use the `title` prop for text-only tab headers
+ * - For more custom tab headers use `<Tab.Button>` which uses a render prop with state information
  */
 export const Tab = ({
   children,
   className,
   id,
   // Destructure `title` so it is not applied to the rendered element
```
ahuth, line 50:
```suggestion
  // Destructure `title` so it is not applied to the rendered element
```

src/components/Tab/Tab.tsx
```diff
@@ -52,7 +61,27 @@ export const Tab = ({
...
-      {children}
+      {withoutTypes(children, TabButton)}
     </div>
   );
 };
+
+const TabButton = ({ children }: TabButtonProps) => {
+  const context = useContext(TabsContext);
+  return (
+    <div>{typeof children === 'function' ? children(context) : children}</div>
```
ahuth, line 72: Could `<Tab>` take the render prop instead? Or do we need the additional component/div layer?

src/util/utility-types.ts
```diff
@@ -42,3 +42,11 @@ export type EitherInclusive<T, U> = EitherExclusive<T, U> | (T & U);
 export type ForwardedRefComponent<T, P> = React.ForwardRefExoticComponent<
   React.PropsWithoutRef<P> & React.RefAttributes<T>
 >;
+
+/**
+ * Utility type used when defining custom render props. Uses template for the
+ * members of the possible
```
ahuth, line 48: What does "Uses template for the members of the possible" mean? 🤔

ahuth (approved):

src/util/utility-types.ts
booc0mtaco: Ahh, i forgot to finish the thought, and i keep calling type params "templates". 

`Uses a type param for the members of the possible render prop` or something like that.

src/components/Tab/Tab.tsx
booc0mtaco: I created `Tab.Button` to mimic the structure of some of our other components with render props, like `Accordion` and a few of the other HeadlessUI offshoots. But in testing this I found something PRETTY peculiar. I am gonna submit a simplified update to this PR to demonstrate what i mean.

Besides that, my plan is to eventually add something like a `Tab.Panel` to contain the non-button children to make the rest of the API match our other components:

```jsx
<Tab title="Some Title">
    <Tab.Button>...</Tab.Button>
    <Tab.Panel>...</Tab.Panel>
</Tab>
```

booc0mtaco: I think, long-term, the final step to make all these internally consistent would be to add the render prop to `Tab` as you're saying, and it would contain `Tab.Button`s and `Tab.Panel`s. Based on how `Tabs` is implemented, it might be a breaking change. I'm for it, tho... the component is a little weird right now.

PR #1769, opened 2023-09-29
feat(TextareaField): support recommendedMaxLength prop for display-only errors

booc0mtaco:
### Summary:

- new prop `recommendedLength` allows for showing an error but letting the user continue typing content
- `maxLength` still works, and can be a different (larger or equal) value
- update snapshots and tests

### Test Plan:

### DIFF
--- a/src/components/TextareaField/TextareaField.tsx
+++ b/src/components/TextareaField/TextareaField.tsx
@@ -36,6 +36,11 @@ type TextareaFieldProps = React.TextareaHTMLAttributes<HTMLTextAreaElement> & {
    * Error state of the form field
    */
   isError?: boolean;
+  /**
+   * Behaves similar to `maxLength` but allows the user to continue typing more text.
+   * Should not be larger than `maxLength`, if present.
+   */
+  recommendedMaxLength?: number;
 } & EitherInclusive<
     {
       /**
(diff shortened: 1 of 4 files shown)


booc0mtaco:
TBD: 

- unit tests to pressure test the input patterns
- verify when to show/hide the field note

booc0mtaco:
https://github.com/chanzuckerberg/edu-design-system/assets/3056447/bcd8a15c-4f7a-4494-afc5-c0b55c4837e5

src/components/TextareaField/TextareaField.tsx
```diff
@@ -111,6 +116,21 @@ const TextArea = forwardRef<HTMLTextAreaElement, TextAreaProps>(
...
+ */
+function getSmallest(lengthA?: number, lengthB?: number): number | undefined {
+  if (lengthA !== undefined && lengthB !== undefined) {
+    return Math.min(lengthA, lengthB);
+  } else if (lengthA && lengthB === undefined) {
+    return lengthA;
+  } else if (lengthB && lengthA === undefined) {
+    return lengthB;
+  }
+}
```
ahuth, line 132: Code golf time 🏌️‍♀️

```suggestion
function getSmallest(...args: Array<number | undefined>): number | undefined {
  if (args.length === 0) {
    return;
  }
  return Math.min(...args.filter(Boolean) as number[]);
}
```

or with type predicate on the filter:

```suggestion
function getSmallest(...args: Array<number | undefined>): number | undefined {
  if (args.length === 0) {
    return;
  }
  return Math.min(...args.filter((n): n is number => Boolean(n)));
}
```

ahuth (approved):

src/components/TextareaField/TextareaField.tsx
booc0mtaco: I like these ⛳! Since the other PR builds upon this one and moves this into a utility, will commit this as-is, and take some strokes off in #1771

PR #1771, opened 2023-10-02
feat(InputField): support recommendedMaxLength prop for display-only errors

booc0mtaco:
### Summary:

- allow handling for `maxlength` and `recommendedMaxLength` like TextareaField
- update snapshots and tests

### Test Plan:

### DIFF
--- a/src/components/InputField/InputField.test.tsx
+++ b/src/components/InputField/InputField.test.tsx
@@ -33,4 +33,54 @@ describe('<InputField />', () => {
 
     expect(onChange).toHaveBeenCalledTimes(testText.length);
   });
+
+  it('will not fire when maxLength is reached', async () => {
+    const user = userEvent.setup();
+    const onChange = jest.fn();
+    const testText = 'typing';
+
+    render(
+      <InputField
+        aria-label="label"
+        data-testid="test-input"
+        defaultValue={testText}
+        maxLength={6}
+        onChange={onChange}
+      />,
+    );
+    const input = screen.getByTestId('test-input');
+
+    input.focus();
+
+    await act(async () => {
+      await user.keyboard(testText);
+    });
+
+    expect(onChange).toHaveBeenCalledTimes(0);
+  });
+
+  it('will fire when recommendedMaxLength is reached', async () => {
+    const user = userEvent.setup();
+    const onChange = jest.fn();
+    const testText = 'typing';
+
+    render(
+      <InputField
+        aria-label="label"
+        data-testid="test-input"
+        defaultValue={testText}
+        onChange={onChange}
+        recommendedMaxLength={6}
+      />,
+    );
+    const input = screen.getByTestId('test-input');
+
+    input.focus();
+
+    await act(async () => {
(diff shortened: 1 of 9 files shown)


booc0mtaco:
Screenshot 2023-10-02 at 16 17 41

src/components/InputField/InputField.test.tsx
```diff
@@ -33,4 +33,54 @@ describe('<InputField />', () => {
...
+        recommendedMaxLength={6}
+      />,
+    );
+    const input = screen.getByTestId('test-input');
+
+    input.focus();
+
+    await act(async () => {
+      await user.keyboard(testText);
+    });
```
ahuth, line 82: (Asked this on another PR, but don't remember the reponse)

What happens if we remove the `act`? IIRC, `user-event` is supposed to handle that for us.

booc0mtaco: Yah, something throws warnings in the test runner when `act` is missing, tho the test itself seems to past. 🤔 Mostly adding it to keep the output clean.

src/components/InputField/InputField.tsx
```diff
@@ -175,6 +186,19 @@ export const InputField: InputFieldType = forwardRef(
       fieldNote && styles['input-field--has-fieldNote'],
     );
 
+    const fieldLength = fieldText?.toString().length;
+
+    const textExceedsMaxLength =
+      maxLength !== undefined && fieldLength ? fieldLength > maxLength : false;
```
ahuth, line 192: Nit, couple ideas:

```ts
const fieldLength = fieldText?.toString().length ?? 0;
const textExceedsMaxLength = maxLength && fieldLength > maxLength;
```

or give `maxLength` a default and simplifyt the check even more

```ts
maxLength = Infinity

...

const textExceedsMaxLength = fieldLength > maxLength;
```

src/components/InputField/InputField.tsx
```diff
@@ -183,6 +207,14 @@ export const InputField: InputFieldType = forwardRef(
       ? ariaDescribedBy || generatedAriaDescribedById
       : undefined;
 
+    const fieldLengthCountClassName = clsx(
+      (textExceedsMaxLength || textExceedsRecommendedLength) &&
+        styles['input-field--invalid-length'],
+    );
+
+    // Pick the smallest of the lengths to set as the maximum value allowed
+    const maxLengthShown = getMinValue(maxLength, recommendedMaxLength);
```
ahuth, line 216: If `maxLength` and `recommendedMaxLength` are numbers (e.g. if they have default values or fallbacks), may be able to use `Math.min`?

ahuth (approved):

src/components/InputField/InputField.tsx
booc0mtaco: Will apply a few things where possible. One thing I wanted to avoid were any props being added if the consumer didn't specify them. `maxLength` is also weird in that it changes how the base field behaves when it has a value, and it treats `Infinity` as invalid if it gets to `<textarea>`

PR #1775, opened 2023-10-04
feat(Tailwind): add config to expose primary and secondary fonts

booc0mtaco:
### Summary:

expose more tokens used in EDS to tailwind, to allow for utility classes for the preconfigured font families, or for any custom font families added by consumers. Can be used in isolation like `font-primary` or `font-secondary`, or as an occasional override to any typography components like `Heading` or `Text`.

- add the example config to tailwind.config.js with brief documentation
- add example to Text component
- update snapshots

### Test Plan:

### DIFF
--- a/tailwind.config.ts
+++ b/tailwind.config.ts
@@ -53,6 +53,13 @@ export default {
       medium: edsTokens['font-weight'].medium,
       bold: edsTokens['font-weight'].bold,
     },
+    fontFamily: {
+      // provide values for the configured font family tokens
+      // Useful when your app imports '@chanzuckerberg/eds/fonts.css' or
+      // if you have custom token values for primary and secondary fonts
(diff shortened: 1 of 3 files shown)


jeremiah-clothier (approved):

tailwind.config.ts
jeremiah-clothier, line 1: do `chore:` commits bump versions appropriately and get logged in the changelog? Just double checking because eds does export its tailwind.config.ts for other apps to import

booc0mtaco: Good callout. `chore`s do NOT edit the changelog, and this wouldn't appear. I think i will bump this to `feat` even tho its pretty minor.... so i guess it should make it a minor release :)

I'll also tag this with `Tailwind` in the commit message

PR #1777, opened 2023-10-05
test(Menu): add delay to menu open user event trigger

booc0mtaco:
### Summary:

Attempt nr. 2 at making the open menu test less flaky. Add a delay to the user event that triggers the open menu. 

Trying this, as the snapshot likely happens right after the event completes, when the positioning library is calculating the final location and width of the popover. 

### Test Plan:

- [x] ~Wrote~ modified [automated tests](https://czi.atlassian.net/wiki/x/Hbl1H)


jeremiah-clothier (approved):
🤞

PR #1779, opened 2023-10-06
feat(Menu): allow href and onClick to co-exist on a menu item

booc0mtaco:
### Summary:

Remove this artificial constraint on `Menu.Item` components so that additional behavior can be attached to menu items along with an `href` with `onClick`. Like with an `<a>`, it will be up to the consumer to avoid undesired behavior.

### Test Plan:


jeremiah-clothier (approved):
Thanks for looking at this and fixing it!

PR #1781, opened 2023-10-12
fix(Select): expose render prop and default value param.s

booc0mtaco:
### Summary:

- clean up example using unstyled and styled render props
- remove custom styling and use FPO block instead
- add snapshots
- update existing snapshots

### Test Plan:


booc0mtaco:
https://github.com/chanzuckerberg/edu-design-system/assets/3056447/a04100cd-0321-4225-bc5c-e2d45c7c6b07

booc0mtaco:
**Motivation**: uncontrolled components allow the form to behave in a way that's more compatible with frameworks using standard HTTP methods for data persistence/submission (e.g., Remix). exposing this option allows for simpler implementation

src/components/Select/Select.tsx
```diff
@@ -368,6 +378,7 @@ export const SelectButton = React.forwardRef<
 });
 
 Select.Button = SelectTrigger;
+Select.ButtonWrapper = SelectButton;
```
jeremiah-clothier, line 381: This is a bit confusing to me. What's going on here?

`select.button = trigger` but `select.buttonwrapper = button`

jeremiah-clothier (commented):
Need clarity on the `Select.ButtonWrapper` and how it's used. I don't see if referenced anywhere expect added as a new export

src/components/Select/Select.tsx
booc0mtaco: Yah, not exactly happy with this as-is. It's referenced in [this story](https://github.com/chanzuckerberg/edu-design-system/pull/1781/files#diff-fa4aaf0cf433f9e3d925818cfd47bf418dfc90fcc4693d909480f780f0ff488aR465-R489), and should be used if someone wants to use BOTH the render prop, but also the existing style for `Select`.

The idea is that you can use `<Select.Button>{text}</Select.Button>` which will secretly use that wrapper, or 

```jsx
<Select.Button>
{(props} => (
    <span>{props.label}</span> {/* or something similar */}
)}
</Select.Button>
```
Which will render the "button" as a plain span. If someone wants to use the wrapper, it wasn't previously available, hence the export. This allows the button to be "headless" AND allows folks who want to do something with the render prop to also use the regular styling. Naming suggestions?

I can add documentation to explain this use case, but the way things work with storybook makes documenting the details here a little cumbersome. Will think about it a bit more.

jeremiah-clothier (approved):

PR #1782, opened 2023-10-12
feat(Popover): allow focus upon popover panel open

booc0mtaco:
### Summary:

Merge in the props from [HeadlessUI](https://headlessui.com/react/popover#popover-panel) for the panel, to support `focus`. This allows consumers to send focus to a particular piece of content within the content container. `onFocus` also allows for hooking into the internals in case there are some arguments to trigger upon focus.

### Test Plan:

### DIFF
--- a/src/components/Popover/Popover.stories.tsx
+++ b/src/components/Popover/Popover.stories.tsx
@@ -5,6 +5,7 @@ import React from 'react';
 import { Popover } from './Popover';
 import type { PopoverProps } from './Popover';
 import Button from '../Button';
+import Hr from '../Hr';
 
 export default {
   title: 'Components/Popover',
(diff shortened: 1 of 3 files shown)


src/components/Popover/Popover.stories.tsx
```diff
@@ -147,3 +148,31 @@ export const LeftEnd: StoryObj<PopoverProps> = {
...
 };
+
+export const FocusClickableElement: StoryObj<PopoverProps> = {
+  play: async ({ canvasElement }) => {
+    // We want to test visual regression for the Popover.Content as well as the button,
+    // but don't want the drawer open initally outside Chromatic.
+    if (isChromatic()) {
+      const canvas = within(canvasElement);
+      const filtersTrigger = await canvas.findByRole('button');
+      userEvent.click(filtersTrigger);
```
jeremiah-clothier, line 159: should this be `awaited`?

src/components/Popover/Popover.stories.tsx
```diff
@@ -147,3 +148,31 @@ export const LeftEnd: StoryObj<PopoverProps> = {
...
+    }
+  },
+  render: (args) => {
+    return (
+      <Popover {...args}>
+        <Popover.Button as={Button} data-testid="popover-trigger-button">
+          Open Popover
+        </Popover.Button>
+        <Popover.Content data-testid="popover-content" focus>
```
jeremiah-clothier, line 168: I may be missing something, why does the content need `focus` set?

booc0mtaco: [`focus`](https://headlessui.com/react/popover#popover-panel) enables the behavior where it will focus on the element once the popover is generated

src/components/Popover/Popover.stories.tsx
booc0mtaco: It does look like `.click` returns a promise, so will prefix with `await` to be safe.

PR #1788, opened 2023-10-17
chore: upgrade testing-library/react to v14

booc0mtaco:
### Summary:

re: https://github.com/testing-library/user-event/issues/1104#issuecomment-1759110918 update version of testing-library for react to latest major v14.

After validating stable tests, try removing/reducing calls to `act()` that might now be superfluous.

### Test Plan:

  - validate build with existing `act()` tests
  - validate build after removing some superfluous `act()` calls in test

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,7 +134,7 @@
     "@storybook/testing-library": "^0.2.0",
     "@storybook/testing-react": "^2.0.1",
     "@testing-library/jest-dom": "^5.17.0",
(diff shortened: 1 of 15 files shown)


booc0mtaco:
@ahuth was able to scrub away all of the extraneous `act()` calls, and it seems like the test runs might complete slightly faster than before 🤔. v14 seems to show fewer slow test runs, so we get to benefit from some improvements with less code ++

timzchang (approved):

PR #1792, opened 2023-10-25
chore: update omlet config

booc0mtaco:
### Summary:

- add TS and TSX files to `include`
- make sure `exports` is defined to mark the package entry point
- confirm package analytics match expectations

### Test Plan:

- n/a (dev-only change affecting component analytics)

### DIFF
--- a/.omletrc.json
+++ b/.omletrc.json
@@ -1,5 +1,8 @@
 {
   "ignore": ["lib/**", ".storybook/**"],
-  "include": ["src/components/**/*.tsx"],
-  "tsconfigPath": "tsconfig.json"
+  "include": ["src/components/**/*.{tsx,ts}", "src/index.ts"],
+  "tsconfigPath": "tsconfig.json",
+  "exports": {
+    ".": "./src/index.ts"
+  }
 }


booc0mtaco:
More information on config settings: https://docs.omlet.dev/cli/config-file

Before:
Screenshot 2023-10-25 at 13 55 45

After:
Screenshot 2023-10-25 at 13 54 20

jeremiah-clothier (approved):

jeremiah-clothier:
> More information on config settings: https://docs.omlet.dev/cli/config-file
> 
> Before: Screenshot 2023-10-25 at 13 55 45
> 
> After: Screenshot 2023-10-25 at 13 54 20

Out of curiosity what is this depicting? There isn't any text labels for the bars so I'm not sure this means but more green seems good

booc0mtaco:
> > More information on config settings: https://docs.omlet.dev/cli/config-file
> > Before: Screenshot 2023-10-25 at 13 55 45
> > After: Screenshot 2023-10-25 at 13 54 20
> 
> Out of curiosity what is this depicting? There isn't any text labels for the bars so I'm not sure this means but more green seems good

Yah i removed the bar graph labels since those projects aren't public-facing at this time. Each graph represents "of the components in the project, how many are our core, EDS components". A few extra things to note here:

* EDS itself (the top bar) isn't 100% EDS 🤔 (this makes sense once you consider the dependencies we use to build our components)
* All projects will have some custom components, like those provided by frameworks like remix, custom recipes that wrap EDS components, etc.
* "The higher the better" generally applies. another thing omlet lets us do is figure out which things people are building on their own, which may be useful for pulling into EDS itself. The UI provides a way to show only custom components in each project, or all projects

PR #1794, opened 2023-10-30
feat: allow typography components to specify any preset values

booc0mtaco:
### Summary:

Components involved:
- `Heading`
- `Text`

Work:
- create a new prop for the relevant type presets (tier 2+)
- mark the previously-defined props as deprecated
- update examples to apply these props in code examples and stories (instead of the previous examples)
- update/remove relevant snapshots

Some clear benefits to this approach:
- makes it so that the typography prop values match those in the requisite designs
- makes it so that the typography prop values match the token names in code
- avoids the complexity of baking in the color tokens as `variant` prop values
- simplifies components and snapshots (once the deprecated props are removed)

Validation steps:
- [x] applying the code to allow this new prop should not cause any snapshot changes
- [x] swap any existing token usages with the new `preset`-based variants does not cause a visual snapshot change

### Test Plan:


booc0mtaco:
ex: `Heading`
Screenshot 2023-10-31 at 10 46 20

booc0mtaco:
Before with `Heading` you had to try and grok how to map what was requested in design with what the code was using:

* sometimes the names would match between the two, and other times you'd have to try and figure out what token `h1` meant
* some of the variant tokens matched existing color tokens, and other times they did not
* one could theoretically set a variant token, then use a tailwind utility or other class to override the token, which bifurcates the API

Screenshot 2023-10-31 at 10 55 46

timzchang (approved):
<3

booc0mtaco:
doing a merge commit so that we get two lines in the release notes, one for each component modified (and the docs PR doesn't increment the version on its own)

PR #1798, opened 2023-11-06
Node 20 + Yarn 4

ahuth:
This PR updates Node to v20 and Yarn to v4.


.yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
```diff
@@ -1,28 +0,0 @@
-/* eslint-disable */
-//prettier-ignore
-module.exports = {
-name: "@yarnpkg/plugin-workspace-tools",
```
ahuth, line 4: The workspace tools are now built in to Yarn itself.

.yarnrc.yml
```diff
@@ -1,9 +1,7 @@
-nodeLinker: node-modules
+compressionLevel: mixed
```
ahuth, line 1: The `yarn set version latest` command made these changes.

Interestingly, the default for `compressionLevel` is 0, not mixed, per https://yarnpkg.com/configuration/yarnrc#compressionLevel 🤔

booc0mtaco (approved):
I forget the command to use to install the latest version of `node`. for `nodenv` users, what should we run?

ahuth:
Thanks 🙏

PR #1801, opened 2023-11-07
chore: add a bit of delay to reduce Modal test flakiness

ahuth:
Noticed [some flakiness with the Modal's interactive examples tests](https://github.com/chanzuckerberg/edu-design-system/actions/runs/6788858470/job/18454751418). Fixed here by adding a small delay.

Alternatively could try to wait for a specific class. But I don't think there's a good one. The classes in question seem to be generated, and I don't want to rely on them.


booc0mtaco:
Thanks for this. It started happening after some version update last week, which is peculiar 🤔

booc0mtaco (approved):

ahuth:
Yeah, kind of odd. Maybe headless updated something.

PR #1802, opened 2023-11-07
chore: update rollup to v4

ahuth:
Update Rollup to v4.

Tested by running `yarn build`, looking at the lib/ dir, and linking into edu-stack and running it there.


booc0mtaco:
@ahuth Nice! I suppose one use case of the size-limit action is that it is telling us here that the resulting lib files didn't change with the rollup upgrade.

booc0mtaco (approved):

ahuth:
True

PR #1816, opened 2023-12-06
fix: remove duplicate addon-docs loading

booc0mtaco:
### Summary:

The docs addon comes pre-installed with storybook 7.x, so we don't need to install it or configure it directly. This clears up that duplicate install and use (which also cleans up some small output from the console when starting storybook. Slightly quicker now.

As it's now pre-installed, we also don't have to worry about keeping it up to date anymore.

Screenshot 2023-12-05 at 16 12 06

### Test Plan:

  - verify the "Docs" tab works normally on build in CI

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -25,13 +25,6 @@ const config: StorybookConfig = {
     '@storybook/addon-interactions',
     '@geometricpanda/storybook-addon-badges',
     {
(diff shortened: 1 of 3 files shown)


ahuth (approved):

booc0mtaco:
@ahuth thanks !

PR #1819, opened 2023-12-16
refactor(tokens): adjust tokens to be import/export compatible

booc0mtaco:
### Summary:

Part 1 of n (hopefully only need to adjust a few individual token values and formats)

- update token file usages in storybook
- add in missing tokens attributes to close needed gaps on import
- updates to usages NOT resulting in visual changes
- add "group" values to exportable tokens to mark types
- unlink usages for deprecated tokens in deprecated components

### Test Plan:

  - verified that no snapshots changed for component usages


.storybook/components/DesignTokens/Tier1/Colors.tsx
```diff
@@ -29,9 +29,6 @@ export const Tier1Colors = () => {
           listItems={getListItems('eds-color-other-yellow', 'yellow')}
         />
         <ColorList listItems={getListItems('eds-color-other-ruby', 'ruby')} />
-        <ColorList
```
booc0mtaco, line 32: Removing as this token breaks the convention. Hiding the docs in preparation for removal

.storybook/components/DesignTokens/Tier3/TypographyComponents.stories.tsx
```diff
@@ -3,12 +3,13 @@ import { at, capitalize, forEach, merge } from 'lodash';
 
 import React from 'react';
 import { Grid, Section } from '../../../../src';
-import breadcrumb from '../../../../src/design-tokens/tier-3-component/breadcrumb.json';
-import button from '../../../../src/design-tokens/tier-3-component/buttons.json';
-import form from '../../../../src/design-tokens/tier-3-component/forms.json';
-import link from '../../../../src/design-tokens/tier-3-component/link.json';
-import tab from '../../../../src/design-tokens/tier-3-component/tab.json';
-import tag from '../../../../src/design-tokens/tier-3-component/tag.json';
+import breadcrumb from '../../../../src/design-tokens/tier-3-component/breadcrumb-typography.json';
```
booc0mtaco, line 6: breaking apart the per-component token files to the types of content. style-dictionary rolls up all the token files together so there's nothing relying on the input files to have all the contents like before.

Will continue this pattern, as there can be different pipelines for token generation based on the token content (in this case, colors and references).

.storybook/data/tokens.json
```diff
@@ -88,6 +89,7 @@
   "eds-size-10": "calc(var(--eds-size-base-unit) * 10)",
   "eds-size-11": "calc(var(--eds-size-base-unit) * 11)",
   "eds-size-12": "calc(var(--eds-size-base-unit) * 12)",
+  "eds-size-20": "calc(var(--eds-size-base-unit) * 20)",
```
booc0mtaco, line 92: Adding a value that was discussed with design. Will eventually revert these back to literal values instead of this clever factoring, as we want to allow support of tokens in other formats that aren't parsed by CSS

src/components/Badge/Badge.module.css
```diff
@@ -22,7 +22,7 @@
    */
   --badge-height: calc(var(--eds-size-2-and-half) + 2 * var(--eds-border-width-sm));
 
-  z-index: var(--eds-z-index-100);
+  z-index: 100;
```
booc0mtaco, line 25: Generally KISS'd away any token usages where the token had the value in the token name. What good is a token where the name is NOT the value?

src/components/Layout/Layout.module.css
```diff
@@ -84,8 +84,8 @@
     top: 0;
     left: 0;
     height: 100vh;
-    width: var(--eds-l-sidebar-width);
-    z-index: var(--eds-z-index-500);
+    width: 13.5rem;
```
booc0mtaco, line 87: removing token usages for layout tokens that are deprecated. last step before actual removal

src/components/Skeleton/Skeleton.module.css
```diff
@@ -30,7 +30,7 @@
 }
 
 .skeleton.skeleton-circle {
-  border-radius: var(--eds-border-radius-round);
+  border-radius: var(--eds-border-radius-full);
```
booc0mtaco, line 33: `-round` only works when the object is square. Else it generates an oblique spheroid! `-full` always does what we intended round to do... form a pill-shaped rounded corner

src/design-tokens/tier-1-definitions/layout.json
```diff
@@ -3,15 +3,15 @@
     "l": {
       "max-width": {
         "value": "71.25rem",
-        "comment": "@deprecated Layout tokens are deprecated and will be removed in a future release"
+        "comment": "@deprecated This should not be used in code or design. It will be removed in a future version of EDS."
```
booc0mtaco, line 6: Named all the deprecation comments with the same format, so that we can later find/replace these at the next major

ahuth (approved):
LGTM

PR #1824, opened 2024-01-08
chore(DragDrop): mark component as deprecated

booc0mtaco:
### Summary:

- in its current form, it is far too application-specific to be generally usable
- also, the underlying library is now deprecated and no longer supported
- marking this as not recommended for current use, to move on to another codebase

### Test Plan:

- n/a (component metadata change)


ahuth (approved):

booc0mtaco:
[dnd-kit](https://dndkit.com/) looks like a leading candidate to replace `react-beautiful-dnd`. Running a trial of this internally to check if its adequate for our needs, and what EDS components are needed for a friendly/accessible drag/drop interface.

PR #1825, opened 2024-01-09
fix(Select): update incorrect documentation

booc0mtaco:
### Summary:

- add in proper description for generated form fields based on key/value info
- hide a field from the documentation table that is currently non-fuctional

### Test Plan:

### DIFF
--- a/src/components/Select/Select.stories.tsx
+++ b/src/components/Select/Select.stories.tsx
@@ -6,14 +6,43 @@ import type { OptionsAlignType, VariantType } from './Select';
 import { Select } from './Select';
 import Icon from '../Icon';
 
-export default {
+const meta: Meta<typeof Select> = {
   title: 'Components/Select',
   component: Select,
   parameters: {
     badges: ['1.2'],
     layout: 'centered',
   },
-} as Meta;
+  argTypes: {
+    multiple: {
+      table: {
+        disable: true,
+      },
+    },
+    value: {
+      table: {
+        description: 'The value of the select field (when controlled)',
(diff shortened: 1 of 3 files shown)


src/components/Select/Select.stories.tsx
```diff
@@ -13,6 +13,13 @@ export default {
     badges: ['1.2'],
     layout: 'centered',
   },
+  argTypes: {
+    multiple: {
+      table: {
+        disable: null,
+      },
+    },
+  },
```
jeremiah-clothier, line 22: Was this breaking the storybook from rendering? Or just disabling because that arg doesn't work well with storybook UI?

jeremiah-clothier (approved):

src/components/Select/Select.stories.tsx
booc0mtaco: Hiding this one b/c we will have a forthcoming component for handling multiple selection and didn't want to have this lingering prop visible. More work to do, but closing a documentation mismatch for the moment.

booc0mtaco:
I'll try to coerce storybook into showing a better type table so that one can see the types for `defaultValue` and `value`. Some additional feedback:

- we might export sub-types for any dependent components so they can be viewed/extended properly (this would apply to other components that have subcomponents with expected types so worth thinking about)

booc0mtaco:
Screenshot 2024-01-09 at 18 32 28
Screenshot 2024-01-09 at 18 32 39
Screenshot 2024-01-09 at 18 32 48

PR #1827, opened 2024-01-10
feat(tokens): output token literal values in css-variables-nested.json

booc0mtaco:
### Summary:

- add a new token file to lib and export for tailwind config
- the new file will allow tailwind IDE plugin to show in the UI, use the value instead of a var
- clean up the command interfacing with style-dictionary to support this change
- add clearer internal docs for the token generation helpers in style-dictionary.config.js
- update config parameters for theming tooling
- organizational and documentation updates

### Test Plan:

  - verify no snapshots are changing
  - verify function and format of tailwind config in a consuming project


booc0mtaco:
Screenshot 2024-01-09 at 19 43 00
Screenshot 2024-01-09 at 19 43 27

booc0mtaco:
Here's the big problem with this:

* This will attach the tailwind classes to the raw values of the tokens, bypassing the css variables they come from
* As a result, the theming tools break down (if someone were to set a new value for a token, the tailwind classes are now detached)
* This converts the change from a small feature, into a breaking change, regardless of solutions (existing usage of this file will lead to the problem at hand IF implemented like so)
* The plugin is incapable of parsing the values associated with the variables since it is not the CSSOM and has no `:root` to refer to

Marking this as breaking allows some flexibility. We can implement changes to `eds-apply-theme` to generate a tailwind file, and document setup changes to read from that. using the local theme file would work, so long as the style-dictionary formatters for generating the format are also exported.

booc0mtaco:
Addressed the problem by generating a new file using unlinked values, and added tooling and instructions to help people move to the new usage. In the new usage, the tailwind plugin will still have the literal values instead of the `var()` equivalents, but importantly, will be set up to point to the new config file. 

If a team continues to use the built in file, that one is ALWAYS correct (uses `var()`); they just won't get the IDE benefits.

booc0mtaco:
closing this one in favor of #1828 to link to the right resources via branch name

PR #1837, opened 2024-01-26
feat(Select): add support for label prop

booc0mtaco:
### Summary:

- provide a cleaner API for implementing the label similar to other form fields
- add disabled prop should a user want to still detach and use Select.Label
- update styles to support this
- docs: clean up all Select stories and API table in storybook
- docs: greatly simplify the presentation of Select documentation
- update relevant snapshots

### Test Plan:

### DIFF
--- a/src/components/Select/Select.module.css
+++ b/src/components/Select/Select.module.css
@@ -13,10 +13,12 @@
 
 /**
  * Compact variant.
+ * TODO: remove treatment for compact, as it only adjusts the component width, not its overall padding
  */
 .select--compact {
   width: min-content;
 }
+
 .select-button--compact {
   padding: var(--eds-size-half);
 }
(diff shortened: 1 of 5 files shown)


booc0mtaco:
Screenshot 2024-01-26 at 12 56 37

booc0mtaco:
This will also greatly improve the documentation table, and code snippets in the resulting storybook pages.

([old](https://chanzuckerberg.github.io/edu-design-system/?path=/docs/components-select--docs)/[new](https://61313967cde49b003ae2a860-gzbmpdfzks.chromatic.com/?path=/docs/components-select--docs))

* improvements to per-prop documentation
* clearer code examples per story
* simpler storybook implementation for stories
* documented each story with a block describing what the story implements
* added story for event handling props - this functionality does exist, but adding an example story (cc @czibbelvin)

jeremiah-clothier (approved):

PR #1843, opened 2024-02-08
fix(Select): set trigger type to button to prevent submits

booc0mtaco:
### Summary:

- when a button is in a form tag, it automatically becomes able to submit
- our wrapper doesn't include this which is usually fine
- sometimes during a slow load, the handler preventing propagation allows submit
- fixing this by adding a non-editable type

### Test Plan:

  - verify trigger buttons include type in snapshots

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -442,6 +442,7 @@ export const SelectButtonWrapper = React.forwardRef<
           theirOnClick && theirOnClick(ev);
         }}
         ref={ref}
+        type="button"
         {...other}
       >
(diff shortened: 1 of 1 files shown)


booc0mtaco:
It looks like the tag properly includes this attribute, but since its related to some racing condition, not exactly sure we can get a failing state to appear

jeremiah-clothier (approved):

jeremiah-clothier:
> It looks like the tag properly includes this attribute, but since its related to some racing condition, not exactly sure we can get a failing state to appear

Yeah there's not really a ticket with "steps to reproduce". The fix seems good though

PR #1844, opened 2024-02-08
feat(script): add import from figma script

booc0mtaco:
### Summary:

- read in the output from the Export/Import Variables file
- allow user to select which mode to read from
- update the local theme file with the values
- allow for multi-modal function (use internal and external)
- DX improvements for this CLI of the script via new lib.s

 2024-01-23 - how to process the file

 # Overall rules

 - all keys should be normalized
  - trim spaces
  - toLowerCase

 # ~~output target - internal~~

<details>
<summary>Will do this in a separate PR after the file formats are discussed </summary>

 There are two targets for this output: the primitives file, or the themes file. In the case
 of the primitives, they should be written to the primitives.json file. In the case of the
 theme values, they should be written to the themes.json file. This can be detected by reading
 the top-level name field (inputFile.name equals "Primitives" or "Themes"). Once this is
 determined, it is possible to direct the rest of the reading to the value parser.
</details>

 # output target - external

 There are two input files, but one output file when used in consumers. all values should be
 written to the configured app-theme.json file as needed. In this case detection is not strictly
 needed, but can be useful for console output.

 For internal and external targets, we should store the available modes from the file. The
 mode are read from inputFile.modes. For selection, the user can be given the value, while the
 script will store the key for later usage. e.g., "1:2" with a UI presenting "Test Mode".

 Given the read in value for the file contents name, and the mode, the variable parsing can
 begin.

 # variable parsing

 ## Util Parsers

 ### name

 The name formats vary between code and figma. For consistency, the names that port over from
 figma use slashes instead of dashes, so conversions can be don by using the file name, and the
 variable name.

 Example:

 Given inputFile.name = "Themes" and inputFile.variables[0].name = "color/text/neutral/default",
 we want to write the value to an object matching the named path eds.theme.color.text.neutral.default

 Given inputFile.name = "Primitives" and inputFili.variables[0].name = "color/other/ruby/800",
 we want to write the value to an object matching the named path eds.color.other.ruby.800.

 ### color value

 Colors are stored in this file using an object format

```
 "resolvedValue": {
  "r": number,
  "g": number,
  "b": number,
  "a": number
 }
```

 The values are all listed as between 0 and 1. For the color portions, they represent a
 conversion of a fraction represented by toHex($listedValue / 0.00392157) = x. Combining the
 value for each color can yield a viable format (hex, rgba, etc)

 ## Parsing a value

 All values are listed under the inputFile.variables key, which is a list of objects. Before
 drawing values from any of the individual variables, we should use the mode to determine
 which value to read.

 Each variable object has the following component key/value pairs:
 - id (internal tracking id with name-spacing)
 - name (the name appearing in the Figma UI)
 - description (more information about the variable)
 - type (constant describing the internal data type of the field's value, e.g., COLOR)
 - valuesByMode* (object detailing linkages betwwen variables keyed on mode)
 - resolvedValuesByMode (object maping resolved values keyed on mode)
 - scopes (list of named scopes the variable applies to)
 - hiddenFromPublishing (s/e boolean)
 - codeSyntax (n/a or known)

 * - for this process, ignore

 Using the processed name and value, we can write the new value into the theme file configured
 at the destination in the project.

----

### Test Plan:


bin/eds-import-from-figma.js
```diff
@@ -0,0 +1,258 @@
...
+      });
+    })
+    .option('verbose', {
+      describe: 'Print additional details for debugging purposes',
+      type: 'boolean',
+    })
+    .option('dry-run', {
+      describe: 'Run the script without writing any changes',
+      type: 'boolean',
+    }).argv;
```
booc0mtaco, line 35: Screenshot 2024-02-09 at 17 06 13

bin/eds-import-from-figma.js
```diff
@@ -0,0 +1,258 @@
...
+   * TODO: separate all of the below out into separate file(s)
+   */
+  const stats = {
+    skipped: [],
+    updated: [],
+    errored: [],
+  };
+
+  const spinner = ora('Parsing tokens').start();
+  importTheme.variables.forEach((figmaVariable) => {
```
booc0mtaco, line 106: This is basically the "main loop" and may be factored out into some more generic module at some point. mostly does a few things

Given a token / variable from figma... 

* format the token name for code
* format the path to write to
* convert the format of the token value (using the earlier selected mode)

bin/eds-import-from-figma.js
```diff
@@ -0,0 +1,258 @@
...
+  const yargs = require('yargs/yargs');
+
+  // eslint-disable-next-line import/extensions
+  const { hideBin } = require('yargs/helpers');
+  const { prompt } = require('enquirer');
+  const at = require('lodash/at');
+  const set = require('lodash/set');
+
+  const { default: ora } = await import('ora');
+  const chalk = require('chalk');
```
booc0mtaco, line 14: These are the lib.s used in many popular tools like `husky` to print animated spinners, change the color of text, etc.

bin/eds-import-from-figma.js
jeremiah-clothier: Is the first line of the print statement accurate?

If someone is running this script would they use the command `eds-import-from-figma <file>` and not `eds-import-from-figma.js <file>`

.storybook/components/Docs/Guidelines/Theming.stories.mdx
```diff
@@ -70,6 +70,36 @@ Once run, you will have a set of theme files written to the configured `dest` pa
...
+
+* In Figma, install the [Export/Import Variables Plugin][export-import-plugin].
+
+With the prerequisites set up, you can download a JSON file, containing the existing token definitions and values. Here's how:
+
+* Open the design(s) in Figma
+* Open the Resources Panel in the toolbar (or use <kbd>Shift + I</kbd>)
+* Activate the "Export/Import Variables" plugin by first clicking the "Plugins" tab, then the "Export/Import Variables" option.
+* Once it opens, click the "Export..." button for the Listed collections and save to your local machine
+* Finally, us the above command, by passing in the file path to each collection.
```
jeremiah-clothier, line 89: For testing purposes, would it be value for point solutions to have something like this in CI?

CI Check `ensure theme matches design?`
1. run `npx eds-import-from-figma`
2. check diff of generated file to existing file in version control

bin/eds-import-from-figma.js
```diff
@@ -0,0 +1,258 @@
...
+  let localTheme;
+  try {
+    localTheme = JSON.parse(
+      fs.readFileSync(`${config.src}app-theme.json`, 'utf8'),
+    );
+  } catch (e) {
+    console.error(
+      chalk.red('cannot load local theme file', `${config.src}app-theme.json`),
+    );
+    process.exit(2);
```
jeremiah-clothier, line 59: Q for learning purposes: why does the 1st try-catch `process.exit(1)` but the 2nd try-catch `process.exit(2)`. What's the difference between the two

bin/eds-import-from-figma.js
jeremiah-clothier: This is the "main loop", is there value is pulling this code out to test it?

.storybook/components/Docs/Guidelines/Theming.stories.mdx
booc0mtaco: In order to run in CI, we would need to have the theme file being fetched from figma, or stored somewhere in the consuming project. The former depends on the type of licensing with Figma (e.g., some do not allow access to the REST API, which can let you pull down such a figma JSON file). The latter I'm not yet sure about... 

What would be useful is running `eds-apply-theme` in husky, which would be easy enough to forget to do, OR running that after a successful import. It will take the imported values and generate the outputs. 

Let's chat more about this to see if we are aligned. I do have some followup tasks around integration of this tool, but wanted to get the core of it in place for discussion and testing

bin/eds-import-from-figma.js
booc0mtaco: It's mostly a UNIX-y thing. I was going to add in a variable where I would apply [`<<=`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Left_shift_assignment), so that there are "meaningful" exit codes that could be documented or used elsewhere.

ahuth: Yeah, it's a way to distinguish between different types of errors. Maybe not strictly needed, but I'll never say no to Unix-y things

bin/eds-import-from-figma.js
```diff
@@ -0,0 +1,258 @@
...
+  const jsonfile = require('jsonfile');
+  const yargs = require('yargs/yargs');
+
+  // eslint-disable-next-line import/extensions
+  const { hideBin } = require('yargs/helpers');
+  const { prompt } = require('enquirer');
+  const at = require('lodash/at');
+  const set = require('lodash/set');
+
+  const { default: ora } = await import('ora');
```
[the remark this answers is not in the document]
booc0mtaco: Correct. This CLI tool's latest version is just ESM, and it was a hassle to try and move the whole thing over. I'm tempted to use the last version that supports `require` to avoid this one awkward line until this settles down

booc0mtaco:
I posted this a bit early to get some initial feedback. there are a few additional things I'm looking to do before wrapping this up:

* factor out the utility methods and parts of the script, and add some tests
* separate the UI parts of the script from the actual functional parts

bin/eds-import-from-figma.js
booc0mtaco: Yah for sure. what will likely happen is that at some point in the future, the output from figma will change and break some/all of this. Having tests will mean we can update the data used in those instead of poking around the whole script to make updates

bin/eds-import-from-figma.js
booc0mtaco: `yargs` will figure out the command to run dynamically, based on how the command was triggered. It will be a different command once this is packaged, BUT a bit unclear what will show on screen

booc0mtaco:
@jeremiah-clothier @ahuth added tests and cleaned up things. "ready" ready for review now

bin/_util.js
```diff
@@ -140,4 +140,114 @@ module.exports = {
...
   },
+  /**
+   * Determine the write path by taking the collection and variable name, and looking it up in
+   * the existing local theme. If there's a path in the local theme file, we can write there (using lodash/set
+   * or similar).
+   *
+   * @param {object} localTheme JSON file loaded, representing the data for the local theme
+   * @param {string} collectionName Name of the exported collection
+   * @param {string} variableName current variable name from figma (e.g., color/text/neutral/default)
+   * @returns string|null representation of the path to write to in the local theme JSON file
```
ahuth, line 151:
```suggestion
   * @returns {string|null} representation of the path to write to in the local theme JSON file
```

bin/_util.js
```diff
@@ -140,4 +140,114 @@ module.exports = {
       }
     }
   },
+  /**
```
ahuth, line 143: Nit: I'd be tempted to put the utils under `src/` and point the `edu-import-from-figma` script at `build/`, so that I could use TS for these.

package.json
```diff
@@ -92,18 +93,24 @@
     "@popperjs/core": "^2.11.8",
     "@tippyjs/react": "^4.2.6",
     "@types/lodash": "^4.14.202",
+    "@types/yargs": "^17.0.32",
```
ahuth, line 96: Nit: I think the @types can be dev deps, so consumers don't need to install them.

bin/_util.js
booc0mtaco: I don't mind that. While working on these scripts, I have run into a few ideas to improve them, and will add a note for this as well. Some includes updating the script(s) to be TS, and consolidating them such that they are subcommands instead of individual files (e.g., `npx eds-theme {init|apply|import}`, among other refactors.

One thing I was hoping to have (or make, if it doesn't exist) is types for the figma data, so that any tools around that can be safer.

jeremiah-clothier (approved):

PR #1845, opened 2024-02-09
ci: replace commitlint action with our own implementation

jeremiah-clothier:
### Summary:

Our usage of `wagoid/commitlint-github-action@v5` is causing upgrade to commitlint 18.6.0 to fail (source: https://github.com/chanzuckerberg/edu-design-system/actions/runs/7847650438/job/21417042874?pr=1841). I'm updating the workflow to use own implementation that way we aren't held back from updating libraries.

https://czi-tech.atlassian.net/browse/EFI-1617

### Test Plan:

- [x] Workflow lints 1 commit when only 1 commit is pushed.
- [x] Workflow lints >1 commits when >1 commits are pushed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -119,13 +119,12 @@ jobs:
           cache: 'yarn'
       - name: Install Dependencies 🧶
         run: yarn install --immutable
-
       - name: Validate commits commitlint
         # This workflow can also be triggered via "workflow_call".
         # Since it's a push event we have access to these properties https://docs.github.com/en/webhooks/webhook-events-and-payloads#push
         if: github.event_name == 'push'
         run: |
-          COMMIT_COUNT=$(echo '${{ toJson(github.event.commits) }}' | jq length)
+          COMMIT_COUNT=$(echo '${{ toJSON(github.event.commits) }}' | jq length)
           echo "Number of commits in the push: $COMMIT_COUNT"
 
           yarn commitlint --from HEAD~$COMMIT_COUNT --to HEAD --verbose
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -102,10 +102,30 @@ jobs:
       - name: Build design tokens 🛠️
         run: yarn build:tokens
 
-      - name: Lint Commit Messages 👕
-        uses: wagoid/commitlint-github-action@v5
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
       - name: Lint Javascript + SCSS 👕
         run: yarn run lint --max-warnings=0
+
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -102,10 +102,29 @@ jobs:
...
+        run: yarn install --immutable
+      - name: Validate commits commitlint
+        # This workflow can also be triggered via "workflow_call".
+        # Since it's a push event we have access to these properties https://docs.github.com/en/webhooks/webhook-events-and-payloads#push
+        if: github.event_name == 'push'
+        run: |
+          COMMIT_COUNT=$(echo '${{ toJSON(github.event.commits) }}' | jq length)
+          echo "Number of commits in the push: $COMMIT_COUNT"
+
+          yarn commitlint --from HEAD~$COMMIT_COUNT --to HEAD --verbose
```
jeremiah-clothier, line 130: Credit to https://commitlint.js.org/#/guides-ci-setup?id=github-actions but I adjusted code to account for multiple commits being pushed up at once. Instead of using `HEAD~1` we use `HEAD~$COMMIT_COUNT`

.github/workflows/test.yml
```diff
@@ -102,10 +102,29 @@ jobs:
...
+          node-version-file: '.node-version'
+          cache: 'yarn'
+      - name: Install Dependencies 🧶
+        run: yarn install --immutable
+      - name: Validate commits commitlint
+        # This workflow can also be triggered via "workflow_call".
+        # Since it's a push event we have access to these properties https://docs.github.com/en/webhooks/webhook-events-and-payloads#push
+        if: github.event_name == 'push'
+        run: |
+          COMMIT_COUNT=$(echo '${{ toJSON(github.event.commits) }}' | jq length)
```
booc0mtaco, line 127: How/where does `jq` get installed? Mostly out of curiousity...

booc0mtaco (approved):
LGTM. should we not use this provided action across the board now? anyone else using that?

.github/workflows/test.yml
jeremiah-clothier: Great question! Github runners (the github hosted virtual machines we run workflows on) come with a set of preinstalled software. This can be seen looking at the logs of the github action. Here's the link in the image below for reference https://github.com/actions/runner-images/blob/ubuntu22/20240204.1/images/ubuntu/Ubuntu2204-Readme.md


Screenshot 2024-02-09 at 3 12 15 PM

jeremiah-clothier:
> LGTM. should we not use this provided action across the board now? anyone else using that?

I didn't check that, good idea. It doesn't look like the commitlint github action is being used much in the org. Only 3 repos (this repo, 1 archived repo, 1 other)

PR #1847, opened 2024-02-15
feat(Select): enable multi-select

booc0mtaco:
### Summary:

- loosen types to be based more closely match HeadlessUI
- add story for multi select
- update styles for selected and unselected items
- remove need for SelectOption

### Test Plan:

- [ ] Wrote [automated tests](https://czi.atlassian.net/wiki/x/Hbl1H)
- [ ] CI tests / new tests are not applicable
- [ ] Manually tested my changes, but I want to keep the details secret
- [x] Manually tested my changes, and here are the details:
  - Create an [alpha publish](https://github.com/chanzuckerberg/edu-design-system/blob/main/docs/PUBLISHING.md#alpha-release) and try out in `edu-stack` or `traject` as a sanity check if changes affect build or deploy, or are breaking, such as token changes, widely used component updates, hooks changes, and major dependency upgrades.

PR #1848, opened 2024-02-15
feat(Select): enable multi-select

booc0mtaco:
### Summary:

- loosen types to be based more closely match HeadlessUI
- add story for multi select
- update styles for selected and unselected items
- remove need for SelectOption

### Remaining Tasks

- [x] Update storybook doc.s to represent types showing in table (e.g., `SelectOption` is now gone)
- [x] Add details on how form fields manifest when used
- [x] Create alpha version to test this in practice with potential consumers
- [x] Fix glitch with positioning/size of selected and unselected `PopoverListItem` rows, so that when the check is added/removed, the height doesn't change slightly

### Test Plan:

### DIFF
--- a/src/components/Select/Select.stories.tsx
+++ b/src/components/Select/Select.stories.tsx
@@ -430,6 +430,41 @@ export const Multiple: StoryObj = {
 };
 
 /**
+ * The component provides some basic styles to handle long text in the provided field. Use
+ * `shouldTruncate` on `.ButtonWrapper` to truncate the text with an ellipsis.
+ */
+export const MultipleWithTruncation: StoryObj = {
+  args: {
+    ...Default.args,
+    label: 'Favorite Animal(s)',
+    multiple: true,
+    'data-testid': 'dropdown',
+    defaultValue: [exampleOptions[0]],
+    className: 'w-60',
+    name: 'standard-button',
+    children: (
+      <>
+        <Select.Button>
+          {({ value, open, disabled }) => (
+            <Select.ButtonWrapper isOpen={open} shouldTruncate>
+              {value.length > 0 ? value.length : 'none'} long selected
+              description
+            </Select.ButtonWrapper>
+          )}
+        </Select.Button>
+        <Select.Options>
+          {exampleOptions.map((option) => (
+            <Select.Option key={option.key} value={option}>
+              {option.label}
+            </Select.Option>
+          ))}
+        </Select.Options>
+      </>
+    ),
+  },
+};
+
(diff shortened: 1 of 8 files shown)


booc0mtaco:
re: #1847 (contains demo video)

src/components/Select/Select.stories.tsx
```diff
@@ -405,6 +407,44 @@ export const StyledUncontrolled: StoryObj = {
   },
 };
 
+/**
+ * You can select multiple values by passing `multiple` to the parent element. When doing this,
+ * make sure all props that use the value (e.g., `value` and `defaultValue`) should use an array instead
+ * of an object or value for the individual `Select.Option` entries.
+ *
+ * When handling the button text, `value` represents the data for all options selected. This allows for a flexible
+ * layout to fit the needs of the design.
```
booc0mtaco, line 416: TODO: add documentation and/or another story showing what fields get generated when multiple values are selected.

booc0mtaco:
https://github.com/chanzuckerberg/edu-design-system/assets/3056447/40958537-925e-4c05-a008-8f176f18f85c

booc0mtaco:
Verified that this doesn't regress any existing usages, or cause type issues. Worth noting that the type now changes from a provided `SelectOption` internal type to `any` as inherited from HeadlessUI

timzchang (approved):

booc0mtaco:
@jeremiah-clothier Let me know if you noticed any problems in your testing. I think there was one oddity that cropped up, but otherwise things in the alpha worked as expected?

jeremiah-clothier (approved):

booc0mtaco:
will merge as a merge commit to preserve the commit hash for the alpha release

jeremiah-clothier:
> @jeremiah-clothier Let me know if you noticed any problems in your testing. I think there was one oddity that cropped up, but otherwise things in the alpha worked as expected?

Yeah, I remember talking about the shrinking "Expand" icon and how a `flex-shrink: 0` should resolve the issue. Is that going to be added later or already fixed here?

booc0mtaco:
@jeremiah-clothier that's the one. i will add a fix here, since this feature will expose the behavior more obviously

booc0mtaco:
Screenshot 2024-02-20 at 12 00 29

Exposing [optional] truncation handling on `.ButtonWrapper` via `shouldTruncate`. Optional now, so that it doesn't interfere with any overrides people are using currently. In a future version, we can make this default to `true` with other changes.

PR #1852, opened 2024-02-23
Lodash tree shaking

ahuth:
This PR updates our usage of Lodash so that we replace named imports

```ts
import {at} from 'lodash';
```

with deep ones

```ts
import at from 'lodash/at';
```

This should help tree shaking work better. I don't think this is a problem in our existing Remix apps (tree shaking works properly), but I'm running into an issue when trying to use ESM modules in Remix apps.

---

Alternatives:
- Use `lodash-es` instead (which is esm). Downside of that is we'd need to have our existing Remix apps pre-bundle lodash so that it's cjs.
- Maybe there's some Vite config I'm missing to make this work

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -95,11 +95,17 @@
     },
     {
       "files": ["src/**/*.{ts,tsx}"],
-      "excludedFiles": ["src/**/*.stories.{ts,tsx}"],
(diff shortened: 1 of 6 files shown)


booc0mtaco (approved):
Good catch. Looks like we don't have many outliers.

PR #1853, opened 2024-02-23
ESM support

ahuth:
This PR fixes the ESM build so that it works with Vite, which Remix will use by default in v3.

(Alternatively we could have Vite process the CJS build. My first choice is to make the ESM build work, though).

Testing:
- [x] Along 
   along eds icon cjs
- [x] CJS build in a Remix app
   edu-stack eds cjs
- [ ] ~~ESM build in a Remix app~~ (didn't work because of Lodash. The EDS types worked, though, which is a good sign.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@chanzuckerberg/eds",
-  "version": "13.11.0",
+  "version": "13.11.1-alpha.1",
   "description": "The React-powered design system library for Chan Zuckerberg Initiative education web applications",
   "author": "CZI <edu-frontend-infra@chanzuckerberg.com>",
   "homepage": "https://github.com/chanzuckerberg/edu-design-system",
--- a/package.json
+++ b/package.json
@@ -40,9 +40,8 @@
     "eds-init-theme": "bin/eds-init.js"
   },
   "scripts": {
-    "build": "yarn build:clean && yarn build:tokens && yarn build:declarations && yarn build:js && yarn copy-fonts-to-lib",
+    "build": "yarn build:clean && yarn build:tokens && yarn build:js && yarn copy-fonts-to-lib",
     "build:clean": "rm -rf lib/",
-    "build:declarations": "tsc --emitDeclarationOnly --project tsconfig.build.json",
     "build:tokens": "rm -rf src/tokens-dist/ && node ./style-dictionary.config.js && yarn prettier-tokens-dist",
     "build:js": "rollup --config",
     "build:storybook": "storybook build -o storybook-static -s src/design-tokens/tier-1-definitions/fonts",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -7,14 +7,19 @@
...
-      "import": "./lib/index.js",
-      "require": "./lib/index.cjs"
+      "import": {
+        "types": "./lib/index.d.ts",
+        "default": "./lib/index.mjs"
+      },
+      "require": {
+        "types": "./lib/index.d.ts",
+        "default": "./lib/index.cjs"
+      }
```
ahuth, line 17: I got the suggestions from https://publint.dev/@chanzuckerberg/eds@13.11.0

package.json
```diff
@@ -40,9 +45,8 @@
     "eds-init-theme": "bin/eds-init.js"
   },
   "scripts": {
-    "build": "yarn build:clean && yarn build:tokens && yarn build:declarations && yarn build:js && yarn copy-fonts-to-lib",
+    "build": "yarn build:clean && yarn build:tokens && yarn build:js && yarn copy-fonts-to-lib",
     "build:clean": "rm -rf lib/",
-    "build:declarations": "tsc --emitDeclarationOnly --project tsconfig.build.json",
```
ahuth, line 45: Looks like the rollup build already produces type declaration files, so this step is redundant.

rollup.config.mjs
```diff
@@ -10,6 +10,7 @@ export default {
   output: [
     {
       dir: 'lib',
+      entryFileNames: '[name].mjs',
```
ahuth, line 13: Using .mjs for now so bundlers interpret them as esm.

One day EDS will only have one build, and we can go back to .js then.

package.json
```diff
@@ -7,14 +7,19 @@
...
   "exports": {
     ".": {
-      "import": "./lib/index.js",
-      "require": "./lib/index.cjs"
+      "import": {
+        "types": "./lib/index.d.ts",
+        "default": "./lib/index.mjs"
+      },
+      "require": {
+        "types": "./lib/index.d.ts",
```
ahuth, line 15: Technically we're supposed to have different type declaration files for esm (.d.mts) vs cjs (.d.cjs).

I'm hoping to get away with just one set, but will test this out to make sure it's workign as expected.

package.json
```diff
@@ -1,14 +1,20 @@
...
   "description": "The React-powered design system library for Chan Zuckerberg Initiative education web applications",
   "author": "CZI <edu-frontend-infra@chanzuckerberg.com>",
   "homepage": "https://github.com/chanzuckerberg/edu-design-system",
   "license": "MIT",
   "exports": {
     ".": {
-      "import": "./lib/index.js",
-      "require": "./lib/index.cjs"
+      "import": {
+        "types": "./lib/index.d.ts",
```
ahuth, line 11: Need to define `types` here for TypeScript's `moduleResolution: "bundler"` config.

PR #1856, opened 2024-02-23
Only produce one build

ahuth:
Instead of producing 2 builds (cjs + esm), only produce one (cjs).

Why? Well, [supporting dual builds is notoriously hard](https://blog.isquaredsoftware.com/2023/08/esm-modernization-lessons/).

For my specific case, I'm working on making EDS support Vite in Remix
- I'm finding it challenging to make it work
- Especially getting Vite to understand EDS's internal imports
- I think it'll be easier to have just a CJS build and take advantage of [Vite's dependency pre-bundling](https://vitejs.dev/guide/dep-pre-bundling.html) to convert to ESM

ESM is the future, and one day (like this year sometime) we should be able to use ESM in all our Remix apps and then change EDS to export esm modules directly.

TODO:
- [ ] Publish an alpha version
- [ ] Test in edu-stack


jeremiah-clothier:
> Dang, why commitlint failing 🤔

Yeahhh I haven't gotten around to fixing this. I think commit messages with `'` cause a parsing error

package.json
```diff
@@ -40,9 +38,8 @@
     "eds-init-theme": "bin/eds-init.js"
   },
   "scripts": {
-    "build": "yarn build:clean && yarn build:tokens && yarn build:declarations && yarn build:js && yarn copy-fonts-to-lib",
+    "build": "yarn build:clean && yarn build:tokens && yarn build:js && yarn copy-fonts-to-lib",
     "build:clean": "rm -rf lib/",
-    "build:declarations": "tsc --emitDeclarationOnly --project tsconfig.build.json",
```
ahuth, line 45: Rollup is already producing type definitions, so this step is redundant. The types generated by each step are exactly the same.

rollup.config.mjs
```diff
@@ -7,24 +7,14 @@ import postcss from 'rollup-plugin-postcss';
...
-    },
-  ],
+  output: {
+    dir: 'lib',
+    format: 'cjs',
+    preserveModules: true,
+    preserveModulesRoot: 'src',
+    sourcemap: true,
+    interop: 'auto',
+  },
```
ahuth, line 17: For posterity: there's an interesting thing here (including in the current EDS) where rollup is tree shaking out `index.js` files, but leaving the types for them...

So for example, the lib/Button directory ends up looking like this:

```
lib/Button
  Button.js
  Button.d.ts
  index.d.ts <- There's an index.d.ts, but no corresponding index.js!
```

But, crucially, the import/export statements in the output are such that everything still works.

Basically, when doing `import {Button} from 'eds'`
- Node or bundlers load Button.js directly
- TypeScript thinks you're loading Button through index.js (which doesn't exist)
- But everything still works

Just an oddity I noticed and wanted to note.

We could stop that behavior by adding `treeshake: false` to the rollup config. Doing that results in everything lining up, but prints some warnings and I'm not sure if there are any ramifications or not.

booc0mtaco: @ahuth we can create a ticket to disable tree-shaking and to fix the warnings, if that's the right direction to go for now? thoughts?

booc0mtaco (approved):
LGTM. Should be trivial to modify/revert if we discover any problems with this one as well before merging back to `next`

rollup.config.mjs
ahuth: Good idea. I'll create it.

PR #1858, opened 2024-02-26
fix(Select): add max height for Select.Options

booc0mtaco:
### Summary:

before, it was possible for options to extend beyond the full height of the page, or be taller than the page. Make it so that its height is capped at a maximum of 25% screen state. Add tests to show the keyboard nav works properly

### Test Plan:


package.json
```diff
@@ -132,6 +132,7 @@
     "@storybook/addon-interactions": "^7.6.16",
     "@storybook/addon-links": "^7.6.16",
     "@storybook/addon-styling": "^1.3.7",
+    "@storybook/jest": "^0.2.3",
```
ahuth, line 135: Does `@storybook/test` work? `@storybook/jest` is being deprecated in favor of /test

ahuth: Not sure if it'll work or not with our setup. But probably worth checking

booc0mtaco: Sure, I can try that. The linting recommended installing this 🤔

src/components/Select/Select.stories.tsx
```diff
@@ -474,6 +475,57 @@ export const AdjustedWidth: StoryObj = {
...
+  play: async (playOptions) => {
+    const canvas = within(playOptions.canvasElement);
+    const selectButton = await canvas.findByRole('button');
+
+    await openMenu(playOptions);
+    await userEvent.keyboard('{ArrowDown}{ArrowDown}{ArrowDown}{ArrowDown}');
+
+    await waitFor(async () => {
+      await expect(selectButton.getAttribute('aria-expanded')).toEqual('true');
+    });
```
ahuth, line 519: Do we need the wrapping `waitFor` because it takes time for the attribute to get set?

booc0mtaco: I'm doing one final test on this. It might not be needed to wait actually; since the popover isn't in the normal DOM hierarchy, chromatic's snapshot is clamped to just the trigger/button. it was perfectly cropped out looks like.

The tell is that the chevron in the snapshot is pointing in the correct direction for it to be open 😅 . I can probably remove `@storybook/test` again, and remove this `waitFor`

booc0mtaco: `waitFor` wasn't needed afterall, but i will keep `@storybook/test` since that let's us assert in the `play` functions (i think we can do more of that in places)

ahuth (approved):
LGTM

booc0mtaco:
https://github.com/chanzuckerberg/edu-design-system/assets/3056447/8ae390da-5069-4dcf-897f-8fde924e94a5

booc0mtaco:
@ahuth thanks 🙏

PR #1861, opened 2024-02-27
feat(Tooltip)!: remove deprecated props

booc0mtaco:
### Summary:

- remove `align` prop, in place of the better-named `placement` prop

This has been marked as deprecated for a while now, so cleaning up the code associated with this. In order to address in a consuming project, you can replace `align` with `placement` in all use cases (this was happening internally already).

- remove `variant` prop

The prop was used for a partially-implemented handling of light/dark mode, and had been marked as deprecated. Removing now, making usage of the light token values as the current, and only, option.

- clean up documentation, storybook config, and test snapshots

### Test Plan:

- [x] clean up existing tests and usages

### DIFF
--- a/src/components/Slider/Slider.tsx
+++ b/src/components/Slider/Slider.tsx
@@ -158,7 +158,6 @@ export const Slider = ({
       />
       {tooltip && (
         <Tooltip
-          align="top"
           appendTo="parent"
           hideOnClick={false}
           offset={({ reference }) => {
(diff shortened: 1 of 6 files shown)


booc0mtaco:
Screenshot 2024-02-27 at 13 32 13

According to analytics, `variant` was unused so safer to remove. `align` is used just a few times, so will document how to address in release notes.

jeremiah-clothier (approved):

ahuth (approved):

PR #1862, opened 2024-02-27
feat(Tag)!: remove deprecated props and update API

booc0mtaco:
### Summary:

- remove deprecated value "yield" from variant list

We had both "warning" and "yield" as possible values for this component's variants, which was confusing. Going forward, we will start with just "warning", and remove this option from the component list. If using "yield", please consider using "warning" instead, or some other treatment.

- change "icon" prop to be of type `IconName` to be consistent with all other components

This syncs up the headless Icon handling across all components, meaning that the names and possible values are determined by the mapped icons, and not any possible ReactNode instance. For consumers, replace any usages of `<Icon />` with the string provided to this cmoponent. Any other nodes used will have to be removed.

- tidy up usage of `<Text />` and typography handling

simplify the handling of type in this component by moving to use `preset` instead of the more `<font>` like weight and size props, which will also be removed.

- update tests, snapshots, and usages.

### Test Plan:

- [x] update existing stories and tests

### DIFF
--- a/src/components/Tag/Tag.module.css
+++ b/src/components/Tag/Tag.module.css
@@ -41,8 +41,6 @@
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: nowrap;
-
-  font: var(--eds-theme-typography-tag);
 }
 
 /**
@@ -70,12 +68,6 @@
   --tag-primary-color: var(--eds-theme-color-text-utility-warning);
   --tag-secondary-color: var(--eds-theme-color-background-utility-warning);
   --tag-outline-color: var(--eds-theme-color-border-utility-warning-default);
-}
-
-:where(.tag--yield) {
-  --tag-primary-color: var(--eds-theme-color-text-grade-revise);
-  --tag-secondary-color: var(--eds-theme-color-background-grade-revise-subtle);
-  --tag-outline-color: var(--eds-theme-color-border-grade-revise-default);
 }
 
 :where(.tag--brand) {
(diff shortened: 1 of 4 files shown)


booc0mtaco:
@jeremiah-clothier will try using `size` or `distinct_size` ([referenced in the doc.s](https://docs.github.com/en/rest/using-the-rest-api/github-event-types?apiVersion=2022-11-28#event-payload-object-for-pushevent)) to detect the number of commits to see how far back to reach for `commitlint`. Wondering if you tried either of those already, and they had some issue?

jeremiah-clothier:
> @jeremiah-clothier will try using `size` or `distinct_size` ([referenced in the doc.s](https://docs.github.com/en/rest/using-the-rest-api/github-event-types?apiVersion=2022-11-28#event-payload-object-for-pushevent)) to detect the number of commits to see how far back to reach for `commitlint`. Wondering if you tried either of those already, and they had some issue?

Ohh maybe, I can look at this Thursday again. Rn the only workaround is to just not use any `'` characters in commits. I'm pretty sure that's the issue

booc0mtaco:
> > @jeremiah-clothier will try using `size` or `distinct_size` ([referenced in the doc.s](https://docs.github.com/en/rest/using-the-rest-api/github-event-types?apiVersion=2022-11-28#event-payload-object-for-pushevent)) to detect the number of commits to see how far back to reach for `commitlint`. Wondering if you tried either of those already, and they had some issue?
> 
> Ohh maybe, I can look at this Thursday again. Rn the only workaround is to just not use any `'` characters in commits. I'm pretty sure that's the issue

Yep, confirmed on two commits (one working and one not working). I'll try that out separate from this. if we merge a nice solution in `next` i can merge that so that it gets over here for this change

PR #1867, opened 2024-02-28
feat(InlineNotification)!: remove deprecated props from component

booc0mtaco:
### Summary:

- remove `inactive` prop from component

This had been marked as deprecated, so now removing. As of today, it is not used in any components, and likely too specific a use case to warrant the additional complexity. Usages of this should be removed from code and associated designs.

- remove `yield` variant

Removing this variant, as it overlaps with the usage of the "warning" prop, but is a different color and treatment. For users of the value, designs and code should move to the "warning" value.

### Test Plan:


booc0mtaco:
Screenshot 2024-02-28 at 17 11 14

jeremiah-clothier (approved):
v14, coming soon 🚀

PR #1870, opened 2024-02-29
feat(Select)!: remove deprecated props

booc0mtaco:
### Summary:

- remove `variant` from component

`variant` was partially implemented and inconsistent with other components which have control for display width (they use boolean `fullWidth`). Remove this prop, and rely on utility classes for width control.

- remove redundant alignment prop `optionsAlign`

Consumers should use the `placement` prop to control how to display the popover. All popover-components share the same API for placement, as determined by PopperJS.

When using `left`, you can set `placement="bottom-start` and when using `right`, you can set `placement="bottom-end`. See more documentation at

https://popper.js.org/docs/v2/constructors/#options

### Test Plan:

### DIFF
--- a/src/components/Select/Select.module.css
+++ b/src/components/Select/Select.module.css
@@ -12,18 +12,6 @@
 }
 
 /**
- * Compact variant.
- * TODO: remove treatment for compact, as it only adjusts the component width, not its overall padding
- */
-.select--compact {
-  width: min-content;
-}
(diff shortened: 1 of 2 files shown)


ahuth (approved):
LGTM

booc0mtaco:
Screenshot 2024-02-29 at 11 29 54

PR #1871, opened 2024-02-29
feat(Link)!: remove and reset deprecated props

booc0mtaco:
### Summary:

- restore `size` to available

This is used sparingly, but will be used more effectively in the future.

- remove `fullWidth` from component

This is a holdover from linkages to `Button`, so removing now that link is its own component. Use utility classes to alter layout.

### Test Plan:

### DIFF
--- a/src/components/Link/Link.module.css
+++ b/src/components/Link/Link.module.css
@@ -525,11 +525,3 @@ button:where(.link--link) {
   padding: 0 var(--eds-size-2);
 }
 
-/**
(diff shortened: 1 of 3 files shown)


jeremiah-clothier (approved):

ahuth (approved):

booc0mtaco:
Almost there, thanks for the help :) Only two more remain and they are the meatier ones ⛏️

PR #1873, opened 2024-03-01
feat(Text)!: remove deprecated props

booc0mtaco:
### Summary:

- remove `size` prop

The `size` prop, being redundant, is now replaced by the more expressive `preset` prop, which contains all the semantic (tier-2) tokens by name. Each of the `size` values corresponds to a specific and existing `preset` value, listed in the following table:

| `size` | `preset` |
|--------|--------|
| body | body-md |
| xs | body-xs |
| sm | body-sm |
| md | Cell |
| lg | body-lg |
| caption-md | caption-md | 
| caption-lg | caption-lg |
| overline | overline |
| callout | body-sm |

In usages of `size`, you can remove the prop. If there is no preset defined for the instance of `Text`, you can use the table above to replace the `size` prop and value with the equivalent preset.

- remove `weight` prop 

Instead of specifying `weight` on any instance of `Text`, you can instead use utility classes or styles to add an override to `font-weight`. The `preset` values specify weight in addition to `font-size`, `line-height`, and other details so overrides like this should be rare.

One example to apply this weight override is to use [`font-*`](https://tailwindcss.com/docs/font-weight#setting-the-font-weight) or use the relevant CSS properties to adjust the weight.

- remove `variant` prop

`variant` overlapped with the utility classes, and served as a namespaced subset of the actual color tokens, in some cases changing the name of the token in the API. We remove variant in favor of utility classes, similar to weight adjustments. For each `variant` value, the equivalent token to use can be found in the following table:

| `variant` | equivalent EDS token |
|--------|--------|
| error | --eds-theme-color-text-utility-error |
| neutral-subtle | --eds-theme-color-text-neutral-subtle |
| neutral-medium | --eds-theme-color-text-neutral-default |
| neutral-strong | --eds-theme-color-text-neutral-strong |
| brand | --eds-theme-color-text-brand-default | 
| info | --eds-color-info-700 |
| inherit | **this is the default behavior now** |
| success | --eds-theme-color-text-utility-success |
| warning | --eds-theme-color-text-utility-warning |
| white | --eds-theme-color-text-neutral-default-inverse |

[Use the following table](https://chanzuckerberg.github.io/edu-design-system/?path=/story/design-tokens-tier-2-usage-colors--text) to match design intent to code equivalent, referencing the table above. For example, if the code currently uses `variant="neutral-medium"`, convert by removing that, and adding `className="text-neutral-default"` or using `color: var(--eds-theme-color-text-neutral-default)` in a stylesheet.

- remove partial definition of `fontSize` from default tailwind config

The existing tailwind config is redundant, as it only partially overlaps the more expressive `preset` field on `Text`, and lacks the precision of the typography tokens. Along with cleaning up these settings, we also remove a mismatch between `Text` and `Heading` (which did not sync up with the associated `preset`s of either component). This restores the default values provided by tailwind.

- update stories and examples
- update component usages in compositions
- [ ] ~~refactor font usages in component styles~~

This removes many of the inlined usages of font tokens in favor of factoring thru `Text`. This avoids some over-application of typography and reduces the number of potentially cascading treatments in component containers

### Test Plan:


ahuth (approved):
LGTM

ahuth (approved):
🤘

PR #1885, opened 2024-03-07
docs: add version tracking to storybook components

booc0mtaco:
this updates how we badge components so we can keep the existing introduction badges, and add new "current" version badges. This will be used to track the versions of individual components between design and code.

Screenshot 2024-03-07 at 17 32 05

### Test Plan:


ahuth (approved):
LGTM 🤘

PR #1886, opened 2024-03-08
feat(Button)!: introduce button v2.0

booc0mtaco:
Introduce **new** Button component, v2.0

WIP:
- old implementation still here with names changed
- no tokens in use just yet
- some TODOs for work next week
- mostly pushing to safeguard against any nasty computer issues and to see what chromatic does

### Test Plan:


src/components/Button/Button.stories.tsx
```diff
@@ -6,7 +6,7 @@ import { SIZES, STATUSES } from '../ClickableStyle';
 import Icon from '../Icon';
 
 export default {
-  title: 'Components/Button',
+  title: 'Components/Button (v1)',
```
booc0mtaco, line 9: Keeping the v1 versions of components around, so that it is easy to do comparisons and avoid snapshot change noise. Once all the component batches are completed, will remove the old versions and annotate the changes between the two.

PR #1889, opened 2024-03-12
feat(Button)!: introduce v2.0 component

booc0mtaco:
**Recreating the PR to have proper integrations kick off**

### Test Plan:

### DIFF
--- a/src/components/Button/Button-v2.tsx
+++ b/src/components/Button/Button-v2.tsx
@@ -1,0 +1,168 @@
+import clsx from 'clsx';
+import type { ReactNode } from 'react';
+import React, { forwardRef } from 'react';
+import type { Size } from '../../util/variant-types';
+import Icon from '../Icon';
+import type { IconName } from '../Icon';
+import LoadingIndicator from '../LoadingIndicator';
+
+import styles from './Button-v2.module.css';
+
+type ButtonHTMLElementProps = React.ButtonHTMLAttributes<HTMLButtonElement>;
+
+type ButtonV2Props = ButtonHTMLElementProps & {
+  // Component API
+  /**
+   * `Button` contents or label.
+   */
+  children: string;
+  /**
+   * Determine the behavior of the button upon click:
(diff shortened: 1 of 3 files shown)


src/components/Button/Button-v2.tsx
```diff
@@ -0,0 +1,167 @@
...
+        ? children
+        : other['aria-label'];
+
+    return (
+      <button
+        aria-label={ariaLabel}
+        className={componentClassName}
+        disabled={isDisabled}
+        ref={ref}
+        tabIndex={isDisabled ? -1 : undefined}
```
ahuth, line 138: I think a button with `disabled` already isn't focusable. Maybe we don't need this?

booc0mtaco: I think you're right... for some reason this was in the old button and I didn't scrutinize it. I will double-check

src/components/Button/Button-v2.tsx
```diff
@@ -0,0 +1,167 @@
...
+    );
+
+    const ariaLabel =
+      iconLayout === 'icon-only' && !other['aria-label']
+        ? children
+        : other['aria-label'];
+
+    return (
+      <button
+        aria-label={ariaLabel}
```
ahuth, line 134: Not saying it's wrong or bad, but it feels weird to always apply an aria-label. Often we try to avoid aria attributes unless we definitely need them.

Do you think we get the same user experience if we only apply ariaLabel when it's explicitly passed in?

booc0mtaco: Yah, that's fair. Reconsidering, I think it would be better to detach the two and add some logic such that "when the layout is `icon-only`, one should add an appropriate `aria-label` or other label to `Button`"

booc0mtaco:
to double-check: disable state for secondary / tertiary buttons

ahuth (approved):
LGTM 🤘

booc0mtaco:
dope, thank you!

PR #1890, opened 2024-03-13
feat(Link)!: introduce v2.0 component

booc0mtaco:
- add stories and styles to match new design
- update snapshots

### Test Plan:

### DIFF
--- a/src/components/Link/Link-v2.tsx
+++ b/src/components/Link/Link-v2.tsx
@@ -1,0 +1,96 @@
+import clsx from 'clsx';
+import React, { forwardRef } from 'react';
+import type { Size } from '../../util/variant-types';
+import type { IconName } from '../Icon';
+import Icon from '../Icon';
+
+import styles from './Link-v2.module.css';
+
+type LinkHTMLElementProps = Omit<
+  React.AnchorHTMLAttributes<HTMLAnchorElement>,
+  'disabled'
+>;
+
+export type LinkProps = LinkHTMLElementProps & {
+  // Component API
+  /**
+   * Component used to render the element. Meant to support interaction with framework navigation libraries.
+   *
+   * **Default is `"a"`**.
+   */
+  as: string | React.ElementType;
+  /**
+   * The link contents or label.
+   */
+  children: string;
+  // Design API
+  /**
+   * Where `Link` sits alongside other text and content:
+   *
+   * * **inline** - Inline link inherits the text size established within the `<p>` paragraph they are embedded in.
+   * * **standalone** - Users can choose from the available sizes.
+   */
+  context?: 'inline' | 'standalone';
+  /**
+   * (trailing) icon to use with the link
+   */
+  icon?: Extract<IconName, 'chevron-right' | 'open-in-new'>;
+  /**
+   * Extra or lowered colors added to a link
+   */
+  emphasis?: 'default' | 'high' | 'low';
+
+  /**
+   * Link size inherits from the surrounding text.
+   */
+  size?: Extract<Size, 'xs' | 'sm' | 'md' | 'lg' | 'xl'>;
+};
+
+/**
+ * `import {Link} from "@chanzuckerberg/eds";`
+ *
+ * Component for making styled anchor tags. Links allow users to navigate within or between a web page(s) or app(s).
+ *
+ */
+export const Link = forwardRef<HTMLAnchorElement, LinkProps>(
+  (
+    {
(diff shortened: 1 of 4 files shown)


src/components/Link/Link-v2.tsx
```diff
@@ -0,0 +1,96 @@
...
+>;
+
+export type LinkProps = LinkHTMLElementProps & {
+  // Component API
+  /**
+   * Component used to render the element. Meant to support interaction with framework navigation libraries.
+   *
+   * **Default is `"a"`**.
+   */
+  as: string | React.ElementType;
```
booc0mtaco, line 21: Adding this to support links link in Remix. Will likely need some work to support fully, but will be able to test once i get thru the first wave of component implementations

ahuth: Challenge I see: Remix links currently use `to` instead of `href`

So to enable passing in a Remix link to here, we'd probably need to figure out how to support passing the EDS Link's `href` prop to the Remix Link's `to` somehow.

src/components/Link/Link-v2.tsx
```diff
@@ -0,0 +1,96 @@
...
+  size?: Extract<Size, 'xs' | 'sm' | 'md' | 'lg' | 'xl'>;
+};
+
+/**
+ * `import {Link} from "@chanzuckerberg/eds";`
+ *
+ * Component for making styled anchor tags. Links allow users to navigate within or between a web page(s) or app(s).
+ *
+ */
+export const Link = forwardRef<HTMLAnchorElement, LinkProps>(
```
ahuth, line 55: Can't wait until React removes `forwardRef` in React 19

src/components/Link/Link-v2.tsx
```diff
@@ -0,0 +1,96 @@
...
+import React, { forwardRef } from 'react';
+import type { Size } from '../../util/variant-types';
+import type { IconName } from '../Icon';
+import Icon from '../Icon';
+
+import styles from './Link-v2.module.css';
+
+type LinkHTMLElementProps = Omit<
+  React.AnchorHTMLAttributes<HTMLAnchorElement>,
+  'disabled'
```
ahuth, line 11: What's omitting `disabled` all about? I thought `disabled` already wasn't an attribute on anchor elements?

src/components/Link/Link-v2.tsx
booc0mtaco: I want to replicate what we do [using `ClickableStyle`](https://github.com/chanzuckerberg/edu-stack/blob/main/app/components/Link.tsx) and also get rid of that component in the next version :). 

I didn't want to copy how `ClickableStyle` works in here, b/c it seems to lose all the typing information and converts everything to `any` :( What does make sense over there is that it uses a type param to combine the types of Remix's `Link` with the component props.

I'm guessing this was part of why you needed to export the types earlier?

src/components/Link/Link-v2.tsx
booc0mtaco: Good catch, this is probably a carry-over from the `ClickableStyles` stuff that wasn't removed a while back.

src/components/Link/Link-v2.tsx
ahuth: I agree on that goal, and not wanting to do ClickableStyle.

My point isn't that we shouldn't do this. It's that to achieve:

```ts
<EdsLink
  as={RemixLink}
  href={...} // or maybe to={...}?
/>
```

we'll need to figure out how to handle href vs to.

booc0mtaco: > I agree on that goal, and not wanting to do ClickableStyle.
> 
> My point is that to achieve:
> 
> ```ts
> <EdsLink
>   as={RemixLink}
>   href={...} // or maybe to={...}?
> />
> ```
> 
> we'll need to figure out how to handle href vs to.

Yah, I think you misunderstood. The way `ClickableStyle` supports to/href is weird, but "works". So we are saying the same thing basically.

I want to avoid polluting EDS with anything specific to Remix or some other framework, as well. I like the TypeParam approach (e.g., combine `EDSLink` with Framework type, with option or to `Omit`). Once I get thru the re-theming, I'll try a few things.

src/components/Link/Link-v2.tsx
booc0mtaco: Ohh, what will happen instead? I never read the upcoming docs for things besides HTML/CSS/JS :P

PR #1891, opened 2024-03-13
feat(NumberIcon)!: introduce 2.0 component

booc0mtaco:
- add in new styles
- create new stories for future snapshots

### Test Plan:


booc0mtaco:
from design: will drop the small size

PR #1929, opened 2024-05-06
fix: sync color tokens with design

booc0mtaco:
### Test Plan:

- [x] apply tokens to where color tokens should be used


src/design-tokens/themes.json
```diff
@@ -550,13 +561,13 @@
...
                 "@": {
-                  "value": "{eds.color.neutral.450}"
+                  "value": "{eds.color.neutral.550}"
                 },
                 "hover": {
-                  "value": "{eds.color.neutral.550}"
+                  "value": "{eds.color.neutral.650}"
                 },
                 "active": {
-                  "value": "{eds.color.neutral.650}"
```
booc0mtaco, line 559: Snapshot changes come from these three changes. Any components using medium emphasis tokens will be accepted.

PR #1951, opened 2024-05-17
[WIP]: add `eds-migrate` script for running codemods on major version upgrades

jeremiah-clothier:
### Summary

This PR adds a new command `eds-migrate` that is intended to be used to make migrating major version easier for engineers.

### Test Plan:

- [x] Wrote some test for individual `transforms`
- [ ] TODO: Manually tested my changes, by creating a release and invoking the command on a project


src/migrate/index.js
```diff
@@ -0,0 +1,59 @@
...
+
+  // runMigration is called by a CLI we want the directory
+  // the command is ran in and not the directory of this file
+  const tsconfigPath = path.join(process.cwd(), './tsconfig.json');
+  if (isVerbose) {
+    console.log(`Using the following tsconfig.json file: ${tsconfigPath}`);
+  }
+  const project = new Project({
+    tsConfigFilePath: path.join(tsconfigPath),
+  });
```
jeremiah-clothier, line 42: I really like this feature of ts-morph when compared to jscodeshift, it can use a tsconfig file and then call `project.getSourceFiles()`. From memory when we migrated to typescript we had to pass file paths explicitly to jscodeshift

src/migrate/README.md
jeremiah-clothier, line 1: Starting a 🧵 for 💭 on my mind before the week ends

1) What are people's thoughts on the folder structure. My idea is that `transforms/` are more general codemods and `migrations/` are collections of transforms which are exposed and executable from the cli

2) I'm not a huge fan of writing everything in `.js` and using `jsdoc` comments. It does make executing the `bin/eds-migrate.js` command easy but when working with codemods having typescript is really helpful. I actually ended up just writing the code in TypeScript and then converting it to JS with JSDoc comments

src/migrate/migrations/14-to-15alpha.js
jeremiah-clothier, line 1: No tests at the migration level right now, tests are at the `/transform` level. Would probably be worthwhile once the migration codemod is solidified more in terms of "what the transforms are"

src/migrate/transforms/rename-jsx-import.test.js
```diff
@@ -0,0 +1,128 @@
...
+        },
+      ],
+    });
+
+    expect(sourceFile.getText()).toEqual(dedent`
+      import {Button} from '@chanzuckerberg/eds';
+    `);
+  });
+
+  it.todo('renames JsxElements when imports are changed');
```
jeremiah-clothier, line 127: reminder: todo

src/migrate/README.md
jeremiah-clothier, line 1: Todo: docs for writing codemods and helpful links or resources, navigating an AST is kinda confusing but there are tools that make it less so

booc0mtaco (commented):
Great start! excited to have these as a part of the repo finally.

I can see an argument for having the scripts in TS over JS, so it's worth keeping that as an option. Feel free to publish an alpha version containing these migrations and such once you get to a place to confirm function

src/migrate/README.md
ahuth: src/ should be fine, as long as no transforms nor migrations are referenced by src/index.ts.

I could also see an argument for another top-level directory. Not sure what I would do, though.

ahuth: Re: typescript, one option is to do what axe-storybook-testing does and have bin/eds-migrate [reference files in the build folder](https://github.com/chanzuckerberg/axe-storybook-testing/blob/dd3558500bcc33113f1962318cfc28941d98d66e/bin/axe-storybook.js#L4) instead of source files.

That way the transforms and migrations could still be TS.

ahuth: https://astexplorer.net/

src/migrate/transforms/rename-jsx-import.test.js
```diff
@@ -0,0 +1,128 @@
+const { dedent } = require('ts-dedent');
+const { InMemoryFileSystemHost, Project } = require('ts-morph');
+
+const transform = require('./rename-jsx-import');
+
+describe('transform', () => {
```
ahuth, line 6: Can the test be in TypeScript, even though the source is JS? Would be nice to still get type checking at least for the tests.

src/migrate/transforms/rename-jsx-import.test.js
```diff
@@ -0,0 +1,128 @@
...
+    expect(sourceFile.getText()).toEqual(dedent`
+      import {Button} from '@headlessui/react';
+    `);
+  });
+
+  it('renames imports statements', () => {
+    const host = new InMemoryFileSystemHost();
+    const project = new Project({
+      compilerOptions: undefined,
+      fileSystem: host,
```
ahuth, line 40: It's nice that ts-morph provides an in-memory file system for running tests!

booc0mtaco (commented):
Soft approval (once the build passes). We can merge and iterate as we go.

Let's confirm that it's possible to trigger the scripts in a local checkout (so we don't have to do `.alpha` releases for each test.

jeremiah-clothier:
> Soft approval (once the build passes). We can merge and iterate as we go.
> 
> Let's confirm that it's possible to trigger the scripts in a local checkout (so we don't have to do `.alpha` releases for each test.

Confirmed I can just run the file locally by

1) Running `build:bin` in EDS directory
2) Provide file in terminal in an app's directory `/Users/jclothier/Documents/programming/edu-design-system/bin/eds-migrate.js`

I wasn't able to figure out referencing a local file in apps/group-up/package.json but will try more and document it if I figure it out. The method above isn't ideal for testing in apps because you'd also want new component changes so you'd want to not just reference the script but ideally a local version of the package

cc @booc0mtaco

PR #1973, opened 2024-06-03
feat!: move from isWarning and isError to status prop

booc0mtaco:
- update all (sub-)components to use status instead of two sep. props
- refactor stories to use the new prop
- ensure snapshots do not change (not updating class names at the moment)
- add simple-case migration

### Test Plan:

- [ ] Added code mod for change

### DIFF
--- a/src/bin/migrate/migrations/alpha14-to-alpha15.test.ts
+++ b/src/bin/migrate/migrations/alpha14-to-alpha15.test.ts
@@ -1,0 +1,156 @@
+import { dedent } from 'ts-dedent';
+
+import { updateStatusProp } from './alpha14-to-alpha15';
+import { createTestSourceFile } from '../helpers';
+
+describe('alpha14-to-alpha15', () => {
+  it('updates an appropriate isWarning prop', () => {
+    const sourceFileText = dedent`
+    import {FieldNote} from '@chanzuckerberg/eds';
+
+    export default function Component() {
+      return (
+        <FieldNote isWarning />
+      )
+    }
+    `;
+
+    const sourceFile = createTestSourceFile(sourceFileText);
+
+    updateStatusProp({
+      file: sourceFile,
+    });
+
+    expect(sourceFile.getText()).toEqual(dedent`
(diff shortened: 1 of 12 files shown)


src/bin/migrate/migrations/alpha14-to-alpha15.ts
```diff
@@ -0,0 +1,98 @@
...
+  type SourceFile,
+} from 'ts-morph';
+import { isDesignSystemImport } from '../helpers';
+
+export default function migration(project: Project) {
+  const files = project.getSourceFiles();
+  const sourceFiles = files.filter((file) => !file.isDeclarationFile());
+
+  sourceFiles.forEach((sourceFile) => {
+    updateStatusProp({ file: sourceFile }); // https://github.com/chanzuckerberg/edu-design-system/pull/1973
```
booc0mtaco, line 14: @jeremiah-clothier FYI For this one-off, custom migration, i've put this into one migration file for now. We shouldn't need this particular one as a common case. 

The pattern in this case would be that any X-to-Y.ts file would re-use the common cases, and include the one-off transforms there as well to avoid confusion about longevity.

src/bin/migrate/migrations/alpha14-to-alpha15.test.ts
jeremiah-clothier, line 1: Look great! We should use some consistent naming for the migration files that way [`npx eds-migrate --list` returns some a sorted list](https://github.com/chanzuckerberg/edu-design-system/blob/v15-components-2.0/src/bin/migrate/index.ts#L12-L17). Either naming convention is fine, will let you choose

image

@booc0mtaco

booc0mtaco: @jeremiah-clothier agreed. i'll probably delete this alpha14-to-alpha15 file and test once v15 is released. After all this, there'd only be {old}-to-{new} migrations and the list will be neatly and numerically organized :D

If we do this again, might include the full version suffix for completeness. e.g., 15.alphaX-to-15.alphaY or similar

PR #1999, opened 2024-06-24
fix: P in P

ahuth:
Use a `<span>` instead of `<p>` for Notification subtitles

Our apps render markdown for notifications which contain p tags. This results in p tags nested within p tags, which is not valid HTML and causes React runtime validations to complain.


booc0mtaco (approved):
🎣 ! `div` might also do the trick ?

PR #2001, opened 2024-06-26
fix(Table): allow nodes in header cells

booc0mtaco:
### Test Plan:

Expand type usages that are allowed in `.HeaderCell` instances. This will allow for some flexibility needed.

### DIFF
--- a/src/components/Table/Table.tsx
+++ b/src/components/Table/Table.tsx
@@ -79,7 +79,7 @@ type TableHeaderCellProps = React.HTMLAttributes<HTMLTableCellElement> & {
   /**
    * Child node(s) that can be nested inside component
    */
-  children?: string;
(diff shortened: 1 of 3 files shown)


src/components/Table/Table.tsx
```diff
@@ -253,9 +253,7 @@ const TableHeaderCell = ({
               rank="tertiary"
               size="sm"
               title={iconTitle}
-            >
-              {children}
```
booc0mtaco, line 257: `Button`s allow for `children` to be optional, when using "icon-only".

ahuth (approved):

jeremiah-clothier:
Thanks for getting around to this, HQ week piled up and I didn't end up posting about it in the eds channel

PR #2003, opened 2024-06-28
chore: update to storybook 8

booc0mtaco:
- update docs to support storybook 8 semantics
- fix documentation issues for hiding react nodes for children props
- configure plugins to mesh with new conventions

### Test Plan:


booc0mtaco:
> Screenshot 2024-06-28 at 11 12 32
> 
> This has happened the last few times we updated storybook... some change causes tests which ran normally to suddenly start emitting noise.

There is some difference between `@storybook/test` and `@storybook/testing-library` which leads to these act warnings being generated. There is a migration that makes this change in the repo, which is why it has cropped up a few times in the past. Restoring the dependency and using `@storybook/testing-library` fixes the outputted warnings without requiring any tests to be changed

package.json
```diff
@@ -44,7 +44,7 @@
     "build:clean": "rm -rf lib/",
     "build:tokens": "rm -rf src/tokens-dist/ && node ./style-dictionary.config.js && yarn prettier-tokens-dist",
     "build:js": "rollup --config",
-    "build:storybook": "storybook build -o storybook-static -s src/design-tokens/tier-1-definitions/fonts",
+    "build:storybook": "storybook build -o storybook-static",
```
booc0mtaco, line 47: Confirmed that the fonts get fetched in the UI without this being added

ahuth (approved):

PR #2006, opened 2024-07-05
feat(InputField): add show/hide button for password field type

booc0mtaco:
- add tests and snapshots
- implement show/hide button to appear after any inputWithin elements

### Test Plan:

### DIFF
--- a/src/components/InputField/InputField.tsx
+++ b/src/components/InputField/InputField.tsx
@@ -8,6 +8,7 @@ import type {
   ForwardedRefComponent,
 } from '../../util/utility-types';
 import type { Status } from '../../util/variant-types';
+import Button from '../Button';
 import FieldLabel from '../FieldLabel';
 import FieldNote from '../FieldNote';
 import Icon, { type IconName } from '../Icon';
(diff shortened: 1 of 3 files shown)


booc0mtaco:
https://github.com/chanzuckerberg/edu-design-system/assets/3056447/dcf9a3db-3ab1-4634-a834-ee1c7b49b836

src/components/InputField/InputField.tsx
```diff
@@ -292,12 +297,24 @@ export const InputField: InputFieldType = forwardRef(
             ref={ref}
             required={required}
             status={shouldRenderError ? 'critical' : status}
-            type={type}
+            type={isPasswordVisible ? 'text' : type}
```
jeremiah-clothier, line 300: Not having the type="password" may mess with the usage of `autocomplete`, but since we're using Google Login for Render I don't think this is an issue

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/password#allowing_autocomplete

jeremiah-clothier (approved):
LGTM! More a curiosity question, how does copy+paste work when the password is hidden

booc0mtaco:
@jeremiah-clothier paste works as you'd expect, as does copy. We don't do any tracking of what's typed and replacing it with ••• in the field, so the browser handles converting the typed content to the actual text

PR #2008, opened 2024-07-05
fix(Select): expose generic types to allow by to work

booc0mtaco:
- borrow type value determination from HeadlessUI internals
- add test to make sure types used map to params

### Test Plan:

### DIFF
--- a/src/components/Select/Select.stories.tsx
+++ b/src/components/Select/Select.stories.tsx
@@ -306,7 +306,7 @@ export const WithSelectedOption: StoryObj<typeof Select> = {
 };
 
 /**
- * Use the `by` option to determine the selection (when using objects for the value list).
+ * Use the `by` option to determine the selection (when using objects for the value list). This helps when you want to compare by value, not reference.
  * - The type comparison can be by a named key in the object `by={'id'}` or using a comparison function
  *
  * See: https://headlessui.com/v1/react/listbox#listbox
@@ -314,6 +314,7 @@ export const WithSelectedOption: StoryObj<typeof Select> = {
 export const WithSelectedBy: StoryObj<typeof Select> = {
   args: {
     ...WithSelectedOption.args,
+    defaultValue: { ...exampleOptions[1] },
     by: 'key',
   },
 };
--- a/src/components/Select/Select.stories.tsx
+++ b/src/components/Select/Select.stories.tsx
@@ -306,6 +306,19 @@ export const WithSelectedOption: StoryObj<typeof Select> = {
 };
 
 /**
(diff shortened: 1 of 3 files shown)


booc0mtaco:
I need to do some validation checks on this to make sure [`keyof`](https://www.typescriptlang.org/docs/handbook/2/keyof-types.html) is doing the right thing for the use cases we will support. right now, this should allow strings, or associative arrays for the values in a `Select`, and by should either be the allowed type of any of those keys, OR the comparison function which takes the full instance (whether it is a string OR an object)

src/components/Select/Select.tsx
```diff
@@ -25,7 +26,12 @@ import Text from '../Text';
 
 import styles from './Select.module.css';
 
-type SelectProps = ExtractProps<typeof Listbox> &
+// https://github.com/tailwindlabs/headlessui/blob/%40headlessui/react%40v1.7.19/packages/%40headlessui-react/src/components/listbox/listbox.tsx#L349
+type SelectProps = ListboxProps<
+  ElementType,
+  string | { [k: string]: unknown },
+  { [k: string]: unknown }
```
booc0mtaco, line 33: the first line represents what is allowed in things like `defaultValue` or `value`

the second line represents what gets used when `by` is evaluated

* when a simple string, `by` is string
* when an object, `by` is the type of the keys on that object (usually `string | number`), or that eval function
* ...

src/components/Select/Select.stories.tsx
```diff
@@ -305,6 +305,19 @@ export const WithSelectedOption: StoryObj<typeof Select> = {
...
 
+/**
+ * Use the `by` option to determine the selection (when using objects for the value list).
+ * - The type comparison can be by a named key in the object `by={'id'}` or using a comparison function
+ *
+ * See: https://headlessui.com/v1/react/listbox#listbox
+ */
+export const WithSelectedBy: StoryObj<typeof Select> = {
+  args: {
+    ...WithSelectedOption.args,
```
forgeabout, line 316: [non-blocking] It might be a more thorough test to use `defaultValue: {...exampleOptions[1]}` instead of inheriting `defaultValue: exampleOptions[1]` from `WithSelectedOption` since the purpose of `by` is to handle cases where the values are copies rather than the exact same objects.

forgeabout (approved):
LGTM. Thanks @booc0mtaco!

src/components/Select/Select.stories.tsx
booc0mtaco: ++ good call. I suspect it will work properly, but this will prevent some by-ref getting thru down the road

PR #2011, opened 2024-07-09
feat(Modal): add height property to modal API

booc0mtaco:
Bumping `Modal` to 2.1 with the following functionality:

- allow fixed (default), auto, and max
- add stories with doc.s
- update snapshots

### Test Plan:

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -3,6 +3,7 @@ import clsx from 'clsx';
 import type { MutableRefObject, ReactNode } from 'react';
 import React from 'react';
 
+import { assertEdsUsage } from '../../util/logging';
 import type { ExtractProps } from '../../util/utility-types';
 import type { Size } from '../../util/variant-types';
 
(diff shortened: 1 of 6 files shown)


src/components/Modal/Modal.tsx
```diff
@@ -79,17 +79,24 @@ type ModalContentProps = {
   onClose: () => void;
   // Design API
   /**
-   * Max size of the modal, which responds to the viewport
-   *
-   * **Default is `"lg"`**.
+   * Determine how the height of the modal container is calculated:
+   * - `"fixed"` applies the fixed size, which will not adjust
```
ericraymond, line 83: Would be good to refer to the size property if that's where fixed size comes from.

ericraymond (approved):

src/components/Modal/Modal.tsx
booc0mtaco: fair point. I think a better word to use here is s/size/dimensions, since each `size` value has different dimensions as well. Will reword a bit

PR #2015, opened 2024-07-12
fix(Link): apply font weight to standalone sizes

booc0mtaco:
- fix handling of size so it isn't applied where it can't be used
- add assert for using size with context=inline
- update tests and snapshots

### Test Plan:

### DIFF
--- a/src/components/Link/Link.module.css
+++ b/src/components/Link/Link.module.css
@@ -23,11 +23,15 @@
 
   /* Sub-component spacing */
   &.link--size-xl,
-  &.link--size-lg { padding-left: calc(var(--eds-size-1) / 16 * 1rem); }
+  &.link--size-lg {
+    padding-left: calc(var(--eds-size-1) / 16 * 1rem);
+  }
 
   &.link--size-md,
-  &.link--size-sm, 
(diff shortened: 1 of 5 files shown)


src/components/Link/Link.module.css
```diff
@@ -51,22 +55,27 @@
...
   }
 
   &.link--size-sm {
     font: var(--eds-theme-typography-body-sm);
+    font-weight: 500;
   }
 
   &.link--size-xs {
     font: var(--eds-theme-typography-body-xs);
+    font-weight: 500;
```
jgao-czi, line 78: Oh just so 500 is the expected font size for Links?

jgao-czi (approved):

src/components/Link/Link.module.css
booc0mtaco: 500 is the weight for each link size. We have to add them here b/c `font:` will try to set it to 400 which is what we saw. later this year when we revisit the typography tokens, we won't need this anymore (will be transparent to you all when that happens)

timzchang (approved):

booc0mtaco:
Thanks @timzchang @jgao-czi :)

PR #2023, opened 2024-07-22
feat(Fieldset): update styles to support new 2.0 components

booc0mtaco:
- add in new props
- update tests
- add in warnings and assertions
- export new components for checkbox and radio button groups

Screenshot 2024-07-22 at 19 13 13
Screenshot 2024-07-24 at 17 30 47
Screenshot 2024-07-24 at 17 30 51
Screenshot 2024-07-24 at 17 30 58
Screenshot 2024-07-24 at 17 31 02
Screenshot 2024-07-24 at 17 31 07

### Test Plan:


plop-templates/Component/Component.test.ts.hbs
```diff
@@ -1,6 +1,7 @@
 import { generateSnapshots } from '@chanzuckerberg/story-utils';
 import * as stories from './{{pascalCase name}}.stories';
+import type { StoryFile } from '../../util/utility-types';
```
booc0mtaco, line 3: Improve: Update the plop template to do the right things

plop-templates/Component/Component.tsx.hbs
```diff
@@ -3,10 +3,13 @@ import React from 'react';
 import styles from './{{pascalCase name}}.module.css';
 
 export interface Props {
+  // Component API
```
booc0mtaco, line 6: Improve: Add newer templating where we group the props by those needed by eng (this area), from those to match what's in figma (Design API)

plop-templates/Component/Component.tsx.hbs
```diff
@@ -18,13 +21,15 @@ export interface Props {
  */
 export const {{pascalCase name}} = ({
   className,
+  // Add other deferenced props to use
```
booc0mtaco, line 24: Make sure the plopped file generates code that isn't immediately invalid to prettier but also a few more instruction areas

plopfile.js
```diff
@@ -52,37 +52,7 @@ module.exports = (plop) => {
         template:
           "export { default as {{pascalCase name}} } from './components/{{pascalCase name}}';",
       },
-      // From https://github.com/bradfrost/czi-vanilla-storybook
```
booc0mtaco, line 55: this sorting removes the useful comments and sections in the index.ts file. Updated the comment so if we want to restore the sorting, can look at the diff associated with the comment

src/components/Radio/Radio.module.css
```diff
@@ -180,7 +180,7 @@
     border-color: var(--eds-theme-color-border-utility-critical);
   }
 
-  & ~ .radio__selected {
+  &:checked ~ .radio__selected {
```
booc0mtaco, line 183: Fixes bug where selected treatment is applied to ALL radio buttons regardless of checked state

plop-templates/Component/Component.test.ts.hbs
```diff
@@ -1,6 +1,7 @@
...
 
 describe('<{{pascalCase name}} />', () => {
-  generateSnapshots(stories);
+  generateSnapshots(stories as StoryFile);
```
ahuth, line 6: Dang, this is annoying 😞. Shouldn't be necessary, but it's not obvious to me what the deal is.

What's weird is this works without casting in Bento. That project is using Storybook 8.2.4. Maybe there was a type regression in 8.2.5?

ahuth (approved):

plop-templates/Component/Component.test.ts.hbs
booc0mtaco: This started happening sorta randomly when I first upgraded to storybook 7. Not every file threw errors at first, and I also couldn't figure out what made some imports freak out like this ... 🤔 Looks like all the tests now have this cast.

Removing the cast yields this TS error:
```
Argument of type 'typeof import("/**/$Component.stories")' is not assignable to parameter of type 'StoryFile'.
  The types of 'default.play' are incompatible between these types.
    Type 'PlayFunction<ReactRenderer, AccordionProps> | undefined' is not assignable to type 'PlayFunction<ReactRenderer, Args> | undefined'.
      Type 'PlayFunction<ReactRenderer, AccordionProps>' is not assignable to type 'PlayFunction<ReactRenderer, Args>'.
        Types of parameters 'context' and 'context' are incompatible.
          Type 'PlayFunctionContext<ReactRenderer, Args>' is missing the following properties from type 'PlayFunctionContext<ReactRenderer, AccordionProps>': context, canvas, mount ts(2345)
```

PR #2036, opened 2024-08-15
feat(config): add transition durations for tailwind

booc0mtaco:
- add in durations via (transitionDuration)
- add in new toast notification implementation example
- use new medium durations in example

- demo autodismiss behavior with stack of notifications
- use state to track

### Test Plan:


booc0mtaco:
https://github.com/user-attachments/assets/98fac804-73b5-4f81-b99e-841ac20a0a60

src/components/ToastNotification/ToastNotification.stories.tsx
```diff
@@ -0,0 +1,142 @@
...
+        {toasts.map((toast) => (
+          <Transition
+            appear
+            enter="transition-all duration-300"
+            enterFrom="opacity-0 transform-gpu scale-0"
+            enterTo="opacity-100 transform-gpu scale-100"
+            key={toast.id}
+            leave="ease-in-out transition-all duration-300"
+            leaveFrom="opacity-100 transform-gpu translate-x-[0px]"
+            leaveTo="opacity-0 transform-gpu translate-x-[-100%]"
```
booc0mtaco, line 108: you always want to use GPU-optimized transition techniques "don't change positioning using things like `top` and `left`). This allows for high frame rate animations, and avoids issues with positioning, AND recalculating layout in the DOM

ahuth:
Ah nice, I like the debounce idea. Only cleanup when toast stuff isn't happening?

On the other hand, if it's just for stories, maybe it's fine to let the list get super large?

booc0mtaco:
> Ah nice, I like the debounce idea. Only cleanup when toast stuff isn't happening?
> 
> On the other hand, if it's just for stories, maybe it's fine to let the list get super large?

Yah, we have control over the time period for how long it takes to transition out, so could keep them in sync. I like the idea of keeping the example simple and describing it as demo-only (e.g., "in a production implementation, make sure to prune the notification list once dismissal has occurred")

ahuth (approved):

PR #2042, opened 2024-08-23
fix(Link): handle inverse variant for standalone links

booc0mtaco:
- update documentation to explain cases
- add warning for misaligned prop usages
- update snapshots and tests

### Test Plan:

### DIFF
--- a/src/components/Link/Link.stories.tsx
+++ b/src/components/Link/Link.stories.tsx
@@ -26,7 +26,7 @@ type Story = StoryObj<typeof Link>;
 export const Default: Story = {};
 
 /**
- * When using context, you can specify a trailing icon for the link.
+ * When using standalone context, you can specify a trailing icon for the link.
  *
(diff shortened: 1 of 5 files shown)


booc0mtaco:
Fix here is to make cases where we have standalone links with default variant appear with the correct color

https://61313967cde49b003ae2a860-xiicuaigmu.chromatic.com/?path=/story/components-link--standalone-inverse-variant

src/components/Link/Link.stories.tsx
```diff
@@ -26,7 +26,7 @@ type Story = StoryObj<typeof Link>;
 export const Default: Story = {};
 
 /**
- * When using context, you can specify a trailing icon for the link.
+ * When using standalone context, you can specify a trailing icon for the link.
```
ahuth, line 29: What's standalone context, and/or can we clarify that?

Although I'm only looking at the diff, not the storybook. Might be more clear

ahuth (approved):

PR #2052, opened 2024-08-30
feat(Button): support for not-allowed cursor when disabled

booc0mtaco:
### Test Plan:

### DIFF
--- a/src/components/Button/Button.module.css
+++ b/src/components/Button/Button.module.css
@@ -263,14 +263,14 @@
  * Disabled
  */
 .button--disabled {
+  pointer-events: none;
+
   &.button--primary.button--variant-default,
   &.button--primary.button--variant-critical,
   &.button--primary.button--variant-neutral {
     color: var(--eds-theme-color-text-utility-disabled-primary);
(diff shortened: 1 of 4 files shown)


src/components/Button/Button.module.css
```diff
@@ -263,23 +263,21 @@
  * Disabled
  */
 .button--disabled {
+  pointer-events: none;
```
booc0mtaco, line 266: We disable all the pointer events for these buttons to make them truly disabled (and this avoids having to add styles to override the `:hover`, `:active` and other pseudo-selectors

src/components/Button/Button.module.css
```diff
@@ -288,34 +286,30 @@
...
 
   &.button--tertiary.button--variant-inverse {
     color: var(--eds-theme-color-text-utility-disabled-primary);
     border-color: transparent;
-
-    pointer-events: none;
   }
 }
 
+.button__disabled {
```
booc0mtaco, line 309: this is to style a container when disabled. because `pointer-events` is also how the cursor gets styled, we cannot just add `cursor: not-allow

ahuth (approved):

PR #2055, opened 2024-09-05
feat(DataTable): implement TanStack Table

booc0mtaco:
**NOTE**: this merges into the existing branch for EDS-1209. posting to generate the snapshots and such, but will just merge into that in-progress work

- use tanstack as basis for table implementation
- add in relevant stories for styles

### Test Plan:

### DIFF
--- a/src/components/DataTable/DataTable.stories.tsx
+++ b/src/components/DataTable/DataTable.stories.tsx
@@ -1,8 +1,15 @@
 import { BADGE } from '@geometricpanda/storybook-addon-badges';
 import type { StoryObj, Meta } from '@storybook/react';
+
+import {
+  createColumnHelper,
+  getCoreRowModel,
+  useReactTable,
+} from '@tanstack/react-table';
+
 import React from 'react';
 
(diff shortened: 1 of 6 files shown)


src/components/DataTable/DataTable.stories.tsx
```diff
@@ -24,7 +31,172 @@ export default {
...
+};
+
+export const TableSizeSm: StoryObj<Args> = {
+  args: {
+    tableStyle: 'border',
+    size: 'sm',
+  },
+  render: (args) => {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    const [data] = React.useState(() => [...defaultData]);
```
ahuth, line 156: If it's not being changed, why does the data need to be in state?

booc0mtaco: This is just some boilerplate to be used later. We want to support selectable rows and sorting, which state will help with. I'm also going to factor these `render` calls so that they aren't so copy-paste

PR #2063, opened 2024-09-27
fix(Button): specify size for enabled/disabled buttons

booc0mtaco:
- separate out the size metrics from the internal size styles
- regenerate snapshots for comparison
- add story for showing disabled buttons at all sizes

### Test Plan:


src/components/AppNotification/__snapshots__/AppNotification.test.ts.snap
```diff
@@ -53,7 +53,7 @@ exports[`<AppNotification /> LightColor story renders snapshot 1`] = `
           class="button-group button-group--horizontal-align-left"
         >
           <button
-            class="button button--layout-none button--secondary button--lg button--variant-default"
+            class="button button--layout-none button--secondary button--lg button--size-lg button--variant-default"
```
sskikne, line 56: Confirming we want both button--lg and button--size-lg. Seems like yes based on the description.

sskikne (approved):
lgtm

booc0mtaco:
Thanks @sskikne ! (the `size` thing is a quirk on the button component right now, which I wanna clean up. having styles suffixed `--prop-value` is the general pattern, but this was the first component we did when building 2.0 and hadn't figured that out yet)

PR #2068, opened 2024-10-09
fix(Select): remove label checks from component

booc0mtaco:
- remove throw from code that is for a complex a11y check
- update tests

This allows for any external, automated accessibility tests to provide a better contextual error without breaking behavior during development.

### Test Plan:

### DIFF
--- a/src/components/Select/Select.tsx
+++ b/src/components/Select/Select.tsx
@@ -18,7 +18,7 @@ import FieldNote from '../FieldNote';
 import Icon, { type IconName } from '../Icon';
 import {
   defaultPopoverModifiers,
-  PopoverContainerV2 as PopoverContainer,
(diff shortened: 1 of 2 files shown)


timzchang (approved):

src/components/Select/Select.tsx
```diff
@@ -18,7 +18,7 @@ import FieldNote from '../FieldNote';
 import Icon, { type IconName } from '../Icon';
 import {
   defaultPopoverModifiers,
-  PopoverContainerV2 as PopoverContainer,
+  default as PopoverContainer,
```
booc0mtaco, line 21: a little leftover piece from the 2.0 migration. Tidying the import here so that the major change later is smoother

PR #2072, opened 2024-10-15
fix(DataTable): fix cell size and border in FF

booc0mtaco:
- add classes for (header) cell containers
- fix cell height size
- tune up tests, stories, and snapshots

### Test Plan:

### DIFF
--- a/src/components/DataTable/DataTable.tsx
+++ b/src/components/DataTable/DataTable.tsx
@@ -213,6 +213,7 @@ export function DataTable<T>({
               <DataTableRow key={headerGroup.id}>
                 {headerGroup.headers.map((header) => (
                   <th
+                    className={styles['data-table__header-cell-container']}
                     colSpan={header.colSpan}
                     key={header.id}
                     style={{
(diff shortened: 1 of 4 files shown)


booc0mtaco:
There is one remaining issue noticed here, with the right-hand border in the right-most header cell. Will address in a separate PR, with any other firefox discrepancies

src/components/DataTable/DataTable.tsx
```diff
@@ -213,6 +213,7 @@ export function DataTable<T>({
               <DataTableRow key={headerGroup.id}>
                 {headerGroup.headers.map((header) => (
                   <th
+                    className={styles['data-table__header-cell-container']}
```
sskikne, line 216: Not a complaint, but curious about this naming format that mizes underscore and dash

sskikne (approved):
lgtm!

src/components/DataTable/DataTable.tsx
booc0mtaco: Yep yep, [doc.s here](https://chanzuckerberg.github.io/edu-design-system/?path=/docs/documentation-contributor-guidelines-code-guidelines--docs)

In our scheme, `-` has no special meaning and is in place of using camelCase in CSS naming, preferring snake-case. `--` and `__` have more strict meaning, with most of the rest is convention

HOLD ME ACCOUNTABLE to that page :)

booc0mtaco:
thanks @sskikne ✅

PR #2073, opened 2024-10-16
feat(DataTable): add support for table row statuses

booc0mtaco:
- component to support standard EDS statuses per row
- update tests and snapshots (new story for comparison)

### Test Plan:

### DIFF
--- a/src/components/DataTable/DataTable.stories.tsx
+++ b/src/components/DataTable/DataTable.stories.tsx
@@ -11,6 +11,7 @@ import {
   DataTableRow,
   DataTableHeader,
   type DataTableProps,
+  type DataTableWithStatus,
 } from './DataTable';
 
 // We import all of the utilities from tanstack here, and this can contain other custom utilities
@@ -53,13 +54,14 @@ export default {
 type Args = DataTableProps;
 
 // Specifying an example data type for the rows of a table
-type Person = {
+// This column is used for tables that are eligible for status
+type Person = DataTableWithStatus<{
   firstName: string;
   lastName: string;
   age: number;
   visits: number;
   progress: number;
-};
+}>;
 
 // Specifying the example (static) data for the table to use with tanstack primitives
 const defaultData: Person[] = [
@@ -76,6 +78,7 @@ const defaultData: Person[] = [
     age: 40,
     visits: 40,
     progress: 80,
+    status: 'warning',
   },
   {
     firstName: 'Tanner',
@@ -90,6 +93,7 @@ const defaultData: Person[] = [
     age: 45,
     visits: 20,
     progress: 10,
+    status: 'critical',
   },
   {
     firstName: 'Tandy',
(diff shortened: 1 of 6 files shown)


src/components/DataTable/DataTable.stories.tsx
```diff
@@ -700,6 +705,93 @@ export const Grouping: StoryObj<Args> = {
...
+    tableStyle: 'border',
+    rowStyle: 'lined',
+    size: 'sm',
+  },
+  render: (args) => {
+    const columns = [
+      columnHelper.accessor(DataTable.__StatusColumnId__, {
+        header: () => <DataTable.StatusHeaderCell />,
+        cell: (info) => <DataTable.StatusCell status={info.getValue()} />,
+        size: 32,
```
booc0mtaco, line 729: TODO add / use some token here

src/components/DataTable/DataTable.tsx
```diff
@@ -159,129 +176,141 @@ export function DataTable<T>({
...
-              <DataTableRow key={headerGroup.id}>
-                {headerGroup.headers.map((header) => (
-                  <th
-                    className={styles['data-table__header-cell-container']}
-                    colSpan={header.colSpan}
-                    key={header.id}
-                    style={{
-                      width: `${header.getSize()}px`,
-                    }}
+    <DataTableContext.Provider value={{ size }}>
```
booc0mtaco, line 179: Most of this diff is whitespace changes. Use the ⚙️ above to update the view to hide whitespace changes.

src/components/DataTable/DataTable.tsx
```diff
@@ -159,129 +176,141 @@ export function DataTable<T>({
...
+            )}
+            <DataTableHeader>
+              {table?.getHeaderGroups().map((headerGroup) => (
+                <DataTableRow key={headerGroup.id}>
+                  {headerGroup.headers.map((header) => {
+                    // TODO-AH: figure out better way to detect a non-default column size
+                    const columnWidth =
+                      header.getSize() !== 150
+                        ? `${header.getSize()}px`
+                        : undefined;
```
booc0mtaco, line 237: Need to investigate how TanStack Table sets the column width. `.getSize` is always defined, and defaults to 150px. Marking this value as special here b/c all sizings will be based on token unit sizes (16 * x) and 150 doesn't land on one of those boundaries (... 144, 160, 176, ...)

src/index.ts
```diff
@@ -61,6 +61,11 @@ export type { ButtonProps } from './components/Button';
 export type { LinkProps } from './components/Link';
 export type { PageNotificationProps } from './components/PageNotification';
 export type { AppNotificationProps } from './components/AppNotification';
+export type {
+  DataTableProps,
+  DataTableRowProps,
+  DataTableWithStatus,
```
booc0mtaco, line 67: Utility type for augmenting table model data with the details of the status type in the type object.

timzchang (approved):
looks good to me!

PR #2081, opened 2024-10-23
fix(Select): pull z-index of select options to top

booc0mtaco:
Give the select field the highest token value for z-index so that it will appear above modal containers (both portal to the top edge of the DOM, so z-index and sequencing can cause `Select.Options` to appear beneath the modal container). Modals currently use z-index of 1050 and are set manually to that level, so this will go above that.

No stacking context changes are needed.

### Test Plan:

### DIFF
--- a/src/components/Select/Select.module.css
+++ b/src/components/Select/Select.module.css
@@ -32,7 +32,7 @@
  */
 .select__options {
   max-height: 25vh;
-  z-index: 100;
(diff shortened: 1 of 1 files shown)


jeremiah-clothier (approved):

booc0mtaco:
Thanks @jeremiah-clothier !

PR #2082, opened 2024-10-24
WIP: update headlessui from v1 to v2

jeremiah-clothier:
### Test Plan:

Updating HeadlessUI from v1 to v2

* Release Notes: https://github.com/tailwindlabs/headlessui/releases/tag/%40headlessui%2Freact%40v2.0.0

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -329,15 +329,16 @@ export const Modal = (props: ModalProps) => {
         // Passing onClose to the Dialog allows it to close the modal when the ESC key is triggered.
         onClose={onClose}
       >
-        <Dialog.Overlay
+        <div
           className={clsx(
             styles['modal__overlay'],
             overlayEmphasis &&
               styles[`modal__overlay--emphasis-${overlayEmphasis}`],
           )}
         />
-
-        <ModalContent onClose={onClose} {...rest} />
+        <Dialog.Panel>
+          <ModalContent onClose={onClose} {...rest} />
+        </Dialog.Panel>
       </Dialog>
     </Transition>
   );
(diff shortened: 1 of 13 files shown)


src/components/Modal/Modal.tsx
```diff
@@ -329,15 +329,16 @@ export const Modal = (props: ModalProps) => {
...
             styles['modal__overlay'],
             overlayEmphasis &&
               styles[`modal__overlay--emphasis-${overlayEmphasis}`],
           )}
         />
-
-        <ModalContent onClose={onClose} {...rest} />
+        <Dialog.Panel>
+          <ModalContent onClose={onClose} {...rest} />
+        </Dialog.Panel>
```
jeremiah-clothier, line 341: This is the only typescript error I got from the migration. Dialog.Overlay was deprecated in v1.6

https://github.com/tailwindlabs/headlessui/releases/tag/%40headlessui%2Freact%40v1.6.0

src/components/ToastNotification/ToastNotification.stories.tsx
```diff
@@ -100,6 +100,7 @@ const ToastNotificationManager = (args: Args) => {
         {toasts.map((toast) => (
           <Transition
             appear
+            as="div"
```
jeremiah-clothier, line 103: The default `as` for this component is now `as={React.Fragment}`

booc0mtaco:
With headlessUI v2 we can also remove a few other dependencies that v1 needed (like popperJS and tippyJS? maybe just one of those or both... cant recall) But great to see it didn't cause much meaningful jitter; the goal of this repo is to abstract these internals :)

jeremiah-clothier:
> With headlessUI v2 we can also remove a few other dependencies that v1 needed (like popperJS and tippyJS? maybe just one of those or both... cant recall) But great to see it didn't cause much meaningful jitter; the goal of this repo is to abstract these internals :)

+1 headlessUI changed all their imports from like `Dialog.Panel` to `DialogPanel` but it's only a deprecation so could be a separate PR after upgrading.

For posterity, we can also get rid of `ExtractProps` in EDS because headless ui v2 exports the props of each of their components (finally)

@booc0mtaco

PR #2087, opened 2024-11-07
feat(Tag): update to 2.0 styles

booc0mtaco:
- use new render color tokens
- update layout
- update tests and snapshots

### Test Plan:


src/components/Tag/Tag.module.css
```diff
@@ -45,6 +44,7 @@
 
 /**
  * Color variants.
+ * TODO(next-major): remove these status colors
```
ademartini-czi, line 47: Are these TODO tracked as tickets?

src/components/Tag/Tag.tsx
```diff
@@ -1,72 +1,115 @@
...
-
 export type Variant = (typeof VARIANTS)[number];
 
 type Props = {
-  /**
-   * The color variant of the tag. It will update the content colors, background color, and border (when `hasOutline` is set to `true`).
-   *
-   * **Default is `"neutral"`**.
-   */
-  variant?: Variant;
```
ademartini-czi, line 26: Do we allow for breaking API contract with major version bump?

booc0mtaco: Yep, we do. Here, I moved `variant` to another part of `Props` for consistency (and to alphabetize better). those TODO(next-major) comments are for when we do the next major, likely in the new year. 

This one is backwards-compatible with some warnings (those `assertEDSUsage` log messages)

src/components/Tag/Tag.module.css
booc0mtaco: Not currently, tho its not a bad idea to create one and add the detail of "delete all the deprecated things"

src/components/Tag/Tag.tsx
ademartini-czi: OK got it, thanks

ademartini-czi (approved):

src/components/Tag/Tag.module.css
booc0mtaco: Added a tracker to the backlog

PR #2113, opened 2024-12-04
refactor: update non-exported types from interface to type

booc0mtaco:
Change TS convention to consistently use `type` instead of a mix of `type` and `interface` for components. Mark any types that get exported as to be updated at the next major version.

**NOTE**: some changes come from tweaks to `prettier` config, so accepting those

### Test Plan:

### DIFF
--- a/.storybook/components/Docs/Guidelines/CodeGuidelines.mdx
+++ b/.storybook/components/Docs/Guidelines/CodeGuidelines.mdx
@@ -207,7 +207,8 @@ Examples:
 
 ```css
 .dropdown-button__icon {
-  transition: transform calc(var(--eds-anim-move-medium) * 1s) var(--eds-anim-ease);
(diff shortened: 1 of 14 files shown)


timzchang (commented):
> Mark any types that get exported as to be updated at the next major version.

Did your strategy change? I'm seeing some exported interfaces being converted to types, and some non exported interfaces being marked for next major version

Also, why the decision to delay some vs update others?

booc0mtaco:
> > Mark any types that get exported as to be updated at the next major version.
> 
> Did your strategy change? I'm seeing some exported interfaces being converted to types, and some non exported interfaces being marked for next major version
> 
> Also, why the decision to delay some vs update others?

@timzchang Oh, I left out an important word here. exported **externally** (from src/index.ts). We have some internal exports for components and component tests to use, but the ones i wanted to postpone were those affecting the external API used by consumers.

timzchang (approved):

PR #2114, opened 2024-12-06
chore: update component version metadata

booc0mtaco:
The previous version markers were mixing two ideas together, so the version numbers were ambiguous. 

Based on how design is defining the theme versions, we only have two 1.0 (SLP), and 2.0 (Render). This is mostly to normalize what versions show up for which components (there was no 1.3 theme version; that "1.3" was referring to the components API, hence the API version numbers still having the minors show up)

### Test Plan:

  - validate against table api and theme versions per component (see storybook)

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -25,35 +25,35 @@ export const decorators = [
   ),
 ];
 
-function createInitialReleaseConfig(usingLabel: string) {
+function createComponentVersion(usingLabel: string) {
   return {
-    [`intro-${usingLabel}`]: {
+    [`api-${usingLabel}`]: {
       styles: {
         backgroundColor: '#ffffff',
         borderColor: '#000000',
         color: '#000000',
       },
-      title: `Introduced: ${usingLabel}`,
+      title: `Component API Version: ${usingLabel}`,
       tooltip: {
-        title: `Introduced in v${usingLabel}`,
-        desc: `This component was introduced in EDS Design version ${usingLabel}`,
+        title: `Code / API v${usingLabel}`,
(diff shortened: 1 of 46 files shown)


booc0mtaco:
Screenshot 2024-12-06 at 10 57 44

.storybook/preview.tsx
```diff
@@ -79,15 +79,13 @@ export const parameters: Preview['parameters'] = {
...
-    ...createCurrentReleaseConfig('1.3'),
-    ...createCurrentReleaseConfig('2.0'),
-    ...createCurrentReleaseConfig('2.1'),
+    ...createComponentVersion('2.0'),
+    ...createComponentVersion('1.3'),
+    ...createComponentVersion('1.2'),
+    ...createComponentVersion('1.1'),
+    ...createComponentVersion('1.0'),
+    ...createThemeVersion('1.0'),
+    ...createThemeVersion('2.0'),
```
timzchang, line 88: why did theme version 1.3 and 2.1 disappear?

timzchang (approved):

.storybook/preview.tsx
booc0mtaco: Good question. The previous version markers were mixing two ideas together, so the version numbers were ambiguous. 

Based on how design is defining the theme versions, we only have two 1.0 (SLP), and 2.0 (Render). This is mostly to normalize what versions show up for which components (there was no 1.3 theme version; that "1.3" was referring to the components API, hence the API version numbers still having the minors show up)

booc0mtaco: added the above to the description, for that missing clarity!

PR #2122, opened 2025-01-03
fix(ToastNotification): fix typo in prop name

booc0mtaco:
- add in dismissType
- add in tests and console messages so make sure the new field falls back

### Test Plan:

### DIFF
--- a/src/components/ToastNotification/ToastNotification.tsx
+++ b/src/components/ToastNotification/ToastNotification.tsx
@@ -32,7 +32,6 @@ export type ToastNotificationProps = {
    * **Default is `"manual"`**.
    */
   dismissType?: 'manual' | 'auto';
-  dissmissType?: 'manual' | 'auto'; // TODO(next-major): remove this misspelled prop at next major release
   /**
    * Keyword to characterize the state of the notification
    */
@@ -41,6 +40,8 @@ export type ToastNotificationProps = {
    * The title/heading of the notification
    */
   title: string;
+  /** @deprecated Please use `dismissType` instead */
+  dissmissType?: 'manual' | 'auto'; // TODO(next-major): remove this misspelled prop at next major release
 };
 
 /**
@@ -64,7 +65,7 @@ export const ToastNotification = ({
     className,
   );
 
-  // if both are defined, temporarily prefer the original (misspelled) prop for now, but throw a warning
+  // if both are defined, temporarily prefer the original value of dissmissType to avoid accidental overrides
   const tempDismissType =
     dissmissType && dismissType ? dissmissType : dismissType;
 
--- a/src/components/ToastNotification/ToastNotification.tsx
+++ b/src/components/ToastNotification/ToastNotification.tsx
@@ -31,7 +31,8 @@ export type ToastNotificationProps = {
    *
    * **Default is `"manual"`**.
    */
-  dissmissType?: 'manual' | 'auto';
+  dismissType?: 'manual' | 'auto';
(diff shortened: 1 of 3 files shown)


src/components/ToastNotification/ToastNotification.tsx
```diff
@@ -62,22 +64,37 @@ export const ToastNotification = ({
     className,
   );
 
+  // if both are defined, temporarily prefer the original (misspelled) prop for now, but throw a warning
+  const tempDismissType =
+    dissmissType && dismissType ? dissmissType : dismissType;
```
forgeabout, line 69: Won't this always return undefined when `dismissType` is undefined, and thus break existing code that only passes the old `dissmissType` prop?

src/components/ToastNotification/ToastNotification.tsx
```diff
@@ -62,22 +64,37 @@ export const ToastNotification = ({
     className,
   );
 
+  // if both are defined, temporarily prefer the original (misspelled) prop for now, but throw a warning
```
forgeabout, line 67: Why would we want or need to prefer the original (misspelled) prop when both are defined?

src/components/ToastNotification/ToastNotification.tsx
```diff
@@ -62,22 +64,37 @@ export const ToastNotification = ({
...
+      !!timeout &&
+        typeof onDismiss === 'undefined' &&
+        tempDismissType === 'auto',
+    ],
     'When using dismissType=auto, an onDismiss method must be defined',
     'error',
   );
 
+  // TODO(next-major): remove this misspelled prop at next major release
+  assertEdsUsage(
```
forgeabout, line 82: The assert is good, but should we also add a `@deprecated` flag to the prop so that developers can more easily catch it without needing to run the code?

forgeabout (changes requested):
Looks like there's a logic error, but let me know if I'm missing something.

src/components/ToastNotification/ToastNotification.tsx
booc0mtaco: This part was a bit tricky to parse, but tl;dr it won't break in this way:

* `dismissType` has a default value of `'manual'` in the function signature
* b/c of that, it's guaranteed to have some value (and a type check to prevent invalid values)
* previously `dissmissType` had that default valued but no longer has one

| `dismissType` | `dissmissType` | Result |
|--------|--------|--------|
| manual | undefined | manual |
| auto | undefined | auto |
| manual | manual | manual |
| manual | auto | auto | 

The intent is that if someone intentionally used `"auto"` with the old prop, that value would be used (and they'd get that console error). Without a check like this, the old prop's value would be replaced with the new prop's default value

src/components/ToastNotification/ToastNotification.tsx
booc0mtaco: Yah good call. The old deprecations were cleaned up at the last major, so didn't think to add one!

src/components/ToastNotification/ToastNotification.tsx
forgeabout: I see. So the `&& dismissType` doesn't actually do anything since it's always true? It may be worth pointing that out in the comment, but not a big deal if support for the old prop will be removed soon anyway.

forgeabout (approved):

src/components/ToastNotification/ToastNotification.tsx
booc0mtaco: Great point; I add this logic before changing the default value, so the check **is** redundant. Since we are here, i'll remove that truthy value from the expression. thanks for the review!

PR #2133, opened 2025-02-20
feat(DataTable): export sort direction enum with types

booc0mtaco:
- export utility object with sort directions
- export type derived from the sort directions
- use the type and object in the component

### Test Plan:


mjbartho (approved):
LGTM!

PR #2147, opened 2025-03-20
feat!: remove Slider component [EDS-1534]

ajmoreno:
Summary:
Removes the Slider component and all associated files and references

- [x] Manually tested my changes, and here are the details: verified slider removed in storybook


.storybook/components/DesignTokens/Tier3/Sizes.stories.tsx
ajmoreno, line 1: wasn't sure if i should delete this or not but it seemed like the whole purpose of it was to show off parts of the slider so went ahead and deleted it (for now)

booc0mtaco: Good call, that is the general idea!

booc0mtaco (commented):
Thanks! One tweak to the commit message then we should be good to go.

Can you change the prefix to `feat!` so that the release automation will mark it as a [breaking change](https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with--to-draw-attention-to-breaking-change)?

PR #2156, opened 2025-03-26
feat!: remove items marked as deprecated [EDS-1543]

ajmoreno:
Summary:
Remove all items marked as deprecated and associated references.

### Test Plan:

- [x] Manually tested my changes, and here are the details: ran local storybook and things looked good


tailwind.config.ts
```diff
@@ -63,9 +49,6 @@ export default {
       textColor: {
         ...textColorTokens,
       },
-      spacing: {
-        ...sizeTokens,
-      },
```
booc0mtaco, line 68: I forgot about this part 😅 This is good to remove, but we also want to bring in the tokens that have `-spacing-` in their instead ([example](https://github.com/chanzuckerberg/edu-design-system/pull/2156/files#diff-ae098022f6228c869ce589a6d9c92799310df9bb7b6533d95dca208a5fc32f51L29))

What we want to spread here is all of the `eds-spacing-size` values that get generated, so tailwind can have stuff like `px-spacing-size-1` or similar

(& in the migration script, we will add a migration that changes `X-size-3` to `X-spacing-size-3` in consuming projects, for reference)

src/components/Tag/Tag.test.tsx
booc0mtaco, line 1: We should add this file back, but only with contents similar to:

```
import * as stories from './Skeleton.stories';
import type { StoryFile } from '../../util/utility-types';

describe('<Skeleton />', () => {
  generateSnapshots(stories as StoryFile);
});
```

src/components/ToastNotification/ToastNotification.stories.tsx
```diff
@@ -94,7 +94,7 @@ const ToastNotificationManager = (args: Args) => {
         Trigger A Toast Notification
       </Button>
       <div
-        className="dur absolute bottom-0 left-0 flex flex-col gap-size-2"
+        className="dur gap-size-2 absolute bottom-0 left-0 flex flex-col"
```
booc0mtaco, line 97: some of these with `-size-` slide around b/c we now need to update it to things like `gap-spacing-size-2`. Also for this one line ... that `dur` is suspicious. maybe it should be deleted while we are here?

booc0mtaco (commented):
A few remaining ones to adjust but looking GOOD

booc0mtaco:
Also I think most of the UI tests show diff.s because the renaming of the `size` utils to include `spacing` now

booc0mtaco:
@ajmoreno Looking at some of the diffs, and realized the work for fixing `-size-` usages will touch most (all?) of the module.css files thru a big find/replace to make sure they're `-spacing-size-` instead.

Not sure why the build is failing, but I can help debug. Just lemme know

ajmoreno:
> @ajmoreno Looking at some of the diffs, and realized the work for fixing `-size-` usages will touch most (all?) of the module.css files thru a big find/replace to make sure they're `-spacing-size-` instead.
> 
> Not sure why the build is failing, but I can help debug. Just lemme know

yeah, i wasn't sure how pervasive of a change i should be making but will go through and adjust them all

ajmoreno:
also i think the failing build is because i need to update the tag snapshots; i think it let me push the change because the prior version didn't have any snapshots at all to diff against but the ci test is going against a version that still does have the old tag snapshots

booc0mtaco:
@ajmoreno Took another look at the diffs and the last remaining set of visual diffs seem to come from this file  using deprecated https://github.com/chanzuckerberg/edu-design-system/blob/main/src/design-tokens/css/base/body.css (the tokens in here have been removed.

We can create another ticket for figuring out what tokens should be used there, since we need to chat with design on what the base foreground/background tokens are. I'll check thru the diff and mark any that need fixing that i notice

src/components/Avatar/Avatar.module.css
```diff
@@ -33,50 +33,50 @@
...
+  height: calc(var(--eds-spacing-size-12) / 16 * 1rem);
+  width: calc(var(--eds-spacing-size-12) / 16 * 1rem);
 }
 
 .avatar--xxxl {
   font: var(--eds-theme-typography-headline-lg-bold);
 
-  height: calc(calc(var(--eds-size-base-unit) / 16 * 1rem) * 20);
-  width: calc(calc(var(--eds-size-base-unit) / 16 * 1rem) * 20);
+  height: calc(calc(var(--eds-spacing-size-1) / 16 * 1rem) * 20);
```
booc0mtaco, line 78: ++

src/components/LoadingIndicator/LoadingIndicator.tsx
```diff
@@ -29,10 +29,10 @@ export type LoadingIndicatorProps = {
...
 // Pixel sizes corresponding to EDS size units
 const loaderSize = {
-  xs: 16, // --eds-size-1
-  sm: 24, // --eds-size-2-and-half
-  md: 40, // --eds-size-5
-  lg: 56, // --eds-size-7
+  xs: 16, // --eds-spacing-size-1
+  sm: 24, // --eds-spacing-size-2-and-half
+  md: 40, // --eds-spacing-size-5
+  lg: 56, // --eds-spacing-size-7
```
booc0mtaco, line 35: ++

booc0mtaco (approved):
LGTM!

PR #2162, opened 2025-04-01
feat(AppNotification): add variant prop with assert

booc0mtaco:
- add in the new prop to use, instead of color
- add in assertion to check if color is used
- mark color as deprecated
- update tests and snapshots

### Test Plan:


ajmoreno (approved):
lgtm

PR #2163, opened 2025-04-10
fix(storybook): set default foreground/background colors for story pages

booc0mtaco:
Remove token usages for foreground/background colors in storybook CSS. Remove token which is no longer being used. Updates to a few components where the base color to use should have been specified but wasn't (we got it for free in some cases, but the component should have specified it deliberately instead of relying on storybook to do so).

### Test Plan:

### DIFF
--- a/src/components/Modal/Modal.stories.tsx
+++ b/src/components/Modal/Modal.stories.tsx
@@ -206,7 +206,9 @@ export const ContentDefault: Story = {
     children: (
       <>
         <Modal.Header>
-          <Heading as="h2">Modal Title</Heading>
+          <Heading as="h2" className="text-utility-default-primary">
+            Modal Title
+          </Heading>
           <Modal.SubTitle>Modal Sub-title</Modal.SubTitle>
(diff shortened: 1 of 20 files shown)


jgao-czi (approved):

src/components/Modal/Modal.stories.tsx
```diff
@@ -206,7 +206,9 @@ export const ContentDefault: Story = {
     children: (
       <>
         <Modal.Header>
-          <Heading as="h2">Modal Title</Heading>
+          <Heading as="h2" className="text-utility-default-primary">
```
booc0mtaco, line 209: This `Heading` component is used in these tests to work around a problem in HeadlessUI and storybook..

* this should be `Modal.Title`
* HeadlessUI doesn't allow for the underlying `.Title` component to be outside the parent
* Since our `Modal` wrapper uses `Heading` under the hood, we use it here to display the open modal in stories

A follow-up task is to make this abundantly clear in the documentation.

PR #2180, opened 2025-05-07
fix(TabGroup): use proper tier 2+ tokens with component

booc0mtaco:
The component was incorrectly using the tier-1 tokens directly. Change the tokens used to match the ones in design. Also fix the errant usage of the recently removed tier 1 `anim` tokens and replace with `ease-in-out`

https://github.com/user-attachments/assets/38427963-4ec9-4463-824f-ac273a6802fd

### Test Plan:


ajmoreno (approved):
lgtm

PR #2181, opened 2025-05-08
feat(Typography): add in full set of Graphik fonts and weights

booc0mtaco:
Since we changed the font weight tokens and values, we now had a set of weights that didn't exactly correspond to the available font files to use. Add in the full set so that any future weight changes can be supported by the matching font file (for Graphik, the drawn font can differ for each weight, and the browser won't try to automatically draw a weight)

### Test Plan:


ajmoreno (approved):
lgtm

PR #2182, opened 2025-05-08
fix(InputChip): add padding to inside left edge to match design

booc0mtaco:
Add left padding equivalent to 12px to the inside of `InputChip` components to match design

### Test Plan:


ajmoreno (approved):
👍

PR #2183, opened 2025-05-08
refactor(config): export and document tailwind config utility method

booc0mtaco:
Before, anyone using the theming tools would have to occasionally combine keys and values very similar to what comes with EDS, in order to get the same config results using their theme file. Inconvenient!

Instead, move this creation logic into an exported method AND use the method to generate the default config. This way we re-use the logic internally, and provide a simpler interface for using the old method (duplicate logic seen in EDS in your project directly).

Update the documentation to reflect this improvement 

### Test Plan:

  - [x] use the function in the built-in storybook to confirm no snapshots change unexpectedly

### DIFF
--- a/.storybook/components/Docs/Guidelines/Theming.mdx
+++ b/.storybook/components/Docs/Guidelines/Theming.mdx
@@ -153,45 +153,24 @@ npx eds-import-from-figma-api file <file_id> --token <personal_access_token>
 
 When you have your own custom theme, you can use the tokens provided in `app-tailwind-theme.config.json` to do advanced tailwind configuration. This file contains all the tokens in JSON format, mapped to the literal values in your local theme.
 
-You can use similar import values to what is in `@chanzuckerberg/eds/tailwind.config.ts` in your local tailwind configuration file:
+EDS provides a utility method to pull in the theme config and generate the proper tailwind 3.x config object.
+
+**NOTE**: Tailwind 4.x support to come in a future release of EDS.
 
 ```ts
 // in your tailwind.config.ts file
 import type { Config } from 'tailwindcss';
-import baseConfig from '@chanzuckerberg/eds/tailwind.config';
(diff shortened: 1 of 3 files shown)


.storybook/components/Docs/Guidelines/Theming.mdx
```diff
@@ -153,45 +153,24 @@ npx eds-import-from-figma-api file <file_id> --token <personal_access_token>
...
-You can use similar import values to what is in `@chanzuckerberg/eds/tailwind.config.ts` in your local tailwind configuration file:
+EDS provides a utility method to pull in the theme config and generate the proper tailwind 3.x config object.
+
+**NOTE**: Tailwind 4.x support to come in a future release of EDS.
 
 ```ts
 // in your tailwind.config.ts file
 import type { Config } from 'tailwindcss';
-import baseConfig from '@chanzuckerberg/eds/tailwind.config';
+import {applyTailwindConfig} from '@chanzuckerberg/eds/tailwind.config';
```
booc0mtaco, line 163: Previously, one would import the config from EDS, and mix/match it with the local config. Now, the base config can come thru this method. The easy path is to use it directly, but doesn't prevent the previous behavior

ajmoreno (approved):
🚢

src/design-tokens/css/base/reset.css
```diff
@@ -4,14 +4,13 @@
   box-sizing: border-box;
 }
 
-/* TODO-AH: figure out what these tokens should be */
 ::-moz-selection {
   /* Code for Firefox */
   color: var(--eds-theme-color-text-highlight-foreground);
-  background: var(--eds-theme-color-text-highlight-background);
+  background: var(--eds-theme-color-background-utility-default-low-emphasis);
```
booc0mtaco, line 10: noticed this documentation / reset glitch (using undefined token), so setting a token to use in this context

booc0mtaco:
merge with merge commit to capture both commit messages in the change log

PR #2190, opened 2025-05-23
fix(scripts): address handling of color types

booc0mtaco:
[fix(scripts): address handling of color types](https://github.com/chanzuckerberg/edu-design-system/commit/2ba97af688f23a1649c70fc14d375eab732fc31d) 

- truncate decimal precision on rgba colors
- handle wrapping of raw colors in --local runs
- update the values by importing from figma and saving to core-tokens.json

While here, we can now use regular hex values for the white/black tokens, since we don't have to manually map opacities for some tokens' values.

### Test Plan:

  - [x] run import script in dry-run mode to check output
  - [x] run script on EDS itself and check changes to the tokens


booc0mtaco:
Usage doc.s: https://chanzuckerberg.github.io/edu-design-system/?path=/docs/documentation-theming--docs#setting-up-and-using-the-theming-tooling

forgeabout (approved):
I met with Andrew, and other than the js files, these are almost entirely generated from the scripts now. If the last few issues requiring manual edits are resolved, it would be great to formally classify the files as generated-- via file headers, README, etc. If we can't fully remove the need for manual intervention, then it would be good to put the manual portion in separate files, which perhaps the script can then include in the generated files.

PR #2195, opened 2025-06-04
feat(script): add interactive options for selecting token tier information

booc0mtaco:
- use enquirer to prompt for tier information
- add in checks for the correct tier values being used
- remove brittle constants in script
- clean up some comments and documentation
- add better UI for script initializing
- update tests

### Test Plan:

  - [x] see video
  - [x] ran locally on EDS to make sure it works the same as before without any errors

### DIFF
--- a/bin/_util.js
+++ b/bin/_util.js
@@ -2,9 +2,6 @@
 
 // NOTE: Do not use directly. Extend this to support re-writing eds-import-from-figma for non-enterprise use cases
 class AbstractFigmaReader {
-  static TIER_1_MODE = '6181:0';
-  static EDS_COLLECTION_NAME = 'EDS tokens';
-
   constructor(jsonData) {
     this._jsonData = jsonData;
   }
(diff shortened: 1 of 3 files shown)


booc0mtaco:
https://github.com/user-attachments/assets/12860988-8c5d-45b0-905d-5586975c08ef

bin/_util.js
```diff
@@ -2,9 +2,6 @@
 
 // NOTE: Do not use directly. Extend this to support re-writing eds-import-from-figma for non-enterprise use cases
 class AbstractFigmaReader {
-  static TIER_1_MODE = '6181:0';
```
booc0mtaco, line 5: Main goal of this change is to remove these very specific and brittle values for the script. instead of hard-coding them, we pull data from the figma payload which map to the dynamic values stored in figma itself. Guidance provided on which ones mean what.

ajmoreno (approved):
lgtm

PR #2196, opened 2025-06-06
Add sublabel prop for input fields (Select, InputField, TextareaField)

booc0mtaco:
* Add `sublabel` for each of the listed components
* Update snapshots and test descriptions

### TODO
- [ ] update aria-label handling for each component so that (sub-)label is part of each labeled field

### Test Plan:

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -80,6 +80,7 @@ export const parameters: Preview['parameters'] = {
     ],
   },
   badgesConfig: {
+    ...createComponentVersion('3.1'),
     ...createComponentVersion('3.0'),
     ...createComponentVersion('2.1'),
     ...createComponentVersion('2.0'),
--- a/src/components/DataTable/__snapshots__/DataTable.test.ts.snap
+++ b/src/components/DataTable/__snapshots__/DataTable.test.ts.snap
@@ -10589,11 +10589,12 @@ exports[`<DataTable /> WithLongCaption story renders snapshot 1`] = `
           class="input-field__body"
         >
           <input
+            aria-describedby=""
             aria-invalid="false"
             aria-label="search"
             class="input input-field__input--leading-icon"
             disabled=""
-            id=":r16:"
+            id=":r19:"
             placeholder="Search..."
             type="search"
           />
@@ -10760,6 +10761,7 @@ exports[`<DataTable /> WithSearch story renders snapshot 1`] = `
           class="input-field__body"
         >
           <input
+            aria-describedby=""
             aria-invalid="false"
             aria-label="search"
             class="input input-field__input--leading-icon"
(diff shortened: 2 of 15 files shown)


booc0mtaco:
Merging as a merge to preserve history during release

.storybook/preview.tsx
```diff
@@ -80,6 +80,7 @@ export const parameters: Preview['parameters'] = {
     ],
   },
   badgesConfig: {
+    ...createComponentVersion('3.1'),
```
booc0mtaco, line 83: Each documentation page can have badges showing the version of the theme/component API. This adds an entry so we can reference 3.1 versions

src/components/DataTable/__snapshots__/DataTable.test.ts.snap
booc0mtaco, line 1: all the `.snap` files are getting ID updates since we add calls to `useId` in react (essentially they create sequential IDs that react tracks to avoid collisions)

src/components/FieldLabel/FieldLabel.tsx
```diff
@@ -41,8 +41,7 @@ export const FieldLabel = React.forwardRef<HTMLLabelElement, FieldLabelProps>(
   ({ children, className, htmlFor, size = 'lg', disabled, ...other }, ref) => {
     const componentClassName = clsx(
       styles['label'],
-      size === 'md' && styles['label--md'],
-      size === 'lg' && styles['label--lg'],
+      size && styles[`label--${size}`],
```
booc0mtaco, line 44: collapsed this into one entry since `size` is typed and can't be arbitrary. Less is more :)

src/components/DataTable/__snapshots__/DataTable.test.ts.snap
jecortez: What exactly is the format of the ids? Why is one `:r19:` vs `:rt:`?

src/components/TextareaField/TextareaField.tsx
```diff
@@ -187,20 +189,22 @@ export const TextareaField: TextareaFieldType = forwardRef(
...
       styles['textarea-field__overline'],
       !label && styles['textarea-field__overline--no-label'],
       disabled && styles['textarea-field__overline--disabled'],
     );
+
     const labelClassName = clsx(
       disabled && styles['textarea-field__label--disabled'],
     );
 
+    const subLabelClassName = clsx(
```
jecortez, line 203: This is the tiniest of nits, but I noticed `sublabel` tends to not be camelCase throughout, but this const + another one below are camelCased. Technically for consistency it would be
```ts
    const sublabelClassName = clsx(
```

src/components/DataTable/__snapshots__/DataTable.test.ts.snap
booc0mtaco: Definition of HTML IDs here: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/id#syntax (react does something to figure out how to iterate the part after the `r` so it could be as simple as `0-9a-z+` in order)

jecortez (approved):

PR #2201, opened 2025-06-27
fix(scripts): update script wrapper to use proper tier 1 mode

booc0mtaco:
- previously, it was sending the collection ID to FigmaVariable
- now, add an enquirer entry to select that mode from the tier 1 collection
- also update the class to return all the collection information
- use the additional collection information to filter out the unpublished collection data
- doulbe check that tests still pass
- add in additional tests for script logic
- add TODOs to flag refactor opportunities

### Test Plan:

  - [x] run script in local (`--dry-run`) mode to test output and function
  - [x] create alpha release and test against evaluators app via version bump

### DIFF
--- a/bin/_util.js
+++ b/bin/_util.js
@@ -171,25 +171,17 @@ module.exports = {
 
     /**
      * Retrieve the defined variable collections from the specific figma file data
-     * @see https://www.figma.com/plugin-docs/api/VariableCollection/
+     * @see https://www.figma.com/developers/api#variables
      *
      * @returns VariableCollection[]
      */
     getVariableCollections() {
-      // TODO-AH: remove the redundant rows in here by making this a unique set
-      return Object.values(this._jsonData.variableCollections).map(
-        (collection) => {
-          return {
-            id: collection.id,
-            name: collection.name,
-          };
-        },
-      );
+      return Object.values(this._jsonData.variableCollections);
     }
 
     /**
      * Retrieve the set of variables associated with a given collection (by ID)
-     * @see https://www.figma.com/plugin-docs/api/Variable/
+     * @see https://www.figma.com/developers/api#variables
      *
      * @param {string} variableCollectionId
      * @returns Variable[]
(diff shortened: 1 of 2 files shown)


bin/eds-import-from-figma-api.js
```diff
@@ -124,33 +123,58 @@ const { identity } = require('lodash');
...
+    .filter((collection) => !collection.hiddenFromPublishing)
     .find(
       (collection) => collection.id === collectionNameResponses.tier2Collection,
     );
 
   // Determine which of the modes in the file should be used
-  let response;
+  let modeResponses;
   try {
-    response = await prompt({
```
booc0mtaco, line 136: From each collection we should select the mode. in the current tier 1 collection, we only have the one mode to pick, for tier 2, the mode represents the theme to use

bin/eds-import-from-figma-api.js
```diff
@@ -167,14 +191,15 @@ const { identity } = require('lodash');
...
 
+  // Main loop: work thru each detected variable, and convert
   figmaApiReader
-    .getVariablesByCollectionId(edsCollection.id)
+    .getVariablesByCollectionId(tier2Collection.id)
     .forEach((figmaVariable) => {
       stats.total.push(figmaVariable.id);
       const variable = new FigmaVariable(
         figmaVariable,
-        collectionNameResponses.tier1Collection,
```
booc0mtaco, line 176: The heart of the bug. I clearly documented this should be the tier 1 mode in the `FigmaVariable` API, but somewhere along the line forgot that the collection ID and the mode ID aren't synonymous

Future refactor should factor the main loop into a separate call, so that inputs to that loop can be tested (and the CLI for the script is separated out as its own thing)

bin/_util.js
```diff
@@ -171,25 +171,17 @@ module.exports = {
...
     /**
      * Retrieve the defined variable collections from the specific figma file data
-     * @see https://www.figma.com/plugin-docs/api/VariableCollection/
+     * @see https://www.figma.com/developers/api#variables
      *
      * @returns VariableCollection[]
      */
     getVariableCollections() {
-      // TODO-AH: remove the redundant rows in here by making this a unique set
-      return Object.values(this._jsonData.variableCollections).map(
```
booc0mtaco, line 180: no longer picking some data from `VariableCollecton` to return. Instead, give the caller the full-fat data

PR #2222, opened 2025-07-22
fix(tailwind): add in settings to use EDS border radii tokens with tailwind

booc0mtaco:
Also, minor refactoring since all reducers are exactly the same now.

In their wisdom, the Tailwind folk named their utility classes `rounded-*` so these won't map to the property name like most of the other util.s, unfortunately.

### Test Plan:

  - [x] checked by using VSCode intellisense to see if the generated config was formatted properly


ajmoreno (approved):
lgtm

PR #2223, opened 2025-07-22
fix: remove lingering usages of depecated animation token

booc0mtaco:
The token was removed, and this cleans up the few lingering usages of that token that were missed.

### Test Plan:

  - [x] see attached media


ajmoreno (approved):
lgtm

PR #2224, opened 2025-07-23
fix: sync type declarations for sublabel and fieldNote

booc0mtaco:
- use ReactNode to seamlessly allow use of null/undefined values
- verify all existing tests work

### Test Plan:


ajmoreno (approved):
👍

PR #2232, opened 2025-08-06
fix(Card): use a11y-compliant HTML tags for component

booc0mtaco:
- `<article>`, `<header>`, and `<footer>` have specific semantic meaning
- use role-less containers (`<div>`) instead
- also sonsider using `<h3>` or higher for the main card header

### Test Plan:


booc0mtaco:
Main test here is that there are no visual changes in the snapshots 📸

ajmoreno (approved):
lgtm

PR #2234, opened 2025-08-08
fix: fix typography token and usages

booc0mtaco:
- add in eds-theme-typography-font-style-headline-decorative-inline
- make proper use in Heading component
- make proper use in Text component

- run `yarn build:tokens` to build generated files

### Test Plan:

### DIFF
--- a/src/components/Heading/Heading.module.css
+++ b/src/components/Heading/Heading.module.css
@@ -13,11 +13,11 @@
   --heading-font-weight: var(--eds-typography-font-weight-normal);
   --heading-font-family: var(--eds-typography-font-family-1);
   --heading-letter-spacing: var(--eds-typography-letter-spacing-normal);
-  --heading-text-transform: ;
 
   font: unset;
   font-family: var(--heading-font-family);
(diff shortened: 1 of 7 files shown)


src/components/Heading/Heading.module.css
```diff
@@ -13,11 +13,11 @@
...
   --heading-font-family: var(--eds-typography-font-family-1);
   --heading-letter-spacing: var(--eds-typography-letter-spacing-normal);
-  --heading-text-transform: ;
 
   font: unset;
   font-family: var(--heading-font-family);
   font-size: calc(var(--heading-font-size) * 1px);
+  font-style: var(--heading-font-style, initial);
   line-height: calc(var(--heading-line-height) * 1px);
   font-weight: var(--heading-font-weight);
```
booc0mtaco, line 22: Before , i was setting a default value of ` ` to the CSS variable, which prevented the fallback from being seen and used, removing this line means it will properly use the fallback values for both `*-font-style` and the previous `*-text-transform`

ajmoreno (approved):
🚢

PR #2241, opened 2025-08-21
feat(Skeleton): add in new tokens and usages

booc0mtaco:
### Summary

* add in new tokens: `eds-theme-color-background-skeleton` and `eds-theme-color-background-skeleton-stripe`
* change animation direction
* use new tokens in component

https://github.com/user-attachments/assets/15443e7f-e2e7-4430-8d41-9cbacd767ae1

### Test Plan:


src/components/Skeleton/Skeleton.module.css
```diff
@@ -4,24 +4,24 @@
 
 @keyframes bg-wipe {
   0% {
-    background-position: 0% 0%;
+    background-position: right 0%;
   }
   100% {
-    background-position: 100% 0%;
+    background-position: left 0%;
```
forgeabout, line 10: What's the benefit of this change?

src/components/Skeleton/Skeleton.module.css
```diff
@@ -4,24 +4,24 @@
...
   }
 }
 
 /**
  * Skeleton
  */
 .skeleton {
-  border-radius: calc(var(--eds-border-radius-lg) * 1px);
-  background-color: var(--eds-theme-color-background-neutral-medium);
+  border-radius: calc(var(--eds-border-radius-150) * 1px);
```
forgeabout, line 18: Did EDS change from using symbolic sizes (e.g. lg) to numeric (e.g. 150)? What does 150 mean?

.storybook/data/tokens.json
forgeabout, line 1: Which files are generated?

forgeabout (approved):
Some comments/questions, but looks reasonalble.

.storybook/data/tokens.json
booc0mtaco: In this pr:
* .storybook/data/tokens.json
* src/tokens-dist/*.*

There's a ticket in the backlog to add headers for such files, tho it can't be applied to the .json examples

src/components/Skeleton/Skeleton.module.css
booc0mtaco: Good question. I use a linear gradient for the background image, which is always 5x the width of the container.  the way `background-position` works is that it will position the background within the container.

since the bg is bigger, we want it to start as pinned to the **right** edge (and go off screen on the left side), then transition to pinned to **left** edge (and go off screen to the right side). It's a little counter-intuitive b/c of the larger bg size (hence the bug lol).

https://developer.mozilla.org/en-US/docs/Web/CSS/background-position#regarding_percentages

PR #2242, opened 2025-08-21
feat(Modal): add in full size modal

booc0mtaco:
Add in support for `size` value `"full"`. Update stories to account for this new functionality

### Test Plan:

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -102,7 +102,7 @@ type ModalContentProps = {
    *
    * **Default is `"lg"`**.
    */
-  size?: Extract<Size, 'sm' | 'lg'>;
+  size?: Extract<Size, 'sm' | 'lg'> | 'full';
 };
 
 type ModalProps = ModalContentProps & {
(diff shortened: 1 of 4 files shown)


src/components/Modal/Modal.tsx
```diff
@@ -102,7 +102,7 @@ type ModalContentProps = {
    *
    * **Default is `"lg"`**.
    */
-  size?: Extract<Size, 'sm' | 'lg'>;
+  size?: Extract<Size, 'sm' | 'lg'> | 'full';
```
forgeabout, line 105: I guess were okay with the type here being sort of a hybrid-- some values coming from the `Size` type and the new "full" option being its own thing.

booc0mtaco: Yah, there are a small set of cases where we extend [utility types](https://github.com/chanzuckerberg/edu-design-system/pull/2242/files#diff-766208c965d34c3bf67b6cf8dc749036c694359ca121365e8e7e27c08454383eL13) with component-specific entries. This helps with keeping it clear which values are cross-component, and which are not

PR #2255, opened 2025-09-02
chore: update case-ing of sublabel in datatable [EDS-1642]

ajmoreno:
Summary:
modifies the datatable component to have properly camel-cased `subLabel` propery

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@ All notable changes to this project will be documented in this file. See [standa
 * apply the border radius tier 2 tokens ([#2215](https://github.com/chanzuckerberg/edu-design-system/issues/2215)) ([160bcc1](https://github.com/chanzuckerberg/edu-design-system/commit/160bcc17cb3775edf3c42f25f02831dfac87f161))
 * **FieldNote:** align field note icon with text ([#2225](https://github.com/chanzuckerberg/edu-design-system/issues/2225)) ([43846fb](https://github.com/chanzuckerberg/edu-design-system/commit/43846fbcc87a52c7a2ed3fa213c8ae64705970c1))
 * remove lingering usages of depecated animation token ([#2223](https://github.com/chanzuckerberg/edu-design-system/issues/2223)) ([64049e3](https://github.com/chanzuckerberg/edu-design-system/commit/64049e3b4d24c65725fccf733b995c5ccab4149b))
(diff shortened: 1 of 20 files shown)


ajmoreno:
i didnt know if i should update the `-sublabel` used in the styles or not, left them as all lower case but can change that if desired

booc0mtaco:
A few other components have the wrong case. I did a case-sensitive search for `sublabel` and found the following:

* [`InputField`](https://github.com/chanzuckerberg/edu-design-system/blob/main/src/components/InputField/InputField.tsx#L130)
* [`TextareaField`](https://github.com/chanzuckerberg/edu-design-system/blob/main/src/components/TextareaField/TextareaField.tsx#L61)
* [`Select`](https://github.com/chanzuckerberg/edu-design-system/blob/main/src/components/Select/Select.tsx#L96)
* [`Toggle`](https://github.com/chanzuckerberg/edu-design-system/blob/main/src/components/Toggle/Toggle.tsx#L41)

Changing the case in the class names is optional; we don't expose those currently, so they aren't part of any documented API. BUT if it's dead simple to find/replace those, that could be sync'd here as well (we may eventually need to offer CSS classes as part of EDS)

ajmoreno:
> A few other components have the wrong case. I did a case-sensitive search for `sublabel` and found the following:
> 
> * [`InputField`](https://github.com/chanzuckerberg/edu-design-system/blob/main/src/components/InputField/InputField.tsx#L130)
> * [`TextareaField`](https://github.com/chanzuckerberg/edu-design-system/blob/main/src/components/TextareaField/TextareaField.tsx#L61)
> * [`Select`](https://github.com/chanzuckerberg/edu-design-system/blob/main/src/components/Select/Select.tsx#L96)
> * [`Toggle`](https://github.com/chanzuckerberg/edu-design-system/blob/main/src/components/Toggle/Toggle.tsx#L41)
> 
> Changing the case in the class names is optional; we don't expose those currently, so they aren't part of any documented API. BUT if it's dead simple to find/replace those, that could be sync'd here as well (we may eventually need to offer CSS classes as part of EDS)

oh weird, i ran a case-sensitive query and only this component turned up. Maybe my branch was stale when i ran it or something, ill update

booc0mtaco:
Cool cool. There may be a few others too but I think it's just these remaining ones

Maybe there was an include/exclude filter applied?

booc0mtaco (commented):
Looks good. Let's change the base branch from `main` to `next` pls. If no conflicts, etc. will approve

PR #2264, opened 2025-09-03
chore: use CHROMATIC_PROJECT_TOKEN as environment variable

booc0mtaco:
See [docs](https://www.chromatic.com/docs/cli/#continuous-integration)

It looks like the previous config was using an unsupported (?) `=` syntax for the value, which may be tripping things up with the latest version of the `chromatic` dependency. Switch to using the environment variable approach.

Result: https://github.com/chanzuckerberg/edu-design-system/actions/runs/17445020400/job/49537232470?pr=2264#step:6:3

### Test Plan:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -79,13 +79,14 @@ jobs:
       - name: Publish to Chromatic 🦄
         run: |
           yarn run chromatic \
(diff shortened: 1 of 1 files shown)


ajmoreno (approved):
lg

PR #2279, opened 2025-09-29
chore: add permissions checks to workflows

booc0mtaco:
Add granular permissions to all workflows which do not have them currently

- `docs.yml` needs to be able to write to github pages
- `test.yml` needs to read contents from github
- `update-dependencies.yml` needs to read contents and allow renovate to create pull requests
- (I already added perm.s to the GH Registry release job)

See: https://github.com/peaceiris/actions-gh-pages/tree/v3/?tab=readme-ov-file#getting-started (`peaceiris/actions-gh-pages@v3`)

### Test Plan:

  - [x] verify that CI jobs pass for test jobs in this PR
  - [x] TBD at next release, verify that Pages publish successfully
  - [ ] TBD at next renovate run, verify that it ran without failing (update number of weekly runs to increase frequency)

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,4 +1,4 @@
-name: Deploy Docs to GH Pages
+name: Deploy Documentation to GitHub Pages
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


booc0mtaco:
> lgtm, for the most part when i did this i just kept crashing into git hub actions failing due to insufficient permissions and would increase the scope as i hit each new failure but it looks like this built cleanly so nice work!

@ajmoreno yah, i expect that it may glitch out with the GH pages publish or renovate run, so will make a release that has ONLY this, to test those out without having to hack anything :P

PR #2301, opened 2025-10-24
fix(AppHeader): check for breakpoint when calculating orientation

booc0mtaco:
Use the `window`s dimensions to override the orientation of `AppHeader` when at the small breakpoint. 

### Test Plan:


czibbelvin (approved):
Approved, but I'm still leery of relying on JS for this behavior

PR #2436, opened 2026-03-30
feat(AppHeader): update app layout based on feedback

booc0mtaco:
### Test Plan:

### DIFF
--- a/src/components/AppHeader/AppHeader.module.css
+++ b/src/components/AppHeader/AppHeader.module.css
@@ -76,7 +76,7 @@
   }
 
   & > div {
-    padding: calc(var(--eds-spacing-size-1) * 1px);
+    padding: calc(var(--eds-spacing-size-4) * 1px) calc(var(--eds-spacing-size-1) * 1px);
   }
 
   &.app-header--style-floating > div {
--- a/src/components/AppHeader/AppHeader.module.css
+++ b/src/components/AppHeader/AppHeader.module.css
@@ -76,17 +76,13 @@
   }
 
   & > div {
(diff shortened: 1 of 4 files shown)


src/components/AppHeader/AppHeader.module.css
```diff
@@ -76,25 +76,30 @@
...
   .app-header__content {
     align-items: flex-end;
     flex-direction: column;
     height: 100%;
   }
-}
+
+  .app-header__nav-group ul {
+    gap: 0;
+  }
```
jecortez, line 96: Will this only gap 0 in vertical? Or also horizontal?

jecortez (commented):
Curious about the gap:0 change

src/components/AppHeader/AppHeader.module.css
booc0mtaco: @jecortez Not visible is [L61](https://github.com/chanzuckerberg/edu-design-system/pull/2436/changes#diff-c9f6c5bf038586d21971d78860d3d654f49c09c3a1405ee5db0bcca96f50fc44L69), which scopes this to just when orientation is set to vertical

jecortez (approved):
I had noticed somewhat noticeable changes in the storybook test when clicking the menu popover in horizontal, as long as that was intentional this seems good

booc0mtaco:
> I had noticed somewhat noticeable changes in the storybook test when clicking the menu popover in horizontal, as long as that was intentional this seems good

Yep, as a part of this, changed the content of those stories to align with designed components. contents of those approved by design so we're solid 🏁

PR #2461, opened 2026-04-15
fix(AppHeader): do not show menu when there are no nav groups

booc0mtaco:
Use the existence / length of `navGroup` to determine if we need to handle the hamburger menu behavior.

### Test Plan:


ajmoreno (approved):
lgtm

PR #2466, opened 2026-04-20
fix: direct postcss rollup plugin to emit lib/index.css always

booc0mtaco:
use `postcss()` to ALWAYS emit index.css when bundling EDS. because we have multiple inputs now (index.ts and tokens.ts), rollup would sometimes resolve to emit the css file to match either of the inputs. Almost always, it would choose `index` but on some occasions it would choose `tokens`, even tho that file does not have any imported non-JS content.

Tested this locally by first trying to change the asset file name using `assetFileNames` provided by rollup itself, then by configuring the `extract` value used by the plugin, which overrides the css content emissions.

### Test Plan:

  - [x] make sure bundle report doesn't show renamed css files

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -1,3 +1,4 @@
+import path from 'node:path';
 import { codecovRollupPlugin } from '@codecov/rollup-plugin';
 import commonjs from '@rollup/plugin-commonjs';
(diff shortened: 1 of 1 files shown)


ajmoreno (approved):
lgtm

PR #2498, opened 2026-05-18
fix(PopoverListItem): align sizing for leading and trailing content slots

booc0mtaco:
The leading and trailing content areas in popovers were not using the same exact layout ergonomics. In particular, `trailingContent` was not using flexbox while `leadingContent` was. Change this, and also fix how trailing content aligns when sub labels are in use.

### Test Plan:

### DIFF
--- a/src/components/PopoverListItem/PopoverListItem.stories.tsx
+++ b/src/components/PopoverListItem/PopoverListItem.stories.tsx
@@ -44,6 +44,17 @@ export const Default: Story = {
   },
 };
 
+export const DefaultWithTrailingContent: Story = {
+  args: {
+    children: 'Favorite',
+    leadingContent: (
+      <Icon name={'heart-filled'} purpose="decorative" size="24px" />
+    ),
+    trailingContent: '5',
+    subLabel: 'Save posts for later',
+  },
+};
+
 /**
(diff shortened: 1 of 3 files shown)


src/components/PopoverListItem/PopoverListItem.stories.tsx
```diff
@@ -44,6 +44,17 @@ export const Default: Story = {
   },
 };
 
+export const DefaultWithTrailingContent: Story = {
+  args: {
+    children: 'Favorite',
+    leadingContent: (
+      <Icon name={'heart-filled'} purpose="decorative" size="24px" />
+    ),
+    trailingContent: '5',
```
michael-berger-czi, line 53: Optional: should leadingContent and trailingContent have storybook controls in these new stories?

michael-berger-czi (approved):

src/components/PopoverListItem/PopoverListItem.stories.tsx
booc0mtaco: Both of these prop.s can have strings or React components in there, and storybook doesn't handle the React component control situation nicely. So I disabled them. That said... I **could** configure the controls to use Icon names, and render the component 🤔 I'll add that to the backlog

PR #2503, opened 2026-05-20
feat: allow programmatic close on Menus and AppHeader

booc0mtaco:
Support changes to the component interface to address a few functionalities:

* `Menu` now exposes the rest of the API from HeadlessUI's Dropdown Menu, including all the sub-components
* Each of these is mapped to the corresponding EDS component (e.g., `PopoverListItem`)
* Allow usage of render prop for `Menu` body as well, which includes ability to trigger `close()`
* Hook up `AppHeader` to use this functionality thru a `shouldClose` value on relevant `NavItem` entries
* Default behavior still exists (without specifying `shouldClose` all sub nav items of a navItem type `"menu"` will have an optional prop to set

### Test Plan:

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -1,8 +1,11 @@
 import {
   Menu as HeadlessMenu,
-  MenuButton as HeadlessMenuButton,
   MenuItem as HeadlessMenuItem,
   MenuItems as HeadlessMenuItems,
+  MenuButton as HeadlessMenuButton,
+  MenuSection as HeadlessMenuSection,
+  MenuHeading as HeadlessMenuHeading,
+  MenuSeparator as HeadlessMenuSeparator,
 } from '@headlessui/react';
 import type { AnchorProps } from '@headlessui/react/dist/internal/floating';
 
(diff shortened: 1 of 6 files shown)


src/components/Menu/Menu.tsx
```diff
@@ -117,10 +121,41 @@ const MenuButton = ({
...
+  <HeadlessMenuSection {...props} />
+);
+
+/**
+ * Separates two `Menu.Section` components, with proper accessibility semantics.
+ *
+ * @see https://headlessui.com/react/menu#menu-separator
+ */
+const MenuSeparator = (props: MenuSeparatorProps) => (
+  <HeadlessMenuSeparator __type="separator" as={PopoverListItem} />
```
GregSilvia: props declared but never spread here, same on MenuHeading below. Intentional?

booc0mtaco: Good callout. `MenuSeparator` only has `as` in it's API in headlessUI, so i could actually spread that onto the underlying component, but it's basically inert in the current version of HeadlessUI.

I'll add `{...props}` used so that we always override it with `PopoverListItem` and it's future-proofed

GregSilvia (approved):
LGTM ✔️

PR #2508, opened 2026-05-22
feat(Select): use controls to mark selection state

booc0mtaco:
Add controls to each select option, such that we see radio buttons when single select, and check boxes when multi-select

### Test Plan:

### DIFF
--- a/src/components/AppHeader/AppHeader.stories.tsx
+++ b/src/components/AppHeader/AppHeader.stories.tsx
@@ -13,6 +13,9 @@ export default {
 
   parameters: {
     chromatic: {
+      // Using this motion preference for components where they trigger animations on mount
+      prefersReducedMmotion: 'reduce',
+      delay: 500,
       viewports: [
         chromaticViewports.googlePixel2,
         chromaticViewports.ipadMini,
--- a/src/components/AppHeader/AppHeader.stories.tsx
+++ b/src/components/AppHeader/AppHeader.stories.tsx
@@ -14,7 +14,7 @@ export default {
   parameters: {
     chromatic: {
       // Using this motion preference for components where they trigger animations on mount
-      prefersReducedMmotion: 'reduce',
+      prefersReducedMotion: 'reduce',
       delay: 500,
       viewports: [
         chromaticViewports.googlePixel2,
--- a/src/components/Menu/Menu.stories.tsx
+++ b/src/components/Menu/Menu.stories.tsx
@@ -16,6 +16,8 @@ export default {
   component: Menu,
   parameters: {
     layout: 'centered',
+    // Using this motion preference for components where they trigger animations on mount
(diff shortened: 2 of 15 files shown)


src/components/Select/__snapshots__/Select.test.tsx.snap
booc0mtaco, line 1: Snapshot / React fun! all of these changed b/c the internal counter now gets 1-2 more entries due to the presence of more components using `useId`

booc0mtaco:
https://github.com/user-attachments/assets/e205adf2-34d8-46cb-bbbb-19822d091924

src/components/AppHeader/AppHeader.stories.tsx
```diff
@@ -13,6 +13,9 @@ export default {
 
   parameters: {
     chromatic: {
+      // Using this motion preference for components where they trigger animations on mount
+      prefersReducedMmotion: 'reduce',
```
jecortez: is `Mmotion` a typo or is this an acronym / field I don't know 😓 ?

src/components/Menu/Menu.stories.tsx
```diff
@@ -16,6 +16,8 @@ export default {
   component: Menu,
   parameters: {
     layout: 'centered',
+    // Using this motion preference for components where they trigger animations on mount
+    chromatic: { delay: 500, prefersReducedMmotion: 'reduce' },
```
jecortez: So I did find this chromatic option [here](https://www.chromatic.com/docs/playwright/configure/#chromatic-options:~:text=prefersReducedMotion), but it only has one M. Is this working as expected atm?

jecortez (commented):
Have a question about that chromatic option

src/components/AppHeader/AppHeader.stories.tsx
booc0mtaco: typo 😅 . I'm surprised this passes type checks!

src/components/Menu/Menu.stories.tsx
booc0mtaco: > So I did find this chromatic option [here](https://www.chromatic.com/docs/playwright/configure/#chromatic-options:~:text=prefersReducedMotion), but it only has one M. Is this working as expected atm?

This is tricky to verify b/c of how `animation` works. This will ensure it DOESN'T play the animation and as a result doesn't potentially produce flaky snapshots

booc0mtaco: Looked into the storybook types for this and it's 

```
interface Parameters {
    [name: string]: any;
}
```

Oh well gotta be way more careful

booc0mtaco:
From design, should add treatment such that there's the proper border color when hovered and not hovered

jecortez (approved):

PR #2516, opened 2026-06-02
feat(ToastNotification): update theming and presentation

booc0mtaco:
Screenshot 2026-06-02 at 11 07 58

### Test Plan:

### DIFF
--- a/src/components/ToastNotification/ToastNotification.tsx
+++ b/src/components/ToastNotification/ToastNotification.tsx
@@ -113,7 +113,8 @@ export const ToastNotification = ({
           iconLayout="icon-only"
           onClick={onDismiss}
           rank="tertiary"
-          variant="neutral"
+          size="md"
+          variant="inverse"
         />
       )}
     </div>
--- a/src/components/ToastNotification/ToastNotification.tsx
+++ b/src/components/ToastNotification/ToastNotification.tsx
@@ -35,7 +35,7 @@ export type ToastNotificationProps = {
   /**
    * Keyword to characterize the state of the notification
    */
-  status?: Extract<Status, 'favorable' | 'critical'>;
+  status?: Status;
   /**
    * The title/heading of the notification
    */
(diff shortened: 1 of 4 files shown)


src/components/ToastNotification/ToastNotification.tsx
```diff
@@ -46,6 +46,18 @@ export type ToastNotificationProps = {
  * `import {ToastNotification} from "@chanzuckerberg/eds";`
  *
  * Toasts display brief, temporary notifications. They're meant to be noticed without disrupting a user's experience or requiring an action to be taken.
+ *
+ * ## Sizing
+ *
+ * Toast notifications use a fixed width of 384px and their height depends on the length of the notification message. As noted in the content guidelines, limit toast notifications to two lines of text.
```
michael-berger-czi: Is the fixed width comment still true? Seems like it scales down now with the screen size.

src/components/ToastNotification/ToastNotification.tsx
```diff
@@ -46,6 +46,18 @@ export type ToastNotificationProps = {
...
+ *
+ * ## Dismissal
+ *
+ * Toast notifications can be manually dismissed or auto-dismissed using the dismissType prop. Auto toasts are automatically dismissed after 8 seconds.
```
michael-berger-czi: It doesn't seem related to this PR, but when I set the manual control for the "Example Dismissing Toasts" story, the toasts still auto-dismiss.

src/components/ToastNotification/ToastNotification.tsx
```diff
@@ -46,6 +46,18 @@ export type ToastNotificationProps = {
...
+ *
+ * ## Placement and Behavior
+ *
+ * Toast notifications slide in and out from the borrom right of the screen. The toast should be placed 24 px away from the top and side of the viewport. The toast notification fades in with the token eds-anim-fade-long and fade out with token eds-animation-fade-quick.
```
michael-berger-czi, line 60: nit: Doesn't it slide out to the left with the leaveTo `translate-x-[-100%] ` prop? Also doesn't it grow instead of sliding?

michael-berger-czi (commented):
A few Qs, will approve after discussion.

src/components/ToastNotification/ToastNotification.tsx
booc0mtaco: I copied these implementation notes from the design files but didn't double-check they matched what the code was doing 😅 . Lemme do a sweep of the text and update where appropriate. Some changes might be to this language, and others might be to the code itself

src/components/ToastNotification/ToastNotification.tsx
booc0mtaco: I'll update the code example to match these instructions. The code example was built originally to show how to use `<Transition>` with this component, but the design guidance describes a different behavior.

src/components/ToastNotification/ToastNotification.tsx
booc0mtaco: Good catch. there was an override in the story which ignored the control

michael-berger-czi (approved):
LGTM

ohjphan:
@booc0mtaco I appreciate how the icon is now aligned with the first line of text. Im wondering if we should do the same for the X as well for consistency to refine it. cc @janachoi  to explore/discuss.

PR #2518, opened 2026-06-02
fix(InputField): add check for value when calculating character length

booc0mtaco:
* add in a controlled story which reflects the function of character counting

### Test Plan:

### DIFF
--- a/src/components/InputField/InputField.tsx
+++ b/src/components/InputField/InputField.tsx
@@ -182,8 +182,9 @@ export const InputField: InputFieldType = forwardRef(
     ref,
   ) => {
     const shouldRenderOverline = !!(label || required);
+    // Default to `.value` to match React's preference order
     const [fieldText, setFieldText] = useState(
-      other.defaultValue || other.value,
+      other.value || other.defaultValue,
     );
 
     // Handling of behavior when field type is password. Show/hide button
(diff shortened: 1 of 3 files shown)


src/components/InputField/InputField.tsx
```diff
@@ -182,7 +182,9 @@ export const InputField: InputFieldType = forwardRef(
     ref,
   ) => {
     const shouldRenderOverline = !!(label || required);
-    const [fieldText, setFieldText] = useState(other.defaultValue);
+    const [fieldText, setFieldText] = useState(
+      other.defaultValue || other.value,
```
michael-berger-czi: Curious why defaultValue takes priority over value if they're both truthy

booc0mtaco: Looked into this. My initial thinking was to keep the previous behavior as the default, but I think this should match React's priority order. If both exist, it'll make the component controlled and ignore `defaultValue`

* By definition, using both values at the same time results in React ignoring `defaultValue` ([ref](https://github.com/facebook/react/issues/20599))
* `value` is the controlled value, meaning an `onChange` method is needed to properly handle the values
* setting `defaultValue` as the primary state source would conflict with react behavior

michael-berger-czi (approved):
LGTM

PR #2528, opened 2026-06-11
fix: move parabolica-text offsets to Text component

booc0mtaco:
`parabolica-text` renders pushed down by ~1px for some reason, which causes obvious misalignments in smaller elements. add the GPU-accelerated offsets in `Text` instead of spreading the contagion of offets across more and more components.

Stop-gap: we may need to add additional tokens to track future positioning offsets so this behavior is themeable.

### Test Plan:

### DIFF
--- a/src/components/Button/Button.module.css
+++ b/src/components/Button/Button.module.css
@@ -29,16 +29,12 @@
   width: 100%;
 
   .button--sm & {
-    /* Offset to better align text in button */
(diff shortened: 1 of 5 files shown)


booc0mtaco:
Abandoning this since it causes an unapetizing number of diff.s and follow up changes. The one-off adjustments already in the codebase produce fewer changes, and need less review

PR #2549, opened 2026-07-03
feat(TabGroup): scroll tabs into view upon selection

booc0mtaco:
Add the following behavior:

* when activating a tab, track the previously selected one in state
* use this to show an additional tab with a smooth scroll, so that users can pick more tabs if needed
* properly handle the underline such that it spans the view properly

https://github.com/user-attachments/assets/38d98320-1781-426b-84ba-9e33489204d2

### Test Plan:

### DIFF
--- a/src/components/TabGroup/TabGroup.tsx
+++ b/src/components/TabGroup/TabGroup.tsx
@@ -1,4 +1,5 @@
 import clsx from 'clsx';
+import clamp from 'lodash/clamp';
 import debounce from 'lodash/debounce';
 import React, {
   type ReactNode,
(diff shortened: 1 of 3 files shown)


src/components/TabGroup/TabGroup.tsx
```diff
@@ -178,11 +182,31 @@ export const TabGroup = ({
...
+    if (previousIndexState !== null) {
+      const desiredIndex = clamp(
+        previousIndexState > activeIndexState
+          ? activeIndexState - 1
+          : activeIndexState + 1,
+        0,
+        tabRefs.length - 1,
+      );
+
+      if (tabRefs[desiredIndex].current?.scrollIntoView) {
```
booc0mtaco: Added this check b/c jest / jsdom do not stub out `.scrollIntoView`, which causes errors in test. Once we move to `vitest` we get better support here.

booc0mtaco:
Behavioral check:

- [x] should only scroll into view horizontally, and when there is scrollable content to reveal

jgao-czi (approved):

PR #2604, opened 2026-08-12
docs(VisualPageIndicator): add usage, interaction, and accessibility guidance

booc0mtaco:
Fills in the `VisualPageIndicator` docblock following the pattern established in #2567. Content comes from [EDS-2108](https://czi.atlassian.net/browse/EDS-2108).

**Every Interaction claim in the ticket checked out**, so that section is carried over close to verbatim. The dots are `<li>` elements with no handlers, no `tabIndex`, and no role, so "not interactive" is literally true. The active dot swaps a background-color token via `.visual-page-indicator--active`.

Two additions beyond the ticket, both verified:

- **The `activePage` contract.** It's zero-based, so the last page is `totalPageCount - 1`. Fewer than two dots warns, and an out-of-range `activePage` logs an error. All three behaviors have existing tests (`errors when active page is above range`, `below range`, `warns when total page count is too small`), so the docs now match what's already asserted.
- **A Content & Accessibility lead-in stating this is a visual cue only.** The dots render as empty list items with no text and no `aria-current`, so a screen reader gets "list, N items" and nothing about position. The recommendation is to state the position in text nearby, e.g. a "Step 2 of 5" label.

On that last point: [EDS-1587](https://czi.atlassian.net/browse/EDS-1587) proposed adding a Storybook example for exactly this ("we should not just use the page indicators as the only way to indicate position in a flow") and was closed **Won't Do**. I read that as declining the *story*, not disputing the guidance, so I documented the guidance in prose rather than proposing new stories. Worth a second opinion if that reading is wrong.

**No Type/Use table.** The component has only `activePage` and `totalPageCount` and no variants, so there's no axis to tabulate. Same call as #2591, #2595, and #2599.

No Resources section; the ticket had none and there's no underlying library.

The story had no `parameters` block, so one was added for the subtitle.

### Test Plan:

Docs-only change, no runtime behavior touched.

- [x] Manually tested my changes, and here are the details: `tsc --noEmit` clean, eslint clean, prettier clean. All 6 VisualPageIndicator tests and 3 snapshots pass unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[EDS-2108]: https://czi.atlassian.net/browse/EDS-2108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EDS-1587]: https://czi.atlassian.net/browse/EDS-1587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/VisualPageIndicator/VisualPageIndicator.stories.ts
+++ b/src/components/VisualPageIndicator/VisualPageIndicator.stories.ts
@@ -6,6 +6,12 @@ import { VisualPageIndicator } from './VisualPageIndicator';
 export default {
   title: 'Components/VisualPageIndicator',
   component: VisualPageIndicator,
+  parameters: {
+    docs: {
+      subtitle:
+        'Static visual cue to help users understand their current position within a series of content or pages.',
+    },
+  },
   tags: ['autodocs', 'version:1.0'],
 } as Meta<Args>;
 
--- a/src/components/VisualPageIndicator/VisualPageIndicator.tsx
+++ b/src/components/VisualPageIndicator/VisualPageIndicator.tsx
@@ -22,7 +22,40 @@ export type VisualPageIndicatorProps = {
 };
 
 /**
- * Static visual cue to help users understand their current position within a series of content or pages.
+ * ## Usage
+ *
+ * Show to users which step they are on, in a given flow. Help to illustrate current position on
+ * pagination experiences.
+ *
+ * ## Interaction
+ *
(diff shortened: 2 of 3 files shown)


GregSilvia (approved):
LGTM

